Issue #197, opened 2026-06-22
Regression (#190): field-level FK drift synthesis mishandles embedded/inline-relation fields — emits the mixin struct name as the table (+ duplicates), and strips a field-annotated table-level FK's ON DELETE


denisvmedia:
## Summary

The field-level FK drift synthesis added in #190 (`synthesizeFieldLevelForeignKeyConstraints`, `migration/schemadiff/internal/compare/compare.go`) produces **broken migrations** when run against a model set that uses **embedded inline-relation mixins** (`//migrator:embedded mode="inline"` structs that carry `foreign=` fields). Discovered by bumping the downstream consumer denisvmedia/inventario to the #190 ptah and running `./scripts/generate-migration.sh`.

The library builds fine against the consumer (API change is additive); the problem is purely in **generated migration SQL** (diff path), and only surfaces against the embedded-mixin pattern — which ptah's own test models don't exercise, so CI stayed green.

## Symptom 1 — embedded mixin struct name emitted as a table, duplicated

The synthesis iterates `generated.Fields` including fields contributed by embedded inline mixins (e.g. inventario's `TenantGroupAwareEntityID` / `TenantUserAwareEntityID`, which carry `created_by_user_id`, `group_id`, `tenant_id`, `user_id` FKs). It synthesizes FK constraints whose **table is the mixin's Go struct name**, not each concrete host table that embeds it:

```sql
-- repeated ~13× (once per embedding table, all collapsed to the bogus name), re-ADD with no DROP and no ON DELETE:
ALTER TABLE TenantGroupAwareEntityID ADD CONSTRAINT fk_entity_created_by FOREIGN KEY (created_by_user_id) REFERENCES users(id);
ALTER TABLE TenantGroupAwareEntityID ADD CONSTRAINT fk_entity_group      FOREIGN KEY (group_id) REFERENCES location_groups(id);
ALTER TABLE TenantUserAwareEntityID  ADD CONSTRAINT fk_entity_user       FOREIGN KEY (user_id) REFERENCES users(id);
ALTER TABLE TenantGroupAwareEntityID ADD CONSTRAINT fk_entity_tenant     FOREIGN KEY (tenant_id) REFERENCES tenants(id);
```

`TenantGroupAwareEntityID` is not a table — it's a base struct embedded inline into many tables. The CREATE path resolves embedded relations to their host tables (`processEmbeddedRelationMode`), but the new diff-synthesis path does not, so it emits the struct name and duplicates it per embedding table.

[...]

Issue #189, opened 2026-06-22
Bug: schema-diff does not detect on_delete / on_update changes on existing field-level foreign keys (silent no-op migration)


denisvmedia:
## Problem

When a foreign key is declared on a `//migrator:schema:field` annotation and its `on_delete` / `on_update` action is **changed** on an **already-applied** schema (the table and FK already exist), `migrate generate` reports **"No schema changes detected"** and emits no migration. The referential action drift is silently ignored, so the only way to actually change the action downstream is to hand-edit the SQL.

This is **distinct from #117** (closed by PR #122). #117 fixed CREATE/ALTER-time *emission* of field-level FK actions, and that works correctly at current `master` (`TestPlanner_FieldLevelForeignKeyActions`, `TestParseField_ForeignKeyActions`, `TestFromField_ForeignKeys` all pass; a fresh `CREATE`/`ADD CONSTRAINT` now carries `ON DELETE`). What remains broken is the **drift/diff detection** of a *change* on a pre-existing field-level FK — the follow-up PR #122 itself deferred in its "Known pre-existing gaps" section ("Drift detector skips field-level FK constraints … Tracked alongside #89").

## Reproduction

Model carries `on_delete="SET NULL"`; the live DB FK is `NO ACTION`:

```go
//migrator:schema:field name="file_id" type="TEXT" foreign="files(id)" foreign_key_name="fk_export_file" on_delete="SET NULL"
FileID *string
```

`schemadiff.Compare(generated, dbSchema)` → `HasChanges() == false`, `ConstraintsAdded == []`, `ConstraintsRemoved == []`, `TablesModified == 0`.

For comparison, the **table-level** form (`//migrator:schema:constraint type="FOREIGN KEY" … on_delete="SET NULL"`) DOES detect the same change (drop+add) — so the comparator logic itself is fine; only the field-level path never reaches it.

## Root cause

In `migration/schemadiff/internal/compare/compare.go`, two independent suppressions keep field-level FKs out of the comparison entirely:

1. **Generated side never synthesizes a FK constraint.** `Constraints()` only iterates `generated.Constraints` (table-level `//migrator:schema:constraint`) plus *synthesized* field-level **CHECK** constraints (`synthesizeFieldLevelCheckConstraints`, compare.go:~866-880). There is **no analogous synthesis for field-level FOREIGN KEY** — `Field.Foreign` / `Field.OnDelete` / `Field.OnUpdate` are never turned into a comparable `goschema.Constraint`.
[...]

PR #205, opened 2026-06-23
Fix(planner): scope modified-FK DROP to its known host table (#199)

### LINKED ISSUE #199: Generated DROP-constraint DO-block resolves the FK by name only — should scope by table_name (ambiguous if a constraint name is reused across tables)
## Problem

When the postgres planner emits a DROP for a modified field-level FK (the DROP-then-ADD in `addNewConstraints` / `removeConstraints`), it renders a runtime DO-block that resolves the owning table by **constraint name alone**:

```sql
SELECT table_name INTO target_table
FROM information_schema.table_constraints
WHERE constraint_name = 'fk_export_file'
  AND table_schema = current_schema()
LIMIT 1;

IF target_table IS NOT NULL THEN
    EXECUTE format('ALTER TABLE %I DROP CONSTRAINT IF EXISTS %I', target_table, 'fk_export_file');
```

PostgreSQL constraint names are unique **per table**, not per schema — the same name can legitimately exist on two tables. With `LIMIT 1` and no `table_name` filter, this can resolve (and drop) the **wrong** table's constraint, leaving the intended one in place.

The planner already knows the owning table at emit time (it's emitting the matching `ADD CONSTRAINT … ON <host>`), so it can scope the lookup precisely. Suggested generated form:

```sql
SELECT table_name INTO target_table
FROM information_schema.table_constraints
WHERE constraint_name = 'fk_export_file'
  AND table_schema = current_schema()
  AND table_name = 'exports'      -- the host the planner is emitting for
LIMIT 1;

IF target_table IS NOT NULL THEN
    EXECUTE format('ALTER TABLE %I.%I DROP CONSTRAINT IF EXISTS %I', current_schema(), target_table, 'fk_export_file');
```

## Why it matters now

Surfaced by CodeRabbit on a downstream migration (denisvmedia/inventario#2181) generated by the post-#198 ptah. In that schema `fk_export_file` happens to be globally unique, so there's no live risk — but the generated SQL is robustness-fragile for any schema that reuses a constraint name across tables, and downstream can't hand-fix it (the `_sqldata/*.sql` files are generator output).

## Fix

In the postgres planner's modify-DROP DO-block emission (and the `removeConstraints` table-qualified path, where the table is already known via `ConstraintsRemovedWithTables`), include `AND table_name = '<host>'` and a schema-qualified `ALTER TABLE %I.%I`. Where the host is already known (the common case after #197/#198), prefer emitting a direct `ALTER TABLE <schema>.<host> DROP CONSTRAINT IF EXISTS <name>` without the lookup at all.

## Acceptance

- [ ] The generated DROP DO-block (or direct drop) is scoped to the specific host table.
- [ ] A test with the same constraint name on two tables drops only the intended one.

Found via downstream #2181 review.

denisvmedia:
Fixes #199.

## Problem

When the postgres planner emits a DROP for a modified field-level FK (the DROP-then-ADD in `addNewConstraints`), `emitModifyDrop` had a two-branch split:

- the **shared-name** (>1 host) branch already emitted a direct table-qualified `ALTER TABLE <host> DROP CONSTRAINT IF EXISTS <name>`;
- the **single-host** branch fell back to `dropConstraintNode` — the name-only `information_schema` DO-block (`WHERE constraint_name = X AND table_schema = current_schema() LIMIT 1`) — **even though `add.TableName` was already known.**

PostgreSQL constraint names are unique *per table*, not per schema. With `LIMIT 1` and no `table_name` filter, that lookup can resolve and drop a same-named constraint on the **wrong** table, leaving the intended one in place. Surfaced by CodeRabbit on downstream `denisvmedia/inventario#2181`.

## Fix

`emitModifyDrop` now **always** emits a direct table-qualified drop (`ast.AlterTableNode{Name: add.TableName}` + `DropConstraintOperation{IfExists: true}`), deduped per `(table, name)`, since the concrete host is carried on `ConstraintAdditionInfo.TableName` and is always known at emit time. The two branches collapse into one.

- Removed the now-dead `sharedNameHosts` map and the `sharedNameHosts` parameter (it only selected the drop *form*, which no longer varies).
- The name-only DO-block (`dropConstraintNode`) is **retained, unchanged**, for the genuinely-unknown-host paths: `removeConstraints` over bare `diff.ConstraintsRemoved`, and the legacy bare-`ConstraintsAdded` modify path — there the diff layer has discarded the owning table, so there is nothing to scope by.
- The `ConstraintsRemovedWithTables` host-known path in `removeConstraints` was already a direct drop and is unchanged.
- Doc comments rewritten to capture the #189 / #197 / #199 rationale.

## Acceptance criteria

- [x] The generated DROP is scoped to the specific host table (direct `ALTER TABLE <host> DROP CONSTRAINT IF EXISTS <name>`, no name-only lookup when the host is known).
- [x] A test with the same constraint name on two tables drops only the intended one.

## Tests

- **New** `TestPlanner_GenerateMigrationAST_ModifiedFK_ScopesDropToHostTable` (`exclude_constraints_test.go`):
  - same name `fk_customer` on `orders` + `invoices`, only `orders` modified → asserts `ALTER TABLE orders DROP CONSTRAINT IF EXISTS fk_customer;`, **no** `information_schema.table_constraints`, `invoices` untouched, and DROP precedes the re-ADD;
  - both hosts modified → each gets exactly one scoped drop and its own distinct action.
- Updated the #189 golden assertion in `fk_action_drift_test.go`: the single-host modify now emits the direct `ALTER TABLE exports DROP CONSTRAINT IF EXISTS fk_export_file;` instead of the name-only DO-block, and asserts the DO-block is absent.

`go build ./...`, `go vet ./migration/...`, `gofmt -l`, and full `go test ./...` all pass.

## Notes

The rendered direct drop is schema-unqualified and uses unquoted identifiers — consistent with every other direct `ALTER TABLE` drop ptah already emits (the #197 multi-host path, `removeConstraints`' `WithTables` path). Adding schema qualification / identifier quoting to `AlterTableNode` across dialects is a reasonable follow-up but out of scope for #199.

### DIFF
--- a/migration/planner/dialects/postgres/exclude_constraints_test.go
+++ b/migration/planner/dialects/postgres/exclude_constraints_test.go
@@ -186,6 +186,109 @@ func TestPlanner_GenerateMigrationAST_ConstraintsAdded(t *testing.T) {
 	}
 }
 
+// TestPlanner_GenerateMigrationAST_ModifiedFK_ScopesDropToHostTable guards the
+// issue #199 fix: a MODIFIED field-level FK (dropped + re-added) must have its
+// DROP scoped to the concrete host table that the comparator recorded, never the
+// name-only information_schema DO block. PostgreSQL constraint names are unique
+// per table, so when the same name exists on two tables and only one is modified,
+// the name-only `... WHERE constraint_name = X ... LIMIT 1` lookup could resolve
+// and drop the constraint on the WRONG table. The planner already knows the host
+// (ConstraintAdditionInfo.TableName), so it emits a direct table-qualified drop.
+func TestPlanner_GenerateMigrationAST_ModifiedFK_ScopesDropToHostTable(t *testing.T) {
+	t.Run("same constraint name on two tables, only one modified, drops only the intended host", func(t *testing.T) {
+		c := qt.New(t)
+
+		// Both `orders` and `invoices` carry an FK literally named `fk_customer`
+		// (e.g. via an embedded inline-relation mixin). Only the one on `orders`
+		// drifted (its on_delete changed), so the comparator reports a modify for
+		// orders alone: orders appears in BOTH the additions and removals, while
+		// invoices appears nowhere.
+		diff := &types.SchemaDiff{
+			ConstraintsAdded:   []string{"fk_customer"},
+			ConstraintsRemoved: []string{"fk_customer"},
+			ConstraintsAddedWithTables: []types.ConstraintAdditionInfo{
+				{
+					Name:          "fk_customer",
+					TableName:     "orders",
+					Type:          "FOREIGN KEY",
+					Columns:       []string{"customer_id"},
+					ForeignTable:  "customers",
+					ForeignColumn: "id",
+					OnDelete:      "CASCADE",
+				},
+			},
+			ConstraintsRemovedWithTables: []types.ConstraintRemovalInfo{
+				{Name: "fk_customer", TableName: "orders", Type: "FOREIGN KEY"},
+			},
+		}
+
+		nodes := postgres.New().GenerateMigrationAST(diff, &goschema.Database{})
+		sql, err := renderer.RenderSQL("postgres", nodes...)
+		c.Assert(err, qt.IsNil)
+
+		// The DROP is scoped to the intended host with a direct ALTER TABLE.
+		c.Assert(strings.Contains(sql, "ALTER TABLE orders DROP CONSTRAINT IF EXISTS fk_customer;"), qt.IsTrue,
+			qt.Commentf("modified FK must be dropped from its known host table; got:\n%s", sql))
+
+		// The unsafe name-only DO block resolution must NOT be used for a
+		// known-host modify — that is what could hit the wrong table.
+		c.Assert(strings.Contains(sql, "information_schema.table_constraints"), qt.IsFalse,
+			qt.Commentf("known-host modify must not use the name-only DO block; got:\n%s", sql))
+
+		// The unrelated host (invoices) must be left completely alone: no drop,
+		// no add.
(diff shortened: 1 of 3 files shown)


denisvmedia:
## Deep self-review + scope follow-up

Ran a deep adversarial self-review of this change (multi-lens review → adversarial verification of every finding → completeness critic). Outcome: the FK fix is correct and genuinely guarded (the new test and the updated #189 golden were verified to **fail** against the pre-change code), with **no blockers**.

The review surfaced that the same per-table DROP ambiguity #199 describes also affected **non-FK** modifies (table-level `UNIQUE` / `CHECK` / `EXCLUDE`), which are reached via the bare `ConstraintsAdded` loop and still emitted the name-only `information_schema` DO-block. Since the comparator populates `ConstraintsRemovedWithTables` in lockstep with the bare list, the host is in fact known there.

**Pushed a follow-up commit (e6346b0)** generalizing the fix to non-FK constraints:
- the legacy modify-DROP now routes through `emitModifyDropForName`, which emits a direct table-qualified `ALTER TABLE <host> DROP CONSTRAINT IF EXISTS <name>` per recorded host (shared `appendScopedDrop` helper, deduped per `(table, name)`);
- a synthetic diff with no recorded host still falls back to the name-only DO-block defensively;
- new test `TestPlanner_GenerateMigrationAST_ModifiedNonFKConstraint_ScopesDropToHostTable` (UNIQUE name reused on two tables, only one modified → only that host dropped; plus the fallback case).

`go build ./...`, `go vet ./...`, `gofmt -l`, and the full `go test ./...` (uncached) are all green.

**Out of scope here, tracked in #206:** the *removal* side (`removeConstraints`) has the same bare-name keying — a constraint name modified on table A but purely removed on table B skips B's drop. It is **pre-existing** (byte-identical to `master`, equally broken before), so it does not block this PR.

Issue #207, opened 2026-06-23
MySQL/MariaDB planner: same (table,name) constraint-scoping bugs as #199/#206 on BOTH add and remove sides


denisvmedia:
## Problem

#199 (PR #205) and #206 fixed the **postgres** planner so modified/removed constraints are scoped to their concrete host table by `(table, name)`. The **MySQL/MariaDB** planner (`migration/planner/dialects/mysql/mysql.go`) has the analogous defect class on **both** sides, so the same mixed scenario produces broken DDL there.

Found while verifying #206: an attempt to mirror only the *remove*-side keying into MySQL produced an inconsistent half-fix (broken SQL for the mixed non-FK case), so the MySQL change was dropped from the #206 PR and split out here. MySQL must be fixed on **both** sides together.

### Failure scenario (mixed modify + pure-removal of a shared constraint name)

A constraint name (e.g. `shared_check`) shared across tables `articles` and `pages`: **modified** on `articles` (lands in `ConstraintsAdded`), **purely removed** on `pages`. On MySQL the generated migration is wrong:

1. **Add side** (`addNewConstraints`, ~`mysql.go:626-637`): the modify-drop resolves the owning table via `removalByName`, keyed on the **bare name** (last-write-wins across removal hosts). For a name with ≥2 removal hosts it collapses to one arbitrary host — emitting `ALTER TABLE pages DROP CONSTRAINT shared_check;` (the **wrong**, pure-removal host) and then `ALTER TABLE articles ADD CONSTRAINT shared_check ...` with **no preceding drop** of the stale `articles` constraint → ADD fails with a duplicate-name error in a real DB.
2. **Remove side** (`removeConstraints`, ~`mysql.go:786-805`): keys its skip-modifies on the **bare name** (same root cause as #206), so it then emits a second `ALTER TABLE pages DROP CONSTRAINT shared_check;` **without `IF EXISTS`** → errors "constraint does not exist", migration aborts.

(The MySQL FK mixed case is closer to correct, but the non-FK case is broken as above.)

## Fix

Mirror the postgres approach from #205/#206, adapted to MySQL DDL (`DROP FOREIGN KEY` vs `DROP {CHECK|INDEX}`):

- [ ] **Remove side:** key the modify-skip on `(table, name)` from `ConstraintsAddedWithTables` (not bare `addedNames`); emit table-qualified drops per `ConstraintsRemovedWithTables` entry.
[...]

PR #208, opened 2026-06-23
Fix(planner): table-scope postgres removeConstraints + mixed modify/removal (#206)

### LINKED ISSUE #206: removeConstraints skips a pure removal when a shared constraint name is also modified on another table (mixed modify + remove)
## Problem

Surfaced by the deep self-review of #199 / PR #205. The fix in #199 made the **add side** (`addNewConstraints` / `emitModifyDrop`) key its modify-DROP on `(table, name)` instead of the bare constraint name. The **removal side** — `removeConstraints` in `migration/planner/dialects/postgres/postgres.go` — still keys its "skip modifies" check on the **bare name**:

```go
addedNames := make(map[string]struct{}, len(diff.ConstraintsAdded))
for _, name := range diff.ConstraintsAdded {
    addedNames[name] = struct{}{}
}
...
for _, info := range diff.ConstraintsRemovedWithTables {
    ...
    if _, modified := addedNames[info.Name]; modified {   // <-- bare name
        handled[info.Name] = struct{}{}
        continue   // skipped as "owned by addNewConstraints as a modify"
    }
    ...
}
```

### Failure scenario

A constraint name shared across two tables (e.g. a field-level FK `fk_x` contributed by an embedded inline-relation mixin), where:

- table **A**: the constraint is **modified** (drop + re-add) → `fk_x` lands in `ConstraintsAdded`
- table **B**: the constraint is **purely removed** → `(B, fk_x)` lands in `ConstraintsRemovedWithTables`

Because `fk_x` is in `addedNames` (thanks to A), `removeConstraints` treats **B's** removal as a modify and **skips it** — so B's stale constraint is never dropped. The skip should be keyed on `(table, name)`, not the name alone.

### Severity / scope

- **Pre-existing**, not a regression from #199/#205 — `removeConstraints` is byte-identical to `master`, and master was equally broken here (via the nondeterministic name-only `LIMIT 1` DO-block). So this does **not** block #205.
- Reproduced end-to-end through the real planner + renderer during the #205 self-review.

## Fix

Mirror what #199 did on the add side: key the modify-skip on `(table, name)` using `ConstraintsRemovedWithTables`, and drive the whole function off the table-qualified list.

Since the comparator populates `ConstraintsRemovedWithTables` / `ConstraintsAddedWithTables` **in lockstep** with the bare `ConstraintsRemoved` / `ConstraintsAdded` lists (`migration/schemadiff/internal/compare/compare.go`, the only append sites), the owning table is always available. We do **not** need backwards compatibility with the bare-name path — it can be retired:

- [ ] In `removeConstraints`, build the modify-skip set as `(table, name)` from `ConstraintsAddedWithTables`, not bare `addedNames`.
- [ ] Emit the pure-removal DROP as a direct table-qualified `ALTER TABLE <host> DROP CONSTRAINT IF EXISTS <name>` for every `ConstraintsRemovedWithTables` entry.
- [ ] Consider removing the name-only `dropConstraintNode` DO-block path entirely (or reducing it to a guarded fallback), once `removeConstraints` no longer relies on the bare list. The `addNewConstraints` legacy path already fell back to it defensively only for synthetic diffs with no recorded host (see PR #205).
- [ ] Optionally generalize the same `(table, name)` keying to MySQL/MariaDB planners if they share the pattern.

## Acceptance

- [ ] A planner test with a shared constraint name where it is **modified on table A and purely removed on table B** drops B's constraint (table-qualified) and re-adds only A's.
- [ ] No remaining name-only constraint-owner resolution where the host is known.

## Notes

- Related: #199 (PR #205) fixed the add-side modify-DROP scoping (FK in #205's first commit, non-FK in the follow-up commit).
- The same per-table ambiguity also applies to non-FK constraint removals; the table-qualified path covers all constraint types.

Found via the PR #205 self-review.

denisvmedia:
Fixes #206.

## Problem

`removeConstraints` (postgres planner) keyed its "skip modifies" check on the **bare constraint name**. A constraint name shared across two tables that is **modified on table A** (its name lands in `ConstraintsAdded`) but **purely removed on table B** was wrongly treated as a modify owned by `addNewConstraints` on B too — so **B's stale constraint was never dropped**.

## Fix

- **Remove side:** key the modify-skip on `(table, name)` from `ConstraintsAddedWithTables`, and emit a direct table-qualified `ALTER TABLE <host> DROP CONSTRAINT IF EXISTS <name>` per `ConstraintsRemovedWithTables` entry (deduped per `(table, name)` via the `appendScopedDrop` helper from #205). The comparator fills the `*WithTables` lists in lockstep with the bare lists, so real diffs are always table-scoped; the name-only `dropConstraintNode` DO-block is now used **only** as a defensive fallback for a synthetic diff that carries no recorded host (the existing `TestPlanner_GenerateMigrationAST_ConstraintsRemoved` unit test still exercises that fallback and stays green).
- **Add side (counterpart):** `emitModifyDropForName` now restricts the modify pre-drop to the hosts actually being **re-added** (`addedHostsByName`). So a shared name modified on A and purely removed on B is dropped **once on A** (add side, drop-then-re-add) and **once on B** (`removeConstraints`), never twice on either, and B is never skipped.

## Acceptance

- [x] A shared constraint name modified on table A and purely removed on table B drops B (table-qualified) and re-adds only A; B is not skipped and no name-only DO-block is used.

## Tests

- **New** `TestPlanner_GenerateMigrationAST_SharedConstraintName_ModifiedOnOneTablePurelyRemovedOnAnother`: `shared_check` on `articles` (modified) + `pages` (purely removed) → `ALTER TABLE pages DROP CONSTRAINT IF EXISTS shared_check;` is emitted (B not skipped), `articles` is dropped-then-re-added, only A is re-added, each host dropped exactly once, no `information_schema` DO-block.
- `go build ./...`, `go vet ./...`, `gofmt -l`, and the full uncached `go test ./...` (33 packages) all pass.

## Scope / follow-ups

- **MySQL/MariaDB** share this bug class on **both** the add and remove sides and need a coordinated both-sides fix (adapted to `DROP FOREIGN KEY` / `DROP {CHECK,INDEX}` and an `IF EXISTS`-equivalent). A remove-side-only mirror produced broken DDL for the mixed non-FK case, so MySQL was deliberately **left untouched** here and split out into **#207**. This PR stays postgres-scoped and fully verified.
- Noted pre-existing (not addressed here): for a multi-host **non-FK** modify where the *same* constraint name is modified on two tables, the re-add resolution (`addConstraintNodeFor`) matches by name and re-adds only the first host — flagged in #207's notes.

This change was developed in an isolated worktree and went through a multi-agent implement→verify workflow; the MySQL blocker above was caught by the adversarial verification pass.

### DIFF
--- a/migration/planner/dialects/postgres/exclude_constraints_test.go
+++ b/migration/planner/dialects/postgres/exclude_constraints_test.go
@@ -386,27 +386,139 @@ func TestPlanner_GenerateMigrationAST_ModifiedNonFKConstraint_ScopesDropToHostTa
 // addNewConstraints and skipped — leaving the stale constraint on B. The fix
 // keys the skip on (table, name), so B is dropped table-qualified, A is dropped
 // and re-added, and no name-only DO block is used.
+//
+// The bug only reproduces for the FOREIGN KEY shape: an FK modify on A is dropped
+// by the ConstraintsAddedWithTables loop (which never touches B), so a buggy
+// bare-name skip in removeConstraints drops B zero times. The CHECK shape is a
+// weaker guard — the old add-side modify-drop already dropped B as a phantom
+// pre-drop, so the bug is masked there except for statement ordering. We cover
+// both: the FK subtest is the load-bearing #206 regression guard.
 func TestPlanner_GenerateMigrationAST_SharedConstraintName_ModifiedOnOneTablePurelyRemovedOnAnother(t *testing.T) {
+	t.Run("foreign key (the load-bearing #206 guard)", func(t *testing.T) {
+		c := qt.New(t)
+
+		// `shared_fk` exists on both `articles` and `pages`. It drifted on
+		// `articles` (a modify: present in both additions and removals) and was
+		// removed outright on `pages` (removal only). The FK modify on articles is
+		// owned by the ConstraintsAddedWithTables loop, so removeConstraints alone
+		// is responsible for dropping pages — which the bare-name skip got wrong.
+		diff := &types.SchemaDiff{
+			ConstraintsAdded:   []string{"shared_fk"},
+			ConstraintsRemoved: []string{"shared_fk"},
+			ConstraintsAddedWithTables: []types.ConstraintAdditionInfo{
+				{Name: "shared_fk", TableName: "articles", Type: "FOREIGN KEY", Columns: []string{"author_id"}, ForeignTable: "users", ForeignColumn: "id", OnDelete: "CASCADE"},
+			},
+			ConstraintsRemovedWithTables: []types.ConstraintRemovalInfo{
+				{Name: "shared_fk", TableName: "articles", Type: "FOREIGN KEY"},
+				{Name: "shared_fk", TableName: "pages", Type: "FOREIGN KEY"},
+			},
+		}
+
+		nodes := postgres.New().GenerateMigrationAST(diff, &goschema.Database{})
+		sql, err := renderer.RenderSQL("postgres", nodes...)
+		c.Assert(err, qt.IsNil)
+
+		// The pure removal on B (pages) is dropped table-qualified and NOT skipped.
+		// This is the assertion that fails against the pre-fix bare-name skip.
+		c.Assert(strings.Contains(sql, "ALTER TABLE pages DROP CONSTRAINT IF EXISTS shared_fk;"), qt.IsTrue,
+			qt.Commentf("purely-removed host must be dropped table-qualified, not skipped; got:\n%s", sql))
+		// The modify on A (articles) is dropped table-qualified and re-added once.
+		c.Assert(strings.Contains(sql, "ALTER TABLE articles DROP CONSTRAINT IF EXISTS shared_fk;"), qt.IsTrue,
+			qt.Commentf("modified host must be dropped from its known table; got:\n%s", sql))
+		c.Assert(strings.Contains(sql, "ALTER TABLE articles ADD CONSTRAINT shared_fk FOREIGN KEY (author_id) REFERENCES users(id) ON DELETE CASCADE;"), qt.IsTrue,
+			qt.Commentf("modified FK must be re-added on its host; got:\n%s", sql))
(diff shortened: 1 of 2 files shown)


denisvmedia:
## Deep self-review — two issues found and fixed (commit 5f4aeb0)

Ran a deep adversarial self-review (6 review lenses → 2 skeptics per finding → completeness critic). The production `removeConstraints` up-path fix was confirmed correct, but two real problems surfaced:

**1. The shipped #206 test guarded nothing (major).** It used a CHECK shape, which never exhibited the bug — the old add-side `emitModifyDropForName` already dropped the pure-removal host as a phantom pre-drop, so the test **passed against the buggy pre-fix code**. The bug only reproduces for a **FOREIGN KEY** shape (the FK modify is owned by the `ConstraintsAddedWithTables` loop, so `removeConstraints` alone must drop the pure-removal host). Reworked the test into an FK subtest (the load-bearing guard) + a CHECK subtest asserting the discriminating statement order. **Verified the FK subtest fails against master** (pure-removal host left undropped).

**2. Down-direction regression (major, caught by the critic).** `emitModifyDropForName`'s `addedHosts` gating fell straight to the name-only `information_schema` DO-block when `addedHosts` was empty. `reverseConstraintAdditions` restores only FOREIGN KEY additions (and nothing without schema context), so a **non-FK modify in a down migration** has an empty `ConstraintsAddedWithTables` → the down migration regressed the table-scoped drop #205 emitted, contradicting #206's own acceptance criterion. Fixed: when `addedHosts` is empty, scope by every recorded removal host before the DO-block fallback. Added `TestPlanner_GenerateMigrationAST_ModifyDrop_ScopesToHostWhenAddedHostsAbsent`, **verified to fail against the previously-committed #206 code**.

Both new guards were checked to fail against the buggy versions (not just pass against the fix). `go build`, `go vet`, `gofmt -l`, and the full uncached `go test ./...` (33 packages) all green. `mysql.go` remains identical to master (tracked in #207).

Issue #164, opened 2026-05-17
Role / GRANT / privilege management as first-class schema objects


denisvmedia:
## Why

Multi-tenant apps using RLS (already supported in Ptah) need to grant `SELECT/INSERT/UPDATE/DELETE` to specific app roles. Today users have to hand-write the GRANT statements in a custom `.up.sql` because Ptah has no `:role` or `:grant` annotation. Atlas covers this; Ent doesn't but ptah's RLS support is the natural place for it.

## Scope

New annotations:

```go
//migrator:schema:role name="app_role" login="false" inherit="true"
//migrator:schema:grant role="app_role" privilege="SELECT,INSERT,UPDATE,DELETE" on_table="users"
//migrator:schema:grant role="app_role" privilege="USAGE" on_schema="public"
```

Diff support:

- Adding/removing a GRANT → emit `GRANT`/`REVOKE`.
- Role attribute changes → `ALTER ROLE`.

Introspection: read from `pg_roles`, `information_schema.role_table_grants`.

## Acceptance
- Round-trip a schema with two roles and per-table grants.
- README example showing the multi-tenant RLS + role recipe end-to-end.

## References
- https://www.postgresql.org/docs/current/sql-grant.html
- https://atlasgo.io/atlas-schema/hcl#roles

PR #282, opened 2026-07-13
fix(goschema): wire Constraints/Views/MaterializedViews/Triggers/Grants through ParseFS/ParseDir (fixes #279)

### LINKED ISSUE #279: ParseFS silently drops views, materialized views, triggers, grants and table-level constraints — every CLI command loses them
## Why

`ParseFS` drops five of the fifteen collections it parses. Views, materialized views, triggers, grants and table-level constraints declared in Go annotations never reach any CLI command.

`core/goschema/walker.go:120-129` — the merge step after parsing each file:

```go
		// Add to result
		result.EmbeddedFields = append(result.EmbeddedFields, database.EmbeddedFields...)
		result.Fields = append(result.Fields, database.Fields...)
		result.Indexes = append(result.Indexes, database.Indexes...)
		result.Tables = append(result.Tables, database.Tables...)
		result.Enums = append(result.Enums, database.Enums...)
		result.Extensions = append(result.Extensions, database.Extensions...)
		result.Functions = append(result.Functions, database.Functions...)
		result.RLSPolicies = append(result.RLSPolicies, database.RLSPolicies...)
		result.RLSEnabledTables = append(result.RLSEnabledTables, database.RLSEnabledTables...)
		result.Roles = append(result.Roles, database.Roles...)
```

Ten appends. `Database` (`core/goschema/types.go:31-45`) has fifteen slice collections. Missing: `Constraints`, `Views`, `MaterializedViews`, `Triggers`, `Grants`.

`ParseSource` populates all fifteen — the parser is fine (`core/goschema/parser.go:365`, `:369`, `:377` dispatch view/trigger/grant comments). The loss happens purely in the per-file merge, so the data is parsed and then thrown away.

This is the path every CLI command takes: `ParseDir` (`walker.go:42`) delegates straight to `ParseFS` (`walker.go:71`), and it is what `cmd/generate/generate.go:144`, `cmd/compare/compare.go:71`, `cmd/migrate/migrate.go:115` and `cmd/internal/schemaops/schemaops.go:54` all call. `ParseFile` — the single-file entry point used by most unit tests — does not go through the merge and returns everything, which is why the parser tests pass.

Repro, same file through both entry points:

```go
//migrator:schema:table name="users"
type User struct {
	//migrator:schema:field name="id" type="SERIAL" primary="true"
	ID int64
	//migrator:schema:field name="email" type="VARCHAR(255)" not_null="true"
	Email string
}

//migrator:schema:view name="active_users" body="SELECT id, email FROM users"
type ActiveUsersView struct{}
```

```
ParseDir  -> Tables=1 Fields=2 Views=0
ParseFile -> Tables=1 Fields=2 Views=1
```

So `ptah generate --root-dir ./models` emits the table and silently omits the view. `ptah compare` reports the schema as synchronized while the view is missing from the database — and if the view exists in the database, `ptah migrate` will happily plan to drop it, since it is absent from the desired schema. Same for materialized views, triggers, grants and table-level constraints.

The features are otherwise implemented: #158 (views/matviews/triggers) and #164 (grants) landed with renderers, planners and diff support, and they work through the YAML frontend (`core/yamlschema/yamlschema.go`), which builds `Database` directly and does not go through the walker. Only the Go-annotation frontend — the project's headline feature — cannot reach them.

Nothing catches this because no Go fixture anywhere in the repo declares one: `rg -l 'migrator:schema:view' --type go` returns only `core/goschema/parser.go`, `core/goschema/types.go` and `core/goschema/parser_test.go`. Every existing test either parses a single file (`ParseFile`, unaffected) or constructs `goschema.Database` by hand.

Related: `Deduplicate` (`core/goschema/walker.go:139`) only handles `Tables`, `Fields`, `Indexes` and `Enums`. Even the collections that *are* appended — extensions, functions, RLS policies, roles — are not deduplicated, so declaring the same extension in two files emits it twice.

## Scope

1. Append the five missing collections in `walker.go:120-129`: `Constraints`, `Views`, `MaterializedViews`, `Triggers`, `Grants`.
2. Extend `Deduplicate` to cover every collection, not just the original four. Keyed on the natural identity of each object (view name + schema, trigger name + table, grant on role + object, constraint name + table).
3. Decide and document what `normalizeTableScopedNames` (`walker.go:140`) must do for the new objects — a view or trigger declared in a schema-scoped file needs the same treatment tables get.
4. Add Go-annotation fixtures under the integration suite for view, materialized view, trigger, grant and table-level constraint, and drive them through `generate` and `migrate` — not just through `ParseFile`.
5. Add a guard so this class of bug cannot recur: a test that reflects over `Database`'s fields and fails if `ParseFS` produces an empty collection for one that `ParseSource` populated. A hand-maintained append list will drift again the next time a schema object is added.

## Acceptance

- The repro above yields `ParseDir -> Views=1`.
- `ptah generate --root-dir <fixtures> --dialect postgres` emits `CREATE VIEW`, `CREATE MATERIALIZED VIEW`, `CREATE TRIGGER`, `GRANT` and a table-level `CONSTRAINT` from Go annotations.
- `ptah compare` against a database that already has those objects reports zero diff (today it reports them as missing from the desired schema and plans drops).
- Declaring the same extension in two files under `--root-dir` produces one `CREATE EXTENSION`, not two.
- The reflection guard fails if a sixteenth collection is added to `Database` and not wired into the walker.

## References

- #158 — views, materialized views and triggers.
- #164 — roles / grants.
- #267 — round-trip conformance harness; the `desired -> apply -> introspect -> diff == 0` loop over Go-annotation fixtures is what would have caught this at the point the features merged.

denisvmedia:
## Summary

Fixes #279: `ParseFS`/`ParseDir` (the path used by `generate`/`compare`/`migrate` and `migration/generator`) was silently dropping five collections from Go `//migrator:schema:*` annotations:

- `Constraints`
- `Views`
- `MaterializedViews`
- `Triggers`
- `Grants`

(Only `Roles` had been wired in the recent #164 work; the rest + dedup/guard were missing.)

## Changes

- `core/goschema/walker.go`: explicit inits + appends for the five in `ParseFS` (after the `Roles` line).
- `core/goschema/utils.go`: extend `Deduplicate` + `deduplicateSchemaObjects` to cover Grants/Constraints/Roles with appropriate keys; added helpers. Updated `normalizeTableScopedNames` + decision comment for Views/MaterializedViews.
- New fixture `integration/fixtures/entities/023-go-annotations-objects/` (table + all five + cross-file dup view case).
- Unit tests + reflection guard in `walker_test.go` (future-proofs against 16th collection drift).
- Sub-test in gonative to drive the real `ParseDir` path.

## Verification (per plan)
- Unit + `-race` + lint (double pass) clean.
- `ptah generate` on fixture emits `CREATE VIEW`, `MATERIALIZED VIEW`, `TRIGGER`, `GRANT` (x2), `ROLE`, table-level `CONSTRAINT`.
- Reflection guard + ParseDir assertions pass and would catch drift.
- 3 full rounds of multi-agent self-review (feature/logic, bugs/gaps/tests, security/completeness) performed via `spawn_subagent`; all findings addressed. Artifacts in scratch.

Refs #279

## 3-round multi-agent self-check
1. Feature + logic (core wiring + dedup + normalize)
2. Bugs + gaps + tests (all kinds) + edges
3. Security (fs/input/panics/paths/cleanup/secrets) + completeness + final signoff

All green. Ready for merge when pipelines pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,8 +30,8 @@ capabilities include:
 - 🧱 **Schema Generation (DDL)**
   Builds platform-specific `CREATE TABLE`, `CREATE INDEX`, and other DDL statements.
 
-- 🔐 **Row-Level Security (PostgreSQL)**
-  Supports PostgreSQL RLS policies and custom functions for multi-tenant data isolation.
+- 🔐 **Row-Level Security And Access Control (PostgreSQL)**
+  Supports PostgreSQL RLS policies, roles, grants, and custom functions for multi-tenant data isolation.
 
 - 🔍 **Database Introspection**
   Reads the current schema directly from PostgreSQL, MySQL, MariaDB, or ClickHouse for comparison and analysis.
@@ -347,6 +347,25 @@ type User struct {
 - `using` - USING clause expression for row filtering
 - `with_check` - WITH CHECK clause expression for INSERT/UPDATE validation
 - `comment` - Policy description
+
+#### Role Attributes (PostgreSQL only)
+- `name` - Role name
+- `login` - Whether the role can login
+- `password` - Encrypted password hash
+- `superuser` - Whether the role has superuser privileges
+- `createdb` or `create_db` - Whether the role can create databases
+- `createrole` or `create_role` - Whether the role can create roles
+- `inherit` - Whether the role inherits privileges
+- `replication` - Whether the role can initiate replication
+- `comment` - Role description
+
+#### Grant Attributes (PostgreSQL only)
+- `role` - Role receiving the privilege
+- `privilege` or `privileges` - One privilege or a comma-separated list such as `SELECT,INSERT,UPDATE,DELETE`
+- `on_table` - Target table for table privileges
+- `on_schema` - Target schema for schema privileges such as `USAGE`
+- `with_option` - Whether to emit `WITH GRANT OPTION`
+- `comment` - Grant description
 
 #### Constraint Attributes
 - `name` - Constraint name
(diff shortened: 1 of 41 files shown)


denisvmedia:
Superseded by #283.

This branch (`agent/fix-279-parsefs`) was cut from the pre-squash `agent/roles-grants-164` commit (`0d4773a`), not from master. Grants have since landed on master as `846c5a4` (#259), so the merge base here is the commit *before* grants — which is why the diff re-applies the entire #164 feature (41 files, +1785) on top of work master already has.

#283 carries the same #279 fix cut from master: 6 files, +283/-1, only the walker/dedup delta. Reviewing that one.

PR #283, opened 2026-07-13
fix(goschema): wire Constraints/Views/MaterializedViews/Triggers/Grants through ParseFS/ParseDir (fixes #279)

### LINKED ISSUE #279: ParseFS silently drops views, materialized views, triggers, grants and table-level constraints — every CLI command loses them

denisvmedia:
## Summary

Fixes #279 (isolated clean PR).

`ParseFS`/`ParseDir` (CLI + generator path) was dropping 5 collections from Go annotations.

## Scope (minimal, 6 files)
- walker.go: inits + appends for Constraints/Views/MaterializedViews/Triggers/Grants
- utils.go: full dedup coverage + normalize decision doc
- walker_test.go: ParseDir test + reflection guard
- new fixture 023-go-annotations-objects (with columns for valid SQL) + dup for dedup test
- gonative subtest driving ParseDir

## Verification
- All unit/race/lint pass.
- generate emits the CREATE statements for all 5 + role + constraint.
- Guard + dedup tests pass.
- 3 detailed multi-agent review rounds executed and captured.

Refs #279. This PR is clean (only the 279 delta vs master).

### DIFF
--- a/core/goschema/utils.go
+++ b/core/goschema/utils.go
@@ -1,11 +1,13 @@
 package goschema
 
 import (
+	"fmt"
 	"log/slog"
 	"maps"
 	"regexp"
 	"slices"
 	"sort"
+	"strconv"
 	"strings"
 	"sync"
 )
@@ -247,18 +249,11 @@ func normalizeTableScopedNames(r *Database) {
 		}
 	}
 	// Views and MaterializedViews: no table-scoped normalization applied here.
-	// Decision: unlike Triggers/Constraints/Grants/Indexes/RLS which reference a .Table,
-	// Views/MaterializedViews declare a standalone .Name (which may include schema prefix
-	// e.g. "public.foo" or be unqualified). They are not "table scoped" from a host table
-	// struct in the same manner. If view names need struct-to-name resolution or schema
-	// inference in future (e.g. schema-scoped Go files), extend this loop similarly.
-	// Current behavior preserved for compatibility with YAML and existing parser paths.
-	for i := range r.Views {
-		_ = &r.Views[i] // name left as-parsed
-	}
-	for i := range r.MaterializedViews {
-		_ = &r.MaterializedViews[i]
-	}
+	// Unlike Triggers/Constraints/Grants/Indexes/RLS, which reference a .Table,
+	// Views/MaterializedViews declare a standalone .Name (which may include a
+	// schema prefix, e.g. "public.foo", or be unqualified). They are not table
+	// scoped from a host table struct in the same manner. Current behavior is
+	// preserved for compatibility with YAML and existing parser paths.
 }
 
 func resolveTableReference(tables []Table, structName, tableName string) *Table {
@@ -822,8 +817,8 @@ func isFunctionInSorted(function Function, sorted []Function) bool {
 //   - Embedded Fields: Deduplicated by struct name + embedded type name combination
 //   - Views and materialized views: Deduplicated by name
 //   - Triggers: Deduplicated by table name + trigger name combination
-//   - Constraints: Deduplicated by table + constraint name
-//   - Grants: Deduplicated by role + privileges + (table or schema) target
+//   - Constraints: Deduplicated by explicit table + name, or declaring struct + name when table is omitted
+//   - Grants: Deduplicated by role + privileges + grant option + (table or schema) target
 //   - Roles: Deduplicated by role name
 //
 // All 15 Database slice collections are now covered (previously only a subset
@@ -996,15 +991,15 @@ func deduplicateTriggers(triggers []Trigger) []Trigger {
 	return deduplicated
 }
 
-// deduplicateConstraints dedups table-level constraints by (StructName + name).
-// Uses StructName (the declaring Go type) so dedup happens before normalizeTableScopedNames
-// qualifies .Table (which may be empty when annotation omits `table=` and relies on struct assoc).
-// This matches the pattern used for indexes (StructName + Name).
+// deduplicateConstraints dedups table-level constraints by their declared
+// table scope when table= is explicit, otherwise by the declaring Go type.
+// The fallback is needed before normalizeTableScopedNames fills .Table for
+// annotations that rely on struct association.
 func deduplicateConstraints(constraints []Constraint) []Constraint {
 	seen := make(map[string]bool)
 	deduplicated := make([]Constraint, 0, len(constraints))
 	for _, c := range constraints {
-		key := c.StructName + "." + c.Name
+		key := constraintDedupKey(c)
 		if !seen[key] {
 			seen[key] = true
 			deduplicated = append(deduplicated, c)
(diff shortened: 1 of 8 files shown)


core/goschema/utils.go
```diff
@@ -971,3 +995,50 @@ func deduplicateTriggers(triggers []Trigger) []Trigger {
...
+
+// deduplicateGrants dedups by role + privileges + target (table or schema).
+// Uses Canonicalize for stable privilege list.
+func deduplicateGrants(grants []Grant) []Grant {
+	seen := make(map[string]bool)
+	deduplicated := make([]Grant, 0, len(grants))
+	for _, g := range grants {
+		g.Canonicalize()
+		privs := strings.Join(g.Privileges, ",")
+		key := g.Role + "|" + privs + "|t:" + g.OnTable + "|s:" + g.OnSchema
```
denisvmedia: **Blocking: `WITH GRANT OPTION` is silently swallowed.**

The key omits `WithOption`. Two grants that differ *only* by the grant option collapse into one, and first-wins means the survivor is whichever the file walk hit first:

```go
db := &Database{Grants: []Grant{
    {Role: "app", Privileges: []string{"SELECT"}, OnTable: "users", WithOption: false},
    {Role: "app", Privileges: []string{"SELECT"}, OnTable: "users", WithOption: true},
}}
Deduplicate(db)
// grants after dedup: 1
//   role=app privs=[SELECT] table=users WITH GRANT OPTION=false
```

So a declared `GRANT ... WITH GRANT OPTION` is silently downgraded to a plain grant. That is the same class of silent privilege loss this PR exists to fix, just moved one layer down. `WithOption` is part of the grant's identity for rendering, so it belongs in the key:

```go
key := g.Role + "|" + privs + "|t:" + g.OnTable + "|s:" + g.OnSchema + "|o:" + strconv.FormatBool(g.WithOption)
```

(`GrantedBy` is introspection-only and should stay out; `Comment` is cosmetic and should stay out — but then see the note below about which duplicate wins.)

---

**Also on this line: the key is order-sensitive on privileges.**

The comment above says "Uses Canonicalize for stable privilege list", but `Canonicalize` (`core/goschema/types.go:731`) trims, uppercases and de-dupes — it does **not** sort. So:

```go
{Role: "app", Privileges: []string{"SELECT", "INSERT"}, OnTable: "users"},
{Role: "app", Privileges: []string{"INSERT", "SELECT"}, OnTable: "users"},
// grants after dedup: 2
```

Same logical grant, not deduped. Harmless compared to the above (a false negative, not data loss), but either sort the privileges when building the key, or drop the word "stable" from the comment.

integration/fixtures/entities/023-go-annotations-objects/duplicate.go
```diff
@@ -0,0 +1,4 @@
+package entities
+
+//migrator:schema:view name="active_users" body="SELECT id, email FROM users WHERE deleted_at IS NULL" comment="Duplicate should be ignored by dedup"
```
denisvmedia: **The dedup fixture proves the opposite of what it intends.**

Files are walked in lexical order, so `duplicate.go` is parsed *before* `users.go`, and dedup is first-wins. The declaration that survives is this one — the one the fixture itself labels as the duplicate. Running `generate` on the fixture:

```
-- Statement 3/7
-- Duplicate should be ignored by dedup
CREATE VIEW active_users AS
SELECT id, email FROM users WHERE deleted_at IS NULL
;
```

The comment on the emitted view is `Duplicate should be ignored by dedup`, not `Active users view`. The test passes only because both declarations happen to carry an identical `body=`; it asserts `HasLen 1` and `Name == "active_users"` and never looks at what actually won.

Two consequences worth deciding on explicitly:

1. **Conflicting duplicates are silently resolved by filename.** If the two declarations had different `body=`, one definition would win at random (well — alphabetically) and the other would vanish with no diagnostic. For a schema tool that is the wrong default: same object name with a *different* definition is a user error and should be an error, not a coin flip. Same-name-same-definition can stay a silent no-op.
2. If you keep silent first-wins, at least make the fixture deterministic and assert it: give the two declarations distinguishable comments and assert which one survives, so the behavior is pinned rather than accidental.

This applies to views/matviews/triggers/roles/constraints alike, not just this fixture.

core/goschema/walker_test.go
```diff
@@ -1234,6 +1236,117 @@ type Product struct {
...
+	fvMerged := reflect.ValueOf(merged)
+	fvDir := reflect.ValueOf(*dirDb)
+	typ := fvMerged.Type()
+	for i := 0; i < fvMerged.NumField(); i++ {
+		if fvMerged.Field(i).Kind() != reflect.Slice {
+			continue
+		}
+		name := typ.Field(i).Name
+		mLen := fvMerged.Field(i).Len()
+		if mLen == 0 {
```
denisvmedia, line 1344: **The reflection guard does not do what its doc comment claims.**

Line 1296 says it "catches if a future 16th collection is added without wiring the append in walker". It cannot, because of this `continue`: a collection that the *fixture* does not declare has `mLen == 0` and is skipped entirely.

The fixture (`023-go-annotations-objects`) declares tables, fields, constraints, views, matviews, triggers, roles and grants — 8 of the 15 slices. `Indexes`, `Enums`, `EmbeddedFields`, `Extensions`, `Functions`, `RLSPolicies` and `RLSEnabledTables` are never exercised, so the guard would not notice if their appends were deleted from `walker.go` tomorrow. And a 16th collection added without a fixture entry passes green — which is exactly the drift this test is supposed to prevent.

The guard needs to fail on *fixture* gaps, not skip them. Roughly:

```go
if mLen == 0 {
    c.Fatalf("%s is not exercised by the fixture — add it to 023-go-annotations-objects "+
        "so the walker append stays covered", name)
}
```

That inverts the failure mode: adding a collection to `Database` without extending the fixture becomes a loud test failure, and the person who adds collection 16 is forced to cover it. It does mean the fixture must grow to declare an index, an enum, an embedded field, an extension, a function and an RLS policy — which is worth doing anyway, since those are the collections nothing else drives through `ParseDir`.

Nit while you are here: the test is named `TestParseFS_ReflectionGuard` but calls `ParseDir`.

core/goschema/utils.go
```diff
@@ -246,6 +246,19 @@ func normalizeTableScopedNames(r *Database) {
...
+	// Views/MaterializedViews declare a standalone .Name (which may include schema prefix
+	// e.g. "public.foo" or be unqualified). They are not "table scoped" from a host table
+	// struct in the same manner. If view names need struct-to-name resolution or schema
+	// inference in future (e.g. schema-scoped Go files), extend this loop similarly.
+	// Current behavior preserved for compatibility with YAML and existing parser paths.
+	for i := range r.Views {
+		_ = &r.Views[i] // name left as-parsed
+	}
+	for i := range r.MaterializedViews {
+		_ = &r.MaterializedViews[i]
```
denisvmedia: These two loops are dead code — they iterate and discard a pointer. `golangci-lint` does not flag them (I ran it: 0 issues), which is exactly why they will survive forever if they land.

The explanatory comment above them is genuinely useful and should stay. The loops should go:

```go
// Views and MaterializedViews: no table-scoped normalization applied here.
// Unlike Triggers/Constraints/Grants/Indexes/RLS, which reference a .Table,
// Views/MaterializedViews declare a standalone .Name ...
```

and nothing else. A `for` loop whose body is `_ = &x[i]` reads like an unfinished edit, and the next person will either delete it or, worse, try to figure out what it was guarding.

core/goschema/utils.go
```diff
@@ -809,8 +822,16 @@ func isFunctionInSorted(function Function, sorted []Function) bool {
 //   - Embedded Fields: Deduplicated by struct name + embedded type name combination
 //   - Views and materialized views: Deduplicated by name
 //   - Triggers: Deduplicated by table name + trigger name combination
+//   - Constraints: Deduplicated by table + constraint name
```
denisvmedia: Doc says "Deduplicated by table + constraint name". The code (line 1007) keys on `StructName + "." + Name`:

```go
db := &Database{Constraints: []Constraint{
    {StructName: "User",      Name: "chk_email", Table: "users"},
    {StructName: "UserMixin", Name: "chk_email", Table: "users"},
}}
Deduplicate(db)
// constraints after dedup: 2
```

Two constraints with the same name on the same table survive, so the documented contract is not the implemented one. The `StructName` key is defensible — it matches how indexes are keyed and it works before `normalizeTableScopedNames` has filled in `.Table` — so I would keep the code and fix the doc line rather than the reverse.

Worth a thought though: with mixins/embedded structs, two structs mapping to the same table *can* legitimately declare the same constraint name, and that would emit a duplicate `CONSTRAINT` into one `CREATE TABLE`. If that is reachable today it is a real bug; if it is not, say so in the comment so the next reader does not have to work it out.

core/goschema/walker_test.go
```diff
@@ -1234,6 +1236,117 @@ type Product struct {
...
+func TestParseDir_AllCollectionsCoveredByReflectionGuard(t *testing.T) {
+	c := qt.New(t)
+	// Simple smoke: ensure after ParseDir we have non-nil slices for the key ones
+	// (detailed guard test below uses reflection against ParseSource)
+	fixtureDir := "../../integration/fixtures/entities/023-go-annotations-objects"
+	result, err := goschema.ParseDir(fixtureDir)
+	c.Assert(err, qt.IsNil)
+	_ = result.Views
+	_ = result.Grants
+}
```
denisvmedia: This test asserts nothing. `_ = result.Views` and `_ = result.Grants` are discards; the only real assertion is `err == nil`, which `TestParseDir_SchemaObjectsAndGrants` already covers with the same fixture.

Its own comment concedes it ("the real guard is in `TestParseFS_ReflectionGuard`"). Delete it — a test that cannot fail is worse than no test, because it makes the guard count look higher than it is.

denisvmedia (commented):
Keeping this one over #282 — #282 was cut from the pre-squash `agent/roles-grants-164` commit, so its merge base predates grants landing on master (`846c5a4`, #259) and its diff re-applies all of #164 on top of work master already has. #282 is closed.

**Do not merge until 1 is fixed.** (GitHub will not let me set Request Changes on my own PR, so: treat this as one.)

The core of the PR is right, and I verified it rather than trusting the description: `go build`, `go test -race ./core/... ./migration/...` and `golangci-lint run` are all clean on the branch, and `generate` on the new fixture emits the table, the role, the view, the matview, both grants, the table-level constraint and the trigger — with a valid auto-generated wrapper function and `EXECUTE FUNCTION`, so the DDL is applicable, not merely present in the output. The five missing appends in `walker.go` and the three missing dedup paths are exactly the right fix.

**Blocking**

1. `deduplicateGrants` omits `WithOption` from the key, so `GRANT ... WITH GRANT OPTION` is silently collapsed into the plain grant (first-wins by file order). Proven with a two-grant repro. This is the same class of silent privilege loss the PR exists to fix, one layer down.

**Should fix before merge**

2. Conflicting duplicates are resolved silently by filename order. The dedup fixture demonstrates the hazard by accident: `duplicate.go` sorts before `users.go`, so the *duplicate* wins and its comment is what lands in the generated `CREATE VIEW`. Same-name-different-definition should be an error, not an alphabetical coin flip.
3. The reflection guard cannot catch what its comment says it catches — it `continue`s past any collection the fixture does not declare, which is 7 of the 15. A 16th collection added without a fixture entry passes green. It needs to fail on fixture gaps instead of skipping them.
4. `deduplicateConstraints` is documented as keyed on table + name; it is keyed on struct + name. Fix the doc (the code is fine), and confirm whether a mixin declaring the same constraint on the same table can emit a duplicate `CONSTRAINT`.

**Nits**

5. The two `_ = &r.Views[i]` loops in `normalizeTableScopedNames` are dead code. Keep the comment, drop the loops.
6. `TestParseDir_AllCollectionsCoveredByReflectionGuard` asserts nothing — delete it.

One correction on my side, not yours: the "Related" paragraph in #279 claimed `Deduplicate` only covered four collections and that extensions get emitted twice. Both were wrong — master already deduped 12 of 15, and the real gap was exactly `Constraints` / `Roles` / `Grants`, which is what this PR closes. I have corrected the issue.

core/goschema/utils.go
denisvmedia: Fixed in 4111d4e. The grant dedup key now includes WithOption and uses a sorted privilege copy for identity; regression coverage keeps plain grants, grant-option grants, and reordered privilege duplicates distinct/collapsed as expected.

integration/fixtures/entities/023-go-annotations-objects/duplicate.go
denisvmedia: Fixed in 4111d4e. The duplicate fixture is now an identical no-op that sorts after the primary declaration, and the test asserts the surviving comment. Conflicting duplicate schema objects now fail ParseFS instead of being resolved by file order.

core/goschema/walker_test.go
denisvmedia: Fixed in 4111d4e. The reflection guard is now TestParseDir_ReflectionGuard, the fixture exercises every Database slice collection, and the guard fails when any slice has no fixture coverage.

core/goschema/utils.go
denisvmedia: Fixed in 4111d4e. Removed the dead loops and kept the explanatory comment.

core/goschema/utils.go
denisvmedia: Fixed in 4111d4e. The docs now state the real constraint identity. I also tightened the implementation: explicit table= constraints dedupe/validate by table scope, while omitted table= keeps the declaring-struct fallback before normalization. Added regression coverage for the mixin-style same-table case.

core/goschema/walker_test.go
denisvmedia: Fixed in 4111d4e. Deleted the no-op test; the reflection guard now carries the coverage.

PR #692, opened 2026-07-25
Detect column widening in schema diff (INTEGER->BIGINT, VARCHAR length)

### LINKED ISSUE #691: schema apply misses column precision changes (INTEGER->BIGINT, VARCHAR length): diff normalizer over-collapses
## Problem

`ptah atlas schema apply` (and the underlying schema diff) does **not** detect a change to a column type's **precision/width**, so it never plans the modification and reports "Schema is synced, no changes to be made":

- integer width: `INTEGER` → `BIGINT`
- varchar length: `VARCHAR(50)` → `VARCHAR(100)`

Repro (PostgreSQL):

```
# DB has: CREATE TABLE users (id SERIAL PRIMARY KEY, age INTEGER);
$ ptah atlas schema apply --url "$PG" --to file://desired.sql --auto-approve
#   where desired.sql = CREATE TABLE users (id SERIAL PRIMARY KEY, age BIGINT);
Schema is synced, no changes to be made.     # WRONG — age stays integer
```

The end state after applying the `BIGINT` desired schema onto an `INTEGER` column is still `integer`, diverging from a database built directly from the desired schema (which is `bigint`). Atlas plans and applies the `ALTER COLUMN ... TYPE` here.

Changes Ptah **does** detect correctly (for contrast): nullability (`SET`/`DROP NOT NULL`) and cross-category type changes (`INTEGER` → `TEXT`).

## Cause

`migration/schemadiff/internal/normalize/normalize.go` canonicalizes column types before comparison and collapses all integer variations (`INT`, `INTEGER`, `BIGINT`, `SMALLINT`, ...) to a single `"integer"` (see the file's leading doc comment), and similarly drops `VARCHAR` length. Two columns that differ only in integer width or varchar length therefore compare equal, so the differ emits no change.

This normalization avoids false-positive churn from dialect type aliases (`INT` vs `INTEGER`, `INT4` vs `integer`), but it is too aggressive: `SMALLINT` / `INTEGER` / `BIGINT` are distinct types with different storage and range, and `VARCHAR(50)` vs `VARCHAR(100)` is a real, applicable change. The fix must distinguish integer width (and varchar length) while keeping the alias-folding (`INT`==`INTEGER`, `INT8`==`BIGINT`) that prevents spurious diffs — so it needs care to avoid regressing existing diff/conformance behavior.

## Surfaced by

The paired-schema planning conformance matrix (stokaro/ptah-atlas-conformance, #652), which applies schema A then B to one database and B to another and compares the canonical introspected end state. The `modify-column-type` (width) and varchar-length cases fail this end-state equivalence; add/drop table/column, nullability, and cross-category type changes pass.

denisvmedia:
## Summary

`ptah atlas schema apply` did not detect column **widenings** — `INTEGER -> BIGINT` and `VARCHAR(50) -> VARCHAR(100)` — so it reported *"Schema is synced, no changes to be made"* and never planned the `ALTER`. A database built directly from the desired schema ends up with `bigint` / `varchar(100)`, while an incrementally-applied one keeps the old width, so the two diverge.

## Root cause

The schema-diff type normalizer (`migration/schemadiff/internal/normalize`.`Type`) deliberately folds integer widths together (`smallint`/`int`/`integer`/`bigint` → `"integer"`) and drops `VARCHAR` length, so that harmless dialect aliases (`int` vs `integer` vs `int4`) do not surface as spurious changes. That same folding hid genuine width/length increases.

## Fix

The normalizer is already paired with a within-category detector — `migration/internal/typechange`.`IsNarrowing` — that catches the *data-losing* direction the normalizer hides (used by the column comparison's narrowing check). Widenings are simply the other direction:

- Add the symmetric `typechange.IsWidening`, reusing the existing `integerRank` so aliases still fold (width is compared by **range, not spelling**: `int == integer == int4`), plus string-length and decimal precision/scale increases.
- The column comparison now reports a within-category change in **either** direction (`shouldReportNarrowingTypeChange` → `shouldReportSizedTypeChange`, returning `IsNarrowing(...) || IsWidening(...)`).
- The SQLite type-affinity guard is preserved, so SQLite — where these width distinctions do not exist — is unaffected.

The migration planner already renders a `"type"` change as `ALTER COLUMN ... TYPE <desired>` regardless of direction (widening reuses the exact path narrowing and cross-category changes already use), and `migration/safety` keeps using `IsNarrowing` alone (widenings are non-destructive).

## Verification

Live PostgreSQL apply (the scenario #691 reported as broken):

- `INTEGER -> BIGINT`: now plans `type: int4 -> BIGINT`; column becomes `bigint`.
- `VARCHAR(50) -> VARCHAR(100)`: now plans the change; length becomes `100`.
- `BIGINT -> INTEGER` (narrowing): still works — no regression.
- `INTEGER -> INTEGER` and dialect aliases: still *"Schema is synced"* — no false positives / non-convergence.

Unit tests added for `IsWidening` (including alias-safety and widening/narrowing symmetry) and for the column comparison (integer, varchar, postgres length-metadata, and decimal widening cases). Full unit suite, `go vet`, `gofmt`, test-style, and `golangci-lint` all pass.

Fixes #691.

### DIFF
--- a/integration/fixtures/entities/004-field-rename/post.go
+++ b/integration/fixtures/entities/004-field-rename/post.go
@@ -13,7 +13,7 @@ type Post struct {
 	//migrator:schema:field name="content" type="TEXT" not_null="true"
 	Content string
 
-	//migrator:schema:field name="user_id" type="BIGINT" not_null="true"
+	//migrator:schema:field name="user_id" type="INTEGER" not_null="true"
 	UserID int64
 
 	//migrator:schema:field name="status" type="ENUM" enum="draft,published,archived" not_null="true" default="draft"
--- a/integration/fixtures/entities/005-field-type-change/post.go
+++ b/integration/fixtures/entities/005-field-type-change/post.go
@@ -13,7 +13,7 @@ type Post struct {
 	//migrator:schema:field name="content" type="TEXT" not_null="true"
 	Content string
 
-	//migrator:schema:field name="user_id" type="BIGINT" not_null="true"
+	//migrator:schema:field name="user_id" type="INTEGER" not_null="true"
 	UserID int64
 
 	//migrator:schema:field name="status" type="ENUM" enum="draft,published,archived" not_null="true" default="draft"
--- a/integration/fixtures/entities/006-field-drop/post.go
+++ b/integration/fixtures/entities/006-field-drop/post.go
@@ -13,7 +13,7 @@ type Post struct {
 	//migrator:schema:field name="content" type="TEXT" not_null="true"
 	Content string
 
-	//migrator:schema:field name="user_id" type="BIGINT" not_null="true"
+	//migrator:schema:field name="user_id" type="INTEGER" not_null="true"
 	UserID int64
 
 	//migrator:schema:field name="status" type="ENUM" enum="draft,published,archived" not_null="true" default="draft"
--- a/integration/fixtures/entities/007-index-add/post.go
+++ b/integration/fixtures/entities/007-index-add/post.go
@@ -13,7 +13,7 @@ type Post struct {
 	//migrator:schema:field name="content" type="TEXT" not_null="true"
 	Content string
 
-	//migrator:schema:field name="user_id" type="BIGINT" not_null="true"
+	//migrator:schema:field name="user_id" type="INTEGER" not_null="true"
 	UserID int64
 
 	//migrator:schema:field name="status" type="ENUM" enum="draft,published,archived" not_null="true" default="draft"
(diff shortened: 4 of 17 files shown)


denisvmedia:
Pushed a follow-up commit correcting incidental `user_id` drift in the `dynamic_basic_evolution` integration fixtures: versions 002-003 declared the FK column `posts.user_id` as `INTEGER` and 004+ as `BIGINT`, so version 004 (a field-rename) also carried an unintended INTEGER→BIGINT widening of a foreign-key column. The pre-existing normalizer bug silently dropped that widening; with detection fixed it becomes a real `ALTER` that MySQL/MariaDB reject on a FK column (errors 3780/1832). Pinned `user_id` to a stable `INTEGER` across versions (dedicated type-change coverage already lives in 005-field-type-change). Verified locally: the scenario now passes on MySQL and MariaDB, and the full suite is 195/195 on postgres+mysql+mariadb. The underlying MySQL/MariaDB planner gap (retyping a FK column needs drop/recreate of the constraint) is tracked in #694.

PR #769, opened 2026-07-27
Recreate foreign keys around MySQL/MariaDB column type changes

### LINKED ISSUE #694: MySQL/MariaDB: column type change on a foreign-key column fails (needs drop/recreate FK)
## Problem

On MySQL and MariaDB, ptah generates a bare `ALTER TABLE ... MODIFY <col> <newtype>` when a column's type changes. If that column participates in a foreign key — as either the referencing or the referenced column — the statement is rejected:

- MySQL: `Error 3780 (HY000): Referencing column 'user_id' and referenced column 'id' in foreign key constraint 'fk_posts_user_id' are incompatible.`
- MariaDB: `Error 1832 (HY000): Cannot change column 'user_id': used in a foreign key constraint 'fk_posts_user_id'`

PostgreSQL applies the same `ALTER COLUMN ... TYPE` without complaint, so this is a MySQL-family planner gap.

## Reproduction

Widen a FK-participating column, e.g. `posts.user_id INTEGER` (referencing `users.id`) → `BIGINT`, and apply on MySQL/MariaDB. The migration fails at the `MODIFY` step. Surfaced by the dynamic_basic_evolution integration scenario once column widening detection landed (#691 / #692): the scenario's incidental `user_id` widening was pinned back to a stable `INTEGER` to keep the suite green, which sidesteps but does not fix this limitation.

## Expected (Atlas parity)

Atlas plans a type change on a FK column by dropping the affected foreign key(s), altering the column (and, where required, the referenced column so the types stay compatible), then recreating the foreign key(s) — all within the migration. Ptah's MySQL/MariaDB planner should do the same: detect that the column is referenced by or referencing a FK and emit `DROP FOREIGN KEY` → `MODIFY` → `ADD CONSTRAINT ... FOREIGN KEY` in dependency order (and keep referencing/referenced column types in lockstep).

## Notes

- PostgreSQL is unaffected.
- Care needed to alter both ends of the FK when only one side's type changes, or MySQL will reject the recreated constraint as incompatible.

denisvmedia:
## Summary

Fixes #694. On MySQL and MariaDB, a column-type change on a column that participates in a foreign key was emitted as a bare `ALTER TABLE ... MODIFY <col> <newtype>`, which the server rejects while the constraint exists:

- MySQL: `Error 3780 ... Referencing column ... and referenced column ... in foreign key constraint ... are incompatible.`
- MariaDB: `Error 1832 ... Cannot change column ...: used in a foreign key constraint ...`

The canonical reproduction is widening a FK column such as `posts.user_id INTEGER` (referencing `users.id`) to `BIGINT`. PostgreSQL applies the same change in place, so this was purely a MySQL-family planner gap.

## Root cause

`modifyExistingColumns` in the MySQL-family planner (`internal/planner/dialects/mysql`) emitted a single `MODIFY COLUMN` per changed column and never accounted for foreign keys. MySQL/MariaDB refuse to alter the type of a column that is referenced by or referencing a foreign key until the constraint is dropped.

## Fix

For every diff, the MySQL-family planner now determines which foreign keys a column-type change touches (as the **referencing or referenced** column, enumerated from field-level `foreign=` references, self-referencing keys, and table-level constraints), and:

1. Drops each affected key that **already exists** BEFORE any column modification.
2. Runs the column modifications.
3. Recreates the affected keys AFTER all modifications.

All affected keys are dropped before any `MODIFY` and recreated only after all of them, so a valid widening that changes **both ends** of a key stays type-compatible when the constraint is recreated. The recreation is driven from the schema the planner is handed — the target schema on the up path, the introspected pre-change schema on the down path — so the generated down migration is the exact inverse of the up migration.

### Ownership split with the constraint machinery — keyed on (table, name)

A column-type change can coincide with a foreign-key **definition** change (an `ON DELETE` drift) on the same key, and a foreign-key **name can be shared across tables** (an embedded inline-relation mixin, #197/#207). Ownership is therefore resolved **per (table, name)**, never the bare name:

- **Added-only** (in the diff's FK additions but not its removals): a brand-new key on an existing column. Not pre-dropped (it does not exist yet); `addNewConstraints` adds it after the modifications.
- **Modified** (in both additions and removals): pre-dropped by the column-type bracketing **before** the `MODIFY`; the re-add (with the new definition) is left to `addNewConstraints`, which runs after the `MODIFY`.
- **Removed-only** (in the removals): pre-dropped before the `MODIFY`; no re-add.
- **Not in the FK constraint diff at all**: a pure column-type change; the bracketing owns both the pre-drop and the post-`MODIFY` re-add.

The planner threads the set of pre-dropped `(table, name)` pairs into `addNewConstraints` and `removeConstraints` so they suppress their own now-redundant drop of the same key — MySQL has no `IF EXISTS` guard on constraint drops, so a duplicate would abort the migration. The owning-table identifier uses the **qualified** table name (matching the `MODIFY` statement and the constraint comparator); the conventional constraint name still derives from the **bare** table name (matching `fromschema.GenerateForeignKeyName` on the FK-creation path).

Example (MySQL, up — widen + `ON DELETE` change on the same key):

```sql
ALTER TABLE posts DROP FOREIGN KEY fk_posts_user_id;                              -- bracketing, before MODIFY
ALTER TABLE posts MODIFY COLUMN user_id BIGINT;                                   -- column change
ALTER TABLE posts ADD CONSTRAINT fk_posts_user_id FOREIGN KEY (user_id)           -- constraint machinery, after MODIFY
    REFERENCES users(id) ON DELETE SET NULL;                                      -- with the new definition
```

Down is the inverse, restoring the prior definition.

### Scope and safety

- **PostgreSQL is unchanged** — it keeps its single `ALTER COLUMN ... TYPE` with no constraint churn (regression test included).
- **SQL Server** shares this planner and gains the same bracketing, rendered as valid `DROP CONSTRAINT` / `ALTER COLUMN` / `ADD CONSTRAINT` T-SQL (SQL Server has the same restriction). Verified locally.
- Only **type** changes trigger the bracketing; nullability/default-only changes do not disturb the referential type match and keep the bare `MODIFY`.

## Testing

New unit and end-to-end tests, all green:

- `internal/planner/dialects/mysql/fk_coltype_test.go` (MySQL + MariaDB): referencing-column change, referenced-column change, both-ends, table-level composite FK, self-referencing FK, down-migration inverse, **coincident type + `ON DELETE` change (asserts DROP → MODIFY → ADD order)**, **co-hosted shared FK name (both hosts dropped and recreated)**, **removed-only FK**, non-type change keeps the FK, no-FK bare `MODIFY`, and the MariaDB `IF EXISTS` guard.
- `migration/planner/fk_coltype_test.go`: end-to-end via the public API — MySQL/MariaDB bracket the change; PostgreSQL unchanged; SQL Server rendered as valid T-SQL in order.
- `migration/generator/fk_coltype_down_internal_test.go`: the real generator up/down path (`schemadiff.CompareWithDialect` → `generateUpMigrationSQL` / `generateDownMigrationSQL`) for both the unchanged-key widening and the coincident `ON DELETE` change, asserting DROP → MODIFY → ADD order and exact inverse on MySQL and MariaDB.

Gates run locally: `go build ./...`, `gofmt -l` (clean), `go vet`, `golangci-lint run` (0 issues), `go tool teststyle ./...` (OK), `go tool qtlint ./...`, full `go test ./...` (pass), and `scripts/check-public-api-snapshot.sh` (no exported-API change).

Live MySQL/MariaDB integration tests require Docker and were not run locally; they run in CI. The `dynamic_basic_evolution` scenario still pins `user_id` to `INTEGER` (pinned in #691/#692); un-pinning it to exercise this fix against live databases is a natural follow-up, intentionally out of scope here to keep the change reviewable and decoupled from the live-DB CI matrix.


denisvmedia:
Addressed the two review blockers in commit 74595c6.

- **Blocker 1 (coincident type + FK-definition change → MODIFY before DROP):** the column-type bracketing now OWNS the pre-MODIFY `DROP FOREIGN KEY` for every affected existing key, so the order is DROP → MODIFY → ADD. The re-add of a modified key is left to `addNewConstraints`, which now suppresses its own drop for keys the bracketing already dropped (pre-seeded `droppedForModify`). `removeConstraints` suppresses likewise.
- **Blocker 2 (name-only keying leaves a co-hosted key undropped):** ownership is now resolved per `(table, name)` via the `ConstraintsAdded/RemovedWithTables` host sets, never the bare name. A shared FK name that is a modification on one host and a pure widening on another now drops and recreates both hosts.
- **SHOULD-FIX 3:** the owning-table identifier now uses `QualifiedName()` (matching the MODIFY and the comparator); the constraint name still derives from the bare table name (matching `GenerateForeignKeyName`).

Added-only vs modify/removed-only are distinguished per `(table, name)`: added-only keys are not pre-dropped. New tests assert DROP → MODIFY → ADD order (not just counts) for the coincident case up and down, the co-hosted shared-name case, the removed-only case, and the SQL Server rendering. All gates green locally including `go tool teststyle ./...` and the public-API snapshot.

PR #876, opened 2026-07-29
[smoke] testkit-sync end-to-end test — do not merge

denisvmedia:
Temporary smoke test of the #865/#875 auto-sync: root go.mod touched + one line removed from testkit/go.sum. Expected: the testkit-sync workflow mints the app token, restores go.sum via tidy, pushes the sync commit, and CI re-runs on the synced head. Will be closed without merging.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -94,3 +94,5 @@ tool (
 	golang.org/x/exp/cmd/apidiff
 	golang.org/x/vuln/cmd/govulncheck
 )
+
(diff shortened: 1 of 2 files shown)


denisvmedia:
Smoke test passed: the workflow minted the app token, restored testkit/go.sum byte-identically, pushed the sync commit, and CI re-ran automatically on the synced head.

Issue #276, opened 2026-07-13
[Atlas C3] Read a limited `atlas.hcl` subset (env / url / dev / migration.dir / exec_order / lint) into the Ptah project config


denisvmedia:
## Why

#250 lists "read `atlas.hcl` `env` blocks" as one bullet of the atlas-mode strategy. It is the *last* of the Atlas compatibility levels, not the first, and it needs its own issue so the epic can close on a decision instead of dragging an implementation behind it.

Today Ptah has no HCL of any kind. `rg -n -i atlas --glob '*.go'` returns five hits and every one is a doc comment (`migration/migratesum/sum.go:16-18`, `:37`, `migration/lint/lint.go:2`). There is no `atlas.hcl` loader, no HCL dependency in `go.mod`, and HCL is explicitly refused where a schema file could plausibly be one — `cmd/generate/generate.go:168-172`:

```go
switch strings.ToLower(filepath.Ext(absPath)) {
case ".yaml", ".yml":
default:
	return nil, fmt.Errorf("unsupported schema file extension %q: only .yaml and .yml are supported", filepath.Ext(absPath))
}
```

That refusal is correct and stays (see the non-goals below). The gap is the *project config*, not the schema.

The second half of the problem is that there is currently **nothing to load into**. Ptah has two unrelated single-purpose config files and no project config:

- `ptah.yaml` — `migration/onlineddl/config.go:29`, whose envelope has exactly one section (`migration/onlineddl/config.go:54-57`):

  ```go
  // ptahConfig is the ptah.yaml envelope.
  type ptahConfig struct {
  	OnlineDDL Config `yaml:"online_ddl"`
  }
  ```

  It is loaded through a plain `yaml.Unmarshal` (`migration/onlineddl/config.go:94`) with no `KnownFields`, so any other key in `ptah.yaml` — `url:`, `dev:`, `dir:` — is silently discarded. Only `migrate-up` and `migrate-down` read it at all (`cmd/migrateup/migrateup.go:175`, `cmd/migratedown/migratedown.go:164`).
- `.ptah-lint.yaml` — `migration/lint/config.go:15`, two keys, `dialect` and `disabled-rules` (`migration/lint/config.go:25-32`), read only by `ptah lint` (`cmd/lint/lint.go:180-188`).

[...]

PR #885, opened 2026-07-29
fix(config): preserve explicit values across merges

### LINKED ISSUE #884: Project config: preserve explicit field presence across source merges
## Why

Project config precedence is documented as CLI flags > `atlas.hcl` > `ptah.yaml` > defaults. The shared `projectconfig.Config` IR currently preserves “absent” versus “explicitly empty/zero” for only a subset of fields through ad hoc `configPresence` booleans.

Other fields are merged with non-zero checks. For example, an explicit `atlas.hcl` value such as `migration { dir = "" }`, `revisions_schema = ""`, or `lock_timeout = ""` does not override a non-empty `ptah.yaml` value. The lower-precedence value is silently restored, so validation and command behavior observe the wrong effective configuration.

This was found while re-auditing #276. Fixing individual migration fields with more one-off booleans would leave the same architecture bug in format, lint, external-schema, and other scalar/list/map fields.

## Scope

1. Make source-field presence a first-class part of the project-config IR for every mergeable field.
2. Preserve presence through Atlas global/env inheritance, Ptah root/env inheritance, and cross-format merging.
3. Make `Merge` apply an explicitly present empty string, zero, false, empty list, or empty map rather than falling back to a lower-precedence value.
4. Keep programmatic `Config` construction ergonomic: non-zero values without loader metadata must continue to behave as explicit overrides.
5. Centralize field-path identifiers or typed presence groups so adding a config field requires one obvious presence/merge implementation path, not unrelated ad hoc booleans.
6. Validate the final effective config after merging so explicit invalid empty values fail with the source-aware error expected by the consuming command.
7. Update project-config documentation to state how explicit empty values participate in precedence.

## Acceptance

- [ ] `ptah.yaml` sets every supported project-config field; an overriding `atlas.hcl` explicitly clears representative string, numeric, boolean, list, map, and nested fields; the effective config reflects the higher-precedence explicit values.
- [ ] Explicitly empty Atlas migration fields do not silently inherit Ptah values.
- [ ] Atlas global/env and Ptah root/env merges preserve the same presence semantics.
- [ ] Programmatically constructed non-zero overrides remain compatible without requiring callers to manipulate internal presence state.
- [ ] Black-box declarative tests cover absent versus explicit zero-value behavior for all config sections.
- [ ] Existing project-config, command, integration, lint, and test-style checks pass.

## Follow-up

After this prerequisite lands, #276 hardening should add an Atlas-independent apply oracle and correct the multiple-env documentation rule.

denisvmedia:
## Summary

- preserve per-field source presence through Ptah root/env, Atlas global/env, and cross-file merges
- carry explicit empty/zero/list presence through native Ptah commands, `ptah atlas`, and `ptah-compat`
- apply command defaults only when project fields are absent while keeping explicit CLI flags highest priority
- resolve `lint.latest` versus Git selection atomically across config and CLI sources
- keep merge results detached from input slices, maps, and pointers
- document exact precedence and native `diff.skip` replacement semantics

## Self-review

- pass 1: audited every project-config field, loader presence, merge propagation, pointer and collection aliasing, and lint selector invariants
- pass 2: audited native Ptah, `ptah atlas`, and `ptah-compat` command boundaries for default resurrection and CLI precedence
- independent review 1: found command-default presence loss, unlabeled Atlas env handling, and same-level lint selector conflicts
- independent review 2: found empty Git selector forwarding, explicit empty `schema.src`, atomic CLI selector precedence, public API safety/snapshot, and deep-copy coverage gaps
- all findings were fixed before the final verification pass

## Verification

- `gopls check` on all changed Go files
- `go test ./... -count=1`
- `go test -race -count=1 ./config/projectconfig ./cmd/internal/dbcli ./internal/migrationlintreport ./cmd/atlas ./cmd/lint`
- `golangci-lint run ./...`
- `go tool qtlint ./...`
- `scripts/check-test-style.sh`
- `scripts/check-public-api-snapshot.sh`
- docs core-link, page-health, link, redirect, exit-code, and versions checks
- `git diff --check`

Fixes #884
Prerequisite for #276

### DIFF
--- a/cmd/atlas/atlas.go
+++ b/cmd/atlas/atlas.go
@@ -616,7 +616,10 @@ func atlasMigrateDirFormatFlag(nativeName string) atlasargs.Flag {
 
 func atlasMigrateDirFormatValue(value string) (string, error) {
 	normalized := strings.ToLower(strings.TrimSpace(value))
-	if normalized == "" || normalized == atlasDirFormatDefault {
+	if normalized == "" {
+		return "", fmt.Errorf("migration directory format must not be empty")
+	}
+	if normalized == atlasDirFormatDefault {
 		return atlasDirFormatDefault, nil
 	}
 	if slices.Contains(unsupportedAtlasDirFormats, normalized) {
--- a/cmd/atlas/migrate_apply.go
+++ b/cmd/atlas/migrate_apply.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/stokaro/ptah/cmd/internal/cmdutil"
 	"github.com/stokaro/ptah/cmd/internal/dbcli"
+	"github.com/stokaro/ptah/config/projectconfig"
 	"github.com/stokaro/ptah/dbschema"
 	"github.com/stokaro/ptah/internal/atlasargs"
 	"github.com/stokaro/ptah/internal/atlasmigrate"
@@ -78,6 +79,7 @@ func atlasMigrateApplyArgs(cmd *cobra.Command, args []string) error {
 
 func runAtlasMigrateApply(cmd *cobra.Command, opts atlasMigrateApplyOptions, args []string) error {
 	formatOutput := cmd.Flags().Changed("format")
+	projectDirFormatPresent := false
 	projectCfg, loaded, err := loadOptionalAtlasProjectConfigForCommand(cmd)
 	if needsAtlasMigrateApplyConfig(cmd) {
 		projectCfg, loaded, err = loadRequiredAtlasProjectConfigForCommand(cmd)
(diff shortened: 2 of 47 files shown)


denisvmedia:
Self-review update: the first independent pass found command-boundary gaps, so this PR is not ready yet. In particular, merged source presence is currently lost when command defaults are applied, and an unlabeled Atlas env does not clear a lower-precedence EnvName. I am extending the patch with typed presence queries, command-level regression coverage, and the lint-selector fixes from the other review pass; full verification and CI will be rerun after that commit.

denisvmedia:
Progress update: typed source presence is now propagated through native Ptah commands, `ptah atlas`, and `ptah-compat`. Local integration review additionally found and fixed two command-boundary issues: empty lint Git selectors were being materialized as changed CLI flags, and an obsolete Atlas lint helper mutated a loaded config value without clearing its provenance. Focused tests, qtlint, test-style checks, docs checks, and targeted lint are green. Two fresh independent reviews and the final full-suite/race/lint pass are still in progress; I will push the consolidated follow-up commit only after those complete.

denisvmedia:
Pushed review-fix commit `0cb2b5d`. It closes every blocker from the two fresh independent reviews: command-boundary default resurrection, atomic latest/Git selector precedence, explicit-empty schema.src handling, unlabeled Atlas env precedence, safe invalid StringField handling, complete collection aliasing coverage, and the public API snapshot. The PR body now contains the final verification matrix. Local full tests, focused race, full golangci-lint, qtlint, teststyle, public API guard, docs checks, and diff check are green; waiting for the new GitHub checks now.

denisvmedia:
Final status: all GitHub checks are green on `0cb2b5d`, including integration-tests, unit-tests, coverage, fuzz, ptah-go-lint, golangci-lint, gosec, govulncheck, GoReleaser, Windows publication, smoke, API export acceptance, and documentation build/checks. The PR is mergeable (`CLEAN`), no inline review threads are open, and the two independent self-review passes have been fully addressed.

Issue #951, opened 2026-08-01
Drop-in for Atlas Pro: expose every implemented Pro capability through ptah-compat


denisvmedia:
## Principle

Every Atlas Pro/Cloud capability Ptah implements must be reachable through `ptah-compat` — the Atlas-shaped surface — not only through native `ptah` verbs. The compat surface is the migration path: Ptah aims to be a drop-in replacement for Atlas **Pro** users' scripts and configs, not just for CE.

Two standing constraints shape how this is done:

- **No non-Atlas flags on compat.** The conformance `cli-surface` tier asserts flag parity with the pinned CE binary. Where a capability needs a safety gate (executing repository-controlled code, destructive operations), the gate is a `PTAH_*` environment variable, invisible to the help surface — precedent: `PTAH_ALLOW_EXTERNAL_SCHEMA` for `atlas.hcl` `data "external_schema"`.
- **Pro-surface flags and verbs are fair game.** Where Atlas *Pro* registers flags/verbs that CE aborts or lacks, compat accepting and implementing them makes Pro pipelines portable; the ce-gating tier documents what CE itself does with each.

## Current violations (from the feature-matrix evidence, all reproduced)

Places where compat fails loudly on a Pro-surface spelling although Ptah has (or could trivially have) the capability:

- [x] `migrate apply --skip-checks` — **resolved by design, not by adding the flag.** Atlas registers no `--skip-checks` on `migrate apply` at all (CE answers `unknown flag`, and the upstream help surface registers it only on `migrate down`), so growing it here would make compat *wider* than Atlas. The bypass is the `PTAH_SKIP_CHECKS` environment variable, per the no-non-Atlas-flags constraint. Verified: compat still answers `unknown flag: --skip-checks`, which is the correct answer.
- [x] `migrate down --plan`, `--to-tag`, `--skip-checks` — all three registered. Verified on a compat build from master: `--plan` (Force dynamic down planning), `--to-tag`, `--skip-checks` all appear in `migrate down --help`.
- [x] `schema plan` output flags — `--edit`, `--format`, `--name-format`, `-o/--output`, `--save`, `--skip-lint` all registered and working. `--directive` and the registry-bound `--push`/`--pending`/`--repo` stay recorded waivers. The local sub-verbs `new` and `validate` landed in #1020; `lint` and `test` stay boundary stubs with their reasons recorded.
[...]

PR #968, opened 2026-08-01
Keep dry-run output quiet on the Atlas-compatible surface

### LINKED ISSUE #967: migrate apply --dry-run writes progress logs to stderr, breaking --format json for 2>&1 consumers
Found by the conformance harness while re-pinning to master (`atlas-cli-report-format` probe flipped ok → gap).

`ptah-compat migrate apply --dry-run --format json` now writes three progress lines to **stderr** before the JSON report:

```
2026/08/01 15:21:06 INFO [DRY RUN] Would begin transaction
2026/08/01 15:21:06 INFO [DRY RUN] Would execute SQL sql="CREATE TABLE users (id INTEGER PRIMARY KEY)" args=[]
2026/08/01 15:21:06 INFO [DRY RUN] Would commit transaction
```

Anyone piping `2>&1 | jq` — the shape the probe uses and a normal CI idiom — gets `invalid character '/' after top-level value`.

**Measured Atlas comparison:** the pinned Atlas CE v1.2.0 binary emits **nothing at all on stderr** for the same dry-run. This is a genuine drop-in divergence, not probe fragility.

**Bisected** across four builds of `cmd/ptah-compat`:

| Build | stderr |
| --- | --- |
| `0642515` (pre-#963 branch pin) | clean |
| `fec2303` (#961) | clean |
| **`a609f81` (#963, "Fix migration dry-run revision state")** | **logs appear** |
| `be310e5` (#962) | logs present (inherited) |

So #963 introduced it; #962 is not at fault.

Fix direction: the dry-run progress logging should not reach stderr on the Atlas-compatible surface (or at minimum must be suppressed whenever a machine-readable `--format` is requested). Decide deliberately what native `ptah` should do — verbose dry-run narration is useful there, but it belongs behind the existing log-level/format controls rather than unconditional stderr.

Blocks: the conformance repo's offline tier goes red on master, which currently blocks stokaro/ptah-atlas-conformance#242 from leaving draft (that repo keeps `waivers.txt` deliberately empty).

Refs #951, #963.

denisvmedia:
Closes #967. Refs #963, #951. Follow-up filed as #969.

## What broke

#963 made `ApplyPlan.Execute` / `DownPlan.Execute` stop short-circuiting on `DryRun` and run the plan through the migrator with the writer in dry-run mode. The dialect writers narrate each simulated statement with the **package-level** `slog` functions:

```go
// internal/dbschema/sqlite/writer.go
if w.dryRun {
    slog.Info("[DRY RUN] Would execute SQL", "sql", sqlExpr, "args", args)
    return nil
}
```

`internal/atlasmigrate` already injects a discard logger into the *migrator* (`WithLogger(slog.New(slog.NewTextHandler(io.Discard, nil)))`), but the writers bypass any injected logger and resolve `slog.Default()` at call time. The Atlas-compatible surface wires no logger at all, so those records went straight to the process's stderr.

Measured on the probe shape:

```console
$ ptah-compat migrate apply --url sqlite://apply.db --dir file://migrations \
    --dry-run --format '{{ json . }}' 2>&1
2026/08/01 15:35:43 INFO [DRY RUN] Would begin transaction
2026/08/01 15:35:43 INFO [DRY RUN] Would execute SQL sql="CREATE TABLE users (id INTEGER PRIMARY KEY)" args=[]
2026/08/01 15:35:43 INFO [DRY RUN] Would commit transaction
{"Driver":"sqlite","URL":{...},"Dir":"file://migrations","Pending":[...],"Target":"1",...}
```

After:

```console
$ ptah-compat migrate apply --url sqlite://apply.db --dir file://migrations \
    --dry-run --format '{{ json . }}' 2>&1 | jq -c '{Target, Pending: (.Pending|length)}'
{"Target":"1","Pending":1}
```

Default format, before → after: three `[DRY RUN]` lines on stderr → stderr empty, stdout unchanged.

## The fix

`cmd/ptah-compat/main.go` installs a quiet default `slog` logger before any command runs (`cliobs.QuietDefaultLogger`) — a text handler at `LevelWarn`.

This makes the contract a **process property** instead of a per-call-site convention. It is not only the writers that log through the package-level `slog` functions — `dbschema.CloseAndWarn`, the `goschema` annotation diagnostics, and the enum-creation notices do too. Suppressing at the call sites, or gating on "a machine-readable `--format` was requested", would leave that class open and need re-litigating every time a library package gains a log line.

The threshold sits on a real seam, not at zero:

- **Info and below is narration** — all 22 `[DRY RUN] Would ...` sites and the two Postgres enum notices. Atlas CE prints none of it. This is what broke #967, and it is dropped.
- **Warn and above is a diagnostic with no other channel** — `core/goschema/utils.go:78` (circular FK ordering), `utils.go:962` (circular function ordering), `dbschema/connection.go:394` `CloseAndWarn` (49 call sites, including `atlasschema/simulate.go` and `migrationreplay/replay.go` — a dev/shadow database that would not close), and `convert/fromschema/fromschema.go:2225` (unrecognized embedding mode silently inlines). None are returned as errors, so dropping them would let a real apply report success while quietly degrading. These are kept.

Atlas parity survives the split because a clean run emits nothing at either level. Verified on the built binary: a circular-FK schema emits `level=WARN msg="Circular dependency detected in foreign key relationships..."` on both `schema apply --dry-run` and `schema apply --auto-approve`, while the JSON probe under `2>&1` still pipes to `jq` and the default-format dry run still leaves stderr byte-empty.

Quiet is not silent in the other direction either: command failures are reported through the command's own error stream, so a dry run that cannot proceed still prints `Error: …` and exits `1`.

Deliberately **not** an exported public API change: `dbschema/types.SchemaWriter` is covered by the released-API baseline and the `docs/public_api.snapshot` guard, so plumbing a logger through the writer interface would have been a breaking change for a regression fix.

## Native surface: keeps the narration, unchanged

Investigating the actual mechanism turned up that the native surface was never the problem. `cliobs.Start` calls `slog.SetDefault(logger)`, so the writers' package-level `slog.Info` already resolves to the command's own handler — the records carry `correlation_id` and `command` attributes, honor `--log-level`, and follow `--log-format` onto stdout as JSON.

```console
$ ptah migrations up ... --dry-run 2>&1 >/dev/null | head -1
time=… level=INFO msg="[DRY RUN] Would execute SQL" correlation_id=… command=migrations.up sql="CREATE TABLE …"

$ ptah migrations up ... --dry-run --log-level warn 2>&1 >/dev/null
(nothing)
```

So the decision is to keep it as-is: the narration is useful when reviewing a plan, it is already behind the documented `--log-level` / `--log-format` controls, and changing its default would be a gratuitous native regression in a compat fix. What was missing was tests and docs pinning that, both of which are added.

## Also fixed: the repeated `Initialize` narration

The #964 reviewer's observation — `[DRY RUN] Would initialize migrations metadata` logged four times per dry run — is a **different root cause in the same PR**, not the same one. It goes through the migrator's *injected* logger, which the compat surface already discards, so it never contributed to #967; it is only visible on the native surface.

The cause is that #963 added `metadataAvailable` / `legacyRevisionTable` caches to `initializeDryRun` but never set `initialized`, so every read entry point (`GetCurrentVersion`, `GetAppliedMigrations`, `GetRevisions`, …) re-ran the whole inspection. It now memoizes like the real path — one narration line, and three fewer `information_schema` round-trips per dry run. The memo is keyed on the writer's dry-run mode so a writer that later leaves dry-run mode still gets its table created (pinned by a test).

`initializeDryRun` is also split so the inspection is a pure query and the state mutation happens in one place.

## Other compat verbs

| Verb | Affected by #963? | Status |
| --- | --- | --- |
| `migrate apply --dry-run` (default + `--format`) | yes | fixed, pinned |
| `migrate down --dry-run --format` | yes — same `internal/atlasmigrate` path | fixed, pinned |
| `schema apply --dry-run` | no — `ApplyRuntimePlan.Execute` returns before the writer ever enters dry-run mode | verified clean, pinned anyway |
| `migrate down --dry-run` *without* `--format` | **no** — pre-existing | filed as #969 |

`migrate down` without `--format` forwards to the native `ptah migrations down`, which starts its own `cliobs` runtime and logs to stderr — measured at 8 lines on a successful dry run and 4 on a successful rollback, both exit `0`. `cmd/migratedown/migratedown.go` is byte-identical at `fec2303` (pre-#963) and at `master`, so this is an older divergence from the forward design. Split out as **#969** (suggested fix there: the compat wrapper passes a quiet log level into the forwarded native command). The compat docs now name the exception explicitly instead of implying a guarantee.

## Tests

- `cmd/ptah-compat/dry_run_quiet_test.go` — subprocess-level, the real contract:
  - combined stdout+stderr of `migrate apply --dry-run --format '{{ json . }}'` decodes as exactly one JSON document and then `io.EOF` (the exact probe shape);
  - same for `migrate down --dry-run --format`;
  - default format: stderr byte-equal to `""` for `migrate apply` and `schema apply`, stdout pinned;
  - regression pin naming #963: no `[DRY RUN]` and no `level=INFO` anywhere in the combined output of three dry-run invocations, each with a planned-work marker asserted so the pin cannot pass on an empty plan;
  - **log-only diagnostics survive**: a circular-FK schema still emits its `level=WARN` record on both a dry run and a real `--auto-approve` apply;
  - failures still report on stderr and exit `1` (missing directory, checksum mismatch).
- `cmd/migrateup/dry_run_log_controls_test.go` — native: report on stdout, narration on the log stream, `--log-level` table (`info`/`debug` narrate, `warn`/`error` quiet), `--log-format json` keeps stdout parseable line-by-line and stderr empty, and exactly one metadata record.
- `migration/migrator/dry_run_initialize_once_test.go` — one `Would initialize migrations metadata` record across a status read plus a `MigrateUp`; and a dry-run `Initialize` does not suppress a later real one.
- `cmd/internal/cliobs/silence_test.go` — the level split as a contract (Info dropped, Warn kept, both directions); `Start` nests on top and restores the quiet default on shutdown.

Mutation-verified in both directions:

| Mutation | Result |
| --- | --- |
| revert `QuietDefaultLogger()` from `main` | all three `PinNoWriterNarration` subtests fail (previously only `down_go_template` did) |
| revert `QuietDefaultLogger()` from `main` | `DefaultFormatKeepsStderrEmpty/migrate_apply` fails, `FormatCombinedOutputIsOneJSONDocument` fails on `Decode` |
| swap `LevelWarn` back to `slog.DiscardHandler` | both `KeepsLogOnlyDiagnostics` subtests fail; `warn`/`error` subtests of the cliobs split fail |

## Open question for the conformance repo

On a **failing** apply, `--format` writes the JSON report to stdout and the `Error: …` diagnostic to stderr, so `2>&1 | jq` breaks on the failure path. That is probably correct Atlas parity — Atlas also writes its diagnostic to stderr — but the `atlas-cli-report-format` probe only covers the success path, so the failure path is currently unpinned either way. Worth a probe on the conformance side to lock in whichever behavior is intended.

## Verification

`go vet ./...` clean · full `go test ./...` green · `go test -race` on the four touched packages green · `golangci-lint run` on the four touched packages: 0 issues · `scripts/check-test-style.sh`: OK · `scripts/check-public-api-snapshot.sh`: no drift · docs site `check-links.mjs`, `check-style.mjs`, `check-page-health.mjs`, `check-core-doc-links.mjs`: OK.

## Docs

- `docs/site/src/content/docs/versioned/apply.md` — the dry-run section showed the stdout report and the `[DRY RUN] Would execute SQL` narration as one block; they are separate streams. Now split, with the `--log-level` / `--log-format` behavior spelled out, a new logging bullet under Execution controls, and a cross-reference to the compat surface.
- `docs/site/src/content/docs/atlas/migrate-commands.md` — scoped to `ptah-compat migrate apply` and to `--format`, shows the `2>&1 | jq` idiom, states that failures *and* Warn-level diagnostics still reach stderr, and calls out the `migrate down` exception with a pointer to #969.
- `DRY_RUN.md` — new "Where Dry Run Output Goes" section covering both streams and both surfaces, including what quiet does and does not drop; the pre-existing interleaved example is now labeled as such.

### DIFF
--- a/DRY_RUN.md
+++ b/DRY_RUN.md
@@ -127,7 +127,10 @@ go run . seed --db-dsn postgres://user:pass@localhost/db --dry-run
 
 ### Dry Run Output Format
 
-When dry run mode is enabled, all operations are prefixed with `[DRY RUN]`:
+When dry run mode is enabled, all operations are prefixed with `[DRY RUN]`. The
+example below shows both output streams interleaved, as a terminal would render
+them; see [Where Dry Run Output Goes](#where-dry-run-output-goes) for which line
+lands on which stream.
 
 ```
 [DRY RUN] Would write schema from ./models to database postgres://user:***@localhost/db
@@ -152,6 +155,34 @@ Creating table 2/2...
 [DRY RUN] Would commit transaction
 ✅ [DRY RUN] Schema operations completed successfully!
 ```
+
+### Where Dry Run Output Goes
+
+Dry run produces two different kinds of output, on two different streams:
+
+1. **The command report** — the `=== DRY RUN MODE ===` banner, the summary
+   counts, and the closing `Would have applied N migrations` line — is written
+   to **stdout** by the command itself.
+2. **The per-statement narration** — the `[DRY RUN] Would begin transaction`,
+   `[DRY RUN] Would execute SQL`, and `[DRY RUN] Would commit transaction`
+   records emitted by the dialect writers — is a **log record**, not report
+   output. It goes to the command's log stream, which is stderr by default.
+
+Because the narration is a log record, the native commands' existing
+observability flags control it:
+
+- `--log-level warn` (or `error`) drops it, leaving only the stdout report.
+- `--log-format json` moves the whole run log — report lines included — onto
+  stdout as one JSON object per line, so `ptah migrations up --dry-run
+  --log-format json | jq` is parseable.
+
+The Atlas-compatible binary (`ptah-compat`) deliberately differs: it writes
+**nothing** to stderr for a successful dry run, because the Atlas CE binary it
+mirrors writes nothing either, and because `--format` consumers routinely fold
+stderr into stdout (`2>&1 | jq`), where one stray log line stops the output from
+being a single JSON document. Failures are still reported there — the command's
+`Error: …` diagnostic goes to stderr and the process exits `1`. See
+[stokaro/ptah#967](https://github.com/stokaro/ptah/issues/967).
 
 ### Safety Features
 
--- a/DRY_RUN.md
+++ b/DRY_RUN.md
@@ -176,12 +176,25 @@ observability flags control it:
   stdout as one JSON object per line, so `ptah migrations up --dry-run
   --log-format json | jq` is parseable.
 
-The Atlas-compatible binary (`ptah-compat`) deliberately differs: it writes
-**nothing** to stderr for a successful dry run, because the Atlas CE binary it
-mirrors writes nothing either, and because `--format` consumers routinely fold
+The Atlas-compatible binary (`ptah-compat`) deliberately differs: it drops the
+narration entirely rather than exposing a flag for it, because the Atlas CE
+binary it mirrors prints none, and because `--format` consumers routinely fold
 stderr into stdout (`2>&1 | jq`), where one stray log line stops the output from
-being a single JSON document. Failures are still reported there — the command's
-`Error: …` diagnostic goes to stderr and the process exits `1`. See
+being a single JSON document.
+
+Quiet is not silent. Two things still reach `ptah-compat`'s stderr, neither of
+which appears on a clean run:
+
+- **Command failures** — the `Error: …` diagnostic, with exit status `1`.
+- **Warn-level diagnostics that exist on no other channel** — a circular
+  foreign-key or function ordering, a dev or shadow database that would not
+  close, an unrecognized embedding mode. None of these are returned as errors,
+  so dropping them would let a real apply report success while quietly
+  degrading.
+
+One exception: `ptah-compat migrate down` without `--format` forwards to the
+native rollback command and inherits its run log on stderr. See
+[stokaro/ptah#969](https://github.com/stokaro/ptah/issues/969) and
 [stokaro/ptah#967](https://github.com/stokaro/ptah/issues/967).
 
 ### Safety Features
--- a/cmd/internal/cliobs/cliobs.go
+++ b/cmd/internal/cliobs/cliobs.go
@@ -48,27 +48,41 @@ type Runtime struct {
 	once     sync.Once
 }
 
-// SilenceDefaultLogger installs a discarding default slog logger.
+// QuietDefaultLogger installs a default slog logger that drops narration and
+// keeps diagnostics.
 //
-// Library code narrates through the package-level slog functions — the dialect
-// writers' dry-run statement narration, connection-close warnings, annotation
-// diagnostics — which resolve to slog.Default() and therefore reach stderr even
-// when the caller never wired observability. The Atlas-compatible binary must
-// not write those records: the pinned Atlas CE binary emits nothing on stderr
-// for the same commands, and `--format` consumers redirect stderr into stdout
-// (`2>&1 | jq`), where a single extra line breaks JSON parsing. See
-// stokaro/ptah#967.
+// Library code logs through the package-level slog functions, which resolve to
+// slog.Default() and therefore reach stderr even when the caller never wired
+// observability. Those calls split cleanly by level, and the threshold sits on
+// the seam:
 //
-// Failures are unaffected: the Atlas-compatible commands report them through
-// the command's own error stream, never through slog.
+//   - Info and below is narration — the dialect writers' "[DRY RUN] Would ..."
+//     statement records, the Postgres enum-creation notices. Atlas CE never
+//     prints any of it, and `--format` consumers redirect stderr into stdout
+//     (`2>&1 | jq`), where one extra line stops the output from being a single
+//     JSON document. This is what broke stokaro/ptah#967, and it is dropped.
+//   - Warn and above is a diagnostic that exists nowhere else: circular
+//     foreign-key and function ordering, a dev or shadow database that would
+//     not close, an unrecognized embedding mode that silently inlines. None of
+//     them are returned as errors, so dropping them would let a real apply
+//     report success while quietly degrading. These are kept.
+//
+// Atlas parity survives the split because a clean run emits nothing at either
+// level: the diagnostics only appear when something is genuinely wrong, and a
+// user who sees one is better served than by byte-parity with CE.
+//
+// Command failures are unaffected either way: the Atlas-compatible commands
+// report them through the command's own error stream, never through slog.
 //
 // Call this once, before any command runs. It deliberately does not take
 // globalStateMu: Start holds that lock for the duration of a command and
 // restores whatever default was installed beforehand, so a command that starts
-// its own observability runtime still nests correctly on top of the silenced
+// its own observability runtime still nests correctly on top of the quiet
 // default.
-func SilenceDefaultLogger() {
-	slog.SetDefault(slog.New(slog.DiscardHandler))
+func QuietDefaultLogger() {
+	slog.SetDefault(slog.New(slog.NewTextHandler(os.Stderr, &slog.HandlerOptions{
+		Level: slog.LevelWarn,
+	})))
 }
 
 // Start initializes command logging, tracing, and metrics.
--- a/cmd/internal/cliobs/cliobs.go
+++ b/cmd/internal/cliobs/cliobs.go
@@ -48,6 +48,29 @@ type Runtime struct {
 	once     sync.Once
 }
 
+// SilenceDefaultLogger installs a discarding default slog logger.
+//
+// Library code narrates through the package-level slog functions — the dialect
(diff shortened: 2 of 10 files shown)


denisvmedia:
All three MAJORs and both MINORs addressed in 975fd7a. Thanks — MAJOR 1 in particular is a real correctness catch, not a preference.

## MAJOR 1 — level split, not full discard

Accepted as argued, no tradeoff to write down. I re-derived the level census independently before changing anything: every `#967` narration site is `Info` (all 22 `[DRY RUN] Would ...` sites plus the two Postgres enum notices), and every log-only diagnostic is `Warn`/`Error` with no other channel — `core/goschema/utils.go:78`, `utils.go:962`, `dbschema/connection.go:394`, `convert/fromschema/fromschema.go:2225`. So `LevelWarn` fixes the bug identically and stops the fix from making a real apply lie about success.

`SilenceDefaultLogger` → `QuietDefaultLogger`, installing `slog.NewTextHandler(os.Stderr, &slog.HandlerOptions{Level: slog.LevelWarn})`. The doc comment now carries the seam argument explicitly.

Reproduced your measurement on my build, both modes:

```console
$ ptah-compat schema apply --url sqlite://circ.db --to file://circular.sql \
    --dev-url sqlite://circdev.db --dry-run 2>&1 >/dev/null
time=… level=WARN msg="Circular dependency detected in foreign key relationships; emit foreign keys after table creation to keep DDL executable."

$ ptah-compat schema apply ... --auto-approve 2>&1 >/dev/null
time=… level=WARN msg="Circular dependency detected in foreign key relationships; …"

$ ptah-compat migrate apply ... --dry-run --format '{{ json . }}' 2>&1 | jq -c '{Target, Pending: (.Pending|length)}'
{"Target":"1","Pending":1}
```

Default-format dry run still leaves stderr byte-empty. Added `TestCompatBinaryKeepsLogOnlyDiagnostics` so the diagnostic half is pinned in both modes — it was previously unpinned, which is exactly why the original threshold slipped through.

## MAJOR 2 — vacuous subtests

Confirmed and fixed. `applyForReal` on the shared database left both apply subtests with nothing pending, so the writers were never entered. The apply cases now use their own `applyDB` with pending work; the down case keeps its applied `downDB`. Each case also asserts a planned-work marker (`Would have applied 1 migrations.`, `"Target":"1"`, `"Planned":[{"Name":"1_init.sql"`) so the pin cannot silently go vacuous again.

Mutation re-verified — all three now red, where before only `down_go_template` was:

```
--- FAIL: TestCompatBinaryDryRunPinNoWriterNarration/apply_default_format
--- FAIL: TestCompatBinaryDryRunPinNoWriterNarration/apply_go_template
--- FAIL: TestCompatBinaryDryRunPinNoWriterNarration/down_go_template
```

And the reverse mutation (`LevelWarn` → `DiscardHandler`) turns both `KeepsLogOnlyDiagnostics` subtests and the `warn`/`error` halves of the cliobs split red.

## MAJOR 3 — the doc overpromised

Right, and this one was my error rather than the code's: I scoped it honestly in the PR body and then wrote the doc as a blanket guarantee. Measured your numbers before rewriting — 8 stderr lines on a successful `migrate down --dry-run`, 4 on a successful rollback, exit 0 both times.

The paragraph is now scoped to `migrate apply` and to `--format`, names the `migrate down` exception in a caution block, notes there is no `--log-level` on the Atlas surface to quiet it, and points at **#969**, which I filed with the repro, the scope analysis (only `migrate down` is affected; `migrateup`/`migratestatus`/`migratedown` are the only `cliobs` starters and only `migrations down` is reached by a forward), and your suggested fix — the compat wrapper passing a quiet log level into the forwarded native command, plus the open design question of per-verb `prefixArgs` vs. a tree-wide default. `versioned/apply.md` left alone as you said. `DRY_RUN.md` got the same correction since it made the same claim.

## MINORs

- `silence_test.go` narrowed from mechanism to contract: a table of Info-dropped / Warn-kept in both directions, rather than "every level disabled".
- `initializedDryRun` now cleared in `WithMigrationsTable` and `WithRevisionTableFormat` alongside the fields beside it.

## Also noted in the PR body

Your failing-apply observation is now written up as an open question for the conformance repo: on failure, `--format` puts the report on stdout and the diagnostic on stderr, so `2>&1 | jq` breaks — probably correct Atlas parity, but the `atlas-cli-report-format` probe only covers the success path, so it is unpinned either way.

## Verification

`go vet ./...` clean · full `go test ./...` green · `-race` on all four touched packages green · `golangci-lint` on the four: 0 issues · `check-test-style.sh` OK · `check-public-api-snapshot.sh` no drift · docs site `check-links` / `check-style` / `check-page-health` / `check-core-doc-links` all OK.

Issue #955, opened 2026-08-01
migrate apply does not verify atlas.sum before applying (validate does)


denisvmedia:
Measured 2026-08-01 (fixtures: `the observation study fixtures, checkpoint-dir/`, `m_tampered`).

A tampered Atlas-format directory (file edited after hashing, `atlas.sum` stale):

- Official Atlas `migrate apply`: refuses **before applying anything** — `L4: ...checkpoint.sql was edited ... Error: checksum mismatch`, exit 1.
- `ptah-compat migrate apply`: no checksum verification — applies 001+002 from the tampered dir, fails only when the tampered statement itself errors.
- Native `ptah migrations up`: same, applies without checksum complaint.
- `ptah-compat migrate validate` on the same dir: **byte-identical output to Atlas** — the verification code exists, apply just never calls it.

This is the integrity gate the whole `atlas.sum`/`ptah.sum` convention exists for; apply-time enforcement is what makes it meaningful in CI. Related: #944 tracks the OCI `--verify-sum` self-referential variant of the same theme.

Fix: run directory-integrity verification at the start of `migrate apply` (compat) and `ptah migrations up` for hashed directories, with the Atlas-shaped error text on mismatch.

Refs #951.

PR #972, opened 2026-08-01
Refuse an unhashed migration directory on the Atlas-compatible apply path

### LINKED ISSUE #970: migrate apply accepts an unhashed migration directory; Atlas refuses with "checksum file not found"
Found by an adversarial review of the conformance fixtures for #955. Measured against the pinned Atlas CE v1.2.0 oracle and `ptah-compat` built at `01a39d5`, same directory with **no `atlas.sum`**:

```
Atlas CE v1.2.0:   exit 1, database never created
  You have a checksum error in your migration directory.
  Please check your migration files and run 'atlas migrate hash' to re-hash the contents
  Error: checksum file not found

ptah-compat:       exit 0, database created, migrations applied
  Migrating to version 20260801100335 from 1 pending migrations.
  Migration complete. Current version: 20260801100335
```

The asymmetry #955 was filed about is still present, just moved to the other branch — on the **same** unhashed directory:

```
ptah-compat migrate validate  -> exit 1, "Error: checksum file not found"   # byte-identical to Atlas
ptah-compat migrate apply     -> exit 0, applies                            # Atlas refuses
```

#955 scoped its fix to "hashed directories", so #962 implemented mismatch detection only. The missing-sum-file branch was never gated. Nothing in the conformance repo measures it either: every `checksum file not found` occurrence there is in the simulated Atlas (`txtar_script.go`), never against the real binaries.

Fix: `migrate apply` on the Atlas-compatible surface must refuse a directory with no integrity file, with the Atlas error shape, before touching the target. Decide deliberately what native `ptah migrations up` should do — `--verify-sum` already means "a missing sum file is an error" there, so the native default may reasonably stay permissive; if so, document the split the way the failed-down surface split is documented.

A conformance fixture for this case is being added alongside; it will land red until this is fixed.

Refs #955, #962, #951.

denisvmedia:
Closes #970. Refs #955, #962, #951. Spawned #973, #974 and #976.

> **Corrections.** 2nd commit: the 1st claimed "Atlas does not gate converted formats either" — **measured false**, see [Converted formats](#converted-formats-known-divergence-973). 3rd commit: the 2nd's no-SQL exemption was top-level only and let an unhashed **subdirectory** migration execute — see [The refusal predicate](#the-refusal-predicate-a-sql-file-anywhere-in-the-tree).

## The gap

On a migration directory with **no integrity file**, measured against the pinned oracle:

```
Atlas CE v1.2.0:   exit 1, database never created
ptah-compat:       exit 0, database created, migrations applied
```

#955 scoped its fix to "hashed directories", so #962 gated only the checksum-**mismatch** branch. The missing-file branch was never gated, which left the compat surface asymmetric with *itself* on the same directory: `ptah-compat migrate validate` already refused with the byte-identical Atlas message while `ptah-compat migrate apply` accepted.

### Before / after, same unhashed directory

`ptah-compat migrate apply --url sqlite://out.db --dir file://unhashed`

| | before | after |
| --- | --- | --- |
| exit | 0 | 1 |
| stdout | `Migrating to version 20260801100335 from 1 pending migrations.` / `Migration complete. Current version: 20260801100335` | `You have a checksum error in your migration directory.` / `Please check your migration files and run 'atlas migrate hash' to re-hash the contents` |
| stderr | — | `Error: checksum file not found` |
| target DB file | created | **never created** |

`ptah migrations up --migrations-dir unhashed --dir-format atlas` is unchanged before and after: it applies. With `--verify-sum`, before and after: `error: migration sum verification failed: atlas.sum not found; run 'ptah migrations hash --dir-format atlas' to create it`, exit 2, DB never created.

## The fix

- `cmd/migratevalidate` gains `FailAtlasChecksumFileNotFound`, the sibling of `FailAtlasChecksumMismatch`. `runAtlasValidate` now routes **both** of its checksum refusals through the exported helpers, so validate and the apply gate render each refusal from one place and cannot drift — the property the #962 review asked for, extended to the not-found case.
- `verifyAtlasApplyChecksum` refuses `!hashed` instead of returning nil, before the database connection is opened, so nothing executes and the target file is never created — in a dry run too.
- The directory format is resolved **once** and fed to both the loader and the gate: `ResolveApplyDirFormat` is exported and `ResolveApplySourceForFormat` takes the resolved value. The old `ResolveApplySource` wrapper is deleted — it had no production callers left, and Go's unused check does not flag exported functions even in `internal/`. Tests now make the same two calls the command makes.

### The refusal predicate: a `.sql` file anywhere in the tree

Measured on CE, including two regressions this branch introduced and fixed:

| directory | CE v1.2.0 | master | commit 1 | commit 2 | **head** |
| empty | exit 0, `No migration files to execute` | exit 0 | ❌ exit 1 | ✅ exit 0 | ✅ exit 0 |
| only `.gitkeep` + `README.md` | exit 0 | exit 0 | ❌ exit 1 | ✅ exit 0 | ✅ exit 0 |
| unhashed, one `foo.sql` | exit 1, gated | exit 0 | ✅ exit 1 | ✅ exit 1 | ✅ exit 1 |
| unhashed, only `sub/…_init.sql` | exit 0, ignores subdirs | exit 0, applies | exit 1 | ❌ **exit 0, applies** | ✅ exit 1 |

Two things drive this:

1. CE gates on any `*.sql` file, *not* on parseable versioned migrations (`foo.sql` is refused). Without the exemption, `mkdir -p migrations && ptah-compat migrate apply` in a CI bootstrap breaks.
2. **The scan is recursive, and deliberately unlike CE.** Atlas reads only the top level; Ptah's registrar recurses. A top-level-only scan therefore exempted exactly the unhashed migrations the gate exists to stop — commit 2 executed `sub/…_init.sql` unverified. **This trades a silent execution for an exit 1 where CE exits 0**, which is the safe side of a pre-existing divergence in what the two tools consider a migration: CE ignores subdirectory files entirely, Ptah executes them. That divergence — including that `ptah migrations hash` does not cover them either, so `atlas.sum` gives them no protection at all — is #976, filed against `master`.

Covered by subtests for every row above; making the scan shallow again reddens the nested ones (nothing pinned that before).

## Converted formats: known divergence (#973)

Directories read through `?format=` stay ungated, and **this is a divergence from Atlas, not parity**. Measured on the pinned CE binary:

```
$ atlas migrate apply --dir 'file://goose?format=goose'      # unhashed
Error: checksum file not found                                exit=1

$ atlas migrate hash --dir 'file://goose?format=goose'        exit=0, writes atlas.sum
$ atlas migrate apply --dir 'file://goose?format=goose'       exit=0, applies

# tampered after hashing:
$ atlas migrate apply  --dir 'file://tamper?format=goose'
	L2: 1_init.sql was edited
Error: checksum mismatch                                      exit=1
$ ptah-compat migrate apply --dir 'file://tamper?format=goose'
Migration complete. Current version: 1                        exit=0   ← edited migration executed
```

I attempted the full fix and stopped on two measured blockers, both in #973:

1. `ptah-compat migrate hash`/`validate` reject `?format=` (`URL query parameters are not supported for this command`) while CE supports it — so users cannot produce the `atlas.sum` a gate would demand.
2. Ptah's sum is byte-identical to CE's for goose, dbmate, liquibase and flyway, but **not** golang-migrate: CE hashes only `1_init.up.sql`; Ptah's atlas-format hasher also includes the down file and hashes the up file differently, because in an Atlas directory it treats `X.up.sql`/`X.down.sql` as one migration. Gating on top of that would **refuse a directory CE applies** — a worse divergence than the current one.

So the exemption stays for this PR, but nothing claims Atlas agrees: the test is `TestCompatMigrateApply_ConvertedDirStaysUngated_KnownDivergence`, the docs carry a caution box, and #973 has the transcripts and a suggested shape.

## `migrate status` — follow-up (#974)

Asked in review: `ptah-compat migrate status` on an unhashed directory exits 0 and prints the report while CE exits 1 with the same checksum refusal. Same bug class, one verb over. **Follow-up, not this PR** — it is read-only (reporting fidelity, not execution safety), #970's conformance fixture is waiting on the apply path, and the shared helper this PR adds makes the later fix small. `migrate lint` is correctly ungated on both, which independently justifies the one deliberately unhashed lint fixture here.

## Tests

| Coverage | Test |
| --- | --- |
| unhashed refused, exit 1, Atlas message, DB never created | `..._UnhashedDirRefusesBeforeExecution` + a binary-level subtest asserting the real process exit code |
| dry run refused the same way | `..._UnhashedDirRefusesDryRun` |
| apply/validate output byte-equal | `..._UnhashedDirMatchesValidateOutput` |
| empty and `.gitkeep`-only dirs apply | `..._DirWithoutSQLFilesApplies` (2 subtests) |
| unhashed non-versioned `foo.sql` refused | `..._UnhashedDirWithNonVersionedSQLRefuses` |
| unhashed subdirectory migration refused | `..._UnhashedDirWithNestedSQLRefuses` (2 subtests) |
| converted formats still apply (divergence pinned) | `..._ConvertedDirStaysUngated_KnownDivergence` (5 formats) |
| hashed-valid applies / hashed-tampered refused | existing tests, unchanged |
| native behavior pinned both ways | `TestMigrateUp_UnhashedAtlasDirStaysUngatedUnlessVerifySum` |
| format resolution + per-format read | `TestResolveApplyDirFormat`, `TestResolveApplySourceForFormatReadsEachFormat` |

Mutation-verified per subtest: reverting the refusal reddens all four refusal tests; forcing the gate onto converted directories reddens all five format subtests individually; dropping the no-SQL exemption reddens both empty-dir subtests; making the scan top-level-only reddens both nested subtests. `cmd/ptah-compat` builds the binary at run time, so its cache key excludes `cmd/atlas` — the test file says to use `-count=1` there, and every run above was `-count=1`.

Note on the honest limit of the seam test: a duplicated-but-agreeing format resolution is invisible to any test (the resolver is pure and both calls would take the same inputs). What actually protects the gate is that `verifyAtlasApplyChecksum` receives an already-resolved format rather than the raw `--dir` string and query — so it cannot re-resolve without a visible signature change — plus the converted-format test, which goes red if the gate ever stops seeing `?format=`. The test comment now says exactly that instead of overstating.

## Docs

`versioned/integrity-and-safety.md` (new "A directory that was never hashed" section, the recursive-scan divergence, and a caution box for #973), `atlas/migrate-commands.md`, `reference/atlas-commands.md`, `versioned/apply.md`, `reference/exit-codes.md`, `docs/exit_codes.md`, and the feature-matrix row (rebuilt; evidence records the correct `atlas.sum` `--verify-sum` message and all three open divergences).

## Verification

`go vet ./...`, full `go test ./... -count=1`, `go tool qtlint ./...`, `golangci-lint run ./cmd/... ./internal/atlasmigrate/...` (0 issues), `scripts/check-test-style.sh`, `scripts/check-public-api*.sh`, and every docs checker (`build-feature-matrix.mjs --check`, `check-style.mjs` + `--selftest`, `check-links.mjs`, `check-exit-codes.mjs`, `check-core-doc-links.mjs`, `check-redirects.mjs`). The testify guard inside `check-test-style.sh` silently no-ops on this machine (#975), so it was also run by hand over the touched packages — quicktest only, no testify.


denisvmedia:
Review addressed in 358b6dd. I re-measured everything against the pinned CE v1.2.0 binary (`ptah-atlas-conformance/bin/atlas`) rather than taking the findings on trust — all three reproduced exactly.

**BLOCKER — the converted-format claim.** You are right and I was wrong. CE gates converted directories: unhashed goose exits 1 with `checksum file not found`, `atlas migrate hash --dir 'file://mig?format=goose'` writes an `atlas.sum`, and a tampered file then exits 1 with `L2: 1_init.sql was edited` where ptah-compat applies it. I took the minimum resolution rather than the preferred one, after attempting the preferred one and hitting two measured blockers:

1. our own `migrate hash`/`validate` reject `?format=`, so nothing in Ptah can produce the sum a gate would demand;
2. sum computation is not format-aware, and diverges on golang-migrate — CE hashes only `1_init.up.sql`; ours adds the down file and hashes the up file differently (Atlas dirs treat `X.up.sql`/`X.down.sql` as one migration). Gating on top of that would refuse a directory CE applies, trading the current divergence for a false refusal.

Both are in #973 with the transcripts and the byte-level table (goose/dbmate/liquibase/flyway identical, golang-migrate different). Every "Atlas does not gate them either" and "would make them unappliable" sentence is gone from the commit message, PR body, test comment, `integrity-and-safety.md`, `migrate-commands.md`, `reference/atlas-commands.md` and the matrix evidence; the test is now `..._ConvertedDirStaysUngated_KnownDivergence` and says it pins behavior, not parity. The second commit message opens by correcting the first, so the false claim is not left standing anywhere in history.

**MAJOR — empty / SQL-less directories.** Real regression, now fixed with your predicate and re-measured against CE: empty → exit 0 `No migration files to execute`, `.gitkeep`+`README.md` → exit 0, unhashed `foo.sql` → exit 1. Comment records that the trigger is the file extension, not the planner's view. Two subtests plus the `foo.sql` counterpart; dropping the exemption reddens both.

**MAJOR 2 — duplicated resolution.** Now single by construction: `ResolveApplyDirFormat` exported, `ResolveApplySourceForFormat` takes the resolved value, `ResolveApplySource` is the wrapper, the command resolves once and feeds both, and `verifyAtlasApplyChecksum` takes the format (dead error branch gone). `TestResolveApplySourceMatchesResolvedFormat` pins the seam per format.

**MINORs.** Matrix evidence now quotes the measured `atlas.sum not found; run 'ptah migrations hash --dir-format atlas' to create it`, and its closing sentence is rewritten. `cmd/ptah-compat/main_test.go` now carries a note that it builds the binary at run time, so the cache key excludes `cmd/atlas` and mutation checks there need `-count=1` — thank you for flagging that, all my re-runs used it.

**`migrate status`.** Reproduced (CE exit 1, ptah-compat exit 0 with the full report). Calling it a follow-up, filed as #974: it is read-only, so it is reporting fidelity rather than execution safety, #970's fixture is waiting on the apply path, and the shared refusal helper this PR adds makes the later fix small. `migrate lint` is correctly ungated on both — which is also why the one unhashed lint fixture here stays deliberate.

Re-verified after the changes: `go vet ./...`, full `go test ./... -count=1`, `go tool qtlint ./...`, `golangci-lint` on the touched packages, teststyle, public-API gates, and all docs builders. Five-format mutation re-run against the new construction: all five subtests fail individually when the gate is forced onto converted directories.

denisvmedia:
Fixed in 674cb38.

**MAJOR — the exemption was top-level only.** Reproduced before touching it: on `migrations/` holding only `sub/20260801100335_init.sql` and no `atlas.sum`, 358b6dd printed `Migrating to version 20260801100335 … Migration complete` and created the database, while CE printed `No migration files to execute`. My exemption's premise — "no SQL here means nothing to execute" — was false for the one layout where Ptah and Atlas disagree about what a migration is, and it re-opened the hole the PR exists to close.

`fs.Glob` is now an `fs.WalkDir` that stops at the first non-dir `.sql` path (`fs.SkipAll`). Re-measured with the rebuilt binary: subdirectory case → exit 1, no database; empty and `.gitkeep`-only → exit 0, `No migration files to execute`; `foo.sql` → exit 1. Mutation (e) is now covered — making the scan shallow again reddens both new subtests (`migration in a subdirectory`, `subdirectory migration beside a top-level non-SQL file`).

The PR body now states the trade explicitly, with a per-commit table: this is **exit 1 where CE exits 0** for that layout, deliberately, because Ptah executes those files and CE ignores them. The gate comment, `integrity-and-safety.md`, `reference/atlas-commands.md`, `atlas/migrate-commands.md`, both exit-code pages and the matrix evidence all say "anywhere in the tree" now and point at #976 — thank you for filing it; it is a sharper problem than my fix implies, since `ptah migrations hash` does not cover those files either, so a *hashed* directory gives them no protection at all.

**MINOR 1.** Deleted `ResolveApplySource` rather than documenting it — it was a four-line wrapper with no production callers, and keeping dead exported API for a hypothetical #973 use is how it rots. The nine test call sites now go through a package-local helper that performs the same two calls the command performs, so the tests exercise production wiring instead of a convenience path production does not use.

**MINOR 2.** You are right that the test proved less than its comment claimed, and with the wrapper gone it would have been a tautology. It is now `TestResolveApplySourceForFormatReadsEachFormat` (what each resolved format makes the migrator read, including that a golang-migrate down file never survives into what executes), and its comment states plainly that it does **not** prove the single-resolution wiring — that a duplicated-but-agreeing computation is invisible to any test, and that what actually carries the property is `verifyAtlasApplyChecksum` taking an already-resolved format (it cannot re-resolve without a visible signature change) plus the converted-format test going red if the gate ever stops seeing `?format=`.

**NIT.** Left as is — you are right that it is cosmetic, and moving the resolution back below `openLocal` would trade a CE-ward improvement on one input pair for a CE-ward improvement on another while the underlying `?bogus=1` rejection diverges anyway.

Verification after the change: `go vet ./...`, full `go test ./... -count=1`, `go tool qtlint ./...`, `golangci-lint` on the touched packages (0 issues), teststyle, public-API gates, all docs builders. One note: `scripts/check-test-style.sh` silently skips its testify grep on this machine (#975 — I saw `rg: command not found` on every run and should have flagged it earlier), so I ran that grep by hand over the touched packages: quicktest only.

denisvmedia:
One more push, 8155acf — and a correction to my own verification claim.

`ptah-go-lint` went red on 674cb38: the teststyle baseline rejected a conditional I added inside a table case (`if tt.format == "golang-migrate"` around the down-file assertion). My previous comment listed `scripts/check-test-style.sh` among the checks I ran for that round; I had run it in the round before and did not re-run it after editing the test, so CI caught this rather than me. The claim was wrong, and the lesson is the same one that produced the earlier corrections — a check listed as run must have been run against the exact tree being pushed.

The fix drops the branch: the assertion is now unconditional, which is legitimate because no fixture other than golang-migrate's writes `1_init.down.sql`. It stays load-bearing where it matters — mutating `ResolveApplySourceForFormat` to skip conversion for golang-migrate fails that subtest and only that one. Re-ran the whole set against the pushed tree this time: `go vet ./...`, `go test ./... -count=1`, `go tool qtlint ./...`, `check-test-style.sh` (now OK), and `golangci-lint` on the touched packages.

Side note from that job log: `rg: not found` in CI too, so the testify guard in `check-test-style.sh` is silently skipped on the runners as well, not just locally — #975 is a real gap in the enforced set rather than a local-environment quirk.

Issue #987, opened 2026-08-01
ptah-compat migrate hash prints on success where Atlas CE is silent


denisvmedia:
Measured against the pinned Atlas CE v1.2.0 binary while enumerating branches for #973.

On success, `atlas migrate hash` writes **zero bytes to stdout and stderr** and exits 0. `ptah-compat migrate hash` prints:

```
Wrote /path/to/migrations/atlas.sum
1 migration file(s) hashed
```

## Why this is a divergence rather than a preference

`ptah-compat` exists to be a drop-in for Atlas. A wrapper script that treats any output from `migrate hash` as noteworthy — or that diffs command output in CI — behaves differently against the two binaries, on the happy path, for every directory.

It has not been an explicit decision so far: 14 assertions across 6 files (`cmd/atlas/atlas_test.go`, `project_config_snapshot_test.go`, `migrate_maint_forward_test.go`, and others) pin the current strings, so the behavior is well covered but was never measured against the oracle.

## Resolution: split by surface

Per the standing rule that `ptah-compat` reproduces Atlas exactly while native `ptah` keeps a better behavior as a documented advantage:

- **`ptah-compat migrate hash`** — silent on success, exit 0. Matching Atlas is the whole point of the surface.
- **`ptah migrations hash`** — keeps `Wrote …` and the file count. Knowing what was written is genuinely more useful, and the native surface has no obligation to reproduce Atlas. Document it in the comparison page as a Ptah-side plus rather than leaving it implicit.

## Scope note

Deliberately **not** folded into the #973 PR chain. It is orthogonal to the directory-format axis, and it touches a verb that chain otherwise only extends. Doing it separately keeps both changes reviewable.

## Definition of done

- Compat success path emits nothing on either stream; the 14 assertions move to the native verb or are rewritten to assert silence, whichever matches what each test is really about.
- Byte-comparison against the pinned binary (`ptah-atlas-conformance/bin/atlas`, community v1.2.0, invoked by absolute path — `/usr/local/bin/atlas` is a different build) on: a clean directory, an already-hashed directory, an empty directory, and each `--dir-format`/`?format=` spelling.
- **Failure paths measured separately and not silenced with the success path** — they are a different contract and CE is not silent there.
- Feature-matrix row and note updated via `docs/site/scripts/data/feature-matrix-rows.json`, and the native advantage stated in the comparison docs.
[...]

Issue #991, opened 2026-08-01
A directory named *.sql: migrate hash writes a sum Atlas CE then refuses to read, and Ptah's two paths disagree with each other


denisvmedia:
Noted as a follow-up in #984, measured in full while implementing #983
(#973 PR 2). Measured against the pinned Atlas CE v1.2.0 oracle
(`ptah-atlas-conformance/bin/atlas`, absolute path), 2026-08-02.

## The layout

A migration directory holding a real migration and a **directory** whose name
ends in `.sql`:

```
d/
  1_init.sql          -- +goose Up / CREATE TABLE w (id int);
  weird.sql/          -- a directory
```

## Measured, three ways

Atlas CE's own glob matches the directory, and then fails to read it, so it
refuses the whole hash and writes nothing:

```
$ atlas migrate hash --dir 'file://d?format=goose'
Error: sql/migrate: read file "weird.sql": read d/weird.sql: is a directory
exit=1                          # no atlas.sum written
```

`ptah-compat`'s converted path skips directories — `topLevelNames` in
`internal/atlasmigrateimport/sumfiles.go` filters `entry.IsDir()` — so it
succeeds and writes a sum:

```
$ ptah-compat migrate hash --dir 'file://d?format=goose'
Wrote d/atlas.sum
1 migration file(s) hashed
exit=0
```

`ptah-compat`'s native atlas path globs like CE and fails, differently worded:

```
$ ptah-compat migrate hash --dir file://d
error: failed to read weird.sql: read weird.sql: is a directory
exit=1                          # no atlas.sum written
```

## Why it matters

Only one direction is reachable in practice, and it is this one:

```
$ atlas migrate validate --dir 'file://d?format=goose'      # sum written by Ptah
You have a checksum error in your migration directory.
Please check your migration files and run 'atlas migrate hash' to re-hash the contents

Error: sql/migrate: read file "weird.sql": read d/weird.sql: is a directory
exit=1
```

So `ptah-compat migrate hash` can produce an `atlas.sum` that Atlas CE refuses
to read. The converse cannot happen: CE never writes such a sum, because it
fails before writing one, so there is no CE-authored sum for Ptah to
misinterpret. This is a usability trap, not a wrong-sum risk.

The sharper half is internal: **Ptah's two paths disagree with each other on the
same directory.** `--dir file://d` refuses it and `--dir 'file://d?format=atlas'`
also refuses it (both go through `computeAtlas`'s glob), while
[...]

Issue #990, opened 2026-08-01
migrate hash/validate refuse three directory-format inputs Atlas CE exits 0 on


denisvmedia:
Found while implementing #983 (#973 PR 2). Measured against the pinned Atlas CE
v1.2.0 oracle (`ptah-atlas-conformance/bin/atlas`, absolute path; the
system-wide `/usr/local/bin/atlas` is a different build and was not used), 2026-08-02.

Three inputs make `ptah-compat migrate hash` and `migrate validate` exit 1 where
CE exits 0. All three are loud refusals, none can produce a wrong sum, and all
three predate the format work — but a script copied from a working CE invocation
still breaks on them.

## 1. An empty `--dir-format` value

```
$ atlas migrate hash --dir file://d --dir-format ""
exit=0                                    # treated as the atlas layout

$ ptah-compat migrate hash --dir file://d --dir-format ""
error: atlas migrate hash --dir-format: migration directory format must not be empty
exit=1
```

The rejection comes from `atlasMigrateDirFormatValue` in `cmd/atlas/atlas.go`,
shared by nine verbs (`hash`, `validate`, `lint`, `new`, `set`, `status`,
`edit`, `rebase`, `rm`, plus a checkpoint variant). It was left in place
deliberately in #973 PR 2: fixing it on two of the nine would leave the compat
surface less consistent than it is now. Fix it for all of them or none.

It is reachable from an ordinary shell mistake — `--dir-format "$FORMAT"` with
`FORMAT` unset — which is why it is worth closing rather than waiving.

## 2. A query parameter other than `format`

```
$ atlas migrate hash --dir 'file://d?format=goose&other=1'    exit=0, hashes as goose
$ atlas migrate hash --dir 'file://d?other=1'                 exit=0, hashes as atlas

$ ptah-compat migrate hash --dir 'file://d?format=goose&other=1'
error: atlas migrate hash --dir: unsupported migration directory URL query parameter "other"
exit=1
```

CE ignores query keys it does not recognize.

## 3. A repeated `format` parameter

```
$ atlas migrate hash --dir 'file://d?format=goose&format=flyway'   exit=0
$ atlas migrate hash --dir 'file://d?format=flyway&format=goose'   exit=0
```

Measured on a layout the two formats disagree about (`1_init.sql`,
`V1__x.sql`, `U1__undo.sql`): the **first** value wins in both orders — the
`goose&flyway` order covered all three files, the `flyway&goose` order covered
`V1__x.sql` alone.

[...]

Issue #989, opened 2026-08-01
CI does not fetch the pinned Atlas oracle, so oracle-only coverage is a property re-established by hand each round


denisvmedia:
Split out of #973 PR 2. Not a behavior gap — a coverage-durability one.

## The situation

Two test surfaces compare Ptah against the pinned Atlas CE v1.2.0 binary:

- `TestSumFileNamesDifferentialFuzz` and its two siblings in
  `internal/atlasmigrateimport/sumfiles_fuzz_test.go`, which skip unless
  `PTAH_ATLAS_ORACLE` points at a binary reporting
  `atlas community version v1.2.0`.
- `internal/atlasmigrateimport/testdata/ce-sums/regenerate.sh`, which rewrites
  the 88-shape corpus from the same binary and refuses any other build.

Neither runs in CI. `PTAH_ATLAS_ORACLE` is unset there, so the fuzz skips (loudly,
by design) and nothing re-derives the corpus.

That is survivable only because every rule the fuzz has ever uniquely caught was
afterwards given a corpus shape that pins it without the oracle. #984 verified
that by hand for five rules, with `PTAH_ATLAS_ORACLE` unset, listing which
oracle-less test reddens for each. #973 PR 2 did the same for its own rules
through `scripts/mutation-check.py`.

**That has now been re-established by hand three rounds running.** It is a
maintained property, not a guaranteed one. The first round where somebody adds a
fuzz-only rule and does not think to add the corresponding corpus shape, CI goes
on passing and the rule quietly stops being held.

## What would fix it

A CI job that fetches the pinned oracle, exports `PTAH_ATLAS_ORACLE`, and runs:

- the differential fuzz over a modest iteration count (`PTAH_ATLAS_FUZZ_N`),
- `regenerate.sh` followed by a `git diff --exit-code` on the corpus, which
  catches both a drifted rule and a drifted generator.

## Why it was not done in that PR

Two things a maintainer should decide rather than an implementer:

1. **Pinning.** A fetch worth having is pinned by checksum, and the checksum has
   to be the one of the *linux/amd64* release artifact. The oracle available
   locally is the darwin build, so its digest is the wrong one to commit; an
   unpinned `curl | install` in CI is a supply-chain regression traded for a
   coverage gain.
2. **Posture.** The project's position on the Atlas artifact is deliberately
   black-box: measure the binary, do not read the source (`ariga.io/atlas`'s
   licensing posture is unresolved). Pulling the binary into every CI run on
[...]

PR #992, opened 2026-08-01
Read foreign migration layouts on compat migrate hash and validate

### LINKED ISSUE #983: migrate hash/validate --dir-format goose is refused with a waiver; Atlas CE accepts it and it is the same code path as ?format=
Found while measuring #973. Pre-existing on `master` @ `4680f02`.

Measured against the pinned Atlas CE v1.2.0 oracle (`ptah-atlas-conformance/bin/atlas`, absolute path; the system-wide `/usr/local/bin/atlas` is a different build and was not used), 2026-08-01.

## The divergence

CE accepts **two** spellings for the source format on `migrate hash` and `migrate validate`, and they produce the identical `atlas.sum`:

```
$ atlas migrate hash --dir 'file://d' --dir-format goose
exit=0
h1:GrzxJ07R56OC5oottJ7wgbW7r8VOOeHaCJJdeEJA+e4=
1_init.sql h1:+rmUcws+9hthURrF7o/ArM9yUHUDSMVvMFGrA04Xyic=

$ atlas migrate hash --dir 'file://d?format=goose'
exit=0                                    # byte-identical sum

$ atlas migrate validate --dir 'file://d' --dir-format goose
Error: checksum file not found            exit=1
```

Both are documented in CE's own help text: `--dir-format string   select migration file format (default "atlas")` appears on `migrate hash --help` and `migrate validate --help`.

Ptah refuses both, with two different messages:

```
$ ptah-compat migrate hash --dir 'file://d?format=goose'
error: atlas migrate hash --dir: migration directory URL query parameters are not supported for this command

$ ptah-compat migrate hash --dir 'file://d' --dir-format goose
error: Atlas accepts --dir-format=goose, but Ptah does not implement that directory format yet
```

The second comes from `unsupportedAtlasDirFormats` in `cmd/atlas/atlas.go`, routed through `atlasMigrateDirFormatValue`.

`migrate apply` correctly registers no `--dir-format` at all — CE's `apply` answers `Error: unknown flag: --dir-format` — so this is specific to `hash` and `validate`, where the flag is genuine Atlas surface.

## Why it is one issue and not two

Both spellings resolve to the same thing: which foreign tool's convention the directory follows. Implementing `?format=` while leaving `--dir-format` on its waiver would leave a user who copied a working CE invocation still broken, so the two land together or not at all.

## Relationship to #973

#973's second PR — dedicated `cmd/atlas/migrate_hash.go` and `migrate_validate.go` — implements exactly this and is expected to close it, asserting directly that the two spellings produce identical sums. Filed separately because it is an independently reproducible parity gap that predates #973 and is worth tracking even if that plan changes.

Refs #973, #951.

denisvmedia:
PR 2 of 3 for #973, on top of #984. **No apply-path change**: `verifyAtlasApplyChecksum` is untouched, `migrate apply` still registers no `--dir-format`, and a converted directory is still ungated. That is PR 3.

Closes #983.

## The gap

Atlas CE accepts a foreign tool's directory layout on `migrate hash` and `migrate validate` under two spellings that mean the same thing, and writes a byte-identical `atlas.sum` from either. Ptah refused both, with two different messages:

```
$ ptah-compat migrate hash --dir 'file://d?format=goose'
error: atlas migrate hash --dir: migration directory URL query parameters are not supported for this command

$ ptah-compat migrate hash --dir file://d --dir-format goose
error: Atlas accepts --dir-format=goose, but Ptah does not implement that directory format yet
```

So a user could not produce, through Ptah's own tooling, the `atlas.sum` that PR 3's gate will demand. Landing one spelling and not the other would leave anyone who copied a working CE invocation still broken, which is why #983 is one issue.

## What lands

`cmd/atlas/migrate_hash.go` and `cmd/atlas/migrate_validate.go` wrap their table-driven adapter commands with a converted-source path, following the `newAtlasMigrateDownCommand` precedent. Both spellings resolve through `atlasmigrate.ResolveApplyDirFormat` — the resolver `migrate apply` already uses — so hash, validate and apply cannot drift on which spelling wins. The shared resolution is `resolveAtlasVerbProject`, lifted out of `atlasArgMapper` rather than reimplemented, and the environment and default layers come from running the real `atlasargs.Map` over a copy of the flag set whose `--dir` and `--dir-format` keep their raw values. One resolution, not two that have to agree by inspection.

**The atlas layout keeps today's native path.** When it was named through a query, the query is stripped and `--dir-format` normalized so exactly the same native invocation runs; the `Wrote …/atlas.sum` / `N migration file(s) hashed` output is byte-identical to `--dir file://d`, asserted directly. Every other layout is covered by a different set of source files, which no value of the native `--dir-format` axis (`{auto, ptah, atlas}`) can express, so those execute in the compat command: `atlasmigrateimport.SumFileNames` selects the file set, `migratesum.ComputeAtlasFiles`/`VerifyAtlasFiles` hash and verify over it, and every refusal renders through the `migratevalidate.FailAtlasChecksum*` helpers the native path and the apply-time gate already share.

`--dev-url` replays a converted directory through `ResolveApplySourceForFormat`, the same conversion apply executes, so what validate proves runnable is what apply would run. It is measured behavior, not an extra: CE replays a converted directory exactly as it replays a native one, and ignoring the flag would turn a failing validate into a passing one.

## Measured, pinned CE v1.2.0, absolute path

All transcripts below come from `~/Work/denis/ptah-atlas-conformance/bin/atlas`, reporting `atlas community version v1.2.0`. The system-wide `/usr/local/bin/atlas` is a different build and was never invoked. Measured 2026-08-02. `scripts/probe-atlas-integrity-verbs.sh` reproduces the whole differential run: it verifies the oracle's version before comparing anything, works under `mktemp`, and fails the run outright if a directory that was never named ends up hashed.

### Both spellings, every layout, byte-identical to CE

One fixture read differently by every rule — goose, golang-migrate pairs, Flyway `V`/`B`/`R`/`U`, a nested migration, a non-SQL sibling — hashed by CE and by ptah-compat:

```
format            spelling  result     dir hash
atlas             query     IDENTICAL  h1:MV7A9g8s4lH6C3RJ0o7ISFEVEP96I6Zfs2ZIuu9Y6Y4=
atlas             flag      IDENTICAL  h1:MV7A9g8s4lH6C3RJ0o7ISFEVEP96I6Zfs2ZIuu9Y6Y4=
goose             query     IDENTICAL  h1:MV7A9g8s4lH6C3RJ0o7ISFEVEP96I6Zfs2ZIuu9Y6Y4=
goose             flag      IDENTICAL  h1:MV7A9g8s4lH6C3RJ0o7ISFEVEP96I6Zfs2ZIuu9Y6Y4=
dbmate            query     IDENTICAL  h1:MV7A9g8s4lH6C3RJ0o7ISFEVEP96I6Zfs2ZIuu9Y6Y4=
dbmate            flag      IDENTICAL  h1:MV7A9g8s4lH6C3RJ0o7ISFEVEP96I6Zfs2ZIuu9Y6Y4=
liquibase         query     IDENTICAL  h1:MV7A9g8s4lH6C3RJ0o7ISFEVEP96I6Zfs2ZIuu9Y6Y4=
liquibase         flag      IDENTICAL  h1:MV7A9g8s4lH6C3RJ0o7ISFEVEP96I6Zfs2ZIuu9Y6Y4=
flyway            query     IDENTICAL  h1:q0FI/KQtDgidAGw6lzG/bY7D4InePOOJ0x0u5qKddH8=
flyway            flag      IDENTICAL  h1:q0FI/KQtDgidAGw6lzG/bY7D4InePOOJ0x0u5qKddH8=
golang-migrate    query     IDENTICAL  h1:ULlaQQbEB403a26QK01XPfa2Hqzvd/EsnRJDGp/L0u8=
golang-migrate    flag      IDENTICAL  h1:ULlaQQbEB403a26QK01XPfa2Hqzvd/EsnRJDGp/L0u8=
```

**The identical-sums invariant is asserted directly**, not inferred from both succeeding: `TestCompatMigrateHashSpellingsAgree_HappyPath` compares the two files byte for byte for all six layouts.

### validate, 15 rows, CE vs ptah-compat

Exit code and stdout/stderr compared verbatim:

```
goose            unhashed  MATCH   exit=1 You have a checksum error…|Error: checksum file not found
goose            clean     MATCH   exit=0 (no output)
goose            tampered  MATCH   exit=1 …|	L2: 1_init.sql was edited|…|Error: checksum mismatch
flyway           tampered  MATCH   exit=1 …|	L3: V1__x.sql was edited|…
golang-migrate   tampered  MATCH   exit=0        # the tampered file was the DOWN file
dbmate/liquibase all three  MATCH
```

The golang-migrate row is the covered-set rule seen from the verification side: CE never hashes the down file of a pair, so editing it is invisible. Editing the up file is not:

```
$ atlas migrate validate --dir 'file://e?format=golang-migrate'    # after editing 1_init.down.sql
exit=0
$ atlas migrate validate --dir 'file://e?format=golang-migrate'    # after editing 1_init.up.sql
	L2: 1_init.up.sql was edited
Error: checksum mismatch                                            exit=1
```

The mismatch line names the **source** file, which is what the shared `migratevalidate` helpers give for free.

### Precedence, measured in both directions

A resolver that always preferred the flag and one that always preferred the query agree on every input where the two name the same format, so both directions were measured on a layout they disagree about (`1_init.sql`, `V1__x.sql`, `U1__undo.sql`):

```
$ atlas migrate hash --dir 'file://d?format=goose'  --dir-format flyway
  -> 1_init.sql U1__undo.sql V1__x.sql          (goose won)
$ atlas migrate hash --dir 'file://d?format=flyway' --dir-format goose
  -> V1__x.sql                                  (flyway won)
```

The query wins. It wins even when empty, which is the row that forces the forwarding rewrite to *replace* `--dir-format` rather than append beside it:

```
$ atlas migrate hash --dir 'file://d?format=' --dir-format flyway
  -> 1_init.sql U1__undo.sql V1__x.sql          (atlas, the flag was irrelevant)
```

An earlier round measured this on a single-file directory where goose and flyway agree, which separated nothing. The discriminating layout is what the tests use.

### `--env`, and what outranks it

```
$ atlas migrate hash --env local                            # migration { format = flyway }
  -> V1__x.sql
$ atlas migrate hash --env local --dir-format goose
  -> 1_init.sql U1__undo.sql V1__x.sql                       (the flag outranks the config)
$ atlas migrate hash --env local --dir 'file://migrations?format=atlas'
  -> 1_init.sql U1__undo.sql V1__x.sql                       (the query outranks the config)
```

ptah-compat reproduces all three, and the third is the input the mutation sweep proved nothing was holding (below).

### `--dev-url`

```
$ ptah-compat migrate validate --dir 'file://d?format=goose' --dev-url sqlite://ok.db     exit=0
$ atlas       migrate validate --dir 'file://d?format=goose' --dev-url sqlite://ok2.db    exit=0

# same directory, one statement replaced with THIS IS NOT SQL;
$ ptah-compat …  exit=1  error: error validating migration SQL on dev database: …near "THIS": syntax error
$ atlas       …  exit=1  Error: replaying the migration directory: …near "THIS": syntax error
```

Integrity precedes the replay on both tools: an unhashed directory with an unreachable `--dev-url` reports `checksum file not found`, never a connection error.

## #980 does not become reachable, confirmed

`SumFileNames` returns an empty slice for an empty covered set and `ComputeAtlasFiles` writes the empty-set digest, so `LoadFS` — the only source of `no importable migration files found` — is never called from the integrity path. Measured on all three shapes #980 names:

```
                     ptah hash            CE hash    ptah validate  CE validate
empty directory      exit=0, 0 hashed     exit=0     exit=0         exit=0
no .sql files        exit=0, 0 hashed     exit=0     exit=0         exit=0
subdirectory only    exit=0, 0 hashed     exit=0     exit=0         exit=0
```

Both tools write the identical empty-set sum, `h1:47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU=`. It stays PR 3's problem. `TestCompatMigrateHashEmptySourceSet_HappyPath` pins it across five shapes, including golang-migrate with only a down file and Flyway with only an undo file.

## MINOR 7: a directory named `weird.sql`

Measured in full, and it is sharper than the note in #984 suggested — **Ptah's two paths disagree with each other**. Atlas CE's own glob matches the directory and then fails to read it:

```
$ atlas migrate hash --dir 'file://w?format=goose'
Error: sql/migrate: read file "weird.sql": read w/weird.sql: is a directory
exit=1                                   # no atlas.sum written
```

`SumFileNames` skips directories, so the converted path writes a sum:

```
$ ptah-compat migrate hash --dir 'file://w?format=goose'
Wrote w/atlas.sum
1 migration file(s) hashed               exit=0
```

The native atlas path globs like CE and fails, differently worded:

```
$ ptah-compat migrate hash --dir file://w
error: failed to read weird.sql: read weird.sql: is a directory        exit=1
```

And the punchline — CE refuses to read the sum Ptah wrote:

```
$ atlas migrate validate --dir 'file://w?format=goose'
You have a checksum error in your migration directory.
Please check your migration files and run 'atlas migrate hash' to re-hash the contents

Error: sql/migrate: read file "weird.sql": read w/weird.sql: is a directory     exit=1
```

Not a wrong-sum risk in the other direction: CE can never write such a sum, because it fails first. Filed as **#991** with three options and a preference, and pinned by `TestCompatMigrateHashDirectoryNamedSQL_KnownDivergence`, which asserts *both* Ptah behaviors so any fix is a visible test change.

## Divergences that remain, all loud

Every one of these is an exit-1 refusal where CE exits 0. None can produce a wrong sum. All are filed as **#990** with transcripts, and all are pinned by tests so closing them is a visible change.

| input | CE | ptah-compat |
| --- | --- | --- |
| `--dir-format ""` | exit 0, atlas | `migration directory format must not be empty` |
| `?format=goose&other=1` | exit 0, goose | `unsupported migration directory URL query parameter "other"` |
| `?format=goose&format=flyway` | exit 0, **first** wins | `migration directory URL contains multiple format parameters` |
| `?format=flyway;x=1` | exit 0, **atlas** — the whole pair is dropped | `parse migration directory URL query: invalid semicolon separator in query` |
| `--dir 'file://a?format=goose' --dir file://b` | exit 0, last `--dir` wins | query rejected on the overridden `--dir` |
| a stray positional, including after `--` | exit 0 | `unexpected positional arguments [...]` |

The reasoning, in short: the empty `--dir-format` rejection is shared by nine verbs through one mapper, so fixing it on two of them makes the surface *less* consistent; the query rules live in `applyDirURLFormat` and `ParseLocalDir`, both of which `migrate apply` uses, and relaxing a permissive-parser rule there silently widens what PR 3's integrity gate accepts, which is the wrong order of operations; the repeated-`--dir` case would need a second notion of "which `--dir` counts" inside the rewrite for an input with no plausible caller; and the stray positional is refused identically on the forwarding path, so it is not new.

The semicolon row was added in review, and it is the one the first enumeration was structurally unable to find: the others all come from `applyDirURLFormat`, while this one arrives earlier, through `ParseLocalDir`'s `url.ParseQuery`. It is also the row with the strongest case for staying refused — CE does not fail on it, it silently hashes a *different set*, because `url.Parse().Query()` discards the whole pair and the format selection goes with it. The reference doc no longer states a count of these, only the list; a count in prose is a claim that goes stale silently, and this chain has been bitten by one twice.

Two shapes moved *toward* CE as a side effect of using the strict shared resolver. `--dir-format ATLAS` and `--dir-format " goose "` used to be lowercased and trimmed into `atlas` and `goose`; CE matches format names exactly and exits 1 on both, and now so do these two verbs. The other seven metadata verbs keep the lenient mapper — that inconsistency is with a pre-existing bug, not with this change, and is covered by #990.

Out of scope by construction: **#987**, the `migrate hash` success output. CE prints nothing; ptah-compat prints `Wrote …/atlas.sum` and `N migration file(s) hashed`, pinned by 14 assertions across 6 files. Those assertions are untouched, and the converted path deliberately reproduces the same two lines so naming a layout does not change the reporting.

## Mutation table — 31/31 killed

`-count=1`, restored by byte-copy, `git status --porcelain` clean afterwards. The last three rows were added in review. Reproduce with `python3 scripts/mutation-check.py scripts/mutations/atlas-compat-integrity-verbs.json`.

| mutation | reddened |
| --- | --- |
| **the query no longer wins over the flag** | precedence, per-format, spellings-agree, env, project (22 subtests) |
| **the flag is ignored, only the query selects a layout** | per-format, spellings-agree, project (17) |
| **an empty query value falls back to the flag** | atlas-layout-unmoved, precedence (2) |
| the resolved layout is discarded and everything reads as atlas | 13 tests (38) |
| atlas.hcl and PTAH_* layers are dropped from the resolution | env, project, project-precedence (7) |
| the raw `--dir` keeps its rejecting mapper, so `?format=` never resolves | 16 tests (51) |
| the raw `--dir-format` keeps its rejecting mapper | 11 tests (35) |
| an omitted `--dir` does not fall back to the native default | default-directory (0) |
| the forwarded arguments are not rewritten for an atlas query | atlas-layout-unmoved, per-format, env (9) |
| the rewrite keeps the `?format=` query on `--dir` | atlas-layout-unmoved, per-format, env (9) |
| **the rewrite leaves a disagreeing `--dir-format` in place** | atlas-layout-unmoved, precedence (3) |
| the rewrite never appends a missing `--dir` | env (1) |
| **the rewrite never appends a missing `--dir-format`** | project-precedence, env (2) |
| the rewrite drops the value token of the flag it replaced | atlas-layout-unmoved, project (5) |
| unknown flags and stray positionals are silently ignored | bad-args (4) |
| a converted layout is forwarded to the native command anyway | 15 tests (53) |
| hash covers the atlas glob instead of the per-layout file set | 11 tests (23) |
| hash writes `ptah.sum` instead of `atlas.sum` | 13 tests (48) |
| hash does not stat the directory first | missing-directory (1) |
| hash reports the covered name count instead of the entry count | output, empty-set, divergence pin (6) |
| validate verifies the atlas glob instead of the per-layout file set | 5 tests (11) |
| validate accepts an unhashed converted directory | validate failure, dev-url failure (2) |
| validate accepts a tampered converted directory | validate failure (4) |
| validate reports a malformed sum as a usage error | validate failure (1) |
| validate drops the mismatch line from the refusal | validate failure (4) |
| validate ignores `--dev-url` on a converted directory | dev-url failure (1) |
| validate replays before checking integrity | dev-url failure (1) |
| validate does not stat the directory first | missing-directory (1) |
| **the rewrite drops the `--` argument terminator** | argv-terminator (2) |
| **`--help` no longer short-circuits before resolution** | help (4) |
| a semicolon query is parsed instead of refused | semicolon-query (3) |

The bolded row **"the rewrite never appends a missing `--dir-format`"** survived the first sweep. It is only observable when a *lower* layer would otherwise supply a value the forwarded native command refuses — with nothing below it, `atlasargs.Map` appends the atlas default anyway and the mutation is a no-op. Two inputs separate them, and neither was covered: `atlas.hcl migration.format = flyway` and `PTAH_DIR_FORMAT=goose`, each beside `--dir 'file://d?format=atlas'`. Both were then measured against CE (`atlas migrate hash --env local --dir 'file://migrations?format=atlas'` with `format = flyway` covers the atlas set, byte-identically) and added in `1513f17`.

One piece of the resolver is deliberately **not** mutation-tested: `atlasVerbNativeName`, which maps an Atlas flag name to the native name it forwards to. On these two verbs every pair is identical (`dir`→`dir`, `dir-format`→`dir-format`, `dev-url`→`dev-url`), so no input separates it from a hard-coded lookup. It is defensive against a future verb that maps them differently, and saying so is better than pretending a test holds it.

## The mutation harness is committed

`scripts/mutation-check.py` plus `scripts/mutations/{atlas-sum-file-set,atlas-compat-integrity-verbs}.json`. The sweep that proves a rule is held has been rebuilt by hand for three PRs running; this commits it with the two things the throwaway versions got wrong.

**It restores by byte-copy, never `git checkout --`.** That restore silently discarded uncommitted work in the mutated files and then reported `0/N killed` — which reads as "every mutation survived" rather than "the sweep ate the code under test". The incident is in the module docstring so the next reader does not reintroduce it. It also refuses a dirty tree before starting and re-checks the tree after, because an interrupted run cannot otherwise tell its own damage from work in progress. The repository root comes from the script's own location, and machine-specific values (the oracle path) stay in the caller's environment rather than the committed spec.

Re-running #984's 32 mutations through the committed harness reproduces **32/32 killed**, tree clean — which is the harness validating itself against a known-good result.

## The case for a CI job that fetches the pinned oracle

Filed as **#989** rather than done here, and the reason is specific.

The differential fuzz and `regenerate.sh` both skip without `PTAH_ATLAS_ORACLE`, which CI never sets. That is survivable only because every rule the fuzz has uniquely caught was afterwards given a corpus shape that pins it without the oracle — verified by hand in #984 for five rules, and again here. **Three rounds running, by hand.** It is a maintained property, not a guaranteed one, and the first round somebody adds a fuzz-only rule without the matching corpus shape, CI goes on passing while the rule stops being held.

It was not done in this PR because a fetch worth having is pinned by the sha256 of the **linux/amd64** release artifact, and the only oracle available here is the darwin build — committing its digest would pin the wrong thing, and an unpinned `curl | install` in CI trades a supply-chain regression for the coverage. The second reason is posture: the project's position on the Atlas artifact is deliberately black-box, and pulling the binary into every CI run on every fork is a different footprint from an operator running it locally. Both are maintainer calls. #989 carries the suggested workflow shape.

## Conformance

For `stokaro/ptah-atlas-conformance`, not pushed. A `ce-integrity/source-dir-formats` scenario should pin, for each of the six layouts and both spellings: `migrate hash` writes an identical `atlas.sum`; `migrate validate` exits 0 clean, 1 `checksum file not found` unhashed, and 1 `checksum mismatch` tampered with the `L<n>:` line naming the source file. Plus the rows that stop a future implementation over-refusing: golang-migrate editing only the down file (exit 0), an empty covered set on all five formats (exit 0, empty-set sum), and the precedence pair on a layout where the two formats disagree.

## Verification

`go vet ./...`, full `go test ./... -count=1`, `golangci-lint run` (0 issues), `go tool qtlint ./...`, `scripts/check-test-style.sh` (its testify guard silently skips without ripgrep, #975, so it was also run by hand: no matches), all three public-API scripts, and every docs checker. `check-responsive.mjs` needs a built site (`dist not found`) and was not run; it is unaffected. `feature-matrix-rows.json` was edited and the page regenerated; `build-feature-matrix.mjs --check` passes.

Branched from `e04e9d4`; `master` has since gained the unrelated #988.

Refs #973, #980, #981, #982, #984, #987, #989, #990, #991.


denisvmedia:
One correction to the body, on a number I repeated rather than measured.

The body says the `migrate hash` success output (#987) is "pinned by 14 assertions across 6 files". Measured on this branch:

```
$ grep -rn "migration file(s) hashed" --include="*_test.go" cmd/ | grep -v migrate_integrity_
cmd/migratehash/migratehash_test.go        2
cmd/atlas/project_config_snapshot_test.go  7
cmd/atlas/atlas_test.go                    4
```

**13 assertions across 3 pre-existing test files**, none of them touched here. The claim that matters — that #987's surface is untouched and this PR reproduces the same two lines on the converted path rather than changing them — holds; the count was carried over from the planning note without checking, and 14/6 is not what is there.

This branch adds 5 more assertions on the same two lines (`migrate_integrity_formats_test.go`, `migrate_integrity_resolution_test.go`), so whoever closes #987 has 18 to update, not 13.

Issue #980, opened 2026-08-01
Converted (?format=) directories with nothing to execute: Ptah errors where Atlas CE reports "No migration files to execute"


denisvmedia:
Found while measuring #973. Pre-existing on `master` @ `4680f02` — not introduced by #972.

Measured against the pinned Atlas CE v1.2.0 oracle (`ptah-atlas-conformance/bin/atlas`, invoked by absolute path; `/usr/local/bin/atlas` is a different build and was not used), 2026-08-01.

## The divergence

Three directory shapes that a converted format has nothing to execute in. CE treats all three as a clean no-op and exits 0; Ptah exits 1 with a load error.

| directory, `?format=goose` | Atlas CE v1.2.0 | ptah-compat @ 4680f02 |
| --- | --- | --- |
| holds only `README.md` | `No migration files to execute`, exit 0 | `error: atlas migrate apply --dir: no importable migration files found in <dir> for format "goose"`, exit 1 |
| completely empty | `No migration files to execute`, exit 0 | same error, exit 1 |
| holds only `sub/1_init.sql` | `No migration files to execute`, exit 0 | same error, exit 1 |

```
$ atlas migrate apply --url sqlite://empty.db --dir 'file://emptydir?format=goose'
No migration files to execute
exit=0

$ ptah-compat migrate apply --url sqlite://empty.db --dir 'file://emptydir?format=goose'
error: atlas migrate apply --dir: no importable migration files found in /…/emptydir for format "goose"
exit=1
```

The same shapes measured on formats whose file set is narrower than `*.sql` behave identically on CE — exit 0 — including golang-migrate holding only `1_init.down.sql` and flyway holding only `U1__init.sql`.

## Why it matters

`mkdir -p migrations && migrate apply` in a CI bootstrap is a real path, and #972 already had to add a no-SQL exemption to the native-atlas integrity gate for exactly that reason (CE exits 0 there too). The converted-format branch never got the equivalent treatment because `LoadFS` rejects an empty entry set outright:

```go
if len(entries) == 0 {
    return nil, fmt.Errorf("no importable migration files found in %s for format %q", display, format)
}
```

The error is arguably useful when a user typos a format, but it is indistinguishable from the legitimate empty-directory case, and it diverges from the oracle on all three shapes.

## Scope note

[...]

Issue #976, opened 2026-08-01
atlas.sum gives subdirectory migrations no protection; Ptah executes them while Atlas CE ignores them


denisvmedia:
Found while verifying #972. Pre-existing on `master` — not introduced by any recent PR.

## The divergence

Atlas CE reads migration files from the top level of a migration directory only. Ptah's migration registrar recurses. So a directory holding `1_a.sql` plus `sub/2_b.sql` means two different things to the two tools, and the integrity file only covers what the hasher saw.

`ptah migrations hash` reports `1 migration file(s) hashed` for that directory, and the resulting `atlas.sum` lists only `1_a.sql`. The subdirectory migration is invisible to hashing and to validation, but not to execution:

```
# tamper sub/2_b.sql after hashing
$ ptah-compat migrate validate --dir file://mig                  exit=0
$ ptah-compat migrate apply --url sqlite://t2.db --dir file://mig  exit=0
$ sqlite3 t2.db .tables
a  atlas_schema_revisions  b  pwned
```

Atlas CE on the same directory applies only `1_a.sql` and never sees the subdirectory at all.

## Why it matters

`atlas.sum` is the mechanism #955, #962 and #970 exist to enforce. On this directory shape it is not a weaker guarantee — it is no guarantee, because the file that executes is not among the files that are hashed. Anyone who assumes "validate passed, so nothing was edited" is wrong.

It also undermines the checksum gate added in #972: the "does this directory contain any `.sql` file" exemption there has to reason about a file set that the loader and the hasher disagree about.

## Two candidate resolutions

1. **Stop recursing** — match CE exactly: only top-level `.sql` files are migrations, and a subdirectory is ignored. Simplest to reason about, matches the oracle, and makes `atlas.sum` cover exactly what runs. Breaking for any project that currently relies on nested migration files.
2. **Hash what we execute** — keep recursion but make the hasher walk the same tree the registrar does. Keeps existing layouts working, but the resulting `atlas.sum` would then diverge from CE's for the same directory, which breaks byte-parity — the one property #973 needs.

These pull in opposite directions, and per the surface-split principle the answer may differ per surface: `ptah-compat` should match CE (option 1) because drop-in fidelity is the point, while native `ptah` could keep recursion as a documented capability provided its own sum covers everything it runs (option 2). If that split is chosen, it must be stated loudly in the docs rather than left implicit.

## Related divergence found alongside it

[...]

Issue #975, opened 2026-08-01
check-test-style.sh silently skips its testify guard when ripgrep is absent


denisvmedia:
## What

`scripts/check-test-style.sh` uses `rg` for its first guard (the testify-import check). When `rg` is not on `PATH` as a **binary**, that guard does not fail and does not warn — it silently no-ops, and the script still prints `OK` and exits 0.

## Why this is worse than a missing check

Two independent reviewers hit this in the same week, on two different PRs (#965 and #972), and in both cases the script reported `OK`. A gate that reports success when it did not run is more harmful than no gate: it is cited as evidence in review ("check-test-style passed") while having checked nothing. The other guards in the script did run, which makes the partial silence harder to notice than a total failure would be.

The specific trap on macOS: `rg` is commonly installed as a **zsh function or alias** rather than a binary. It resolves fine in an interactive shell, so the check appears to work when a developer runs it by hand, but the script executes under `/bin/sh`, where the function does not exist. Local runs and CI can therefore disagree, and the direction of the disagreement is the dangerous one — locally silent, and silent in any environment without the binary.

## Reproduce

```sh
# with rg on PATH as a binary
./scripts/check-test-style.sh          # testify guard runs

# without it (or with rg only defined as a zsh function)
env PATH=/usr/bin:/bin ./scripts/check-test-style.sh
# still prints OK, still exits 0, testify guard never executed
```

## Expected

Missing tooling must be loud. Either:

1. **Fail closed** — if `rg` is not executable, exit non-zero with a message naming the guard that could not run. This is the right default for a CI gate.
2. **Fall back to `grep -rE`** for this guard. The pattern is simple enough that portable `grep` covers it, which removes the dependency entirely and is probably the better fix.

Whichever is chosen, the script must never print `OK` for a guard it did not execute.

## Test

The regression test is a run under a stripped `PATH`: assert the script exits non-zero (option 1) or that the guard still detects a planted `testify` import (option 2). Assert both directions — that it fires on a violation and stays silent on a clean tree — so the test cannot itself become vacuous.

## Related

- Surfaced during review of #965 and #972.
- Worth auditing the other `scripts/*.sh` gates for the same shape: any `command -v`-free use of a non-POSIX tool inside a script that reports success.

denisvmedia:
[...]

Issue #973, opened 2026-08-01
migrate apply does not gate converted (?format=) directories; Atlas CE refuses both unhashed and tampered ones


denisvmedia:
Found while reviewing #972 (the #970 fix). #972 exempts directories read through `?format=` from the apply-time integrity gate. That exemption was justified in the PR with the claim "Atlas does not gate them either". **That claim is measured FALSE.** #972 keeps the exemption because closing it is larger than that PR, and renames its test to `..._ConvertedDirStaysUngated_KnownDivergence` so the gap is visible rather than certified as parity. This issue tracks closing it.

## Measured, pinned Atlas CE v1.2.0 (`ptah-atlas-conformance/bin/atlas`), 2026-08-01

Unhashed goose directory:

```
$ atlas migrate apply --url sqlite://ce1.db --dir 'file://goose?format=goose'
You have a checksum error in your migration directory.
Please check your migration files and run 'atlas migrate hash' to re-hash the contents

Error: checksum file not found
exit=1                        # ce1.db was never created
```

CE can hash a converted directory, and then applies it:

```
$ atlas migrate hash --dir 'file://goose?format=goose'
exit=0
$ cat goose/atlas.sum
h1:GrzxJ07R56OC5oottJ7wgbW7r8VOOeHaCJJdeEJA+e4=
1_init.sql h1:+rmUcws+9hthURrF7o/ArM9yUHUDSMVvMFGrA04Xyic=

$ atlas migrate apply --url sqlite://ce2.db --dir 'file://goose?format=goose'
Migrating to version 1 (1 migrations in total):
  -- migrating version 1
    -> CREATE TABLE widgets (id INTEGER PRIMARY KEY);
  -- ok (354.708µs)
exit=0
$ atlas migrate validate --dir 'file://goose?format=goose'
exit=0
```

**The worse half — a tampered converted directory executes on `ptah-compat` where CE refuses:**

```
# hashed by CE, then one migration edited without re-hashing
$ atlas migrate apply --url sqlite://t-ce.db --dir 'file://tamper?format=goose'
You have a checksum error in your migration directory.

	L2: 1_init.sql was edited

Please check your migration files and run 'atlas migrate hash' to re-hash the contents

Error: checksum mismatch
exit=1

$ ptah-compat migrate apply --url sqlite://t-ptah.db --dir 'file://tamper?format=goose'
Migrating to version 1 from 1 pending migrations.
Migration complete. Current version: 1
exit=0                        # the edited migration executed
```

[...]

PR #1001, opened 2026-08-02
Converge the Flyway importer on Atlas CE's file selection

### LINKED ISSUE #982: Flyway importer executes migrations no atlas.sum covers: lower baselines are kept and lowercase prefixes are accepted
> **Priority raised 2026-08-02, after #994 (PR 3 of #973) landed the apply-time integrity gate.**
>
> This is no longer an internal-consistency issue. It is the reason `ptah-compat migrate apply` cannot promise, for Flyway, what it now promises for the other four layouts: that everything it executes was covered by the checksum it verified.
>
> The gate verifies the set Atlas CE covers. The importer runs a **wider** set. So for Flyway, and only for Flyway, `executed ⊄ verified`, and the excess executes on a directory **both tools call clean**:
>
> ```
> B1__one.sql  B2__two.sql  V3__three.sql        # hashed by the pinned CE binary
>   atlas.sum covers B2__two.sql and V3__three.sql — B1 is not covered
>   append `CREATE TABLE pwned (id int);` to B1__one.sql
>
>   $ atlas       migrate validate --dir 'file://d?format=flyway'   exit=0
>   $ ptah-compat migrate validate --dir 'file://d?format=flyway'   exit=0
>   $ atlas       migrate apply    …  -> two, three
>   $ ptah-compat migrate apply    …  -> two, three, one, PWNED
> ```
>
> ```
> V1__init.sql hashed by CE, then v2__evil.sql added
>   validate exits 0 on both; atlas applies 1 migration, ptah applies 2
> ```
>
> Re-running `ptah-compat migrate hash` does not close either shape: the covered set is the same set, so the file stays outside it. Measured controls confirm goose, dbmate, liquibase and golang-migrate each refuse the corresponding file at import — **Flyway is the only format with this hole.**
>
> **The fix is convergence, not refusal.** Making the gate refuse when the importer's consumed set is not a subset of the covered set would refuse whole directories that Atlas CE applies happily, which is an over-refusing gate — the failure mode #973's whole chain is built to avoid. The importer has to adopt CE's selection so consumed equals covered.
>
> #994 does not fix this. It documents it (`versioned/integrity-and-safety.md` carries a scoped caution naming both shapes), states it in `verifyConvertedAtlasApplyChecksum`'s doc comment, and pins it in `scripts/probe-atlas-apply-gate.sh` section 9a, which fails if either shape stops reproducing. Reproduce with that script.

---

Found while measuring #973. Pre-existing on `master` @ `4680f02`.

Measured against the pinned Atlas CE v1.2.0 oracle (`ptah-atlas-conformance/bin/atlas`, absolute path; the system-wide `/usr/local/bin/atlas` is a different build and was not used), 2026-08-01.

Unlike the sibling follow-ups from #973, this one changes **which migrations execute**, not just which are reported or hashed.

## Divergence 1 — lower baselines are not squashed

`B2__baseline.sql` squashes history at version 2. CE drops everything at or below that version, *including a lower baseline*; Ptah's importer only drops non-baseline files.

```
dir: B1__one.sql  B2__two.sql  V3__three.sql

$ atlas migrate hash --dir 'file://two-baselines?format=flyway'
h1:n+akKfr0d1eU9jirCvmOSX8OjgGOC9XxTK+Z39Fc3lM=
B2__two.sql   h1:/hq6iJA19Ii0BWP5L0OJdl5mKtkOHwqHBC7Q/S5plL8=
V3__three.sql h1:cEkYD+HiwITshuvY4j/ATiTwsmtBpuvg1fK7/GX7uZ4=
```

`B1__one.sql` is absent from CE's view of the directory. `loadFlywayEntries` in `internal/atlasmigrateimport/importer.go` keeps it:

```go
if !entry.baseline && baseline != nil && compareFlywayVersions(entry.version, *baseline) <= 0 {
    continue
}
```

The `!entry.baseline` guard exempts every baseline from the cut, so `B1__one.sql` is converted and **executed** where CE never sees it. Baseline files are usually cumulative-schema snapshots, so replaying a superseded one before the current one is not a harmless extra statement.

## Divergence 2 — lowercase prefixes are accepted

CE's prefix match is case-sensitive; Ptah's `flywayFileRe` is `(?i)`:

```go
flywayFileRe = regexp.MustCompile(`(?i)^([vbru])([0-9][0-9._]*)?__(.+)\.sql$`)
```

```
dir: v1__one.sql  V2__two.sql

$ atlas migrate hash --dir 'file://lower-prefix?format=flyway'
h1:4djhMkolNRBLZ83O+vwssF8HYiquL0qTScFU7VlSLiQ=
V2__two.sql h1:tZj8SvsBcwT0kKIWrPQpbU8hci4nQTI4xEM/762MvuQ=
```

CE ignores `v1__one.sql` entirely. Ptah imports and executes it. The same asymmetry applies to a lowercase `u`, which Ptah treats as an undo file and skips while CE treats it as an ordinary unrecognized file — the two happen to agree on the outcome there, for different reasons.

## Why this surfaced now

#973 needs an integrity file set byte-identical to CE's per format. That file set is derived from CE's selection rules, and building it exposed both differences: a hasher reusing `flywayFileRe` or the importer's baseline filter would produce a sum CE does not, and would false-refuse a directory CE hashed.

The #973 hasher therefore implements CE's rules directly rather than reusing the importer's, which is correct for hashing but leaves the importer's *execution* selection unchanged and now visibly inconsistent with the sums the same tool writes: a flyway directory holding two baselines will hash the CE way and execute the Ptah way.

## Definition of done

- Decide whether the importer adopts CE's selection rules (baseline squashing including lower baselines, case-sensitive prefixes) or keeps its own as a documented divergence. Given that the hasher must match CE byte for byte, keeping two different selection rules inside one tool needs an explicit justification. **Since #994, "keeps its own" also means accepting that the apply-time gate does not cover everything apply executes for this one layout.**
- Measure both shapes above against the pinned CE binary and pin them with tests that fail in both directions.
- If the rules converge, the #973 hasher and the importer should share one selection function rather than agreeing by coincidence. `atlasmigrateimport.SumFileNames` is that function; `loadFlywayEntries` is what has to move to it.
- Delete the scoped caution #994 added to `versioned/integrity-and-safety.md`, and the `#982` paragraph in `verifyConvertedAtlasApplyChecksum`'s doc comment, and turn section 9a of `scripts/probe-atlas-apply-gate.sh` into a parity assertion.

Refs #973, #994.

denisvmedia:
Closes #982. On top of #984, #992 and #994.

## The gap

The apply gate #994 added verifies `SumFileNames`; the importer ran `flywayFileRe`/`loadFlywayEntries`, a **wider** set. So `executed ⊄ verified` on a directory both tools call clean:

```
1. superseded baseline, tamper on the uncovered B1
   atlas [B2, V3]   base [B1, PWNED, B2, V3]   head [B2, V3]
2. lowercase prefix, V1__init.sql hashed by CE, attacker adds v2__evil.sql
   atlas [V1]       base [V1, v2__evil]        head [V1]
```

`loadFlywayEntries` and `SumFileNames` now share `flywayCoveredFiles`, which returns the parsed, ordered records. Not a call to `SumFileNames`: that returns `[]string`, and the importer needs each file's kind, version token and description, so it would parse them a second time — the same two-implementations shape one level down.

## Atlas CE's Flyway revision model, measured

The projection forced by convergence turned out to interact with revision tracking, so the model was measured end to end rather than assumed. Full transcripts — seven scenarios with revision-table dumps plus four isolating the ordering claim — are in [#982](https://github.com/stokaro/ptah/issues/982#issuecomment-5155984990). The summary:

1. **Pending** = covered files whose version token is **string**-greater than the current (max recorded) version. Not "not in the recorded set".
2. **Execution order** = `atlas.sum` order — surviving baseline first, then versioned by components, then repeatables — and it holds **across runs**.
3. **Identity** = the version token string alone; no file identity.
4. A pending baseline is **executed**, not recorded, and can fail loudly.
5. Revision rows are never deleted; baselines get `type = 2`, like any migration.

The decisive measurement:

```
G. V1 applied; then B10__base.sql (squashes V1) + V2__x.sql added
   covered: B10__base.sql V2__x.sql       sum order and version order DISAGREE
   -> migrating version 10, then version 2      tables: old,base,x
```

Two pending files on an existing database, and CE followed **sum order**. That establishes "surviving baseline first" as a property of the applied sequence rather than of one conversion, and it is why the band stays.

And execute-vs-record, settled:

```
B. B3__base.sql = CREATE TABLE p;  (p already exists)
   -- migrating version 3 -> CREATE TABLE p (id int);
      table p already exists
   Error: ... from version "3": table p already exists      exit=1
```

## What the convergence costs

Atlas CE keys a migration on an opaque version **string** and executes in sum order; Ptah's migrator keys on an `int64` and executes in version order. Three measured facts force the projection:

| measured | consequence |
| --- | --- |
| the surviving baseline executes FIRST whatever its version, across runs | separate int64 **bands** |
| every repeatable executes as version `""` whatever its token | one **reserved slot** at the top |
| `Vx__a.sql` and `Vy__b.sql` both score `{0}` and both run, in walk order | a small **tie budget** |

### Migration IDENTITY changes, and that needs a migration

The converted version is the key `atlas_schema_revisions` stores, so a database migrated through `?format=flyway` by **v0.1.0–v0.1.2** records versions this build matches to no file: every migration reads as pending. Measured on one database, base then head:

```
base: revisions 10000, 20000                     rows in seeded: 1
head: CREATE TABLE fails -> dirty revision, every later apply blocked
      or, where the body is a seed: exit 0, "Migration complete.", rows: 2
```

`migrate apply` now reconstructs the old encoding for the files it still covers and refuses before executing anything:

```
error: this database was migrated by a Ptah build older than the one that fixed
stokaro/ptah#982 ... 2 already-applied migration(s) would run a second time and
nothing has been applied

recorded version -> version this build uses:
  10000                -> 4611686018427469511  V1__init.sql
  20000                -> 4611686018427510315  V2__seed.sql

to adopt the new encoding, migrate the recorded versions forward and re-run:
  UPDATE atlas_schema_revisions SET version = '4611686018427469511' WHERE version = '10000';
  UPDATE atlas_schema_revisions SET version = '4611686018427510315' WHERE version = '20000';
```

A test extracts those statements from the message, runs them verbatim, applies again (`No migration files to execute.`, seed did not re-run) and then applies a newly added `V3`. The printed repair and the working repair cannot drift. Statements are printed because `migrate set`/`status` reject `?format=` (**#1002**) and `--baseline` refuses on a non-empty revision table.

### The band is right; reading it as a chronology was wrong

Adding a baseline to an already-applied directory is the ordinary Flyway workflow, and CE runs it (A, C, G). The converted baseline lands below every recorded version, and the linear guard read that as "authored earlier". It is not — the position encodes sum order. Exactly that one version is now exempted from the guard, supplied as data by the layer that computed the projection.

The exemption is narrow, and the inputs that prove it:

```
K  baseline after apply            atlas 0:[p,q,r,s]   head 0:[p,q,r,s]   fixed
I  ordinary V2 below high-water    atlas 1 out-of-order   head 1 out-of-order   still refused
L  versioned added after R__       atlas 1 out-of-order   head 1 out-of-order   still refused
```

`I` and `L` are what separate "exempt the surviving baseline" from "exempt anything below the high-water mark" — CE refuses both, so a loose exemption would silently apply them.

### Two divergences kept on purpose

CE's string comparison means a baseline whose token does not sort above the high-water mark **silently never runs**:

```
F  V2 applied, add B10          atlas: "No migration files to execute", base absent
D  V1,V2,V3 applied, add B2     atlas: "No migration files to execute", base absent
```

Ptah runs it in both. Matching CE would import a silent skip, and case B shows CE's baseline execution can itself be destructive, so "match Atlas" is not automatically safe. Filed as **#1003** as a decision, pinned by `..._KnownDivergence`, and stated in the docs.

## Also fixed since first review

**32-bit was compile-verified but semantically wrong.** Components were scored with `strconv.Atoi`, whose bit size is the platform `int`; a 32-bit build clamped `V9000000000` and `V10000000000` to the same ceiling, writing an `atlas.sum` the oracle rejects. Now `int64` via `ParseInt(_,10,64)`, pinned by `testdata/ce-sums/flyway/wide-components` — an oracle-written sum, so CI holds it without the oracle.

**Probe 9a settled for counts.** It now compares applied migrations in application order and checks the count against the covered-set size; `sqlite3` is a hard requirement so it cannot degrade. Verified against the exact mutant: `ran=[base,x]` vs `ran=[x,base]` → DIFFER.

**Editing a repeatable body** fails on a Ptah revision checksum where CE exits 0, so the matrix row is back to **`partial`** with the limitation in the note.

## Corrections to earlier claims

- **"the only input where head refuses and CE applies"** was misleading. The tie-budget class is refused on `master` too, **and CE applies all six**. The fixture is now five ordinary helper files starting with a capital V, which Flyway ignores and CE's loose parse turns into migrations.
- **A pre-existing refusal was undisclosed**: a description ending in `.down` converts to a name the migrator reads as a down half. CE applies both files; base and head both refuse. Now pinned, with a note that it sits after `LoadFS` and so the fuzz's enumeration cannot see it.
- **"CE cannot execute those directories at all"** was too strong: CE executes up to the collision and *then* panics, leaving table `a` behind.
- **The tie-budget impossibility was overstated.** `tie=8` yields a bound holding every `yyyyMMddHHmmss` stamp to roughly year 5651. The budget stays 4 as a stated choice, not a necessity.
- `assertImporterConsumesTheCoveredSet` derives membership and order from the same slice; the comment now says so and points at what does hold ordering.
- `mutation-check.py` compiles separately so BUILD-FAILED is a real diagnostic, reports no-result rows on their own line, and documents running alone.

## Evidence

- **12-shape apply/re-apply parity sweep** against CE: 10 SAME, 2 DIFFER — and the two are exactly F and D above.
- **28 shapes**, execution set and creation order identical to the oracle.
- **Differential fuzz, third assertion**: the importer's consumed set vs the covered set, with refusals matched against an enumerated list rather than skipped.
- **Mutation: 27/27** on the new spec, including the detector and the exemption. Predecessors reproduce **32/32**, **31/31**, **17/17**, tree clean.
- Probe: every gated row matches CE; 9a five rows MATCH on order.

Two rules are documented as having **no separating input** rather than given tests that would not hold them — the legacy reconstruction's `FormatFlyway` guard and its trailing-zero trim. Every other survivor was resolved by finding the discriminating input: both versions recorded at once; a nested guard reachable only when the directory name also parses; a baseline present alongside an ordinary out-of-order migration; and a goose version colliding with a converted baseline slot.

## Verification

`go vet ./...`, full `go test ./... -count=1`, `golangci-lint run` (0 issues), `go tool qtlint ./...`, `scripts/check-test-style.sh` (178 conditional groups, 45 white-box files — baseline unmoved; its testify guard printed `rg: command not found`, #975, so it was run by hand: no matches), the public-API scripts, every docs checker, `shellcheck -S warning` on the probe, and cross-builds for `GOARCH=386` and `GOARCH=arm`.

Branched from `1c70ea1`. Refs #973, #976, #980, #984, #991, #992, #994, #1002, #1003.


denisvmedia:
## Why `Migrator.WithOutOfOrderExempt` is public

This PR widens the public API of `migration/migrator` by one method, for what is a Flyway-compat need. Flagging the decision explicitly because "exported for exactly one caller" is how a surface grows by accident, and this repo has three API gates plus a released baseline precisely to stop that.

**An internal seam was considered first and rejected on cost, not on principle.**

`migration/migrator` already imports repo-root internal packages (`internal/dblock`, `internal/lexer`, `internal/atlasretry`), so `migrator → internal/X` is cycle-free. But the seam needs the *opposite* direction: `internal/atlasmigrate` has to obtain a setter that operates on a `*migrator.Migrator`. For an internal package to offer that, it must either

1. **name the type** — `func Apply(m *migrator.Migrator, versions []int64) *migrator.Migrator` — which makes `internal/X` import `migration/migrator` while `migration/migrator` imports `internal/X`: an import cycle; or
2. **erase the type** — `var Apply func(any, []int64) any`, installed from `migration/migrator`'s `init()`.

Option 2 is the only one that compiles, and it costs: a package that exists solely to hold one mutable function variable, an `init()`, package-level mutable state, and two type assertions on the apply path. The project's own Go guidance calls out `init()`, global state and package-level mutable variables as things to avoid, and it would trade a documented method for an undocumented indirection that is harder to audit, not easier.

**So the method stays public and the doc comment carries the whole contract** — what the linear guard assumes, when a version stops being a chronology, and how to misuse it:

> WHAT THE GUARD ASSUMES. Under `ExecOrderLinear` the migrator refuses a pending migration whose version is below the highest version already applied … Both treat "version below the current one" as evidence that the migration was AUTHORED before what is already in the database … That inference is only sound while the version is a chronology.
>
> WHEN IT IS NOT A CHRONOLOGY. … For Flyway the version is a projection of Atlas CE's `atlas.sum` ORDER, which is not authoring order: a surviving baseline is emitted and executed FIRST whatever its own version …
>
> HOW TO MISUSE IT. The exempt list is taken on trust. Supplying a version that IS chronological — an ordinary migration that really was authored late — silently turns off out-of-order detection for it: no error, no warning … Pass only versions whose position you computed yourself and know to be an artifact of a layout projection. Anything derived from user input, or a whole band of versions rather than the specific ones, defeats the guard for real branching mistakes.

The "whole band rather than the specific ones" warning is not hypothetical: it is precisely the mutation the sweep tests (`the exemption is loosened to every out-of-order version`), and the test that kills it needs a directory holding a baseline *and* an ordinary out-of-order migration at once, because an empty exempt list short-circuits the filter and makes the looser rule unreachable.

Happy to switch to the erased-type seam if you would rather the surface not grow — it is a contained change, and the tests that hold the behavior do not depend on which shape carries it.

Issue #1002, opened 2026-08-02
migrate status and migrate set reject ?format= and --dir-format, so a converted directory has no revision-repair route


denisvmedia:
Atlas CE accepts a foreign layout on `migrate status` and `migrate set` under both spellings. `ptah-compat` rejects both, which leaves an operator with no supported way to inspect or repair the revision table of a directory read through `?format=`.

## Measured, pinned Atlas CE v1.2.0, 2026-08-02

A hashed Flyway directory, already applied:

```
$ atlas migrate status --url sqlite://x.db --dir 'file://d?format=flyway'
Migration Status: OK
  -- Current Version: 1
  -- Next Version:    Already at latest version
  -- Executed Files:  1
exit=0

$ atlas migrate set 1 --url sqlite://x.db --dir 'file://d?format=flyway'
exit=0
```

`ptah-compat`, both verbs and both spellings:

```
$ ptah-compat migrate status --url sqlite://x.db --dir 'file://d?format=flyway'
error: atlas migrate status --dir: migration directory URL query parameters are not supported for this command

$ ptah-compat migrate status --url sqlite://x.db --dir file://d --dir-format flyway
error: atlas migrate status --dir-format: Atlas accepts --dir-format=flyway, but Ptah does not implement that directory format yet

$ ptah-compat migrate set 1 --url sqlite://x.db --dir 'file://d?format=flyway'
Error: atlas migrate set --dir: migration directory URL query parameters are not supported for this command
```

## Why it matters beyond parity

`migrate set` is the verb that repairs a revision table, and #982 created a situation that needs exactly that: it changed the Atlas version a Flyway file converts to, so a database migrated by Ptah v0.1.0–v0.1.2 through `?format=flyway` carries rows keyed on the old encoding. The PR that closed #982 added a refusal on the apply path so those migrations cannot silently re-run, and that refusal has to print raw `UPDATE` statements, because there is no verb that can do the same job.

`--baseline` is not an alternative — measured, it refuses once the revisions table is non-empty:

```
$ ptah-compat migrate apply --baseline <version> --url sqlite://x.db --dir 'file://d?format=flyway'
error: error baselining migrations: schema migrations table is not empty; rerun with force to baseline anyway
```

and the compat surface exposes no force flag for it.

## Suggested shape

[...]

Issue #954, opened 2026-08-01
Atlas checkpoint directive is not honored: double-apply on fresh databases, healthy Atlas databases degraded


denisvmedia:
Measured 2026-08-01 against Atlas (fixtures: `the observation study fixtures, checkpoint-dir/`).

Atlas `migrate checkpoint` writes `NNNN_checkpoint.sql` whose first line is `-- atlas:checkpoint`, and rewrites `atlas.sum` with it as an ordinary `h1:` entry. Measured Atlas semantics: on a **fresh** database only the checkpoint runs (pre-checkpoint files skipped, single revision row `type=2`); on a database that already has pre-checkpoint revisions the checkpoint is **silently skipped** ("No migration files to execute", no revision row).

Ptah ignores the directive entirely:

- `ptah-compat migrate apply` on a fresh db replays 001+002 **and then executes the checkpoint**: `table users already exists`, exit 1, dirty revision row (`applied=0` + error text).
- On a pre-checkpoint db it executes the checkpoint Atlas would skip — same failure. Worse: running this against a **healthy Atlas-checkpointed database degrades Atlas's own view** from OK to PENDING ("Executed Files: 3 (last one partially)").
- Native `ptah migrations up` (`--dir-format auto`) has the same checkpoint blindness into its `schema_migrations` table.
- The producing side knows the convention (`ptah migrations checkpoint`, the `--dir-format atlas` waiver) but the **read side is unguarded**.

Fix: teach the Atlas-format directory reader the `-- atlas:checkpoint` directive with both halves of the semantics (fresh: start from the latest checkpoint; existing pre-checkpoint revisions: skip). A ce-gating-style scenario should pin the measured Atlas behavior.

Related interop-ok facts worth keeping: `migrate validate` is byte-identical to Atlas on tamper detection, and revision rows Ptah writes are fully readable by Atlas.

Refs #951.

PR #1011, opened 2026-08-02
Write Atlas-format checkpoints from `migrate checkpoint --dir-format atlas`

denisvmedia:
Closes the `migrate checkpoint --dir-format atlas` item of #951. It was a recorded waiver: native checkpoints existed, but only the Ptah two-file convention was written.

> **Corrections after review.** Two claims below were wrong when first written and are fixed in place. (1) "No we-fail-where-CE-succeeds divergence is possible on this verb" — true of `checkpoint`, false of the *artifact* it writes: a shallow version bump against a recursive reader produced a directory CE applies at exit 0 and Ptah failed on. (2) "`?format=` needs a per-verb pass, filed separately" — it was not filed; #1013 now covers the untracked half, and the refusal is only correct on `checkpoint`.

## What the oracle says, before any code

Pinned CE v1.2.0, absolute path only.

**CE registers `migrate checkpoint` and gates it — with zero of its own flags.** The verb is absent from `atlas migrate --help`'s command list, but invoking it aborts rather than falling through to group help:

```
$ atlas migrate checkpoint
Abort: 'atlas migrate checkpoint' is not supported by the community version.
exit=1

$ atlas migrate frobnicate          # nonsense control
'atlas migrate' wraps several sub-commands for migration management.
Usage: atlas migrate [command] ...
exit=0
```

Flag parsing runs **before** the gate, which makes the registered flag set measurable on a gated verb:

| probe | result | reading |
| --- | --- | --- |
| `--frobnicate-nonsense` | `Error: unknown flag` | control: unknown flags are rejected here |
| `--dir-format` `--dir` `--dev-url` `--edit` `--format` `--qualifier` `--schema` `--lock-name` `--lock-timeout` | `Error: unknown flag` | **none of them registered on CE** |
| `--config`, `--env` | the community abort | inherited from the `migrate` parent, so they parse |
| `--var=x` | `variables must be format as key=value` | parses, also inherited |

So **there is no CE flag surface to conform to on this verb** — compat's existing `--dir`/`--dev-url`/`--dir-format` were already Pro-surface additions, and adding behavior is #951's second standing constraint. Atlas registers `--dir-format` here with **default `atlas`**.

**CE cannot write a checkpoint but does honor `-- atlas:checkpoint` on read.** That turned this item from guesswork into something the pinned binary certifies — both halves of #954 reproduce on CE v1.2.0:

```
$ atlas migrate apply --dir file://m --url sqlite://fresh.db   # after ptah-compat wrote the checkpoint
Migrating to version 20260802120402 (1 migrations in total)
$ sqlite3 fresh.db 'select version, description, type, applied, total from atlas_schema_revisions'
20260802120402|checkpoint|2|1|1

$ atlas migrate apply --dir file://m --url sqlite://pre.db     # db already at 20250801000002
No migration files to execute
exit=0
```

`atlas migrate validate` exits 0 on the directory we wrote, and `ptah-compat migrate apply` produces the byte-identical revision row and the identical SQLite schema.

Artifact shape is taken from the #954 fixture (`<version>_checkpoint.sql`; `-- atlas:checkpoint`, blank line, SQL; an ordinary `h1:` entry in `atlas.sum`), and the version policy from the same capture: Atlas timestamped `20260801100335` over a history dated `20250801000002`, i.e. **now()**, not newest+1.

## Measured branch table

| input | before | after |
| --- | --- | --- |
| `--dir-format atlas` | waiver, exit 1 | writes `<ts>_<name>.sql` + `atlas.sum`, exit 0 |
| **no `--dir-format`** (compat) | ptah pair | **atlas** — matches the Pro default |
| `--dir-format ""` (explicit) | ptah pair | atlas — measured CE takes empty as the atlas default |
| `--dir-format ptah` | ptah pair | unchanged |
| `--dir-format ATLAS` | waiver | atlas (case-folded) |
| `--dir-format auto` — **native** | `supports only "ptah"` | refused, named message |
| `--dir-format auto` — **compat** | unknown-format | unchanged: `auto` is not an Atlas format, so it draws the ordinary unknown-format rejection |
| `golang-migrate`/`goose`/`flyway`/`liquibase`/`dbmate` | rejected | unchanged |
| `sprocket` | unknown-format | unchanged |
| unhashed **atlas** directory | n/a | writes checkpoint + creates `atlas.sum`; CE validates |
| unhashed **ptah** directory | n/a | **refused** — see the guard below |
| directory already holding a checkpoint | n/a | new checkpoint sorts above all history; newest wins on apply (CE agrees) |
| nested migration newer than the timestamp | n/a | bound raised past it — see the blocker below |
| empty directory | `no migrations found` | unchanged, nothing written |
| `?format=atlas` on `--dir` | query params refused | unchanged — correct here, see below |
| `PTAH_DIR_FORMAT=ptah`, `atlas.hcl` `migration { format = "ptah" }` | n/a | both select the ptah pair |

## The version bound (blocker found in review)

The replay and the reader both recurse; the Atlas timestamp resolver scans only the top level. On `m/20250801000001_init.sql` + `m/sub/29990101000000_future.sql` the checkpoint replayed the nested file into its body and still sorted **below** it, so a fresh apply ran the checkpoint and then re-ran that migration — `table future_tbl already exists`, exit 1, dirty revision row. #954 reproduced by our own writer, and a **we-fail-where-CE-succeeds** divergence, since CE's shallow reader applies the same directory at exit 0.

The version is now the maximum of the timestamp and one above the newest migration from the recursive walk. `ResolveAtlasCheckpointVersion` keeps its signature and documents that its scan is deliberately shallow. The **equality** boundary (`>` vs `>=`) needs its own fixture, since on every other input the two operands differ; `TestCompatCommand_MigrateCheckpointNestedTieDoesNotCollide` supplies it.

## Two refusals that protect the directory

A directory must end with one integrity file and one convention:

- **Both sums.** Writing an atlas checkpoint beside a `ptah.sum` left the directory unreadable to `--dir-format auto`, at exit 0, with the failure surfacing on a later command.
- **Ptah files, hashed or not.** The sum-file check cannot see an *unhashed* ptah directory, and converting one made the checkpoint permanently invisible: discovery reads the ptah pair, while `validate` finds the new `atlas.sum` and calls the directory sound. The check is content-shaped. The converse is deliberately not symmetric — the Atlas name grammar also accepts ptah names, so it would refuse every ptah directory.

## `?format=`, scoped honestly

Refusing it on `checkpoint` is correct: **CE aborts every invocation of this verb, so there is no CE contract to diverge from.** That is not true of the others that share the refusal. On `lint`, `new` and `diff` CE exits 0 and *functionally* honors the parameter (`?format=golang-migrate` turns `L2: 1_init.down.sql was added` exit 1 into `Migration Status: PENDING` exit 0) and validates it (`?format=totally-bogus` → `unknown dir format`). Unknown-key intolerance is a further defect on all eight verbs, including the three that work. Tracking: #1002 (`status`/`set`), #990 (empty `--dir-format`, non-`format` and repeated keys on `hash`/`validate`), **#1013** (the rest).

## Mutation table — 30/30 killed

`-count=1`, byte-copy restore, `git status --porcelain` clean afterwards, no leaked files.

| mutation | reddened |
| --- | --- |
| `--dir-format=atlas` silently writes the ptah convention | 9 tests |
| compat default reverts to ptah | default, unhashed-ptah |
| an unset `--dir-format` falls back to ptah | empty-value |
| external dir formats accepted | rejections (1 subtest) |
| `--dir-format=auto` accepted | auto-rejection |
| the atlas branch is never taken | 13 tests |
| **timestamp trusted without the recursive bound** | nested-outranks, nested-tie |
| **bound applied with the wrong comparison (`>=`)** | nested-tie |
| atlas uses the ptah newest-plus-one counter | timestamp |
| second-integrity-file guard disabled | guard (4 subtests) |
| guard protects only the ptah / only the atlas direction | guard (1 subtest each) |
| guard also refuses the directory's own sum | own-sum, guard (2 subtests) |
| **unreadable directory treated as conflict-free** | unreadable-directory |
| **ptah-content guard disabled** | unhashed-ptah (2 subtests) |
| **ptah-content guard fires on any format** | 7 tests |
| **ptah-content guard probes the atlas format** | 15 tests |
| ptah-width guard dropped for atlas | ptah-width, rendered-value |
| **width guard measures the flag text** | rendered-value |
| ten-digit ceiling also applied to atlas | above-ptah-maximum |
| atlas dry run writes files | dry-run |
| directive appended after the body | 11 tests |
| directive omitted entirely | 12 tests |
| body not trimmed | leading-blank-lines |
| empty description writes bare `<version>.sql` | names (2 subtests) |
| `atlas.sum` not refreshed | 14 tests |
| existing checkpoint overwritten | refuses-to-overwrite |
| **checkpoint survives a failed `atlas.sum` write** | rollback |
| version not bumped past future-dated history | resolve-version |
| non-positive version accepted | non-positive |

Rows that came back SURVIVED, BUILD-FAILED or PATCH-FAILED across rounds are described in the review-response comment rather than quietly repaired.

## Verification

`go vet ./...`, `go test ./... -count=1`, `golangci-lint run` (0 issues), `go tool qtlint ./...`, `scripts/check-test-style.sh` — baseline untouched, no waiver added — plus a manual testify grep (its guard silently no-ops without `rg`, #975), all three public-API scripts, every docs checker, and `check-responsive` after `npm ci && npm run build` in `docs/site`.

## Docs

`docs/site/scripts/data/feature-matrix-rows.json` only, page regenerated by the builder. The "Atlas-format checkpoint output" row flips ❌ → ✅ with the CE-measured evidence, including both refusals and the version bound. Prose updated in `reference/atlas-commands.md`, `versioned/checkpoints.md`, `atlas/docs-coverage.md`, `atlas/comparison.md`, `atlas/migrate-commands.md`, `reference/native-commands.md`, plus `docs/public_api.md` and the regenerated `docs/public_api.snapshot` for the four new exported symbols.

## What a conformance scenario should pin

Not pushed to stokaro/ptah-atlas-conformance. Suggested ce-gating scenario: `migrate checkpoint` aborts at exit 1 **while** `--frobnicate-nonsense` returns `unknown flag` and `migrate frobnicate` returns group help at exit 0 — the triple that establishes "registered, gated, zero own flags". Second, cli-surface: CE honors `-- atlas:checkpoint` on read despite gating the verb, which is what licenses using the pinned binary as the oracle at all. Third, now that it is known: CE's reader is **shallow**, so a nested migration is invisible to it — the asymmetry that produced this PR's blocker.

Refs #951, #954, #984, #992, #994, #1013.


denisvmedia:
Review addressed in `abffeaf4` + `937f9b6e`. Every finding was reproduced before being fixed — the blocker reproduced exactly as described.

## BLOCKER — fixed

Reproduced verbatim on `m/20250801000001_init.sql` + `m/sub/29990101000000_future.sql`:

| | before | after |
| --- | --- | --- |
| checkpoint version | `20260802134121` (below the nested file) | `29990101000001` |
| body contains `future_tbl` | yes | yes |
| `ptah-compat migrate apply`, fresh DB | **exit 1**, `table future_tbl already exists`, dirty row `29990101000000\|2\|0\|failed…` | exit 0, one clean row |
| pinned CE, same directory | exit 0, one clean row | exit 0, one clean row |

Both sharpening points confirmed: the ptah path handled the same shape correctly all along (`sub/0009999999_future.up.sql` → `0010000000`), and CE's shallow reader makes this a **we-fail-where-CE-succeeds** divergence — the direction the PR body called impossible on this verb. That claim was true of `checkpoint` and false of the artifact it writes; the body is corrected.

Fixed at the call site with the recursive `latest` already in hand. `ResolveAtlasCheckpointVersion` keeps its signature and now documents that its scan is deliberately shallow, matching Atlas's own reader, and that supplying the bound is the caller's job — so no breaking change is deferred.

**One thing the suggested fix did not cover.** `if resolved <= latest use latest+1` was reproduced as a mutation and it survived: with `>=` instead of `>`, nothing reddened. On every fixture the two operands differ, so no test could separate strict from non-strict. The separating input needs them to coincide — top-level history at `29990101000000` makes the shallow resolver return `29990101000001`, and a nested migration at exactly that version makes the bound equal it. Accepting the tie writes a checkpoint sharing a version with an existing migration. `TestCompatCommand_MigrateCheckpointNestedTieDoesNotCollide` pins it.

## MAJOR (integrity guard) — fixed, content-shaped

`checkPtahFileConflict` discovers ptah-format files and refuses when any exist and the target is atlas, hashed or not. Measured: unhashed ptah directory now refused both unflagged and explicit; pure-Atlas directory unaffected; `--dir-format ptah` unaffected.

The converse is **deliberately not symmetric**, and I want the reasoning visible in case you disagree: the Atlas name grammar also accepts ptah's `NNNNNNNNNN_name.up.sql`, so probing Atlas-shaped files under `--dir-format=ptah` matches every ptah directory and would refuse the format's ordinary use. A mutation doing exactly that is in the sweep and reddens seven tests.

Both override channels are now measured and pinned as correct: `PTAH_DIR_FORMAT=ptah` and `atlas.hcl` `migration { format = "ptah" }` via `--env` each select the ptah pair.

## MAJOR (`?format=` scope) — reworded, and **#1013 filed**

You were right that "filed separately" was inaccurate. #1002 covers `status`/`set`; #990 covers empty `--dir-format`, non-`format` keys and repeated keys on `hash`/`validate` only. **#1013** now covers `?format=` on `lint`/`new`/`diff` and unknown-key intolerance on every verb, with the measured golang-migrate outcome flip (`L2: … was added` exit 1 → `Migration Status: PENDING` exit 0), the `?format=totally-bogus` validation, and the `?nonsense=1` exit-0 divergence.

The reason for checkpoint's refusal is rewritten to say CE aborts every invocation and so has no contract here — no longer lumping the one gated verb together with five free ones. `migrate diff` added to the refusing list (measured: six verbs reject, three accept).

## MINORs — all four fixed

- **D4** — the `os.Remove` rollback is now reached by making `atlas.sum` a directory, so the atomic replace fails after the checkpoint is written. The test asserts the checkpoint is **gone**, not merely that an error came back.
- **D3** — a non-`ErrNotExist` stat error surfaces. Triggered deterministically with `ENOTDIR` (a regular file as `--migrations-dir`), which Go does not fold into `fs.ErrNotExist`. I first wrote this with `chmod 000` + a root check, but the `t.Skip` tripped the teststyle baseline — `ENOTDIR` needs no conditional, so **no waiver was added**.
- **D5** — the width guard measures `strconv.FormatInt(version, 10)`, not the flag text. `--version 01234567890` is now rejected; previously it wrote `1234567890_checkpoint.sql`, the exact name the guard exists to prevent.
- **D7** — `--dir-format ATLAS` case folding is held by a test.

## Docs

"bumped past the newest existing migration" corrected in both pages to describe the maximum of the timestamp and the recursive bound, with the nesting reason stated. The `--dir-format=auto` row is identified as a **native**-surface message; on compat `auto` is not an Atlas format and draws the ordinary unknown-format rejection (both measured).

## Mutation table — 30/30 killed

Up from 22. Three rows needed repair after the first re-sweep, all reported rather than quietly fixed: one SURVIVED (the `>=` boundary above), one BUILD-FAILED (`case false:` orphaned the `errors` import once `errors.Is` became its only use), one PATCH-FAILED (the newest-plus-one anchor stopped matching once the bound was introduced). Each now kills, via exactly the test written for it.

New rows: second-integrity-file guard disabled · guard protects only ptah / only atlas / also refuses its own sum · **unreadable directory treated as conflict-free** · **ptah-content guard disabled** · **fires on any format** · **probes the atlas format** · **width guard measures flag text** · **timestamp trusted without the bound** · **bound applied with the wrong comparison** · **checkpoint survives a failed atlas.sum write**.

## Gates

`go vet ./...`, `go test ./... -count=1`, `golangci-lint run` (0 issues), `go tool qtlint ./...`, `scripts/check-test-style.sh` — **baseline untouched** — plus the manual testify grep (#975; `line 7: rg: command not found` again, no matches by hand), all three public-API scripts, every docs checker, and `check-responsive` after `npm ci && npm run build`.

Worth knowing for the next agent: `golangci-lint` initially reported 5 revive + 1 staticcheck + 1 unused that were **entirely stale cache**, keyed to a worktree path that no longer exists. After `golangci-lint cache clean` it reported one real `lll` issue in my new error string, now fixed. Phantom issues pointing at deleted paths should not be believed before clearing the cache.

Issue #1014, opened 2026-08-02
atlas.hcl `check "migrate_apply" { drift }`: CE accepts and silently ignores unknown top-level blocks, ptah-compat rejects them with exit 1


denisvmedia:
Measured 2026-08-02. Oracle: **Atlas CE v1.3.0**, community build from the `v1.3.0` tag (`atlas community version v1.3.0`), built the same way `make atlas` builds the pinned binary. Control oracle: pinned Atlas CE v1.2.0 (`~/Work/denis/ptah-atlas-conformance/bin/atlas`). Ptah side: `ptah-compat` built from master `598aa64a`.

Atlas v1.3.0 announces **Pre-Apply Drift Detection** — "Block a deployment when the database drifted from the expected state, with the diff printed before any statement runs" — configured with a new `atlas.hcl` block ([documented](https://atlasgo.io/blog/2026/07/30/atlas-v1-3), retrieved 2026-08-02):

```hcl
check "migrate_apply" {
  drift {
    on_error = FAIL
  }
}
```

**This issue carries two distinct obligations. They are not the same obligation and should not be closed by the same change.**

- **Obligation A — accept the construct.** A CE drop-in obligation. Mandatory.
- **Obligation B — implement drift detection.** A Pro best-effort capability. Optional, and separable.

---

# Obligation A — accept unknown `atlas.hcl` constructs (CE obligation, mandatory)

CE **accepts the block and silently ignores it**. `ptah-compat` **rejects the whole config**.

```text
$ cat drift.hcl
check "migrate_apply" {
  drift {
    on_error = "FAIL"
  }
}
env "local" {
  url = "sqlite://test.db?_fk=1"
}

$ atlas migrate apply -c file://drift.hcl --env local --dir file://mig     # CE v1.3.0
Migrating to version 20260101000002 from 20260101000001 (1 migrations in total):
  -- migrating version 20260101000002
    -> CREATE TABLE orders (id integer PRIMARY KEY);
  -- ok (497.791µs)
[exit 0]

$ ptah-compat migrate apply -c file://drift.hcl --env local --dir file://mig
error: unsupported atlas.hcl construct "check" at drift.hcl:1
[exit 1]
```

This is not specific to `check`. CE ignores **any** unrecognised top-level block; the nonsense control proves it is a general policy, not a `check`-shaped allowance:

```text
$ cat ctrl.hcl
frobnicate_nonsense "zzz" {
  totally_made_up = "yes"
}
env "local" {
[...]

PR #1036, opened 2026-08-02
feat(atlas): read Atlas-format .test.hcl cases on schema test and migrate test

denisvmedia:
Closes the last open item of the drop-in umbrella #951.

## The defect

A directory of Atlas `.test.hcl` files reported `error: no test cases found`. For someone porting an Atlas pipeline that is the worst shape of failure: their entire test suite does not run, and the message says there was nothing to run rather than that we could not read it.

```
$ ptah-compat schema test --dev-url sqlite://... t     # t/ holds basic.test.hcl
error: no test cases found in t                        # before
```

## The translation

```hcl
test "schema" "users_insert_select" {
  exec { sql = "INSERT INTO users (id, name) VALUES (1, 'ada')" }
  exec { sql = "SELECT name FROM users WHERE id = 1", output = "ada" }
}
```

`exec { sql }` is a statement · `exec { sql, output }` is an assertion on the first scalar · `migrate { to }` is a version step.

## Four decisions that are deliberate, not incidental

**Step order is preserved by walking `Body.Blocks`.** A schema-driven `hcl` decode groups blocks by type, which would move a `migrate` step after both `exec` steps and silently change what the case does. Order is load-bearing here: an `exec` that seeds a row must run before the one reading it back. Pinned by `TestParseAtlasTestCasesPreservesStepOrder` with an interleaved fixture — the only shape that separates the two implementations.

**Cases are selected by kind.** Atlas labels each case `test "schema" "..."` or `test "migrate" "..."` and they are not interchangeable — a migrate case drives the migration directory to a version, which a schema test run must not do.

**The exported `LoadCases` keeps its YAML-only contract.** The Atlas reader is reachable through a new `LoadCasesOfKind`. Widening the existing function would have been a public API change for no benefit; its unchanged behavior is pinned by its own test, and the snapshot diff is **22 additions, 0 removals**, so `check-public-api-released.sh` exits 0.

**An unknown attribute inside a step is refused, not ignored.** This surface is our own ingestion, not `atlas.hcl`: a typo in `output` would downgrade an assertion to a bare statement and the case would pass while checking nothing. That is the opposite call from #1014, and for the opposite reason.

## Verification

End to end, **in both directions** — a passing case proves the reader ran, only a failing one proves the assertion is real:

```
matching output    PASS  case "sanity" / scalar "1"                    exit 0
mismatched output  FAIL  expected scalar "999", got "1"                exit 1
```

Three mutations, each killed by exactly one test:

| Mutation | Killed by |
| --- | --- |
| `output` no longer maps to an assertion | `…TranslatesExecAndOutput` |
| match on `filepath.Ext` instead of the compound `.test.hcl` suffix | `…ReadsAtlasFilesAndLeavesSchemaHCLAlone` |
| kind filter dropped | `…SelectsByKind` |

The suffix mutation matters because `filepath.Ext` yields `.hcl` for a *schema* file too — the fixture puts `schema.hcl` and `schema.test.hcl` in one directory, which is the only layout that tells the two rules apart.

```
go build ./...                                  BUILD=0
go vet ./...                                    VET=0
go test ./migration/... ./cmd/... -count=1      TEST=0
scripts/check-public-api.sh                     API_EXIT=0
scripts/check-public-api-snapshot.sh            SNAPSHOT_CHECK=0
scripts/check-public-api-released.sh            RELEASED_EXIT=0
build-feature-matrix.mjs --check                OK (159 rows)
testify imports tree-wide                       0
```

`scripts/check-test-style.sh` exits 1, **identically on master** (`MASTER_STYLE=1`): it scans untracked scratch worktrees inside the repo (`.codex/worktress/…`, `.claude/worktrees/…`) and reports 16 766 violations from them. Violations in tracked source: **0**. That the checker walks untracked directories is a gate defect worth its own issue — anyone with a scratch worktree in the repo gets a red check unrelated to their change.

Refs #951.

### DIFF
--- a/cmd/migrationstest/migrationstest.go
+++ b/cmd/migrationstest/migrationstest.go
@@ -102,7 +102,7 @@ func run(ctx context.Context, out io.Writer, opts options) error {
 		return err
 	}
 
-	cases, err := dbtest.LoadCases(opts.dir)
+	cases, err := dbtest.LoadCasesOfKind(opts.dir, dbtest.AtlasTestKindMigrate)
 	if err != nil {
 		return fmt.Errorf("failed to load test cases: %w", err)
 	}
--- a/cmd/schema/test.go
+++ b/cmd/schema/test.go
@@ -89,7 +89,7 @@ func runSchemaTest(ctx context.Context, out io.Writer, opts testOptions) error {
 		return fmt.Errorf("unsupported report format %q: want text, json, or html", opts.report)
 	}
 
-	cases, err := dbtest.LoadCases(opts.dir)
+	cases, err := dbtest.LoadCasesOfKind(opts.dir, dbtest.AtlasTestKindSchema)
 	if err != nil {
 		return fmt.Errorf("failed to load test cases: %w", err)
 	}
--- a/docs/public_api.snapshot
+++ b/docs/public_api.snapshot
@@ -5628,9 +5628,13 @@ type RowUpdate struct {
 ## go.5x5.cz/ptah/migration/dbtest
 
 type Assertion struct{ ... }
+type AtlasTestKind string
+    const AtlasTestKindSchema AtlasTestKind = "schema" ...
 type Case struct{ ... }
     func FilterCases(cases []Case, pattern string) ([]Case, error)
     func LoadCases(dir string) ([]Case, error)
+    func LoadCasesOfKind(dir string, kind AtlasTestKind) ([]Case, error)
+    func ParseAtlasTestCases(data []byte, filename string, kind AtlasTestKind) ([]Case, error)
     func ParseCases(data []byte) ([]Case, error)
 type CaseResult struct{ ... }
 type Options struct{ ... }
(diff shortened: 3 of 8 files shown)


denisvmedia:
## Second review pass

Three things the first pass had not established. Two came back clean; one is a real gap, and it is pre-existing rather than introduced here.

**`migrate test` end to end — was never actually run.** The first pass verified `schema test` only and reasoned that the other verb "goes through the same loader". It does, but that was an argument, not a measurement. Run:

```
$ ptah-compat migrate test --dir file://migrations --dev-url sqlite://... t
PASS  case "users_table_exists"
    PASS  step "(unnamed step)" — migrated to 20240101000000
    PASS  step "(unnamed step)" — scalar "1"
exit 0
```

Both step kinds execute, including the `migrate { to }` translation that only this verb reaches.

**`attr.Expr.Value(nil)` cannot panic on a reference.** The reader evaluates with a nil `hcl.EvalContext`, which is the kind of thing that panics on an expression the author did not anticipate. Probed with a variable reference, a non-string, and an arithmetic expression — all three return errors, none panics:

```
sql = var.something   -> Variables not allowed; Variables may not be used here.
sql = 42              -> sql must be a string
sql = "a" + "b"       -> Invalid operand; Unsuitable value for left operand
```

**Duplicate case names across files are accepted silently — pre-existing, both paths.** A directory holding `a.yaml` with `name: dup` and `b.test.hcl` with `test "schema" "dup"` loads two cases both named `dup`, no error. Each parser validates only the set it produced; nothing validates the union.

The same is true of the YAML-only path that predates this change:

```
LoadCases(dir with a.yaml + b.yaml, both "dup")  ->  err=<nil>, 2 cases named "dup"
```

So this PR inherits the gap rather than creating it, and fixing it only in `LoadCasesOfKind` would leave the two loaders disagreeing about what is valid. Filed separately as #1038 instead of widening this change.

Worth saying that this reader makes the collision *more likely*, though: two formats can now live in one directory, and "converted the YAML case to HCL, left the old file behind" is an ordinary thing to do. The failure is quiet — `--run dup` matches both, the report shows two rows with the same name, and nothing says why.

PR #1039, opened 2026-08-02
refactor(atlas): split the overloaded atlas.hcl error into three classes

denisvmedia:
Step 1 of #1014. **No tolerance is introduced here** — this is the change that makes step 2 verifiable.

## The problem

`config/projectconfig/atlas.go` funnelled **79 call sites** into a single `unsupported()` helper, and the type-coercion helpers routed evaluation failures and type errors through it too:

```go
func (p atlasParser) stringAttr(name string, attr *hclsyntax.Attribute) (string, error) {
    value, diags := attr.Expr.Value(p.ctx)
    if diags.HasErrors() || value.Type() != cty.String {
        return "", unsupportedAttr(name, attr)     // three failures, one message
    }
```

So `env { url = var.nope }`, `env { url = 42 }` and a genuinely unknown top-level name all produced the byte-identical `unsupported atlas.hcl construct "..."`.

## Why this has to land before any relaxation

Measured against the pinned CE binary: on the 17 positions where **CE exits 1**, `ptah-compat` also exits 1 — 17/17. But only **one** of those agreements is by mechanism. The other sixteen agree because compat refuses on the construct **name** before evaluating anything, while CE fails because it **evaluated** the body and the expression inside did not resolve.

While the three share a message, no test can tell those apart. A relaxation applied first would convert sixteen real agreements into coincidental ones — **the parity suite would go greener while going wronger**.

## What changed

Nine coercion helpers now report their own class. Exit codes are unchanged; all three still fail.

| Helper | Reports |
| --- | --- |
| `evaluationFailed` | carries the HCL diagnostic, which names the offending sub-expression |
| `wrongValueType` | names what was wanted |
| `emptyValue` | a known key given nothing usable |

**Two helpers deliberately keep ignoring `diags`.** `identifierOrStringAttr` and `scopedEnumOrStringAttr` accept a bare identifier, and a bare enum identifier such as `file` *always* fails evaluation — that is the intended path there. Reporting an evaluation failure would reject valid input. Only their wrong-value branches moved.

## The messages got markedly better, and not by accident

A shared string could not say anything specific. Once each class has its own, each can:

```diff
-unsupported atlas.hcl construct "format" at atlas.hcl:3
+atlas.hcl "format" at atlas.hcl:3 must be one of atlas, golang-migrate, goose, flyway, liquibase, dbmate

-unsupported atlas.hcl construct "url" at atlas.hcl:2
+cannot evaluate atlas.hcl "url" at atlas.hcl:2: Call to function "file" failed: path escapes atlas.hcl directory: ../secret.txt
```

That second one matters beyond readability: the `file()`/`fileset()` sandbox diagnostics — `path escapes atlas.hcl directory`, `absolute paths are not supported`, `path escapes from parent` — are **security-relevant** and were entirely hidden behind the word "unsupported".

## The failing tests were the point

Ten existing assertions broke, and every one of them had been asserting "unsupported construct" on what is really an evaluation failure or a wrong value. Each is now updated to assert its actual class, so the test says what it tests.

`TestAtlasParserDistinguishesFailureClasses` pins the invariant directly, and a mutation that merges the classes back into one message is killed by **ten** tests.

## Verification

```
go build ./...                        BUILD=0
go vet ./...                          VET=0
go test ./... -count=1                TEST=0
mutation: re-merge the three classes  10 tests fail; restored -> 0
```

Completeness was checked rather than assumed — the first attempt at this split covered two helpers of nine and looked done. `awk` over the coercion range now reports **zero** remaining collapsed sites; the 66 surviving `unsupportedAttr`/`unsupportedBlock` calls are the genuine unknown-name surface, which is exactly what step 2 will relax.

### DIFF
--- a/config/projectconfig/atlas.go
+++ b/config/projectconfig/atlas.go
@@ -123,6 +123,17 @@ type atlasParser struct {
 	baseDir string
 	// externalSchemas holds the declared data.external_schema sources by name.
 	externalSchemas map[string]externalSchemaDataSource
+	// sensitiveValues holds the resolved values of variables declared
+	// `sensitive = true`, so they can be scrubbed from any diagnostic before it
+	// reaches stderr. HCL renders function arguments into its own error text --
+	// `file(var.secret)` produces `openat <the secret>: no such file` -- so a
+	// diagnostic that is merely passed through leaks exactly what `sensitive`
+	// exists to protect.
+	//
+	// It is a pointer for the same reason ctx is: the parser is passed by
+	// value, so a plain slice field would be appended to on a copy and the
+	// values would never reach the scrubber.
+	sensitiveValues *[]string
 }
 
 func newAtlasParser(fsys fs.FS, rawVars []string, filename string) (atlasParser, error) {
@@ -131,6 +142,7 @@ func newAtlasParser(fsys fs.FS, rawVars []string, filename string) (atlasParser,
 		return atlasParser{}, err
 	}
 	return atlasParser{
+		sensitiveValues: &[]string{},
 		ctx: &hcl.EvalContext{
 			Variables: map[string]cty.Value{},
 			Functions: map[string]function.Function{
@@ -892,6 +904,12 @@ func (p atlasParser) configureVariables(blocks []*hclsyntax.Block) error {
 		if err != nil {
 			return err
 		}
+		if variable.sensitive && p.sensitiveValues != nil &&
+			value.Type() == cty.String && !value.IsNull() {
+			if raw := value.AsString(); raw != "" {
+				*p.sensitiveValues = append(*p.sensitiveValues, raw)
+			}
+		}
 		vars[name] = value
 	}
 	for name, value := range p.varOverride {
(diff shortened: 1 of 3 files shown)


denisvmedia:
## Second review pass found a leak this PR introduced — fixed before merge

CI was green on the previous SHA. The pass that mattered asked a different question: *what does this change now expose that it did not before?*

Carrying the HCL diagnostic through is the whole benefit of the split — it names the offending sub-expression, and for `file()`/`fileset()` it reports the sandbox violation. But **HCL renders evaluated arguments into its own error text**:

```
this PR (before the fix):  Call to function "file" failed: openat SUPERSECRET_TOKEN_12345: no such file or directory
master:                    unsupported atlas.hcl construct "url" at atlas.hcl:7
```

`file(var.secret)` published the value of a `sensitive = true` variable on **stderr**, where it lands in CI logs. Master did not leak it, only because everything was hidden behind one uninformative message.

This is a regression against an invariant the file already states out loud — `redactedAtlasVariableValue` returns `"(sensitive value)"`, and the default-mismatch error carries the comment *"The default value stays out of the message: for sensitive variables it must not leak."* The split honoured that for our own messages and broke it for HCL's.

### Fix

The parser records the resolved values of `sensitive = true` variables and scrubs them from any diagnostic before it is emitted:

```
after:  openat (sensitive value): no such file or directory
```

The diagnostic stays useful; the value is gone. The field is a pointer for the same reason `ctx` is — the parser is passed by value, so a plain slice would be appended to on a copy and never reach the scrubber.

### The test asserts both directions, and the second half is not decoration

- a `sensitive` variable's value is **replaced**;
- a **non**-sensitive variable's value is **preserved**, because that is what makes the message actionable.

Without the control, a scrubber that redacted everything would pass the first assertion while destroying every diagnostic in the package. Mutations confirm both halves do work:

| Mutation | Result |
| --- | --- |
| scrubber disabled | **1** test fails (the leak test) |
| scrub everything | **8** tests fail (the control plus the diagnostic-content assertions) |
| merge the three error classes back | **10** tests fail |

```
go vet ./...            VET=0
go test ./... -count=1  TEST=0
```

Worth naming the shape, because it is not specific to this change: **making an error more informative is a disclosure change.** Anything that starts surfacing text from a layer that previously got swallowed needs asking what that layer puts in its strings. Here the answer was "the arguments you passed it".

PR #1040, opened 2026-08-02
feat(atlas): tolerate unknown top-level atlas.hcl names, as Atlas CE does

denisvmedia:
Step 2 of #1014, **scoped to the top level**. Nested positions (inside `env`, `lint`, `diff`, `locals`) still refuse and remain a known divergence — this is deliberately one increment, not the whole obligation, and saying so is the point: partial tolerance is the kind of thing that looks finished.

## The defect

An `atlas.hcl` carrying the v1.3 drift construct made `ptah-compat` exit 1 before a single migration ran, while CE accepts the file and proceeds:

```
$ ptah-compat migrate apply -c file://drift.hcl --env local
error: unsupported atlas.hcl construct "check" at drift.hcl:1
```

Under the CE drop-in rule that is a defect, not a divergence.

## Three things measured, each contradicting the obvious reading

**Unknown *attributes* are tolerated too, not only blocks.** `frobnicate = "yes"` is exit 0 with 0 bytes on stderr. This is the question the issue left open, and a fix scoped to blocks would still diverge.

**The body of a tolerated construct is still evaluated** — tolerance is name-level, not subtree-level:

```
frobnicate { v = "literal" }         -> exit 0
frobnicate { v = var.undefined_ref } -> exit 1, "Unsupported attribute"
```

Skipping the subtree would accept a file CE **rejects**.

**`atlas` is the one top-level name CE knows and refuses**, rather than not recognizing:

```
cloud { token = "t" }   -> CE exit 0   (unknown, tolerated)
atlas { cloud { … } }   -> CE exit 1   "atlas block is not supported by the community version of Atlas."
```

Measured across nine candidates — `cloud`, `docker`, `remote_dir`, `project`, `check`, `exporter`, `script`, `test` all tolerated; only `atlas` gated, and its bare and labelled spellings fail differently. **The first version of this change folded it into the tolerated path and made compat exit 0 on a config CE rejects** — looser than CE, the more dangerous direction. Caught by measuring rather than by reading, and the tests now prevent it.

## Reporting

Tolerated constructs are recorded on `Config.IgnoredConstructs`. CE reports nothing, and that silence is the footgun the issue describes: a typo'd block name does nothing and looks fine. Exit codes, stdout and every error message stay identical whether or not a caller reports these, so drop-in fidelity does not depend on it. Writing the stderr line at the CLI boundary is the next increment.

The assignment lives at `parse()`'s single exit rather than in `parseEnv` — **`Merge` carries only the fields it knows about and silently dropped it upstream.** The value was recorded correctly and read correctly; the plumbing between lost it, which no single step could show.

## Verification

Mutations, each killed:

| Mutation | Tests failing |
| --- | --- |
| skip the subtree instead of evaluating it | 2 |
| fold the gated `atlas` block into the tolerated path | 7 |
| tolerate attributes but not blocks | 5 |

```
go vet ./...                                  VET=0
go test ./... -count=1                        TEST=0
scripts/check-public-api-snapshot.sh          SNAPSHOT=0   (additive: new type + field)
```

Seven positions were compared directly against the pinned CE **v1.3.0** binary — five tolerated, two fatal — and all seven now match.

One measurement note: an early probe reported `CE=1` on the `check` block, which looked like a contradiction. It was my harness collapsing the block onto one line, which HCL rejects for a nested block. Worth checking *what* failed before adjusting the implementation to match it.

Conformance: stokaro/ptah-atlas-conformance#255 already pins the `variable` exception and the name-level rule, including the literal-value discriminator.

### DIFF
--- a/config/projectconfig/atlas.go
+++ b/config/projectconfig/atlas.go
@@ -134,6 +134,29 @@ type atlasParser struct {
 	// value, so a plain slice field would be appended to on a copy and the
 	// values would never reach the scrubber.
 	sensitiveValues *[]string
+	// ignored records the constructs tolerated under Atlas CE's
+	// unknown-name policy, so the caller can report them. Pointer for the same
+	// reason as sensitiveValues: the parser is passed by value.
+	ignored *[]IgnoredAtlasConstruct
+}
+
+// IgnoredAtlasConstruct is an atlas.hcl name that was accepted and not acted
+// on, matching what Atlas CE does with a name it does not recognize.
+//
+// CE reports nothing at all. Ptah records them so the caller can say one line
+// per construct on stderr: CE's silence is a footgun -- a typo'd block name
+// does nothing and looks fine -- and stdout, the exit code, and every error
+// diagnostic stay byte-identical, so nothing that reads those can tell the
+// difference.
+type IgnoredAtlasConstruct struct {
+	// Name is the construct's name as written.
+	Name string
+	// Kind is "block" or "attribute".
+	Kind string
+	// Filename and Line locate it. The position matters because the same name
+	// is tolerated in one place and fatal in another.
+	Filename string
+	Line     int
 }
 
 func newAtlasParser(fsys fs.FS, rawVars []string, filename string) (atlasParser, error) {
@@ -143,6 +166,7 @@ func newAtlasParser(fsys fs.FS, rawVars []string, filename string) (atlasParser,
 	}
 	return atlasParser{
 		sensitiveValues: &[]string{},
+		ignored:         &[]IgnoredAtlasConstruct{},
 		ctx: &hcl.EvalContext{
 			Variables: map[string]cty.Value{},
 			Functions: map[string]function.Function{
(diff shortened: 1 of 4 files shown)


denisvmedia:
## Second review pass

The question this pass asked: **evaluating an ignored block's body is new execution — what does it execute?**

`file()` and `getenv()` have side effects, and master never reached them in an unknown construct because it refused the construct first. So this change genuinely runs code that used to be skipped.

**No program execution is reachable.** `data "external_schema"` runs an external program, but it is collected from *top-level* `data` blocks in `configureEvalContext`; a `data` block nested inside an ignored block is never collected, only walked for attribute expressions. So the new evaluation reaches `file()`, `getenv()` and the stdlib functions, not process spawning.

**`getenv()` matches CE.** Both exit 0.

**`file()` is where it gets interesting**, and it turned up a standing divergence that was previously masked:

| | CE | compat |
| --- | --- | --- |
| `file("local.txt")` | 0 | 0 |
| `file("/etc/passwd")` | **0** | **1** — `absolute paths are not supported` |
| `file("../../../../etc/passwd")` | **0** | **1** — `path escapes atlas.hcl directory` |

CE reads both. Ptah's `file()` is sandboxed to the config directory.

**This is not a regression from this PR.** Measured both ways:

```
frobnicate { v = file("/etc/passwd") }     master: exit 1   this PR: exit 1
env { url = file("/etc/passwd") }          master: exit 1   this PR: exit 1
```

Master refused the unknown *construct* before evaluating; this PR refuses the *sandbox violation*. Same exit code, and the new message is the accurate one. In a known position master already refused, which is what shows the divergence predates this change.

What did change is visibility: the sandbox refusal now surfaces in a position where CE proceeds, so the divergence is no longer hidden behind an unrelated error. Filed as #1042 — it needs a policy decision (keep the sandbox and record the divergence / surface-split / gate on a `PTAH_*` variable), not an implementation one, and it is orthogonal to this PR.

**On the disclosure question** I flagged for this increment: `Config.IgnoredConstructs` carries construct *names* and file positions, both written by the user in their own config. No values, no evaluated content. When the stderr warning lands in the next increment it will print those names, so the question stays answered — but it has to be re-asked then, not assumed carried over.

CI: 15 pass, 1 skipping, on `0d5a785f` — the amended SHA, checked rather than assumed after the force-push.

PR #1044, opened 2026-08-02
fix(atlas): match Atlas's migrate lint diagnostic format on the compat surface

### LINKED ISSUE #1043: migrate lint diagnostic format diverges from Atlas: inline [CODE], no analyzer URL, no suggested-fix section
Measured against the upstream Atlas txtar fixtures vendored in the conformance repo (`third_party/atlas/upstream/internal/integration/testdata/sqlite/cli-migrate-lint-*.txtar`). Under the standing rule — Atlas CE, full drop-in, no exceptions — this is a defect.

## The divergence

Same input, a migration dropping a table:

```
Atlas:
      -- L1: Dropping table "pets" https://atlasgo.io/lint/analyzers#DS102
    -- suggested fix:
      -> Add a pre-migration check to ensure table "pets" is empty before dropping it

ptah-compat:
      -- L1 [DS102]: DROP TABLE permanently deletes the table and every row in it; take a
         verified backup first and consider a rename-and-retire window instead
```

Everything around it already matches — the `Analyzing changes from version …`, `-- analyzing version …`, `-- destructive changes detected:`, `-- ok (…)`, the summary block with `1 version with errors` / `1 schema change` / `1 diagnostic`. It is the diagnostic line itself, plus a missing section.

**Not a stream problem.** Checked first: `ptah-compat migrate lint` writes 463 bytes to **stdout** with an empty stderr, matching Atlas. The divergence is the message shape.

## Three parts, deliberately separated

`internal/atlasreport/migrate_lint_text.go:79` is the whole of the first two:

```go
content := fmt.Sprintf("L%d [%s]: %s", diag.line, diag.code, diag.message)
```

1. **Line format.** Atlas is `L%d: %s`, with the code carried by the URL rather than inline as `[%s]`.
2. **Analyzer URL.** Atlas appends `https://atlasgo.io/lint/analyzers#<CODE>`. Mechanical, and useful on its own — it is how a reader finds out what the code means.
3. **Message text and the `suggested fix` section.** This is the substantial part. Atlas's per-analyzer wording is short and object-specific (`Dropping table "pets"`); Ptah's is prose explaining the risk. Atlas then emits a separate `-- suggested fix:` / `-> …` block, which Ptah has no per-diagnostic field for.

Parts 1 and 2 without part 3 produce output that is neither Ptah's nor Atlas's — structurally closer, still not fixture-matching, and worse to read than either. **They should land together**, which is why this is one issue rather than a quick format patch.

## Definition of done

- The four `cli-migrate-lint-*.txtar` upstream fixtures pass against `ptah-compat`.
- Each analyzer carries both its Atlas-shaped message and its suggested fix, or an explicit record of why it cannot.
- Where Ptah's wording is genuinely more useful — and it often is; "take a verified backup first and consider a rename-and-retire window" says more than "Dropping table" — apply the surface split rather than losing it: `ptah-compat` matches Atlas, native `ptah migrations lint` keeps the fuller text as a documented advantage.

## Why nobody noticed

The conformance tier that runs these fixtures builds ptah from a **pinned module version**, not from a checkout, and the pin was a pseudo-version dated 2026-08-01. Advancing it is what surfaced this. See stokaro/ptah-atlas-conformance#256 — that pin is also now unadvanceable after the move to `go.5x5.cz/ptah`, so this would have stayed invisible indefinitely.

Refs #285, stokaro/ptah-atlas-conformance#256.

denisvmedia:
Closes #1043.

## The defect

```
Atlas:
      -- L1: Dropping table "pets" https://atlasgo.io/lint/analyzers#DS102
    -- suggested fix:
      -> Add a pre-migration check to ensure table "pets" is empty before dropping it

ptah-compat (before):
      -- L1 [DS102]: DROP TABLE permanently deletes the table and every row in it; take a
         verified backup first and consider a rename-and-retire window instead
```

Everything around the diagnostic already matched — the analyzing header, the group label, `-- ok (…)`, the summary counts. It was the diagnostic line and a missing section. Under the CE drop-in rule that is a defect: anything parsing lint output breaks on it, which is what the vendored upstream txtar fixtures exist to represent.

Checked first that it was not a stream problem — `ptah-compat migrate lint` writes to **stdout** with an empty stderr, matching Atlas. The divergence was the message shape.

## Ptah's wording is kept, not replaced

> "DROP TABLE permanently deletes the table and every row in it; take a verified backup first and consider a rename-and-retire window instead"

says considerably more than `Dropping table "pets"`, and losing it would be a real cost. So this is the **surface split**: native `ptah migrations lint` keeps the fuller prose, only the compat renderer speaks Atlas.

```
native  : DS101: DROP TABLE permanently deletes the table and every row in it; take a
          verified backup first and consider a rename-and-retire window instead
compat  : -- L1: Dropping table "pets" https://atlasgo.io/lint/analyzers#DS102
          -- suggested fix:
            -> Add a pre-migration check to ensure table "pets" is empty before dropping it
```

**This was only possible because `Finding.Context.Subjects` carries structured fields** — `Name`, `Parent`, `DataType`. The Atlas wording is reconstructed in `internal/atlasreport` from data rather than by rewriting the analyzers, so neither surface has to know about the other. Had the finding carried only a rendered string, matching Atlas would have meant changing what the analyzers produce, and the split would have cost an architecture change instead of one file.

Where a code has no Atlas wording here, or the subjects it needs are absent, the native message is used unchanged. Being wordier than Atlas is a smaller divergence than inventing a sentence Atlas never prints.

## Verification against the oracle

DS102 and MF103 match the upstream fixtures **byte for byte**, including the wrap position and the nine-space continuation indent:

```
      -- L1: Adding a non-nullable "int" column "c2" will fail in case table "users" is not empty
         https://atlasgo.io/lint/analyzers#MF103
```

That is what makes rewriting the nine test expectations legitimate — the oracle is the fixture, not this code.

**The expectations were rewritten per case, not by search-and-replace.** The old message carried no table name, so all four `DS102` blocks were textually identical, while the new ones differ (`pets`, `users`, `pets`, `users`). A blanket substitution would have put the wrong name in three of them and left the suite green on it.

```
go build ./...           BUILD=0
golangci-lint run ./...  0 issues
go test ./... -count=1   TEST=0
```

## What this unblocks

stokaro/ptah-atlas-conformance#256: those four `cli-migrate-lint-*.txtar` fixtures were the last blocker to moving the conformance module pin onto `go.5x5.cz/ptah`. The pin is currently stuck on a `github.com/stokaro/ptah` pseudo-version dated 2026-08-01 — a path nothing publishes to any more — so every tier has been measuring a build from before the rename, and would have kept doing so indefinitely.

### DIFF
--- a/cmd/atlas/migrate_lint_default_test.go
+++ b/cmd/atlas/migrate_lint_default_test.go
@@ -80,8 +80,9 @@ func TestCompatCommand_MigrateLintDefaultTextDestructive(t *testing.T) {
 			"\n"+
 			"  -- analyzing version 2\n"+
 			"    -- destructive changes detected:\n"+
-			"      -- L1 [DS102]: DROP TABLE permanently deletes the table and every row in it; take a verified backup\n"+
-			"         first and consider a rename-and-retire window instead\n"+
+			"      -- L1: Dropping table \"users\" https://atlasgo.io/lint/analyzers#DS102\n"+
+			"    -- suggested fix:\n"+
+			"      -> Add a pre-migration check to ensure table \"users\" is empty before dropping it\n"+
 			"  -- ok (DUR)\n"+
 			"\n"+
 			"  -------------------------\n"+
@@ -106,8 +107,8 @@ func TestCompatCommand_MigrateLintDefaultTextDataDependentWarning(t *testing.T) 
 			"\n"+
 			"  -- analyzing version 2\n"+
 			"    -- data dependent changes detected:\n"+
-			"      -- L1 [MF103]: adding a NOT NULL column without a DEFAULT fails or blocks on populated tables; add it\n"+
-			"         nullable, backfill, then enforce NOT NULL in a later migration\n"+
+			"      -- L1: Adding a non-nullable \"int\" column \"c2\" will fail in case table \"users\" is not empty\n"+
+			"         https://atlasgo.io/lint/analyzers#MF103\n"+
 			"  -- ok (DUR)\n"+
 			"\n"+
 			"  -------------------------\n"+
@@ -141,8 +142,8 @@ func TestCompatCommand_MigrateLintDefaultTextAddNotNullFixture(t *testing.T) {
 			"\n"+
 			"  -- analyzing version 2\n"+
 			"    -- data dependent changes detected:\n"+
-			"      -- L1 [MF103]: adding a NOT NULL column without a DEFAULT fails or blocks on populated tables; add it\n"+
-			"         nullable, backfill, then enforce NOT NULL in a later migration\n"+
+			"      -- L1: Adding a non-nullable \"int\" column \"c2\" will fail in case table \"users\" is not empty\n"+
+			"         https://atlasgo.io/lint/analyzers#MF103\n"+
 			"  -- ok (DUR)\n"+
 			"\n"+
 			"  -------------------------\n"+
--- a/internal/atlasreport/migrate_lint_atlas_text.go
+++ b/internal/atlasreport/migrate_lint_atlas_text.go
@@ -1,0 +1,57 @@
+package atlasreport
+
+import (
+	"fmt"
+
+	migrationlint "go.5x5.cz/ptah/migration/lint"
+)
+
+// atlasAnalyzerDocsURL is where Atlas points a reader for an analyzer code. The
+// code travels in the fragment rather than inline as `[CODE]`, which is why the
+// compat renderer does not print it twice.
+const atlasAnalyzerDocsURL = "https://atlasgo.io/lint/analyzers#"
+
+// atlasDiagnosticText renders one finding the way Atlas words it, plus the
+// suggested fix Atlas prints under it.
+//
+// Ptah's own message is deliberately different and usually says more -- compare
+// "Dropping table \"pets\"" with "DROP TABLE permanently deletes the table and
+// every row in it; take a verified backup first and consider a rename-and-retire
+// window instead". Losing that would be a real cost, so it is not lost: this is
+// the compat surface only, and native `ptah migrations lint` keeps the fuller
+// prose. That is the surface split, applied where it belongs.
+//
+// It is reconstructed from the finding's STRUCTURED subjects rather than by
+// rewriting the analyzers, so neither surface has to know about the other. When
+// a code has no Atlas wording here, or the subjects it needs are absent, the
+// native message is used unchanged -- being wordier than Atlas is a smaller
+// divergence than inventing a sentence Atlas never prints.
+func atlasDiagnosticText(code string, finding migrationlint.Finding) (message, fix string) {
+	subject := atlasPrimarySubject(finding)
+	switch code {
+	case "DS102":
+		if subject == nil || subject.Name == "" {
+			return finding.Message, ""
+		}
+		return fmt.Sprintf("Dropping table %q", subject.Name),
+			fmt.Sprintf("Add a pre-migration check to ensure table %q is empty before dropping it",
+				subject.Name)
+	case "MF103":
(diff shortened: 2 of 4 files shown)


denisvmedia:
## Second review pass — CI was green and this was not ready to merge

The pass asked one question: **is the layout right, or did it only look right on the fixture I checked?**

It only looked right.

Atlas collects every suggested fix in a group under **one** header at the end, pluralised when there is more than one. It does not print a fix after each diagnostic:

```
Atlas:
      -- L1: Dropping table "pets" https://atlasgo.io/lint/analyzers#DS102
      -- L2: Dropping table "users" https://atlasgo.io/lint/analyzers#DS102
    -- suggested fixes:
      -> Add a pre-migration check to ensure table "pets" is empty before dropping it
      -> Add a pre-migration check to ensure table "users" is empty before dropping it

this PR, first version:
      -- L1: Dropping table "pets" …
    -- suggested fix:
      -> … "pets" …
      -- L2: Dropping table "users" …
    -- suggested fix:
      -> … "users" …
```

**Why every check passed anyway.** With a *single* diagnostic the two layouts are byte-identical — the fixture I verified against could not distinguish them. And the suite's existing two-diagnostic case puts them in **separate versions**, so each lands in its own group with one fix, and they coincide there too. Not one fixture in the suite could tell interleaving from collecting.

That is the rule this project already writes down: *a fixture where both candidate answers coincide is not a test.* I broke it and caught it only on the second pass, by running a case the suite did not contain.

## What closes it

A fixture with two `DROP TABLE` statements in **one** version, so both diagnostics share a group. Both mutations are killed by it and by nothing else:

| Mutation | Result |
| --- | --- |
| print a fix after each diagnostic (the original layout) | 2 tests fail, including the new one |
| drop the plural `suggested fixes:` header | 2 tests fail, including the new one |

Verified against the pinned CE binary in **both** shapes, diffed rather than eyeballed:

```
single diagnostic  : IDENTICAL
two diagnostics    : IDENTICAL
```

```
go build ./...           BUILD=0
golangci-lint run ./...  0 issues
go test ./... -count=1   TEST=0
```

## Also checked

- **Subject kinds.** `atlasPrimarySubject` takes `Subjects[0]`. Confirmed the analyzers build homogeneous slices — `DS101` emits all `SubjectTable`, the not-null rule all `SubjectColumn` — so index 0 always carries the right kind for the code. Not relying on that would be better, but it is not currently a defect.
- **Native output is untouched.** `ptah migrations lint` still prints `DS101: DROP TABLE permanently deletes the table and every row in it; take a verified backup first and consider a rename-and-retire window instead`. The split holds.

denisvmedia:
## Verified against the fixtures this was meant to unblock — 3 of 4, not 4 of 4

Ran the upstream txtar lint fixtures against a **build of this branch** rather than against the conformance module pin, since the pin is on `v0.2.0` and predates the fix — a green result there would have proved nothing:

```
PTAH_COMPAT_BIN=<this branch> go test ./internal/probe/ -run TestTxtarScriptProbeExecutesSQLiteMigrateLintFixtures

  cli-migrate-lint-destructive.txtar    PASS
  cli-migrate-lint-add-notnull.txtar    PASS
  cli-migrate-lint-minimal-env.txtar    PASS
  cli-migrate-lint-ignore.txtar         FAIL
```

**The remaining one is a different feature.** `cli-migrate-lint-ignore.txtar` exercises `-- atlas:nolint` suppression — "Ignore all diagnostics" and "Ignore specific diagnostics" — and expects `-- no diagnostics found` once everything is suppressed. That is suppression parity, not message formatting, and nothing in this PR touches it.

So this closes #1043 as scoped, and it removes three of the four blockers named in stokaro/ptah-atlas-conformance#256. The fourth needs its own work, and I would rather say that than let "the lint fixtures pass now" stand when one does not.

Issue #1074, opened 2026-08-03
Align migrate lint analyzer semantics for renames and multi-target drops


denisvmedia:
## Problem

Atlas CE v1.3.0 and `ptah-compat migrate lint` still disagree on analyzer semantics for valid DDL. This is not a text-rendering problem and must not be hidden in the compatibility renderer.

Measured on 2026-08-03 against the pinned Atlas CE v1.3.0 binary.

### 1. Column rename is classified differently

Input:

```sql
CREATE TABLE users (id int);
ALTER TABLE users RENAME COLUMN id TO oid;
```

Atlas reports DS103 as an error, prints the dropped logical column `id`, suggests a NULL precheck, and exits 1:

```text
-- destructive changes detected:
-- L1: Dropping non-virtual column "id" https://atlasgo.io/lint/analyzers#DS103
-- suggested fix:
-> Add a pre-migration check to ensure column "id" is NULL before dropping it
```

Ptah reports BC101 as a warning and exits 0:

```text
-- diagnostics detected:
-- L1 [BC101]: renames are not backwards compatible: ...
```

### 2. PostgreSQL multi-target DROP TABLE differs

Input against an isolated PostgreSQL database:

```sql
DROP TABLE app."Users", app.audit_log;
```

Atlas CE v1.3.0 emits no diagnostic and exits 0. Ptah analyzes this shape differently. The renderer must not manufacture multiple Atlas diagnostics or silently choose a primary subject to make the output look compatible; the analyzer/parser semantics need an explicit oracle-backed decision.

## Scope

- Add differential conformance fixtures for both inputs.
- Align rename classification, severity, subject identity, suggested fix, and exit code with Atlas.
- Measure and implement the exact PostgreSQL multi-target behavior at the analyzer layer.
- Keep native `ptah migrations lint` free to provide richer prose, but make `ptah-compat migrate lint` match Atlas semantics.
- Preserve declarative black-box tests and add real PostgreSQL coverage for the multi-target case.
- Update the Atlas feature matrix and lint documentation with the resulting support boundary.

## Definition of done

- The two differential fixtures match Atlas CE v1.3.0 without waivers or conformance reclassification.
- Structured lint output and default text output agree on rule, severity, subjects, and exit status.
- Single-target DS102/DS103/MF103 behavior remains unchanged.
- Full tests, race tests for touched packages, qtlint, teststyle, golangci-lint, and conformance regression budget pass.

[...]

Issue #998, opened 2026-08-02
Match Atlas per-file txmode precedence and validation


denisvmedia:
## Problem

Ptah recognizes `-- atlas:txmode none`, but it currently reduces file mode to a Boolean. That is insufficient to reproduce Atlas file-mode precedence and validation across global `--tx-mode file|all|none`.

The feature matrix must not claim complete transaction-mode parity until the behavior is measured against the pinned Atlas oracle and encoded in conformance.

## Scope

- Capture the current Atlas CE behavior for per-file `txmode` directives without using Atlas implementation code.
- Parse an explicit file transaction-mode enum rather than a Boolean.
- Implement precedence between global and per-file modes.
- Validate unsupported values, duplicate directives, and conflicts before revision bookkeeping.
- Cover plain SQL and txtar migration sections independently for up and down.

## Acceptance

- [ ] Conformance records the global `file|all|none` by file directive absent/`file`/`none`/`all` matrix.
- [ ] Unknown, duplicate, misplaced, and conflicting directives have oracle-backed expected diagnostics and exit codes.
- [ ] Ptah matches the measured matrix for plain and txtar migrations.
- [ ] Up and down sections retain independent file modes.
- [ ] Native `-- +ptah no_transaction` behavior remains explicitly documented and tested.
- [ ] The feature matrix reports the gap honestly until every required conformance cell passes.

## Related

- #152
- #274
- #887

PR #1075, opened 2026-08-03
Restore Atlas migrate lint text in ptah-compat

### LINKED ISSUE #1062: migrate lint prose on the compat surface is Ptah's own, by decision
## Decision

`ptah-compat migrate lint` prints **Ptah's own diagnostic prose and layout**. It does not reproduce another tool's wording, its documentation-link form of the rule code, or its trailing collected-fixes header.

Ptah is an independent implementation. Its output is written for its own readers rather than shaped to match someone else's phrasing.

This reverses #1044, which had made the compat renderer mirror that phrasing. That change is reverted in favor of keeping our prose (#1063).

## Why

Our message says more, and says something actionable:

```
L1 [DS102]: DROP TABLE permanently deletes the table and every row in it; take a
   verified backup first and consider a rename-and-retire window instead
```

The terse equivalent is `Dropping table "pets"` plus a link. Trading the first for the second is a downgrade for every person who reads the output.

Three specifics, so the shape is not re-litigated piecemeal:

- **Rule code stays inline** as `L<n> [CODE]:`. It is legible on its own and needs no external link to be understood.
- **Remediation stays with its diagnostic**, not collected under a trailing header away from the thing it is about.
- **No documentation URL** is emitted.

## What it costs, measured rather than assumed

Four imported txtar fixtures assert the other phrasing and therefore diverge. Verified by pointing the conformance module at the working tree rather than at the published pin:

```
cli-migrate-lint-add-notnull.txtar    FAIL
cli-migrate-lint-destructive.txtar    FAIL
cli-migrate-lint-ignore.txtar         FAIL
cli-migrate-lint-minimal-env.txtar    FAIL
```

These are a **deliberate divergence**, not a regression. Anything consuming lint results programmatically should read a structured format rather than scrape prose; that is the answer to the compatibility argument #1044 made.

## Where this is pinned

- `internal/atlasreport/migrate_lint_text.go` — the decision, with its reasoning, on `writeMigrateLintTextGroup`. It says plainly that reshaping this output to match another tool is reverting a decision rather than fixing a defect.
- `TestMigrateLintProseIsOurs` — guards both directions: our prose present, and the terse phrasing, the docs URL and the trailing fixes header each absent. Confirmed to fail when the renderer is mutated back.

## Open

The conformance repo still reports those four fixtures as `ok`, because its pin predates the revert. When the pin moves they will turn red and need reclassifying as a recorded divergence rather than a failure.

denisvmedia:
## Summary

- restore Atlas-compatible default `migrate lint` text on the `ptah-compat` drop-in surface
- retain Ptah's richer diagnostics on native `ptah migrations lint`
- reproduce measured Atlas CE v1.3.0 wording, analyzer links, grouped fixes, identifier/type normalization, and analyzer-specific wrapping for DS102, DS103, and MF103
- label unmeasured Ptah diagnostics explicitly without fabricating Atlas analyzer links
- document the native/compatibility surface split

## Why

Advancing the companion conformance repository to the Ptah commit after #998 exposed 18 failed assertions across four imported Atlas lint txtars. The previous change deliberately put native Ptah prose on the drop-in binary, but `ptah-compat` is the Atlas command replacement and must preserve that contract. Conformance was not weakened or reclassified.

The exact copy and wrap boundaries in this PR were measured against Atlas CE v1.3.0, accounting for the release adopted by stokaro/ptah-atlas-conformance#251.

Semantic analyzer differences found during review are not hidden in the renderer. They are tracked separately in #1074 with differential and real-PostgreSQL acceptance criteria.

## Validation

- `go test ./... -count=1`
- `go test -race ./cmd/atlas ./internal/atlasreport ./cmd/lint -count=1`
- `go tool qtlint ./...`
- `scripts/check-test-style.sh`
- `golangci-lint run ./...`
- `node docs/site/scripts/check-style.mjs`
- `node docs/site/scripts/check-links.mjs`
- `node docs/site/scripts/build-feature-matrix.mjs --check`
- imported Atlas `cli-migrate-lint-*.txtar` conformance fixture test against the locally built `ptah-compat`

Closes #1062.
Refs #1043, #1074, stokaro/ptah-atlas-conformance#251.


denisvmedia:
Never do that again

denisvmedia:
Understood. I opened this PR before the final independent review had completed, and that review then found two real Atlas v1.3 compatibility blockers. That was the wrong order. I will leave this PR closed and will not repeat it: the replacement PR will be opened only after the final commit has passed the complete local, external-conformance, and independent-review checks.

PR #1114, opened 2026-08-03
feat(atlas): accept and drop unknown schema-HCL names, as Atlas CE does (#1016)

### LINKED ISSUE #1016: Schema HCL: CE silently ignores unknown top-level blocks and column attributes (`annotation`, `invisible`), ptah-compat rejects them with exit 1
Measured 2026-08-02. Oracle: **Atlas CE v1.3.0**, community build from the `v1.3.0` tag (`atlas community version v1.3.0`). Control oracle: pinned Atlas CE v1.2.0. Ptah side: `ptah-compat` built from master `598aa64a`. Dev database: local `mysql:9.7` container (`--dev-url mysql://root:pass@localhost:13399/test`).

Atlas v1.3.0 announces **Schema Annotations** — "Attach type-safe metadata to schema resources and render GraphQL, OpenAPI, or any other format from it with templates" — and **MySQL invisible columns** ([documented](https://atlasgo.io/blog/2026/07/30/atlas-v1-3), retrieved 2026-08-02).

This is the schema-HCL sibling of #1014 (which covers the `atlas.hcl` project-config layer). Different parser, same divergence.

## Measured

| input | Atlas CE v1.2.0 **and** v1.3.0 | ptah-compat |
| --- | --- | --- |
| `annotation "gql" { attr "name" { type = string } }` top-level block | accepted, silently dropped, exit 0 | `unsupported top-level block "annotation"`, exit 1 |
| column attribute `invisible = true` | accepted, silently dropped, exit 0 | `unsupported column attribute "invisible"`, exit 1 |
| **control:** column attribute `zzz_nonsense_attr = true` | accepted, silently dropped, exit 0 | `unsupported column attribute "zzz_nonsense_attr"`, exit 1 |
| `partition { type = HASH, columns = [...], partitions = 4 }` | `Error: Unknown variable; There is no variable named "HASH"`, exit 1 | `unsupported partition attribute "partitions"`, exit 1 |

Transcripts:

```text
$ cat annot.hcl
schema "test" {}
annotation "gql" {
  attr "name" {
    type = string
  }
}
table "t" {
  schema = schema.test
  column "id" {
    type = int
  }
  annotation {
    gql = "Thing"
  }
}

$ atlas schema diff --from file://empty.hcl --to file://annot.hcl --dev-url "$DEV"   # CE v1.3.0
-- Create "t" table
CREATE TABLE `t` (`id` int NOT NULL) CHARSET utf8mb4 COLLATE utf8mb4_0900_ai_ci;
[exit 0]

$ ptah-compat schema diff --from file://empty.hcl --to file://annot.hcl --dev-url "$DEV"
error: load --to schema: parse HCL schema at .../annot.hcl:2,1-11: unsupported top-level block "annotation"
[exit 1]
```

```text
$ atlas schema diff --from file://empty.hcl --to file://invisible.hcl --dev-url "$DEV"   # CE v1.3.0
-- Create "t" table
CREATE TABLE `t` (`id` int NOT NULL, `secret` int NOT NULL) CHARSET utf8mb4 COLLATE utf8mb4_0900_ai_ci;
[exit 0]

$ ptah-compat schema diff --from file://empty.hcl --to file://invisible.hcl --dev-url "$DEV"
error: load --to schema: parse HCL schema at .../invisible.hcl:7,3-9: unsupported column attribute "invisible"
[exit 1]
```

Note what the CE `invisible` transcript actually proves: the emitted DDL has **no `INVISIBLE` keyword**. CE did not implement invisible columns — it ignored an attribute it does not know, exactly as it ignored `zzz_nonsense_attr`. The nonsense control is what makes that readable; without it, the exit-0 could be misread as support.

## Classification

**CE obligation (the rows above).** Measured against the community binary; CE reads these schemas to completion and emits DDL. Under the standing rule CE's behavior is the contract, so `ptah-compat` must not fail closed where CE proceeds. A schema authored for Atlas Pro annotations still *works* under CE (annotations are simply inert) and must therefore still work under `ptah-compat`.

**Pro best-effort (not in this issue's DoD).** Actually rendering annotations: the `exporter "template" "<name>" { template { src, on "<kind>" { name = "<path template>" } } }` block, the `attr` template function, and `schema inspect --export`. All *documented, not measured*. Measured facts about the surface: `schema inspect --export` is `Error: unknown flag: --export` on **both** CE v1.2.0 and v1.3.0, so `--export` is a Pro-only flag that CE never registered; and the `exporter` block in `atlas.hcl` is silently ignored by CE while `ptah-compat` rejects it (`unsupported atlas.hcl construct "exporter"`) — that rejection belongs to #1014's project-config layer.

**No CE gap:** `partition`. Both binaries exit 1. CE has no `HASH` identifier in scope, so CE v1.3.0 does not accept partition HCL either despite the release note; MySQL partitioning is Pro. `ptah-compat` already recognises a `partition` block and fails further in (on the `partitions` attribute), so Ptah is ahead here, not behind. Recorded for completeness only.

## Definition of done

- `ptah-compat` accepts and ignores unrecognised schema-HCL top-level blocks and unrecognised column attributes, matching CE's exit code and emitted DDL byte for byte.
- The accept-vs-warn decision is taken jointly with #1014 so both HCL layers behave the same way; whatever is chosen, it is the same choice in both parsers.
- Native `ptah` behavior is a separate decision and may stay strict — record the split if the surfaces diverge.
- Conformance fixtures asserting, against the pinned CE binary: (i) `annotation` block → exit 0 and DDL identical to the same schema without it; (ii) `invisible = true` → exit 0 and DDL containing **no** `INVISIBLE`, pinning that CE does not implement it; (iii) the `zzz_nonsense_attr` nonsense-attribute control → exit 0, proving the ignore is a general policy rather than per-attribute support; (iv) `partition`/`HASH` → exit 1, pinning that this one is *not* a parity gap so a future refactor cannot "fix" it into divergence.

Refs #951, #1014, #934.

denisvmedia:
Closes #1016.

`ptah-compat` refused a schema file over a name it does not model — a top-level `annotation` block, an `invisible = true` column attribute, an unknown attribute anywhere — where the community binary reads the same file to completion. This makes it accept and drop those names, and does so without opening a single position where `ptah-compat` exits 0 and the pinned community binary exits 1.

The branch has been rebased onto current master and squashed. Two earlier revisions of it were rejected for exactly that loose direction; this revision replaces the mechanism rather than patching it.

## What was measured

Oracle: `/…/ptah-atlas-conformance/bin/atlas`, `atlas community version v1.3.0`, by absolute path. Dev database: `sqlite://file?mode=memory`. Command in every row: `<binary> schema inspect -u file://<fixture> --dev-url "sqlite://file?mode=memory"`.

### 1. The issue still reproduces on master — this is not superseded

`ptah-compat` built from master `da1c4cd2`:

```
----- atlasCE     / annot.hcl      -> exit 0    (emits table "t" { column "id" { … } })
----- ptahMaster  / annot.hcl      -> exit 1
Error: parse HCL schema at …/annot.hcl:2,1-11: unsupported top-level block "annotation"

----- atlasCE     / invisible.hcl  -> exit 0    (DDL contains NO INVISIBLE keyword)
----- ptahMaster  / invisible.hcl  -> exit 1
Error: parse HCL schema at …/invisible.hcl:7,3-9: unsupported column attribute "invisible"

----- atlasCE     / nonsense.hcl   -> exit 0
----- ptahMaster  / nonsense.hcl   -> exit 1
Error: parse HCL schema at …/nonsense.hcl:4,3-9: unsupported column attribute "zzz_nonsense_attr"
```

The `invisible` row is worth reading twice: the oracle's DDL carries no `INVISIBLE` keyword, so it did not implement the attribute — it ignored one it does not know, exactly as it ignored `zzz_nonsense_attr`. The nonsense control is what makes that readable.

### 2. The branch as it stood was loose in 12 positions

`ptah-compat` built from the branch head `060eafe1`, over the fixtures the review named. `LOOSE` = candidate exits 0, oracle exits 1.

```
fixture                  atlasCE  master   candidate verdict
L1_inner_a               1        1        0        LOOSE     annotation { inner { a = 1 }  ref = inner.a }
L2_inner_typo            1        1        0        LOOSE     ref = inner.typo
L3_attr_nested_typo      1        1        0        LOOSE     ref = attr.name.nested.typo
L4_toplevel_unlabeled    1        1        0        LOOSE     unlabeled top-level block, ref = <name>.typo
L5_index_typo            1        1        0        LOOSE     ref = inner["typo"]
L6_primary_key_nope      1        1        0        LOOSE     ref = primary_key.nope
L7_primary_key_columns   1        1        0        LOOSE     ref = primary_key.columns
L8_partition_typo        1        1        0        LOOSE     ref = partition.typo
L9_var_string_attr       1        1        0        LOOSE     var.v.nope, variable "v" { type = string }
L10_var_list_attr        1        1        0        LOOSE     var.v.typo, type = list(string)
L11_var_string_arith     1        1        0        LOOSE     1 + var.v, type = string
L12_partial_ref          1        1        0        LOOSE     ref = table.p.column
```

The oracle's own diagnostics on those, for the record:

```
L1  Unsupported attribute; This object does not have an attribute named "a".
L5  Invalid index; The given key does not identify an element in this collection value.
L9  Unsupported attribute; Can't access attributes on a primitive-typed value (string).
L11 Invalid operand; Unsuitable value for right operand: a number is required.
L12 invalid reference used in ref
```

### 3. After this change: zero loose positions over 56 fixtures

Same command, `ptah-compat` built from this branch, across four fixture sets — the 12 above, 5 refusal controls, 20 literal/keyword expression probes, 2 block-reference probes, 15 probes aimed specifically at the type-keyword value:

```
LOOSE positions: 0
```

Every one of the 12 now exits 1 on both. Six fixtures are `strict` (Ptah exits 1, oracle exits 0) and each is listed under "what this deliberately does not do" below.

## The mechanism, and why it changed

The two earlier revisions built reference roots from the file being parsed, so that `table.t.column.id`, `attr.name` and `var.v` would resolve inside a dropped body the way they do on the oracle. Both had to represent something they had not enumerated — an unlabeled block, a variable of unknown type — and both used `cty.DynamicVal` for it. An unknown value makes member access, indexing and arithmetic evaluate to *unknown* instead of failing, so the expression passes and the file is accepted. That is the whole of finding 2 above.

The scope is now closed. It holds three names — `string`, `int`, `bool`, the only bare identifiers measured to resolve inside a dropped body on every dialect, and the ones `annotation "gql" { attr "name" { type = string } }` needs — plus the previously measured function table. Each keyword is an opaque `cty` capsule, so member access, conversion, iteration and arithmetic all fail on it. Nothing in the scope comes from the file, so no member is ever guessed and no value in it is unknown. There is no way to reach an unknown result from a scope that contains none, which is what makes this hold for expressions nobody thought to write down.

Every number below is from an applied source mutation with the rest of the tree unchanged; the tree is clean again.

| mutation | red |
| --- | --- |
| `tolerant: false` (tolerance off) | **62 leaf subtests, 12 top-level** across `internal/atlashcl`, `internal/schemafile`, `cmd/atlas` |
| `checkDroppedExpr` returns `nil` (body never evaluated) | **48 leaf subtests, 5 top-level** — every refusal row plus the whole oracle refusal table |
| **`typeKeywordValue = cty.DynamicVal`** (the exact regression class, narrow form) | `TestDroppedBodyScopeHoldsOnlyKnownOpaqueValues/{string,int,bool}` + `…StillEvaluatesTheBody/type_keyword_is_not_a_number` — **4 leaf, 2 top-level** |
| **`cty.DynamicVal` roots for `table`/`column`/`var`/`attr`/`inner`/`primary_key`** (the regression class as it actually shipped) | **32 leaf subtests, 6 top-level**, including 9 rows of `TestOracleRefusesUnevaluableDroppedBodies` and all 4 of `TestOracleAcceptsReferencesPtahRefuses` |
| `int`/`bool` removed from the scope | `TestOracleAcceptsAndDropsUnknownSchemaHCLNames/type_keyword_inside_the_dropped_block`, `…AcceptsEvaluableBodies/{two_type_keywords_compared,type_keyword_as_the_whole_value}` — 3 leaf, 4 top-level |
| scope emptied entirely | 7 leaf, 9 top-level — including the issue's headline `top-level_annotation_block` and both `cmd/atlas` surface tests |
| `IgnoreUnknownHCLNames: true` hard-wired in `internal/atlassource/resolve.go` | `TestSchemaDiffUnknownHCLNamesSplitByCommandTree` |
| `IgnoreUnknownHCLNames: true` hard-wired in `internal/atlasschema/inspect_source.go` | `TestSchemaInspectUnknownHCLNamesSplitByCommandTree` |
| `noteIgnored` made a no-op | `TestParseIgnoreUnknownNamesRecordsWhatItDropped` |

The two hard-wire mutations are separate rows on purpose: `schema inspect` and `schema diff` reach the loader through different resolver calls, and a test on one does not catch the other. That is how the tolerance leaked into the native `ptah schema` commands once already.

`TestDroppedBodyScopeHoldsOnlyKnownOpaqueValues` asserts on the scope rather than on a list of expressions, which is why the narrow `DynamicVal` mutation is caught at all — no expression fixture in the file exercises `1 + string` through a dynamic keyword.

## Conformance gate

`internal/atlashcl/oracle_conformance_test.go` re-measures every claim against the pinned binary: **35 leaf subtests, 0 skips**, up from 19 on the branch head. New coverage is the class the old gate could not see — its nine refusal fixtures all used *labeled* reference roots, so the unlabeled-block and typed-variable wildcards were invisible to it while accepting all twelve loose files.

`TestOracleAcceptsReferencesPtahRefuses` is new and fails in **both** directions: it asserts the oracle exits 0 *and* Ptah refuses, so if the oracle stops accepting one of these the row is no longer a divergence and must move, and if Ptah starts accepting one the trade has been reversed without being re-argued.

`.github/workflows/atlas-oracle.yml` lists the tests before running them and fails on a skip; the expected count moved 3 → 4 for the new test.

## What this deliberately does not do

- **It does not model the file as a scope.** Four measured shapes the oracle accepts at exit 0 are refused here: `ref = table.t` (and `table.t.column.id`), `ref = column.id` inside a table body, `ref = var.v` with the variable declared, `ref = attr.name`. So is a call to a function outside the measured table, and `[for k, v in table.p : k]`. That is the strict direction, it is recorded in `TestOracleAcceptsReferencesPtahRefuses` and in `docs/atlas_hcl_schema.md`, and it is the price of not guessing. Making them resolve is what both previous revisions attempted and is what produced the 12 loose positions.
- **It does not close the `env`-block divergence.** A schema file carrying an `env` block makes the oracle refuse the whole file (`cannot parse project file … as a schema file`); Ptah exits 0. That reproduces on master, is a classification-boundary question between the project-config and schema-file layers, and belongs to its own issue.
- **It does not settle `partition`/`HASH`.** Unquoted `type = HASH` is exit 1 on MySQL and exit 0 on PostgreSQL, which models partitions; a parser that never sees the dialect cannot match both. The dialect-independent half is pinned instead: `HASH` is not a reference root anywhere.
- **It does not turn native `ptah` tolerant.** `Parse`/`ParseFile` stay strict. `ptah schema inspect` still refuses the same file `ptah-compat schema inspect` accepts.
- **It does not keep dead plumbing.** `schemafile.Options.RecordIgnoredHCLName` forwarded to the parser's recorder, had no producer anywhere in the tree and no test — making it a no-op left `internal/schemafile` green. Removed. `atlashcl.Options.RecordIgnored` stays; it has a test that goes red when disabled.

## Gates

| gate | result |
| --- | --- |
| `go build ./...` | exit 0 |
| `go vet ./...` | exit 0 |
| `go test ./... -count=1` | exit 0, 176 packages ok |
| `golangci-lint run ./...` | exit 1 — **both findings are foreign paths**: `../../../../ptah-1101/migration/importer/emit.go:130` and `../wf_57b2b3d8-4aa-2/examples/viz/models/schema.go:3`, two other checkouts parked under the repo. Neither is in this diff. Restricted to the packages this branch touches (`./internal/atlashcl/... ./internal/schemafile/... ./internal/atlasschema/... ./internal/atlassource/... ./internal/atlasmigrate/... ./cmd/atlas/...`): **0 issues**. |
| `scripts/check-test-style.sh` | exit 0 — `teststyle: OK (175 conditional groups, 45 white-box files)`. Went red twice on this branch's own new file (file-name rule, then missing justification) before it went green, so the gate demonstrably ran. |
| `node docs/site/scripts/check-style.mjs` | exit 0 — went red first on `docs/atlas_hcl_schema.md:107` (banned filler "simply"), fixed |
| `node docs/site/scripts/build-feature-matrix.mjs --check` | exit 0 |
| oracle run, `PTAH_ATLAS_ORACLE=…/bin/atlas go test -count=1 -v -run '^TestOracle' ./internal/atlashcl` | exit 0, 35 leaf subtests, 0 skips |

One note on `go test`, stated rather than buried: on two earlier runs made while heavy parallel work was in flight, `cmd/viz/TestSVGReportsGraphvizStderrOnFailure` failed with `render SVG with Graphviz dot: signal: killed` at exactly 10.0s. It reproduced identically on a clean `origin/master` checkout at `c0f86693`, and it passes on a quiet run. It is a load-sensitive timeout in a package this branch does not touch.

### DIFF
--- a/.github/workflows/atlas-oracle.yml
+++ b/.github/workflows/atlas-oracle.yml
@@ -57,6 +57,33 @@ jobs:
           -run '^TestSumFileNamesDifferentialFuzz(RealisticFlyway|OtherFormats)?$'
           ./internal/atlasmigrateimport
 
+      # The schema-HCL unknown-name policy (stokaro/ptah#1016) is the only other
+      # thing in this repo whose contract IS the community binary's behavior. It
+      # is listed before it is run so an empty selection fails the job instead of
+      # reporting a green run that measured nothing.
+      - name: Verify schema-HCL conformance selection
+        run: |
+          GOWORK=off go test -list '^TestOracle' \
+            ./internal/atlashcl > "$RUNNER_TEMP/atlas-schema-hcl-tests"
+          cat "$RUNNER_TEMP/atlas-schema-hcl-tests"
+          grep -Fx 'TestOracleAcceptsAndDropsUnknownSchemaHCLNames' "$RUNNER_TEMP/atlas-schema-hcl-tests"
+          grep -Fx 'TestOracleRefusesUnevaluableDroppedBodies' "$RUNNER_TEMP/atlas-schema-hcl-tests"
+          grep -Fx 'TestOracleAcceptsReferencesPtahRefuses' "$RUNNER_TEMP/atlas-schema-hcl-tests"
+          grep -Fx 'TestOraclePartitionIsDialectSplit' "$RUNNER_TEMP/atlas-schema-hcl-tests"
+          test "$(grep -c '^TestOracle' "$RUNNER_TEMP/atlas-schema-hcl-tests")" -eq 4
+
+      - name: Run schema-HCL conformance tests
+        run: |
+          GOWORK=off go test -count=1 -v -run '^TestOracle' ./internal/atlashcl \
+            > "$RUNNER_TEMP/atlas-schema-hcl-run" 2>&1 || {
+              cat "$RUNNER_TEMP/atlas-schema-hcl-run"
+              exit 1
+            }
+          cat "$RUNNER_TEMP/atlas-schema-hcl-run"
+          # A missing oracle skips these silently. That is the exact failure
+          # mode this job exists to prevent, so a skip fails the job.
+          ! grep -q 'SKIPPED' "$RUNNER_TEMP/atlas-schema-hcl-run"
+
       - name: Regenerate Atlas CE corpus
         run: >-
           internal/atlasmigrateimport/testdata/ce-sums/regenerate.sh
--- a/cmd/atlas/schema_apply.go
+++ b/cmd/atlas/schema_apply.go
@@ -226,6 +226,10 @@ func runAtlasSchemaApply(cmd *cobra.Command, opts atlasSchemaApplyOptions) error
 		TxMode:     txMode,
 		DryRun:     opts.dryRun,
 		ProjectEnv: projectEnv,
+
+		// Atlas-compatible surface: a schema file written for another tool
+		// must not be refused over a name this parser does not model.
+		IgnoreUnknownHCLNames: true,
 	})
 	if err != nil {
 		return cmdutil.Fail(cmd, err)
@@ -372,7 +376,10 @@ func runAtlasSchemaApplyPlanFile(cmd *cobra.Command, opts atlasSchemaApplyOption
 
 	var desired *goschema.Database
 	if len(opts.toURLs) > 0 {
-		desired, err = schemafile.LoadAll(opts.toURLs, schemafile.Options{Dialect: conn.Info().Dialect})
+		desired, err = schemafile.LoadAll(opts.toURLs, schemafile.Options{
+			Dialect:               conn.Info().Dialect,
+			IgnoreUnknownHCLNames: true,
+		})
 		if err != nil {
 			return cmdutil.Fail(cmd, fmt.Errorf("load --to schema: %w", err))
 		}
--- a/cmd/atlas/schema_diff.go
+++ b/cmd/atlas/schema_diff.go
@@ -130,6 +130,9 @@ func runAtlasSchemaDiff(cmd *cobra.Command, opts atlasSchemaDiffOptions) error {
 		Include:    opts.include,
 		Policy:     policy,
 		ProjectEnv: projectEnv,
+
+		// Atlas-compatible surface; see cmd/atlas/schema_apply.go.
+		IgnoreUnknownHCLNames: true,
 	})
 	if err != nil {
 		return cmdutil.Fail(cmd, err)
--- a/cmd/atlas/schema_inspect.go
+++ b/cmd/atlas/schema_inspect.go
@@ -131,6 +131,9 @@ func runAtlasSchemaInspect(cmd *cobra.Command, opts atlasSchemaInspectOptions) e
 		Diagnostics:    cmd.ErrOrStderr(),
 		ProjectEnv:     projectEnv,
 		ConnectTimeout: dbcli.DefaultConnectTimeout,
+
+		// Atlas-compatible surface; see cmd/atlas/schema_apply.go.
+		IgnoreUnknownHCLNames: true,
 	})
 	if err != nil {
 		return cmdutil.Fail(cmd, err)
--- a/cmd/atlas/schema_plan.go
+++ b/cmd/atlas/schema_plan.go
@@ -171,6 +171,9 @@ func runAtlasSchemaPlan(cmd *cobra.Command, opts atlasSchemaPlanOptions) error {
 		ToURLs:  opts.toURLs,
 		Exclude: opts.exclude,
 		Policy:  policy,
+
+		// Atlas-compatible surface; see cmd/atlas/schema_apply.go.
+		IgnoreUnknownHCLNames: true,
 	})
 	if err != nil {
 		return cmdutil.Fail(cmd, err)
--- a/cmd/atlas/schema_plan_validate.go
+++ b/cmd/atlas/schema_plan_validate.go
@@ -151,7 +151,10 @@ func runAtlasSchemaPlanValidate(cmd *cobra.Command, opts atlasSchemaPlanValidate
 	}
 	defer dbschema.CloseAndWarn(conn)
 
-	desired, err := schemafile.LoadAll(opts.toURLs, schemafile.Options{Dialect: conn.Info().Dialect})
+	desired, err := schemafile.LoadAll(opts.toURLs, schemafile.Options{
+		Dialect:               conn.Info().Dialect,
+		IgnoreUnknownHCLNames: true,
+	})
 	if err != nil {
 		return cmdutil.Fail(cmd, fmt.Errorf("load --to schema: %w", err))
 	}
--- a/cmd/atlas/unknown_hcl_names_surface_test.go
+++ b/cmd/atlas/unknown_hcl_names_surface_test.go
@@ -1,0 +1,117 @@
+package atlas_test
+
+import (
+	"bytes"
+	"os"
+	"path/filepath"
+	"testing"
+
+	qt "github.com/frankban/quicktest"
+
+	"go.5x5.cz/ptah/cmd/atlas"
+	"go.5x5.cz/ptah/cmd/root"
+)
+
+// unknownNameSchemaHCL carries a top-level block and a column attribute that
+// Ptah's schema-HCL parser does not model. The community binary reads this file
+// to completion and emits DDL identical to the same schema without them.
+const unknownNameSchemaHCL = `
+annotation "gql" {
+  attr "name" {
+    type = string
+  }
+}
+schema "main" {
+}
+table "t" {
+  schema = schema.main
+  column "id" {
+    type      = int
+    invisible = true
+  }
+}
+`
+
+// TestSchemaInspectUnknownHCLNamesSplitByCommandTree pins the split
+// stokaro/ptah#1016 left to this change: the SAME file is accepted by the
+// Atlas-compatible command tree and refused by Ptah's own.
+//
+// One fixture, two command trees, opposite verdicts -- which is the only shape
+// that can catch the tolerance leaking from one tree into the other. It leaked
+// in exactly that direction once already: the option was set unconditionally
+// inside the shared resolver, so `ptah schema inspect` silently dropped typos
+// while the documentation said it named them.
+func TestSchemaInspectUnknownHCLNamesSplitByCommandTree(t *testing.T) {
+	c := qt.New(t)
+	dir := t.TempDir()
+	schemaPath := filepath.Join(dir, "schema.hcl")
+	c.Assert(os.WriteFile(schemaPath, []byte(unknownNameSchemaHCL), 0o600), qt.IsNil)
+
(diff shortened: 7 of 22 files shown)


denisvmedia:
Superseded by #1125.

This branch conflicted with master, which has since taken ten commits including #1117. GitHub reported `mergeable_state: dirty`, so it could not be updated and CI never ran on it. A PR's head branch cannot be repointed through the API, so the rebased work is a fresh PR; `fix/issue-1016-tolerate-unknown-schema-hcl-names` is left untouched.

#1125 is stacked on #1124 → #1119 → master, because this change shares `cmd/atlas/schema_apply.go` with the named-lock branch and `schema_diff.go` / `schema_inspect.go` with the schema-side branch. Merge order: **#1119, then #1124, then #1125.**

All four conflicts were pure unions — both sides adding a field to the same struct or literal — so nothing here was dropped or reduced. Every discriminator was re-proved on the rebased tree and reproduces this description's numbers exactly, including the `cty.DynamicVal` regression guard at 4 leaf / 2 top-level and the oracle conformance at 35 leaf subtests with zero skips.

One measurement worth carrying forward: the tolerance mutation reports 48 leaf failures instead of 62 when `PTAH_ATLAS_ORACLE` is unset, because the three `TestOracle*` tests skip silently rather than failing. The sweep still looks healthy while under-reporting coverage by 14 subtests, so that variable has to be set when this branch's coverage is being measured.

The description here remains the reference for the mechanism and the 56-fixture loose/strict measurement.

Issue #979, opened 2026-08-01
`--include` selectors that match nothing: shape check closes one spelling, not the class


denisvmedia:
Follow-up from #977 (which implemented `schema inspect --include`, closing that item of #951).

## What #977 added, and what it does not cover

#977 added a **shape** check to the shared include validator: an include selector whose glob carries more than one separator outside quotes is rejected before any database is contacted, because the selection matches a resource only as `name` or `schema.name`.

That closed the literal-dot spelling of a child-depth selector. It did **not** close the underlying condition, because `path.Match` treats `.` as an ordinary character — only `/` is a separator there. Measured on the #977 head binary (`95fd2eb`), SQLite, empty `--from` against a `--to` holding `users`:

```text
$ ptah-compat schema diff … --include 'main.users.email'
error: unsupported Atlas include selector "main.users.email": …
exit=1

$ ptah-compat schema diff … --include 'main.users*email'    -> "Schemas are synced, no changes to be made."  exit=0
$ ptah-compat schema diff … --include 'main.users?email'    -> "Schemas are synced, no changes to be made."  exit=0
$ ptah-compat schema diff … --include 'main.users[.]email'  -> "Schemas are synced, no changes to be made."  exit=0
```

Every glob metacharacter that can stand for a dot escapes the check: `*`, `?`, **and character classes**. The character-class case is worth calling out because it is *new* in `95fd2eb`: the intermediate commit `f631bb6` counted raw `.` characters and rejected `main.other[.]id` (exit 1); once depth started skipping bracket runs as identifier quoting, the same input became exit 0. A reader who tests `*` and `?` will reasonably assume `[.]` is caught.

## Two conditions, opposite treatment

Both of the following are "the shape check made the wrong call", but in opposite directions.

### A. Selector matched nothing, and that is silent

| Input | Condition | Today |
| --- | --- | --- |
| `--include 'main.users.email'` | matched nothing | exit 1, named error |
| `--include 'main.users*email'` | matched nothing | exit 0, "Schemas are synced" |
| `--include 'main.users?email'` | matched nothing | exit 0, "Schemas are synced" |
| `--include 'main.users[.]email'` | matched nothing | exit 0, "Schemas are synced" |
| `--include 'no_such_table'` | matched nothing | exit 0, "Schemas are synced" |

A typo'd table name and a child-depth pattern get opposite treatment for the same underlying condition.

### B. Selector *would* have matched, and is refused anyway

[...]

Issue #940, opened 2026-07-31
Declarative apply and diff: source and template gaps


denisvmedia:
Desired-state sources, template helpers and the dev-database rehearsal are not uniform across the declarative commands.

Found by a deep audit of every capability the feature matrix marked as partial.
Each item below was reproduced against a binary built from this repository.

Severity assessed as **incomplete-capability**.

### 1. Report every object schema clean drops, not just tables

`schema clean` builds its plan from cleanupObjects, which enumerates foreign keys (where the dialect supports explicit FK cleanup), tables, PostgreSQL enums, and PostgreSQL sequences. The apply path ignores that plan and calls conn.SchemaWriter().DropAllTables(ctx), which removes more object kinds — views at minimum, and every other kind the writer handles. The result is a --dry-run and a --format report that understate a destructive operation: an object the plan never mentions is destroyed anyway.

Reproduction:

```text
$ go build -o /tmp/pc ./cmd/ptah-compat
$ python3 -c "import sqlite3; c=sqlite3.connect('cl3.db'); c.executescript('CREATE TABLE users (id INTEGER PRIMARY KEY); CREATE VIEW v_const AS SELECT 1 AS one;'); c.commit()"
$ /tmp/pc schema clean --url sqlite://cl3.db --dry-run --format '{{ range .Changes }}{{ .Cmd }}{{ end }}'
DROP TABLE IF EXISTS "users"
$ python3 -c "import sqlite3; print(sqlite3.connect('cl3.db').execute('select type,name from sqlite_master').fetchall())"
[('table', 'users'), ('view', 'v_const')]
$ /tmp/pc schema clean --url sqlite://cl3.db --auto-approve
Planned cleanup changes: 1
- DROP TABLE IF EXISTS "users"
Schema clean completed successfully.
$ python3 -c "import sqlite3; print(sqlite3.connect('cl3.db').execute('select type,name from sqlite_master').fetchall())"
[]

The view v_const never appeared in the plan and was dropped anyway.
```

Evidence: internal/schemaclean/clean.go:75-81 (Apply calls DropAllTables) versus internal/schemaclean/clean.go:109-147 (cleanupObjects enumerates only FKs, tables, enums) and :149-160 (sequences, PostgreSQL only)

### 2. Register the json template helper on schema apply and schema diff

[...]

PR #1115, opened 2026-08-03
feat(compat): schema-side Atlas flags, and a named decision on --export/--web

denisvmedia:
Refs #951 — the schema-side group of the flag checklist (item 3), plus the explicit decision item 4 asked for on the two UI-bound flags.

## Atlas-side source for every flag

The pinned community binary v1.3.0 registers **none** of these. Measured by running the spelling, never by grepping `--help`, and every probe carries both controls:

```
### schema apply                (pinned community binary v1.3.0)
--dry-run             [POSITIVE CONTROL]    present
--frobnicate-nonsense [NEG CONTROL]         MISSING
--skip-lint           [SUBJECT]             MISSING
### schema clean
--url                 [POSITIVE CONTROL]    present
--frobnicate-nonsense [NEG CONTROL]         MISSING
--include / --exclude [SUBJECT]             MISSING
### schema inspect
--format              [POSITIVE CONTROL]    present
--frobnicate-nonsense [NEG CONTROL]         MISSING
--output / --web      [SUBJECT]             MISSING
### schema diff
--format              [POSITIVE CONTROL]    present
--frobnicate-nonsense [NEG CONTROL]         MISSING
--export              [SUBJECT]             MISSING
```

(`schema test` is community-gated there: even its positive control `--run` answers `unknown flag`, so the binary is no oracle for that verb.)

So the source for all six is Atlas's published CLI reference, https://atlasgo.io/cli-reference, quoted verbatim below. The spelling, type, shorthand and default match it.

| verb | flag | reference text |
| --- | --- | --- |
| `schema apply` | `--skip-lint` | `--skip-lint    skip linting migration plan` |
| `schema clean` | `--exclude strings` | `list of glob patterns used to filter resources from applying` |
| `schema clean` | `--include strings` | `list of glob patterns used to select which resources to keep in inspection` |
| `schema inspect` | `-o, --output string` | `output file path` |
| `schema test` | `-s, --schema strings` | `set schema names` |
| `schema diff` | `--export` | `use exporter defined in env` |
| `schema inspect` | `-w, --web` | `open the schema ERD in the browser` |

## Measurement, before and after

Same probe, same controls, against `ptah-compat` built from this branch's base and then from its head:

| verb | flag | before | after |
| --- | --- | --- | --- |
| `schema apply` | `--skip-lint` | MISSING | present |
| `schema clean` | `--include` | MISSING | present |
| `schema clean` | `--exclude` | MISSING | present |
| `schema inspect` | `--output` | MISSING | present |
| `schema inspect` | `--web` | MISSING | present (refuses) |
| `schema test` | `--schema` | MISSING | present |
| `schema diff` | `--export` | MISSING | present (refuses) |
| `schema diff` | `--web` | MISSING | MISSING (not in this batch) |

Every positive control answered present and every `--frobnicate-nonsense` answered MISSING, in both runs.

## What each flag does

**`schema apply --skip-lint`.** Compat's `schema apply` did not lint at all, so registering the flag over nothing would have been the accept-and-ignore failure this issue exists to prevent. The apply now lints the planned statements — edits included, and the pre-approved `--plan` path too — whenever the selected `atlas.hcl` env declares a `lint` policy, using the same severities that decide `migrate lint`. A finding the policy rates as an error refuses the apply, before the `--dry-run` exit so a dry run reports the same verdict. `--skip-lint` runs it anyway.

Only the rules the policy *names* run. The engine ships rules a declarative plan cannot answer — a plan is not a migration directory, so it has no paired down file and no history to be non-linear against — and enabling the whole registry would refuse applies over findings nobody asked about. A project with no `lint` block therefore has no lint pass, which keeps an unflagged CE-shaped `schema apply` byte-identical to what it did before.

**`schema clean --include/--exclude`.** Routed to `internal/atlasfilter`, the selector engine `schema apply`, `schema diff` and `schema inspect` already share — no second glob matcher. The engine is fed a synthetic schema built from the objects the plan already enumerates, so the selection can only narrow what `schemaclean.Inspect` found, and whatever the planner learns to enumerate flows through unchanged. An object kind the mapping does not know is refused rather than silently kept or dropped, which is the guard #940 needs while it widens that enumeration.

The important half is execution. `schemaclean.Apply` hands the whole database to `DropAllTables`, which is correct only for a plan describing the whole database; a narrowed run now executes the plan's own statements via a new `schemaclean.ApplyPlan`, so the printed plan is what is destroyed. The unflagged path is untouched and still calls `DropAllTables`.

Two consequences worth stating: `atlas.hcl` `exclude` now reaches `schema clean` like every other schema verb (ignoring it was the dangerous direction — an operator who excluded a table from their workflow would still have watched this command drop it); and drop statements keep the dialect's cascade behavior, so removing a selected object can still take dependents with it.

**`schema inspect -o/--output`.** Writes the rendered schema to a file instead of stdout, staged beside the destination and renamed over it, so a pipeline that reads the file back (`schema apply --to file://…`) cannot read a truncated schema.

**`schema test -s/--schema`.** Restricts the desired schema through `internal/schemascope`'s existing allow-list before the cases run. A selection that keeps no tables out of a schema that had some is refused rather than provisioning an empty database and failing every case on a missing relation — the zero-match false green from #979. The native `ptah schema test` gains the same flag; the compat arg mapper rewrites every occurrence and the native flag is a string array, so repeated `-s` values accumulate instead of the last one winning.

## The two UI-bound flags: registered refusals, and why

Written into the code at `cmd/atlas/schema_ui_flags.go`, into the two commands' help, and into the feature matrix.

- **`schema inspect -w/--web`** — refused. Its payload is not the problem: Ptah renders the same ERD locally, as `schema inspect --format '{{ mermaid . }}'` and as `ptah viz`. What has no counterpart is *opening a viewer*, and a CLI that launches a browser is the wrong shape for the pipelines this surface serves. The refusal names the local spelling, so the capability stays reachable.
- **`schema diff --export`** — refused. It selects an exporter declared by an `atlas.hcl` `exporter` block. `config/projectconfig` tolerates that block and evaluates nothing from it, so there is no exporter to select; accepting the flag would emit the default diff and report an export that never happened.

Registering the refusal rather than leaving the flags absent is the point: a script that passes either spelling gets one line saying what is missing and a non-zero exit, before any database is contacted. Both are covered by tests that fail if the refusal is turned into a silent accept.

| test | on revert |
| --- | --- |
| `TestSchemaApplySkipLint` | `unknown flag: --skip-lint` on the two `--skip-lint` rows; the "error policy refuses the destructive plan" row fails on `wantErr` because nothing lints, so a `DROP TABLE` the project rated an error applies silently |
| `TestSchemaCleanSelectorsNarrowThePlan` | `unknown flag: --include` / `--exclude` on every row |
| `TestSchemaCleanSelectorsNarrowWhatIsDestroyed` | same unknown flag; and with the selector wired to the plan but execution left on `DropAllTables`, it fails on `sqliteTableCount(audit_log) == 1` — the plan says one table, the database lost two |
| `TestSchemaCleanUnselectedRunStillDropsEverything` | fails if routing the selectors quietly turned a plain `schema clean` into a partial one |
| `TestSchemaCleanRejectsUnsupportedSelectorsBeforeConnecting` | fails if a malformed selector reaches the database before failing |
| `TestSchemaInspectOutputWritesTheFile` | `unknown flag: --output` / `-o`; also fails if `--output` writes the file *and* still prints to stdout |
| `TestSchemaInspectOutputReportsAnUnwritableDestination` | fails if a bad destination drops the inspection instead of failing loudly |
| `TestSchemaTestSchemaSelection` | `unknown flag: --schema`; the repeated-`-s` row fails under last-wins forwarding because the run would be scoped to `nosuch` alone; the zero-match row fails if an empty selection is provisioned instead of refused |
| `TestSchemaUIFlagsAreRegisteredRefusals` | `unknown flag` on revert; and on the exit status if either refusal is turned into a silent accept |

The `schema test` fixture is discriminating on purpose: its two tables live in two schemas, and on SQLite only `main` exists, so the unscoped run fails outright on the other table. A `--schema` that parsed and was then ignored cannot make it pass, and neither can one that kept both tables.

## Gates

| gate | result |
| --- | --- |
| `go build ./...` | clean |
| `go vet ./...` | clean |
| `go test ./... -count=1` | all packages pass |
| `golangci-lint run ./...` | `0 issues.` |
| `node docs/site/scripts/check-style.mjs` | `check-style.mjs: OK (100 documentation files)` |
| `node docs/site/scripts/build-feature-matrix.mjs --check` | `OK (162 rows)` |

Both public-API gates are silent on success, so each was broken first to prove it runs: adding an exported `PtahGateProbeSymbol` to `migration/lint` drove the snapshot gate to exit 1 with a diff, and adding a `gateprobe/` package drove the package gate to exit 1 with `undocumented public package`. Both probes were removed and both gates re-run green.

Two existing assertions in `cmd/atlas/atlas_test.go` encoded the previous decision (`--output`, `--web`, `--export` all unregistered on `schema inspect`) and were updated to the new one, keeping `--export` on inspect as still-unregistered.

## Flags NOT added, and why

- **`schema diff --web`, `schema inspect --export`, `migrate lint --web`** — same Atlas-side source and same verdict as their twins here, left out to stay inside this batch rather than because they differ. `migrate lint --web` in particular overlaps whoever owns the lint rows.
- **`schema apply --lock-name`, `--skip-lock`** — the named-lock family is owned by another agent in parallel; touching them here would produce two implementations of one feature.
- **`schema test --report`, `--seed-dir`** — unchanged verdict: Atlas registers neither on either test verb, so compat must not either. Reachable natively (`ptah schema test --report`), and a compat-side control would go through a `PTAH_*` variable.
- **`schema clean --schema`** — not registered by Atlas on this verb per the reference (`-u`, `--exclude`, `--include`, `--dry-run`, `--format`, `--auto-approve` is the whole set), so the cleanup scope is `--include`/`--exclude` only.

Nothing in this batch was undecidable: every flag had a documented Atlas-side source.

## Note for #940

`internal/schemaclean` is being reworked under #940 for planning less than it destroys. This branch adds `ApplyPlan` beside `Apply` and does not change `Inspect`, `cleanupObjects` or `inspectRuntimeObjects`, so the enumeration that issue is widening stays entirely its own. The selector projection consumes whatever the plan enumerates and refuses an object kind it has no mapping for, so a new kind arriving from #940 surfaces as a loud refusal on a scoped clean rather than as a silently mis-scoped drop.

### DIFF
--- a/cmd/atlas/atlas.go
+++ b/cmd/atlas/atlas.go
@@ -574,6 +574,11 @@ func atlasMigrateVersionValue(value string) (string, error) {
 // native Go-annotation root directory, --dev-url selects the throwaway
 // database (ephemeral SQLite when omitted), and the optional [paths]
 // positional selects the directory of Ptah-native YAML test cases.
+//
+// -s/--schema restricts the desired schema before the cases run. Atlas registers
+// it as a repeatable `strings` flag (atlasgo.io/cli-reference); the arg mapper
+// rewrites every occurrence, and the native flag is a string array, so repeated
+// values accumulate instead of the last one winning.
 func atlasSchemaTestVerb() atlasVerb {
 	return atlasVerb{
 		use:                "test",
@@ -588,6 +593,7 @@ func atlasSchemaTestVerb() atlasVerb {
 			atlasargs.NativeLocalDir("url", "u", "Desired schema URL: local file:// directory with Go schema annotations", "root-dir"),
 			atlasargs.NativeString("dev-url", "", "Dev database URL the test cases run against", "db-url"),
 			atlasargs.String("run", "", "Run only test cases matching a Go regular expression"),
+			atlasargs.String(atlasSchemaFlagName, atlasSchemaFlagShorthand, "Restrict the desired schema to these schema names"),
 		},
 	}
 }
--- a/cmd/atlas/atlas_test.go
+++ b/cmd/atlas/atlas_test.go
@@ -200,11 +200,16 @@ func TestCompatCommand_AdvertisesEssentialAtlasFlags(t *testing.T) {
 		{
 			name: "schema_inspect",
 			path: []string{"schema", "inspect"},
-			// --include is a Pro-surface flag the pinned Atlas CE binary does
-			// not register on inspect; compat implements it so Pro pipelines
-			// port. --output, --web, and --export stay unregistered.
-			flags:     []string{"--url", "--dev-url", "--env", "--schema", "--exclude", "--format", "--include"},
-			forbidden: []string{"--output", "--web", "--export"},
+			// --include and --output are Pro-surface flags the pinned Atlas CE
+			// binary does not register on inspect; compat implements both so
+			// Pro pipelines port. --web is registered as a named refusal
+			// (see schema_ui_flags.go). --export stays unregistered: it is a
+			// separate item of stokaro/ptah#951.
+			flags: []string{
+				"--url", "--dev-url", "--env", "--schema", "--exclude",
+				"--format", "--include", "--output", "--web",
+			},
+			forbidden: []string{"--export"},
 		},
 		{
 			name:  "schema_apply",
@@ -1288,13 +1293,18 @@ func TestCompatCommand_SchemaInspectUsesAtlasProjectFormatAndSchemaMode(t *testi
 }
 
 // TestCompatCommand_SchemaInspectRejectsProOnlyOutputFlags pins the inspect
-// flags Atlas registers that compat deliberately does not:
-// --output, --web, and --export are separate items of stokaro/ptah#951.
-// --include is registered and covered by schema_inspect_include_test.go.
+// flags Atlas registers that compat deliberately does not.
+//
+// The list has shrunk to --export. --output is implemented
+// (schema_inspect_output_test.go) and --web is a registered refusal
+// (schema_ui_flags_test.go); --export on inspect is the twin of the
+// `schema diff --export` decision and stays a separate item of
+// stokaro/ptah#951. --include is registered and covered by
+// schema_inspect_include_test.go.
 func TestCompatCommand_SchemaInspectRejectsProOnlyOutputFlags(t *testing.T) {
 	c := qt.New(t)
 
-	for _, flag := range []string{"--output", "--web", "--export"} {
+	for _, flag := range []string{"--export"} {
 		c.Run(flag, func(c *qt.C) {
 			cmd := NewCompatCommand("atlas")
 			var out bytes.Buffer
--- a/cmd/atlas/output_file.go
+++ b/cmd/atlas/output_file.go
@@ -1,0 +1,61 @@
+package atlas
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"go.5x5.cz/ptah/internal/fsdurable"
+)
+
+// writeAtlasOutputFile publishes a rendered document at path atomically.
+//
+// The document is staged in the destination directory and then renamed over the
+// destination, so a concurrent reader sees either the previous file or the
+// complete new one. A schema inspection redirected into a file is routinely read
+// back by the next pipeline step (`schema apply --to file://…`), and a truncated
+// read there is a schema with objects missing, which is exactly the failure a
+// half-written file would produce.
+func writeAtlasOutputFile(path string, document []byte) error {
+	dir := filepath.Dir(path)
+	staged, err := os.CreateTemp(dir, ".ptah-output-*.tmp")
+	if err != nil {
+		return fmt.Errorf("stage output file: %w", err)
+	}
+	stagedPath := staged.Name()
+	if err := prepareAtlasOutputFile(staged, stagedPath, document); err != nil {
+		return err
+	}
+	if err := fsdurable.ReplaceFile(stagedPath, path); err != nil {
+		return errors.Join(fmt.Errorf("publish output file: %w", err), removeAtlasOutputStage(stagedPath))
+	}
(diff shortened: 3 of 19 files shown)


denisvmedia:
Superseded by #1124.

This branch was cut from `c0f86693` and conflicted with master, which has since taken ten commits including #1117. GitHub reported `mergeable_state: dirty`, so it could not be updated and CI never ran on it. A PR's head branch cannot be repointed through the API, so the rebased work is a fresh PR; `feat/951-schema-side-compat-flags` is left untouched.

#1124 is stacked on #1119 (the rebased named-lock branch) rather than on master, because this change and that one both touch `cmd/atlas/schema_apply.go`, `cmd/atlas/atlas_test.go` and both feature-matrix files. Rebasing onto plain master would have left those two conflicting with each other. **Merge #1119 first**; GitHub retargets #1124 to master automatically.

Three things worth knowing, all detailed in #1124:

- The `schema test` desired-source classification from #1110 and this branch's `--schema` selection met for the first time in the rebase. Merging them exposed a real defect: a live introspection leaves default-schema objects unqualified, so every selection over a database source refused — `--schema main` included. It now scopes with the dialect's default schema, and the combination is covered by a new test that previously had no coverage at all.
- `cmd/atlas/atlas_test.go`'s `schema_apply` row auto-merged wrongly without git reporting a conflict, and was caught by `TestCompatCommand_AdvertisesEssentialAtlasFlags`.
- Two feature-matrix claims in this description have been overtaken by master (`schema clean` over-dropping, and `migrate test --revisions-schema` being missing) and were not carried forward.

The description here remains the reference for the feature rationale and the Atlas-side sources.

PR #1116, opened 2026-08-03
feat(compat): named-lock family on migrate apply and schema apply

denisvmedia:
Refs #951 — the named-lock family, implemented once over the existing `internal/dblock` machinery.

Four of the seven flags in the batch land. Three do not, because no Atlas-side source registers them.

## What landed, and the Atlas-side source for each

| Verb | Flag | Atlas-side source | Behavior change |
| --- | --- | --- | --- |
| `migrate apply` | `--lock-name string` | Published CLI reference, atlasgo.io/cli-reference (retrieved 2026-08-03) | Replaces the session advisory lock name (default `ptah_migrate`); a run naming a held lock waits and then fails on `--lock-timeout` |
| `migrate apply` | `--skip-lock` (bool) | same | No advisory lock is requested at all; a held lock is ignored rather than waited on |
| `schema apply` | `--lock-name string` | same | Replaces the apply lock name (default `ptah_schema_apply`), same wait/timeout semantics |
| `schema apply` | `--skip-lock` (bool) | same | Same: no lock requested |

Spelling, type and default match that source verbatim:

```
--lock-name string   set the name of the advisory lock to use. If not set, the default is used (atlas_migrate_execute)
--skip-lock          skip acquiring a database advisory lock
```

`--lock-name` is a string with no cobra default (unset means "the tool's own default lock name"), `--skip-lock` a boolean defaulting to false — exactly as published.

The pinned community binary v1.3.0 registers **neither flag, on any verb**. That was measured by running each spelling, not by grepping help. So this is Pro surface adopted openly, sourced from the published reference. That is the same shape as `schema inspect --include` (#977).

## What I did NOT add, and why

| Flag | Pinned community binary | Published CLI reference | Verdict |
| --- | --- | --- | --- |
| `migrate diff --lock-name` | `unknown flag` (verb is **not** gated, so this is authoritative) | absent — `migrate diff` lists `--lock-timeout duration` and no other member of this family | **declined**: no Atlas-side source |
| `migrate down --lock-name` | verb is community-gated and registers **none** of its own flags, so the binary is no oracle | absent — `migrate down` lists `--lock-timeout duration` only | **declined**: no Atlas-side source |
| `migrate checkpoint --lock-name` | same: gated, registers none of its own flags | absent — `migrate checkpoint` lists `--lock-timeout duration` only | **declined**: no Atlas-side source |

The issue's checklist asked for `--lock-name` on five verbs. Measured against both sources, it exists on two. Adding it to the other three would put non-Atlas flags on the compat surface, which is exactly what this issue's standing constraint forbids — the same resolution as `migrate apply --skip-checks` and the `--report`/`--seed-dir` request.

Verbatim from the reference for the three declined verbs (whole flag blocks, so the absence is visible rather than asserted):

```
atlas migrate diff [flags] [name]
  --to strings              ...
  --dev-url string          ...
  --dir string              ...
  --dir-format string       ...
  -s, --schema strings      ...
  --lock-timeout duration   set how long to wait for the database lock (default 10s)
  --format string           ...
  --qualifier string        ...
  --edit

atlas migrate down [flags] [amount]
  -u, --url string            ...
  --to-version string         ...
  --to-tag string             ...
  --dir string                ...
  --dev-url string            ...
  --format string             ...
  --dry-run
  --revisions-schema string   ...
  --lock-timeout duration     set how long to wait for the database lock (default 10s)
  --skip-checks
  --plan

atlas migrate checkpoint [flags] [tag]
  --dev-url string          ...
  --dir string              ...
  --dir-format string       ...
  -s, --schema strings      ...
  --lock-timeout duration   set how long to wait for the database lock (default 10s)
  --format string           ...
  --qualifier string        ...
  --edit
```

Controls on that same fetch, so a blank or partially-rendered page cannot read as an absence: the page is non-empty (50,303 plaintext characters), the control term `atlas migrate apply` occurs 10 times, the invented control `frobnicate` occurs 0 times, and `--lock-name` occurs exactly **2** times across the whole page — once under `migrate apply`, once under `schema apply`. Same count for `--skip-lock`.

## Measurement, before and after

Every row was produced by RUNNING the spelling and looking for `unknown flag`. Never by grepping `--help`. Every row carries both controls, and the controls are reported, not assumed.

### The pinned community binary (v1.3.0)

| Verb | positive control | `--frobnicate-nonsense` | `--lock-name` | `--skip-lock` |
| `migrate apply` | `--dry-run` PRESENT | MISSING | MISSING | MISSING |
| `migrate diff` | `--dev-url` PRESENT | MISSING | MISSING | MISSING |
| `migrate down` | `--env` PRESENT (see note) | MISSING | MISSING | MISSING |
| `migrate checkpoint` | `--env` PRESENT (see note) | MISSING | MISSING | MISSING |
| `schema apply` | `--dry-run` PRESENT | MISSING | MISSING | MISSING |

Note on the two gated verbs: my first probe used `--dry-run` / `--dir-format` as the positive control on `migrate down` and `migrate checkpoint` and **both controls failed**, which per the probe rules means the probe was broken, not the row. Investigating rather than believing it: those two verbs are community-gated and register none of their own flags — only inherited `--config`/`--env`/`--var` parse and reach the abort. So the positive control there has to be an inherited flag. With `--env` the control passes and the rows are readable. This is the same shape #1011 recorded for `migrate checkpoint`; it now also holds for `migrate down` on v1.3.0.

### ptah-compat, before (`c0f86693`) and after

| Verb | flag | before | after |
| --- | --- | --- | --- |
| `migrate apply` | `--lock-name` | MISSING | **PRESENT** |
| `migrate apply` | `--skip-lock` | MISSING | **PRESENT** |
| `schema apply` | `--lock-name` | MISSING | **PRESENT** |
| `schema apply` | `--skip-lock` | MISSING | **PRESENT** |
| `migrate diff` | `--lock-name` / `--skip-lock` | MISSING | MISSING (unchanged, deliberate) |
| `migrate down` | `--lock-name` / `--skip-lock` | MISSING | MISSING (unchanged, deliberate) |
| `migrate checkpoint` | `--lock-name` / `--skip-lock` | MISSING | MISSING (unchanged, deliberate) |

Positive control `--dry-run` PRESENT and negative control `--frobnicate-nonsense` MISSING on every verb, in both binaries, on the same runs.

## The semantics, decided and stated

**What happens when `--lock-name` names a lock another process holds.** The run waits for it, bounded by `--lock-timeout` (empty waits indefinitely, which is compat's existing `--lock-timeout` semantics). An elapsed timeout fails the run before anything is applied. Live-verified against PostgreSQL 16 with a competing session holding `pg_advisory_lock(fnv32a("atlas_migrate_execute"))`:

```
$ ptah-compat schema apply --url postgres://… --to file://schema.sql --auto-approve \
    --lock-name atlas_migrate_execute --lock-timeout 2s
Error: acquire schema apply lock: timed out acquiring advisory lock "atlas_migrate_execute" on postgres after 2s
exit=1

$ ptah-compat migrate apply --url postgres://… --dir file://migrations \
    --lock-name atlas_migrate_execute --lock-timeout 2s
Error: error applying migrations: failed to acquire migration lock for migrate up: timed out acquiring migration lock "atlas_migrate_execute" for postgres after 2s
exit=1
```

**What `--skip-lock` does to that path.** It removes it. No lock is requested, so there is nothing to wait on and the held lock is irrelevant. Same competing session still holding the same lock:

```
$ ptah-compat schema apply … --skip-lock
Schema apply completed successfully.
exit=0                        # and the table exists afterwards

$ ptah-compat migrate apply … --skip-lock
Migration complete. Current version: 20260101000001
exit=0
```

That is the point of the flag, and it is why it must not be possible to pass it and still get a lock — or to pass `--lock-name` and silently get none.

**Two inputs are refused rather than interpreted.** Both are the silent-failure shape this family exists to avoid:

- `--lock-name ""` (explicitly blank) → `--lock-name must not be empty`. Passing it through would fall back to the default lock, so a caller who meant to coordinate on a named lock would coordinate on a different one without being told.
- `--lock-name X --skip-lock` → `--lock-name and --skip-lock cannot be used together: --skip-lock takes no lock, so there is no lock to name`. The published reference documents no resolution order for the pair; honouring either would drop the other silently. This is a deliberate, documented divergence for an input Atlas defines no behavior for.

**Defaults are unchanged, on purpose.** Compat still defaults to `ptah_migrate` / `ptah_schema_apply`, not to Atlas's `atlas_migrate_execute`. Moving the default would de-serialize a fleet mid-upgrade: an older ptah-compat holding `ptah_migrate` and a newer one holding `atlas_migrate_execute` do not see each other. `--lock-name atlas_migrate_execute` is the explicit spelling for coordinating with Atlas, which is precisely why the flag has to exist. Whether the compat default should eventually move is a separate decision with its own upgrade story — flagged below, not taken here.

**Unsupported dialects.** SQLite, ClickHouse, CockroachDB and Spanner have no advisory locks. An explicit `--lock-name` there prints a stderr note naming the lock that was not acquired, so a flag that cannot do its job says so instead of reading as serialization. `--skip-lock` prints nothing — a caller who asked for no lock does not need to be told the dialect has none. Output for every invocation that passes neither flag is byte-identical to before.

Each was verified by actually applying the mutation, running the test, and reverting.

| Test | Mutation | What it prints reverted |
| --- | --- | --- |
| `migration/migrator` `TestMigrateUpLockRequest/skipped_lock_requests_nothing` | drop `\|\| m.migrationLockSkipped` from `withMigrationLock` | `acquired=true name="ptah_migrate" spans=[ptah.migrate.up ptah.lock.acquire …]` — got `true`, want `false` |
| `internal/atlasmigrate` `TestPrepareApply_LockRequestReachesMigrator/skip_lock` | same mutation | `acquired=true name="ptah_migrate"` — got `true`, want `false` |
| `internal/atlasmigrate` `…/lock_name` | drop `WithMigrationLockName(opts.migrationLockName)` from `newApplyMigrator` | lock name is `ptah_migrate`, want `atlas_migrate_execute` |
| `cmd/atlas` `TestMigrateApplyNamedLockReachesMigrator/named_lock` | same | prints `the advisory lock "ptah_migrate" is not acquired`, want `"atlas_migrate_execute"` |
| `cmd/atlas` `TestSchemaApplyNamedLockReachesLockMachinery/skip_lock_silences_the_timeout_note` | make `--skip-lock` acquire the lock anyway | the run emits `note: schema apply locking is not supported for dialect "sqlite"; --lock-timeout is ignored …`, proving a lock was taken |
| `internal/atlasschema` `TestAcquireApplyLock_RecordsRequestedName` (3 rows) + `cmd/atlas` `…/named_lock` | stop recording the name on `dblock`'s no-op path | `lock.Name()` is `""` on every row |
| `cmd/atlas` `TestCompatLockFlagContradictionsRefused/*_blank_name` | remove the blank-name refusal | the run proceeds to `open migrations directory: … no such file or directory` / `load --to schema: schema file does not exist` instead of refusing |
| `cmd/atlas` `TestCompatLockFlagContradictionsRefused/*_name_with_skip` | remove the mutual-exclusion refusal | both rows fail: the contradiction is accepted |
| `cmd/atlas` `TestCompatLockFlagRegistration/migrate_apply_*` | unregister both flags on `migrate apply` | both rows fail, and `TestCompatCommand_AdvertisesEssentialAtlasFlags/migrate_apply` fails with them |
| `cmd/atlas` `TestCompatLockFlagRegistration/migrate_diff_lock_name` | register `--lock-name` on `migrate diff` | fails — this row is what keeps the three declined spellings declined |

Registration is measured in the test the same way as on the command line: by running the spelling and reading the error, with a positive and a negative control on every row. No help grepping.

Two live PostgreSQL tests ship env-gated (`POSTGRES_TEST_DSN` / `TEST_DATABASE_URL`), matching the existing lock live tests: `TestAcquireApplyLock_NamedLockContendsLive` (same name contends and times out; releasing the holder unblocks it) and `TestAcquireApplyLock_DistinctNamesDoNotContendLive` (a different name does not contend at all). Both were run and passed against a throwaway `postgres:16-alpine`, along with the three pre-existing live lock tests.

## Gates

| Gate | Result |
| --- | --- |
| `go build ./...` | pass |
| `go vet ./...` | pass |
| `go test ./... -count=1` | pass, no failures |
| `golangci-lint run ./...` | `0 issues.` |
| `scripts/check-public-api.sh` | pass |
| `scripts/check-public-api-snapshot.sh` | pass (snapshot regenerated for the three new `Migrator` methods) |
| `docs/site` `check-style.mjs` | `OK (100 documentation files)` |
| `docs/site` `build-feature-matrix.mjs --check` | `OK (159 rows)` |

## Public API delta

`migration/migrator` gains three methods, snapshot updated:

- `WithoutMigrationLock() *Migrator` — the capability `--skip-lock` needs.
- `MigrationLockName() string` and `MigrationLockSkipped() bool` — so a command can report the lock the machinery resolved rather than echoing its own flag value. The `migrate apply` stderr note reads its lock name from here, which is what makes that note a real observation of the machinery.

`internal/atlasschema.AcquireApplyLock` gains a `name` parameter (internal package; the native `ptah schema apply` call sites pass `""` and keep the default).

## One thing this PR does NOT decide, flagged for #951

**The conformance `cli-surface` tier will go RED until its allowance table is updated**, in the separate conformance checkout. `compareFlags` measures "extra" against the CE flag set plus a per-command `proSurfaceFlags()` allowance, and that table currently has an entry only for `schema inspect`. It needs:

```go
"migrate apply": {"--lock-name", "--skip-lock"},
"schema apply":  {"--lock-name", "--skip-lock"},
```

I did not make that change — it is a different repository and outside this worktree. Without it the tier reports `flag mismatch: extra --lock-name, --skip-lock` on both verbs. Worth pairing with this PR rather than discovering after merge.

Two smaller notes:

- A `migrate apply --dry-run` takes no lock at all (`withMigrationLock` short-circuits on dry-run), so `--lock-name` is inert there. That is pre-existing and applies equally to `--lock-timeout`; a dry run mutates nothing, so there is nothing to serialize.
- Whether compat's *default* lock name should eventually become `atlas_migrate_execute` is a real interop question this PR leaves open, with the upgrade hazard described above. It needs its own decision.

### DIFF
--- a/cmd/atlas/atlas_test.go
+++ b/cmd/atlas/atlas_test.go
@@ -207,9 +207,14 @@ func TestCompatCommand_AdvertisesEssentialAtlasFlags(t *testing.T) {
 			forbidden: []string{"--output", "--web", "--export"},
 		},
 		{
-			name:  "schema_apply",
-			path:  []string{"schema", "apply"},
-			flags: []string{"--url", "--to", "--dev-url", "--dry-run", "--auto-approve", "--format", "--schema", "--exclude", "--include", "--tx-mode", "--plan", "--edit", "--lock-timeout"},
+			name: "schema_apply",
+			path: []string{"schema", "apply"},
+			// --lock-name and --skip-lock are Pro-surface flags the pinned
+			// Atlas CE binary does not register; Atlas's published CLI
+			// reference does register them on this verb, so compat implements
+			// them. --skip-lint stays unregistered.
+			flags:     []string{"--url", "--to", "--dev-url", "--dry-run", "--auto-approve", "--format", "--schema", "--exclude", "--include", "--tx-mode", "--plan", "--edit", "--lock-timeout", "--lock-name", "--skip-lock"},
+			forbidden: []string{"--skip-lint"},
 		},
 		{
 			name:      "schema_diff",
@@ -242,8 +247,12 @@ func TestCompatCommand_AdvertisesEssentialAtlasFlags(t *testing.T) {
 				"--revisions-schema",
 				"--lock-timeout",
 				"--format",
+				// Pro-surface flags Atlas's published CLI reference registers
+				// on this verb; the pinned CE binary registers neither.
+				"--lock-name",
+				"--skip-lock",
 			},
-			forbidden: []string{"--to-version", "--lock-name"},
+			forbidden: []string{"--to-version"},
 		},
 		{
 			name: "migrate_down",
@@ -3088,18 +3097,6 @@ func TestCompatCommand_MigrateApplyRejectsNonAtlasFlags(t *testing.T) {
 		err := cmd.Execute()
 
 		c.Assert(err, qt.ErrorMatches, `unknown flag: --to-version`)
-	})
-
-	c.Run("lock_name", func(c *qt.C) {
-		cmd := NewCompatCommand("atlas")
-		var out bytes.Buffer
-		cmd.SetOut(&out)
-		cmd.SetErr(&out)
-		cmd.SetArgs([]string{"migrate", "apply", "--lock-name", "custom-lock"})
-
-		err := cmd.Execute()
-
-		c.Assert(err, qt.ErrorMatches, `unknown flag: --lock-name`)
 	})
 }
 
--- a/cmd/atlas/lock_flags.go
+++ b/cmd/atlas/lock_flags.go
@@ -1,0 +1,99 @@
+package atlas
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+)
+
+// The named-lock family: `--lock-name` and `--skip-lock`.
+//
+// ATLAS-SIDE SOURCE. Atlas's published CLI reference (atlasgo.io/cli-reference)
+// registers both flags on `atlas migrate apply` and on `atlas schema apply`,
+// and on no other verb:
+//
+//	--lock-name string   set the name of the advisory lock to use. If not set,
+//	                     the default is used (atlas_migrate_execute)
+//	--skip-lock          skip acquiring a database advisory lock
+//
+// The pinned Atlas community binary v1.3.0 registers neither, on any verb —
+// `migrate apply --lock-name x` and `schema apply --skip-lock` both answer
+// `unknown flag`, with `--dry-run` answering present on the same run and
+// `--frobnicate-nonsense` answering unknown. So this is Pro surface adopted
+// openly, sourced from the published reference rather than from the binary.
+//
+// The same reference does NOT register `--lock-name` on `migrate diff`,
+// `migrate down` or `migrate checkpoint`: those three carry `--lock-timeout`
+// and nothing else from this family. Those spellings are therefore absent here
+// on purpose (stokaro/ptah#951 standing constraint: no compat flag without an
+// Atlas-side source).
+//
+// WHY THE DEFAULT NAME IS NOT atlas_migrate_execute. Ptah's own default lock
+// names (`ptah_migrate`, `ptah_schema_apply`) are unchanged, because moving
+// them would silently de-serialize a fleet mid-upgrade: an older ptah-compat
+// holding `ptah_migrate` and a newer one holding `atlas_migrate_execute` do not
+// see each other. `--lock-name atlas_migrate_execute` is the explicit spelling
+// for coordinating with Atlas, and is why the flag has to exist.
+
+const (
+	atlasLockNameFlag = "lock-name"
+	atlasSkipLockFlag = "skip-lock"
+)
+
+// atlasLockOptions holds the raw named-lock flag values for one compat verb.
+type atlasLockOptions struct {
+	name string
+	skip bool
+}
+
+// atlasLockRequest is the resolved named-lock decision handed to the lock
+// machinery. A zero Name means "this verb's default lock name"; Skip means no
+// lock is requested at all.
+type atlasLockRequest struct {
+	Name string
+	Skip bool
+}
+
+// registerAtlasLockNameFlag registers `--lock-name` on a verb that Atlas
+// registers it on. Type and default match the reference: a string with no
+// default, where "unset" means the tool's own default lock name.
+func registerAtlasLockNameFlag(flags *pflag.FlagSet, opts *atlasLockOptions) {
+	flags.StringVar(&opts.name, atlasLockNameFlag, "",
+		"Name of the database advisory lock to acquire; unset uses the default lock name")
+}
+
+// registerAtlasSkipLockFlag registers `--skip-lock` on a verb that Atlas
+// registers it on.
+func registerAtlasSkipLockFlag(flags *pflag.FlagSet, opts *atlasLockOptions) {
+	flags.BoolVar(&opts.skip, atlasSkipLockFlag, false,
+		"Skip acquiring the database advisory lock")
+}
+
+// resolveAtlasLockRequest turns the raw flag values into the lock decision the
+// machinery receives.
+//
+// Two inputs are refused rather than interpreted:
+//
+//   - An explicit but blank `--lock-name`. Passing it through would fall back
+//     to the default lock name, so a caller that meant to name a lock would
+//     silently coordinate on a different one.
+//   - `--lock-name` together with `--skip-lock`. There is no lock to name when
+//     no lock is taken, and Atlas's reference documents no resolution order for
+//     the pair. Honouring either one would drop the other silently, which is
+//     the failure this whole family exists to avoid; refusing says so instead.
+func resolveAtlasLockRequest(cmd *cobra.Command, opts atlasLockOptions) (atlasLockRequest, error) {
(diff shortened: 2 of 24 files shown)


denisvmedia:
Superseded by #1119.

This branch was cut from `c0f86693` and conflicted with master, which has since taken ten commits including #1117. GitHub reported `mergeable_state: dirty`, so the branch could not be updated and CI never ran on it.

The head branch of an open PR cannot be repointed through the API, so the rebased work is a fresh PR rather than a force-push over `agent/951-named-lock-family` — that branch is left untouched in case anyone is reading it.

#1119 is this branch rebased onto `1ae8306a`, with four conflicts resolved (three mechanical merges, one judgment call on `TestCompatCommand_MigrateApplyRejectsNonAtlasFlags`, whose two subtests both became obsolete once #1117 and this change had both landed). All discriminators were re-proved against the rebased tree and all gates are green. The `migrate checkpoint` boundary — `--lock-timeout` from #1117 registered, `--lock-name` still declined here — was confirmed by running the merged binary.

The feature rationale and measurement tables in this description remain the reference for the change; #1119 documents only what the rebase altered.

PR #1122, opened 2026-08-03
docs: mirror the schema test exit-code row onto the site copy

denisvmedia:
One line. `master` is currently red on the **Build versioned site** job and has been since #1110 merged; this makes it green again.

## What is broken

`docs/site/scripts/check-exit-codes.mjs` treats `docs/exit_codes.md` as the source of truth and `docs/site/src/content/docs/reference/exit-codes.md` as a mirror of it, failing when a row exists in the source but not in the mirror:

```js
const sourcePath = join(repoRoot, 'docs', 'exit_codes.md');
const sitePath   = join(siteRoot, 'src', 'content', 'docs', 'reference', 'exit-codes.md');
```

#1110 (`e725ba98`, "Accept a database URL as the schema test desired state") widened the `ptah schema test` usage-error cell in the **source** to name the two new database-source refusals. It did not update the **mirror**. Its diff touches `docs/exit_codes.md`, `docs/conformance.md` and the two feature-matrix files — not the site copy.

Result on master `1ae8306a`, clean worktree:

```
$ node scripts/check-exit-codes.mjs
docs/site/src/content/docs/reference/exit-codes.md is missing 1 exit-code reference row(s):
- | `ptah schema test` | Every schema test case passed. | One or more cases failed. | Usage error (including a `--root-dir` database source whose dialect differs from `--db-url`, or a non-SQLite database source with no `--db-url`), invalid or unreadable cases, connection failure, interrupted run, desired-schema parse/apply failure, or report error. |
exit 1
```

## The fix

Copies the source row verbatim onto the mirror. One line, no behavior change, no other row touched.

## Verification

| Check | Before | After |
| --- | --- | --- |
| `check:exit-codes` | exit 1 | `OK (60 reference rows)` |
| `check:exit-codes:selftest` | PASS | PASS |
| `check:core-doc-links` | PASS | PASS |
| `check:page-health` | PASS | PASS |
| `check:links` | PASS | PASS |
| `check:redirects` | PASS | PASS |
| `check:style` | PASS | PASS |

The gate was proved live rather than assumed: after the fix, mutating one token of the restored row drove `check:exit-codes` back to exit 1, and restoring it returned exit 0.

`check:responsive` reports `dist not found; run "npm run build" first` in a bare checkout both before and after — it needs the site built first, which CI does. Unrelated to this change.

## Why this is its own PR

It was found while rebasing #1114/#1115/#1119, all of which inherit this red check. It is master's defect rather than any of theirs — none of those branches touches either exit-code file — so it is kept separate rather than bundled into a feature PR. Merging it turns the **Build versioned site** job green on master and on all three of those PRs.


denisvmedia:
Superseded by #1121, already merged — same one-row sync of `docs/site/src/content/docs/reference/exit-codes.md` against `docs/exit_codes.md`.

Verified on `origin/master` after #1121 landed: `check-exit-codes.mjs: OK (60 reference rows)`. Nothing in this branch is left to apply.

Closing rather than merging so the second copy does not land as an empty diff. The two branches were written independently within a few minutes of each other, which is worth noting on its own: master was red for every branch cut from it, so more than one line of work hit the same failure at once.

Issue #1098, opened 2026-08-03
migrate apply: a Flyway directory reaching version 10 applies out of order where Atlas CE refuses (no baseline involved)


denisvmedia:
Found while implementing #1003 and deliberately **not** folded into it: this has nothing to do with baselines, and closing #1003 does not close this.

`ptah-compat migrate apply` exits 0 and executes a migration that the pinned Atlas CE refuses with a non-linear error. Every Flyway project that reaches a two-digit version hits it on its tenth migration.

## Measured

Pinned community binary `ptah-atlas-conformance/bin/atlas` — `atlas community version v1.3.0` — sqlite, `--dir file://m?format=flyway`, `atlas.sum` written by each tool separately (byte-identical), exit codes captured immediately after each command. `ptah-compat` built from `master` at bfc7ee29, re-verified on the #1003 branch.

**V2 applied, then `V10__y.sql` added:**

```
$ atlas migrate apply --url sqlite://ce.db --dir 'file://m?format=flyway'
Error: migration file V10__y.sql was added out of order. See: https://atlasgo.io/versioned/apply#non-linear-error
exit=1        tables: s2

$ ptah-compat migrate apply --url sqlite://ptah.db --dir 'file://m?format=flyway'
Migrating to version 4611686018427836747 from 1 pending migrations.
Migration complete. Current version: 4611686018427836747
exit=0        tables: s2,y        <- executed what the oracle refuses
```

This is the looser direction: compat exits 0 where the pinned binary exits 1.

**The mirror shape already matches**, which is what makes it a projection bug rather than a missing guard — V10 applied, then `V2__y.sql` added:

```
atlas:      exit=1  Error: migration file V2__y.sql was added out of order.
ptah-compat exit=1  Error: error applying migrations: out-of-order pending migrations below current version
                    4611686018427836747: [4611686018427510315] (use --exec-order=non-linear ...)
```

## Why

Atlas CE uses two different comparisons on the same directory, and Ptah reproduces only one of them:

- **Sum and execution order is numeric.** A fresh apply of a directory holding `V2__a.sql` and `V10__b.sql` runs version 2 then version 10, exit 0 (measured).
- **The pending / non-linear comparison is on the version STRING.** `"10" < "2"`, so a `V10` added after `V2` is "below the current version" and is refused.

[...]

PR #1123, opened 2026-08-03
Refuse a Flyway baseline that lands under recorded history (#1003)

### LINKED ISSUE #1003: Decision: should ptah-compat reproduce Atlas CE's silent skip of a Flyway baseline whose token does not sort above the high-water mark?
**This is a decision to be made, not a defect.** Both behaviors are defensible and the usual "compat matches Atlas, native may be better" principle does not settle it, because the Atlas behavior here is a footgun rather than a preference.

## What was measured

Pinned Atlas CE v1.2.0 (`ptah-atlas-conformance/bin/atlas`, absolute path), 2026-08-02.

Atlas CE decides pending-ness for a converted Flyway directory by comparing the version token as a **string** against the current high-water mark — not by asking whether the version is in the recorded set. Two consequences, both measured:

**F. A baseline whose token does not sort above the current version never runs.**

```
V2__x.sql applied           revisions: 2=x            tables: x
add B10__base.sql, re-hash  covered: B10__base.sql V2__x.sql
$ atlas migrate apply ...   No migration files to execute        exit=0
                            revisions: 2=x            tables: x     <- base never ran
```

`"10" < "2"` as strings, so CE considers version 10 already covered.

**7/D. A baseline that supersedes already-applied files never runs.**

```
V1__p, V2__q, V3__t applied   revisions: 1=p, 2=q, 3=t     tables: p,q,t
add B2__base.sql, re-hash     covered: B2__base.sql V3__t.sql
$ atlas migrate apply ...     No migration files to execute      exit=0
                              revisions unchanged        tables: p,q,t   <- base never ran
```

`ptah-compat` runs the baseline in both cases, because it keys pending-ness on the recorded set.

**B. And CE's execution of a baseline can itself be destructive**, which is why "match Atlas" is not automatically the safe direction:

```
V1__p (CREATE TABLE p), V2__q applied
add B3__base.sql containing CREATE TABLE p (id int);  -- p already exists
$ atlas migrate apply ...
  -- migrating version 3 -> CREATE TABLE p (id int);
     table p already exists
  Error: sql/migrate: executing statement "CREATE TABLE p (id int);" from version "3": table p already exists
  exit=1        revisions: 1=p, 2=q  (no row 3)      tables: p,q
```

So CE both silently skips baselines in some shapes and destructively re-runs them in others, depending only on how the version token sorts as a string.

## CE's own inconsistency, worth knowing before copying anything

After that failure, `migrate status` reports **`No migration applied yet`** despite two applied rows:

```
$ atlas migrate status --url sqlite://x.db --dir 'file://d?format=flyway'
Migration Status: PENDING
  -- Current Version: No migration applied yet
  -- Next Version:    3
  -- Executed Files:  2
  -- Pending Files:   2
```

`Executed Files: 2` and `Current Version: No migration applied yet` cannot both be right.

## Supporting measurements

- CE never deletes revision rows; superseded files' rows linger indefinitely (rows 1 and 2 above remain after a baseline supersedes them).
- Baselines are recorded with `type = 2`, identical to ordinary migrations. There is no distinct baseline revision type.
- CE has no file identity: renaming `V3__t.sql` to `B3__t.sql` keeps the token `3`, and the next apply is a no-op.

## The decision

1. **Reproduce the skip** — drop-in fidelity for anyone migrating off Atlas; imports a shape where a baseline silently never runs.
2. **Keep the current louder behavior** — the baseline runs, which is what the operator wrote it for; diverges from Atlas on two measured shapes.
3. **Refuse loudly on exactly these shapes** — neither silent nor divergent-by-default, but a third behavior belonging to neither tool.

Pinned meanwhile by `TestCompatMigrateApply_FlywayBaselineBelowHighWaterMark_KnownDivergence`, so whichever way this goes is a visible test change.

Refs #982, #1002.

denisvmedia:
Closes #1003.

`ptah-compat migrate apply` now refuses a converted Flyway directory's surviving
baseline (`B*__*.sql`) when the recorded history has already passed its position,
instead of either silently skipping it (what the pinned community binary v1.3.0
does for that shape) or running it blindly (what compat did, which fails on
`table already exists` and strands a dirty revision). The reasoning, the measured
Atlas CE matrix and the deliberate abstentions are in the first commit's message.

The second and third commits fix a regression that the first one introduced, and
correct two shipped claims that measurement contradicts.

## The regression: zero-padded versions were read as the same version

`flywaySequenceVersion` projected the baseline's token through
`flywayAtlasVersion`, and `flywayOrderingKey` scores `"2"`, `"02"` and `"002"`
**identically** — components `{2}`, one key, tie slot 0. The baseline's
synthesized version therefore collided with the revision recorded for
`V02__q.sql`, and the "a migration of the same version is already applied" clause
refused directories that both the pinned binary and the commit this branch forks
from (`c0f86693`) apply successfully.

Twelve cells, re-measured for this PR against three binaries — the pinned
community build v1.3.0, a compat built from `c0f86693`, and a compat built from
this branch's head. sqlite, `--dir file://m?format=flyway`, `atlas.sum` written by
each tool for its own run, exit codes captured immediately, "executed" read back
from `sqlite_master`:

| Applied | Baseline | pinned v1.3.0 | compat @ c0f86693 | compat @ this branch |
| `V01`, `V02` | `B1__base.sql` | 0, executed | 0, executed | 0, executed |
| `V01`, `V02` | `B2__base.sql` | 0, executed | 0, executed | 0, executed |
| `V01`, `V02`, `V03` | `B1__base.sql` | 0, executed | 0, executed | 0, executed |
| `V01`, `V02`, `V03` | `B2__base.sql` | 0, executed | 0, executed | 0, executed |
| `V01`, `V02`, `V03` | `B3__base.sql` | 0, executed | 0, executed | 0, executed |
| `V001`, `V002` | `B1__base.sql` | 0, executed | 0, executed | 0, executed |
| `V001`, `V002` | `B2__base.sql` | 0, executed | 0, executed | 0, executed |
| `V001`, `V002` | `B01__base.sql` | 0, executed | 0, executed | 0, executed |
| `V001`, `V002` | `B02__base.sql` | 0, executed | 0, executed | 0, executed |
| `V05`, `V06`, `V07` | `B5__base.sql` | 0, executed | 0, executed | 0, executed |
| `V05`, `V06`, `V07` | `B6__base.sql` | 0, executed | 0, executed | 0, executed |
| `V05`, `V06`, `V07` | `B7__base.sql` | 0, executed | 0, executed | 0, executed |

All three agree on every cell. Before the fix, the third column was **exit 1 with
nothing executed on all twelve** — for example `B2__base.sql is a Flyway baseline
and this database already has migration history ... version 4611686018427510315 —
a migration of the same version — is already applied`. The pinned binary treats
those tokens as distinct and says so on its own stdout: `Migrating to version 2
from 02`.

Zero-padded versions are standard Flyway practice, so this was not an edge case.

## The fix: ask identity on the token, order on the key

The ordering key is **not** changed. Collapsing `V02` and `V2` is correct for
order — Atlas CE runs them in one position — and the tie budget in
`flywayAtlasVersion` exists precisely because tokens that score identically still
have to be told apart by walk position. Identity is a different question, and the
refusal was asking it with the ordering answer.

`FlywaySurvivingBaseline` now reports `SameVersion`: the migrations whose version
token is the baseline's own, **byte for byte**, each paired with the Atlas version
a database that ran it recorded. The domain is this directory with the baseline
removed, because that is the only shape a revision row can have come from — a file
carrying the baseline's exact token is squashed out of the covered set by
construction (`file.version <= baseline.version` holds on equality), and
re-running the selection without the baseline puts it back at the tie slot it was
recorded under instead of assuming slot 0. A pre-baseline set this build cannot
convert yields nothing, because no build could have applied it either, so no
revision row of its making exists.

`flywaySequenceVersion` is gone. The reasoning is written out at
`flywaySameVersionMigrations` in `internal/atlasmigrateimport/importer.go` and
pointed at from `checkFlywayBaselineHistory`, which is where the next reader
arrives from a refusal.

## The discriminator, re-proved

A fix that quietly turned the refusal off would pass all twelve cells and fail
nothing else, so this is the check that separates the two. Reverting **only** the
call site in `cmd/atlas/migrate_apply.go` still reddens exactly the four rows it
reddened before, and no others:

```
--- FAIL: .../a_baseline_the_recorded_history_does_not_reach
--- PASS: .../a_baseline_that_squashes_the_whole_recorded_history
--- PASS: .../a_baseline_that_squashes_a_two-digit_recorded_version
--- PASS: .../the_same_directory_against_a_database_with_no_history
--- FAIL: .../a_baseline_whose_token_lands_between_two_applied_versions
--- FAIL: .../a_two-digit_baseline_between_two_applied_versions
--- FAIL: .../a_baseline_carrying_a_version_already_applied
--- PASS: .../a_baseline_beside_two-digit_zero-padded_versions
--- PASS: .../a_zero-padded_baseline_beside_three-digit_zero-padded_versions
--- PASS: .../a_baseline_matching_the_middle_of_a_zero-padded_run
--- PASS: .../the_printed_escape_hatch_runs_the_baseline_and_settles_the_directory
--- PASS: .../an_ordinary_migration_below_the_mark_keeps_the_out-of-order_refusal
--- PASS: .../an_out-of-order_migration_beside_a_baseline_reports_the_guard
```

Three rows are new, one per padding shape, each stating in a comment what it
prints if the fix is reverted:

- `a baseline beside two-digit zero-padded versions` (`V01`, `V02` + `B2`) —
  reverted: `B2__base.sql is a Flyway baseline ... version 4611686018427510315 —
  a migration of the same version — is already applied`, exit 1, tables `p`, `q`
  without `base`.
- `a zero-padded baseline beside three-digit zero-padded versions` (`V001`,
  `V002` + `B02`) — reverted: same message naming `B02__base.sql` and
  `4611686018427510315`, exit 1, tables `p`, `q` without `base`.
- `a baseline matching the middle of a zero-padded run` (`V05`, `V06`, `V07` +
  `B6`) — reverted: same message naming `B6__base.sql` and `4611686018427673531`,
  exit 1, tables `p`, `q`, `t` without `base`.

Mutation sweeps: `scripts/mutations/flyway-baseline-history.json` 11/11 killed
(four rules added, including "the same-version test is answered by the ordering
key, not the token", which is the regression itself and is killed by exactly the
three new rows), `scripts/mutations/flyway-importer-selection.json` 27/27 killed.

## Two shipped claims corrected

1. `docs/site/src/content/docs/versioned/integrity-and-safety.md` claimed *"A
   baseline that squashes away everything the database has applied still runs on
   both tools"*. That went false when the refusal landed: `B2__base.sql` on a
   database at `V1`, `V2` squashes both and is refused. The note now states the
   exception, and pins the zero-padding behavior with Atlas CE's own stdout.

2. The refusal asserted `version 4611686018427510315 — a migration of the same
   version — is already applied`. That number appears in no file name and in no
   Atlas output — the community binary reports the same two migrations as `2` and
   `02`. Both blocking lists now name files:

   ```text
   the baseline carries version "2" and no revision here records it, but this
   database has already applied:
     V2__q.sql — a migration carrying the same version "2"
   ```

   and, for the covered clause:

   ```text
   the baseline carries version "10" and no revision here records it, but this
   database has already applied:
     V2__s2.sql — still covered by this directory
   ```

   Two assertions were added to existing rows so the naming is held, and two
   mutations cover it.

## Where else the conflation was looked for

Every other equality on a converted version is taken **within one conversion**,
where the tie index keeps distinct tokens distinct — `Covered` vs.
`AppliedMigrations`, the `SelectedVersions` membership test, and the
out-of-order exemption all compare versions that came out of the same
`flywayConvertedVersions` call. `flywaySequenceVersion` was the only synthesizer
in the current encoding, and it is gone.

One other place collapses the same tokens and is **deliberately left alone**: the
frozen pre-#982 base-100 encoding in `internal/atlasmigrateimport/flywaylegacy.go`
also scores `"02"` and `"2"` to `20000`. Its comparison is per-file across
encodings, and the old build genuinely collapsed them, so reproducing that is the
whole point of the reconstruction. The one shape where it could misreport —
`V2__a.sql` beside `V02__b.sql`, both recorded by a legacy build — could not have
been applied by that build either, since it converted both to one version. Not
measured, not changed.

## Deliberately left open

Everything the first commit lists (a database with no recorded history still runs
the baseline; a baseline that squashes away the whole recorded history still runs;
`--exec-order=non-linear` executes it; the linear guard speaks first when another
pending migration is out of order), unchanged.

Separate and still open: with `V2` applied, an **ordinary** `V10__y.sql` is
refused by the pinned binary and applied here at exit 0. No baseline is involved —
it is the projection of Flyway tokens onto int64 numerically where Atlas CE
compares the token strings. That is filed as #1098 and is **not** folded into this
PR.

## Gates

`go build ./...`, `go vet ./...`, `go test ./... -count=1`, `golangci-lint run
./...`, `scripts/check-test-style.sh`, `scripts/check-public-api.sh`,
`scripts/check-public-api-snapshot.sh`, and from `docs/site`
`node scripts/check-style.mjs` and `node scripts/build-feature-matrix.mjs
--check` — all green.


denisvmedia:
CI note — the one red check is not this change.

`Build versioned site` fails in `check:exit-codes`:

```
docs/site/src/content/docs/reference/exit-codes.md is missing 1 exit-code reference row(s):
- | `ptah schema test` | Every schema test case passed. | ...
```

Master is already red on the same job with the same message. The docs workflow's
last two runs on `master` — #786 at `1ae8306a`, which is this PR's base sha, and
#785 at `e725ba98` — both fail in `Build versioned site`, and the failing job's
log on #786 (`actions/runs/30816707395`, job `91696177560`) is byte-for-byte the
one quoted above. Nothing in this PR touches `reference/exit-codes.md`,
`ptah schema test`, or any command registration; the diff's only docs file is
`versioned/integrity-and-safety.md`.

Locally the check passes on this branch and on the checkout it forks from:

```
docs/site $ node scripts/check-exit-codes.mjs
check-exit-codes.mjs: OK (60 reference rows)
```

which is expected — `ptah schema test` landed on master after `c0f86693`, and the
row for it was never added to `exit-codes.md`. That belongs to whichever PR added
the command, not here.

`Documentation style` (`check-style.mjs`) and the feature-matrix check, which are
the gates this PR's docs edit is under, both pass.

PR #1124, opened 2026-08-03
feat(compat): schema-side Atlas flags, and a named decision on --export/--web (rebased, stacked on #1119)

denisvmedia:
Replaces #1115, which conflicted with master and never ran CI. The head branch of an open PR cannot be repointed through the API, so this is a fresh PR from the rebased branch. **#1115 carries the feature rationale and the full measurement tables — read it for the "why"; this description covers what the rebase changed.**

Refs #951.

> **Stacked on #1119, not on master.** `base` is `agent/951-named-lock-family-rebased-onto-1117`. This branch and #1119 both touch `cmd/atlas/schema_apply.go`, `cmd/atlas/atlas_test.go` and both feature-matrix files. Rebasing onto plain master would have left a known conflict between the two — the exact condition this rebase exists to remove — so the two are stacked. GitHub retargets this to `master` automatically when #1119 merges. **Merge #1119 first.**

## What conflicted, and how it was resolved

Seven files conflicted. In four of them the merge base proved the *branch's* text was the older copy and master had moved, so master's description was kept and only this branch's genuinely-new material was added:

| File | Conflict | Resolution |
| --- | --- | --- |
| `cmd/atlas/atlas.go` | Master (#1110) reworded the `schema test` verb comment to "desired-schema root"; branch still said "Go-annotation root directory" and added an `-s/--schema` paragraph | Master's wording + the branch's new paragraph |
| `cmd/atlas/schema_apply.go` (help) | Master (#1113) changed empty-selection behavior to *refuse*; the branch text still said it "reports a synced schema" | Master's sentence + the branch's `lint` policy paragraph |
| `cmd/atlas/schema_apply.go` (flags) | Master side carried #1119's `--lock-name`/`--skip-lock`; branch added `--skip-lint` | All three registered |
| `cmd/atlas/schema_diff.go` | Same empty-selection drift | Master's sentences + the branch's `--export` refusal note |
| `cmd/atlas/schema_inspect.go` | Master's selector description vs the branch's older copy | Master's description + the branch's `--output` and `--web` paragraphs |
| `cmd/schema/test.go` | #1110's source classification vs this branch's schema selection | Genuine semantic merge — see below |
| feature-matrix `.json` + `.md` | Row edits on both sides | Merged per row, `.md` regenerated |

### Two stale claims not carried forward

The branch's feature-matrix rows asserted things master has since falsified. Carrying them would have re-published false statements:

[...]

PR #1125, opened 2026-08-03
feat(atlas): accept and drop unknown schema-HCL names, as Atlas CE does (#1016) — rebased, stacked on #1124

### LINKED ISSUE #1016: Schema HCL: CE silently ignores unknown top-level blocks and column attributes (`annotation`, `invisible`), ptah-compat rejects them with exit 1

denisvmedia:
Replaces #1114, which conflicted with master and never ran CI. A PR's head branch cannot be repointed through the API, so this is a fresh PR from the rebased branch. **#1114 carries the full measurement, the 56-fixture loose/strict tables and the mechanism rationale — read it for the "why"; this description covers what the rebase changed.**

Closes #1016.

> **Stacked on #1124 → #1119 → master.** `base` is `feat/951-schema-side-compat-flags-rebased`. This branch shares `cmd/atlas/schema_apply.go` with #1119 and additionally `schema_diff.go` / `schema_inspect.go` with #1124, so all three form one linear stack rather than three branches that would conflict pairwise. **Merge order: #1119, then #1124, then this.** GitHub retargets each automatically as the one below it merges.

## What conflicted, and how it was resolved

Four files, and every one was a pure union — both sides adding a field to the same struct or struct literal. Nothing was dropped and no side was taken wholesale.

| File | Conflict | Resolution |
| --- | --- | --- |
| `internal/atlassource/resolve.go` | Master added `Schemas` to `ResolveOptions`; this branch added `IgnoreUnknownHCLNames` | Both fields |
| `internal/atlasschema/diff.go` (options) | Master added `Diagnostics`; branch added `IgnoreUnknownHCLNames` | Both fields |
| `internal/atlasschema/diff.go` (call) | Master added `Schemas` + its rationale comment; branch added `IgnoreUnknownHCLNames` | Both, comment kept |
| `cmd/atlas/schema_diff.go` | Master added `Diagnostics: cmd.ErrOrStderr()`; branch added `IgnoreUnknownHCLNames: true` | Both |
| `internal/atlashcl/atlashcl.go` | Master added `filename:` to the `parser` literal; branch added `tolerant:` / `recordIgnored:` | Both — the merge base has no `filename` field, which confirms master added it rather than this branch removing it |

## Reverted-hunk sweep

Files changed against the stacked base: **22**, an exact set match with PR #1114's original 22 — no strays, nothing outside scope, neither exit-code file touched.

## Discriminators re-proved after the rebase

Applied to the rebased tree, run, reverted. Tree clean afterwards; nothing staged during the sweep.

| Mutation | Red | Matches #1114 |
| --- | --- | --- |
| `tolerant: false` (tolerance off), oracle enabled | **62 leaf, 12 top-level** across `internal/atlashcl`, `internal/schemafile`, `cmd/atlas` | yes, exactly |
| `typeKeywordValue = cty.DynamicVal` (the exact regression class) | **4 leaf, 2 top-level** — `TestDroppedBodyScopeHoldsOnlyKnownOpaqueValues/{string,int,bool}` + `…StillEvaluatesTheBody/type_keyword_is_not_a_number` | yes, exactly |
| scope emptied entirely | 5 leaf, 7 top-level, including both `cmd/atlas` surface tests | yes |
| `IgnoreUnknownHCLNames: true` hard-wired in `internal/atlassource/resolve.go` | `TestSchemaDiffUnknownHCLNamesSplitByCommandTree` only | yes |
| `IgnoreUnknownHCLNames: true` hard-wired in `internal/atlasschema/inspect_source.go` | `TestSchemaInspectUnknownHCLNamesSplitByCommandTree` only | yes |

The last two stay separate rows for the reason #1114 gives: `schema inspect` and `schema diff` reach the loader through different resolver calls, and each mutation is caught only by its own test.

### One measurement worth recording

The tolerance mutation first reported **48 leaf / 9 top-level**, not 62/12. The difference is not a dropped hunk — it is that the three `TestOracle*` tests **skip silently** when `PTAH_ATLAS_ORACLE` is unset, so their failures never appear and the count quietly drops. With the pinned binary exported the count is exactly 62/12. A mutation sweep run without that variable would have under-reported this branch's coverage by 14 leaf subtests while looking perfectly healthy.

## Oracle conformance

Re-measured against the pinned community binary by absolute path:

```
PTAH_ATLAS_ORACLE=…/ptah-atlas-conformance/bin/atlas go test -count=1 -v -run '^TestOracle' ./internal/atlashcl
ok   go.5x5.cz/ptah/internal/atlashcl
35 leaf subtests PASS, 0 FAIL, 0 SKIP
```

35 leaf subtests with zero skips, matching #1114 exactly. `.github/workflows/atlas-oracle.yml` still names all four `TestOracle*` tests and still asserts the count is 4, so the skip-detection guard survived the rebase.

## Gates

| Gate | Result |
| --- | --- |
| `go build ./...` | exit 0 |
| `go vet ./...` | exit 0 |
| `go test ./... -count=1` | exit 0, 0 FAIL lines |
| `golangci-lint run ./...` | `0 issues.` — full tree, and clean this time; the foreign-path findings #1114 reported are not present |
| `scripts/check-public-api-snapshot.sh` | exit 0, no `--update` needed |
| `build-feature-matrix.mjs --check` | `OK (162 rows)` |
| docs `check:style`, `check:core-doc-links`, `check:page-health`, `check:links`, `check:redirects` (+ selftests) | all PASS |
| docs `check:exit-codes` | **FAIL — inherited from master**, fixed in #1122 |

`check:exit-codes` fails identically on pristine master `1ae8306a` (#1110 edited `docs/exit_codes.md` without updating its site mirror). This branch touches neither file.


denisvmedia:
Correction to the gates table: the `check:exit-codes` failure was fixed by **#1121**, not #1122. The two PRs made the identical one-row sync of the site mirror against `docs/exit_codes.md`; #1121 merged and #1122 was closed rather than landing an empty diff.

This branch initially showed **Build versioned site** red for a purely structural reason rather than any defect of its own: its base is #1124's branch, which did not yet carry the fix, while #1124 itself was green because *its* base (#1119) already did. The fix has now been propagated down the stack — #1119 → #1124 → here — so all three carry it.

denisvmedia:
**A latent defect this branch has been carrying, found by its first-ever CI run.**

The `differential and corpus` job went red on `TestOracleAcceptsAndDropsUnknownSchemaHCLNames/top-level_annotation_block`:

```
c.Assert(withDDL, qt.Equals, withoutDDL)
got:  "Notice: This Atlas edition lacks support for ...\n\ntable \"t\" { ... }"
want: "table \"t\" { ... }"
```

This is not from the rebase. `internal/atlashcl/oracle_conformance_test.go` is byte-identical between #1114's original head `25bc7b4f` and the rebased commit, and the four conflicts I resolved were struct-field unions in `atlassource/resolve.go`, `atlasschema/diff.go`, `cmd/atlas/schema_diff.go` and `atlashcl/atlashcl.go` — none of them touches the conformance harness.

It is also not inherited from master: the failing test is **new in this branch**, which is why the same job is green on #1124. The branch had never run CI before, because it was CONFLICTING the whole time.

**Cause.** Each row runs the oracle twice and asserts the two captures are equal. The binary prints an edition notice on its first `schema inspect` in a fresh environment and on that run only, so it attaches to the first invocation of the pair and not the second. The signature is unmistakable: the **first** subtest failed and every later one passed, because the notice was spent on the job's first invocation. A developer machine has already spent it, which is why this passed locally for everyone who ran it.

`requireSchemaOracle` does not absorb it — it runs `version`, whose stdout it parses, and the notice does not appear there.

**Fix.** A throwaway `schema inspect` runs once before the first measurement. Matching the notice by its wording was rejected: that text is not part of any contract, so a matcher would rot silently the next time it is reworded and this run would go green while comparing whatever the new text happened to be.

**Verified with a test double** that delegates every call to the pinned binary and emits a notice on its first `schema inspect` only:

| | result |
| --- | --- |
| before the fix, against the double | FAIL — first subtest, same assertion, same message as CI |
| after the fix, against the double | 35 leaf PASS, 0 FAIL, 0 SKIP |
| after the fix, against the pinned binary | 35 leaf PASS, 0 FAIL, 0 SKIP |

And the gate was checked for vacuity rather than assumed intact: with the warm-up in place, reverting `tolerant` to `false` still turns the conformance run red at **14 leaf / 3 top-level** subtests, the same delta the oracle-enabled sweep reports.

All other gates re-run after the change: `go build`, `go vet`, `go test ./...` (0 FAIL lines), `golangci-lint` (`0 issues.`), `check-test-style.sh`, both public-API gates — all green.

This is a change beyond the rebase, in this branch's own file. Flagging it explicitly rather than folding it in quietly.

PR #1132, opened 2026-08-03
Decide Flyway linearity on the version token, not the ordering key (#1098)

### LINKED ISSUE #1098: migrate apply: a Flyway directory reaching version 10 applies out of order where Atlas CE refuses (no baseline involved)

denisvmedia:
Closes #1098.

A converted Flyway directory carries two orders and they are not the same order.
`ptah-compat migrate apply` had only one of them, so a Flyway project's tenth
migration applied at exit 0 where the pinned community binary exits 1.

Rebased onto master this branch met #1003, which landed while it waited. Both
changes answer the same question — "was this converted Flyway migration authored
out of order?" — in different places. That interaction is settled below, in
**The #1003 interaction** and **What the token comparison still buys**.

## The reproduction, before

Pinned community binary at `ptah-atlas-conformance/bin/atlas`, which reports
`atlas community version v1.3.0`. sqlite, `--dir file://m?format=flyway`,
`atlas.sum` written by each tool for its own run, exit codes captured
immediately, tables read back from `sqlite_master`. `ptah-compat` built from
`master`.

`V2__s2.sql` applied, then `V10__y.sql` added:

```
$ atlas migrate apply --url sqlite://ce.db --dir 'file://m?format=flyway'
Error: migration file V10__y.sql was added out of order. See: https://atlasgo.io/versioned/apply#non-linear-error
exit=1        tables: s2

$ ptah-compat migrate apply --url sqlite://ptah.db --dir 'file://m?format=flyway'
Migrating to version 4611686018427836747 from 1 pending migrations.
Migration complete. Current version: 4611686018427836747
exit=0        tables: s2,y        <- executed what the oracle refuses
```

## The reproduction, after

Same fixture, `ptah-compat` built from this branch:

```
$ ptah-compat migrate apply --url sqlite://ptah.db --dir 'file://m?format=flyway'
Error: error applying migrations: out-of-order pending migrations for current
version 4611686018427510315 (source version "2"): 4611686018427836747
(source version "10") (use --exec-order=non-linear to apply or
--exec-order=linear-skip to ignore)
exit=1        tables: s2
```

The mirror shape (`V10` applied, `V2__y.sql` added) already matched and still
does, and a fresh apply of a directory holding both still runs version 2 then
version 10 at exit 0 on both tools.

## Why

Two comparisons decide two different questions about the same directory:

- **ORDER**, for `atlas.sum` and for execution, is **numeric** on the version
  components. `flywayOrderingKey` reproduces exactly that, and it is measured to
  match — a fresh apply of `V2__a.sql` beside `V10__b.sql` runs version 2 then
  version 10 on the pinned binary. #982 converged on it and it is right.
- **LINEARITY** — "was this file added after everything already applied" — is
  decided on the version **token as a string**, where `"10"` sorts below `"2"`.

The apply path keyed pending-ness and the linear guard on the ordering int64,
which answers the second question with the first one's answer. This is the same
class of mistake PR #1123 had to fix inside #1003: ordering and identity are
different questions, and here ordering and linearity are too.

The rule is derived from the pinned binary by measurement alone — 20 probe
cells for the rule itself, then the grids below, every cell explained. Nothing
here rests on reading that binary's source.

## What changed

The token is carried **alongside** the executed version, never replacing it —
renumbering would strand every revision row already recorded, which is the cost
`checkLegacyFlywayRevisions` exists to document.

- `atlasmigrateimport.FlywaySourceVersions` maps each converted version to Atlas
  CE's version string for its file. Nil for every other layout.
- `Migrator.WithSourceVersions` hands that map to the linear guard, which now
  refuses a pending migration when its version sorts **below the current one**
  **OR** its token does not sort **above every applied token**.
- The comparison runs inside the migration lock, on the applied set the migrator
  reads there, so a concurrent writer cannot move the mark between the check and
  the run — the same reasoning `--to-version` was built on in #1117.

The two halves are unioned and neither subsumes the other: `V3` added to a
database holding `V2` and `V10` is caught only by the numeric one, `V10` added to
a database holding `V2` only by the textual one.

## The #1003 interaction

#1003 refuses a surviving baseline whose position the recorded history has
already passed. This branch teaches the linear guard to compare the source
token. Rebased, two of this branch's own rows went red, and one of them was the
decision: **`V2` applied, `B10__sqbase.sql` added** was asserted to stay exempt
and apply, and on master it is refused.

The whole space where the two guards can meet, re-measured for this PR against
three binaries — the pinned community build v1.3.0, a compat built from master
at `2a189a46`, and a compat built from this branch's head. Same protocol as
above; "answered by" names the guard that produces compat's verdict.

| applied | added | pinned v1.3.0 | compat @ master | compat @ head | answered by |
| `V2` | `B10` | 0, skipped silently | 1 | 1 | #1003 |
| `V2` | `B1` | 0, skipped silently | 1 | 1 | #1003 |
| `V2` | `B3` | 0, executed | 0, executed | 0, executed | — |
| `V1`,`V2`,`V3` | `B2` | 0, skipped silently | 1 | 1 | #1003 |
| `V1`,`V2` | `B2` | 0, skipped silently | 1 | 1 | #1003 |
| `V1`,`V2` | `B3` | 0, executed | 0, executed | 0, executed | — |
| `V02` | `B10` | 0, executed | 0, executed | 0, executed | — |
| `V2` | `B02` | 0, skipped silently | 1 | 1 | #1003 |
| `V01`,`V02` | `B2` | 0, executed | 0, executed | 0, executed | — |
| `V2`,`V10` | `B11` | 1, added out of order | 1 | 1 | #1003 |
| `V2` | `B10`, `V11` | 1, names `V11` | 1 | 1 | #1003 |
| `V9` | `B10` | 0, skipped silently | 1 | 1 | #1003 |
| `V2`,`V10` | `B3` | 0, executed | 0, executed | 0, executed | — |
| `V2`,`V10` | `B1` | 0, skipped silently | 1 | 1 | #1003 |
| `V10` | `B2` | 0, executed | 0, executed | 0, executed | — |
| `V2` | `B10`, `V1` | 0, skipped silently | 1 | 1 | #1003 |
| `V02`,`V03` | `B10` | 0, executed | 0, executed | 0, executed | — |
| `V2` | `B10` `--dry-run` | 0, skipped silently | 1 | 1 | #1003 |
| `V2` | `B10` `--exec-order=non-linear` | 0, skipped silently | 0, executed | 0, executed | — |
| `V2` | `B10` `--exec-order=linear-skip` | 0, skipped silently | 0, executed | 0, executed | — |
| `V2` | `V10` | 1, added out of order | **0, executed** | 1 | source token |
| `V2` | `R__r` | 1, added out of order | **0, executed** | 1 | source token |
| `V2` | `V1` | 0, skipped silently | 1 | 1 | numeric |
| `V2`,`V10` | `V3` | 1, added out of order | 1 | 1 | numeric |
| `V2` | `V3` | 0, executed | 0, executed | 0, executed | — |
| `V01` | `V2` | 0, executed | 0, executed | 0, executed | — |

### The decision

**The refusal is correct and this branch's test expectation was written against
a world that no longer exists.** The row now asserts the refusal.

The expectation matched **neither** tool. The pinned binary exits 0 having run
nothing; the row asserted exit 0 having *run the squash on top of the history it
squashes*. That is the third behavior #1003 examined and rejected.

The reason the token comparison does **not** subsume #1003's refusal is visible
in the grid and provable from the selection rule. Split the baseline rows by
whether the source comparison would flag the baseline at all — that is, whether
its token sorts at or below the highest applied token:

- **Token at or below the mark**: `V2`+`B10`, `V2`+`B1`, `V1,V2,V3`+`B2`,
  `V1,V2`+`B2`, `V2`+`B02`, `V2,V10`+`B11`, `V2`+`B10,V11`, `V9`+`B10`,
  `V2,V10`+`B1`, `V2`+`B10,V1`. **All ten are refused by #1003's guard**, on
  master and on this branch, and the message is #1003's in every one.
- **Token above the mark**: `V2`+`B3`, `V1,V2`+`B3`, `V02`+`B10`,
  `V01,V02`+`B2`, `V2,V10`+`B3`, `V10`+`B2`, `V02,V03`+`B10`. **All seven
  execute on all three binaries**, and the source comparison never flags them.

That is not a coincidence of the fixtures. Squashing compares the two version
tokens as strings, so a file whose token outranks the baseline's survives into
`Covered` and one carrying the baseline's own token lands in `SameVersion`; if
the baseline's token is at or below the highest applied token, the file holding
that mark must be in one of the two, and #1003's guard fires. So the source
comparison can never be the deciding refusal for a surviving baseline.
Narrowing #1003's guard would therefore have removed a refusal and replaced it
with nothing for those ten cells, and re-opened the shapes it was built for.

The exemption is still load-bearing — for the **numeric** half. A surviving
baseline is converted into the band below every survivor whatever its own token,
so without the exemption the seven executing rows above would be refused,
including the twelve zero-padded cells PR #1123 fixed.

## What the token comparison still buys

Every cell in the table below. None of them has a `B` file in it, and #1003's
guard returns nil for all of them — it requires a surviving baseline. The two
guards partition the space rather than overlapping it.

| applied | added | pinned v1.3.0 | compat before | compat now |
| `V2` | `V10` | 1 | 0, executed | 1 |
| `V3` | `V10` | 1 | 0, executed | 1 |
| `V9` | `V10` | 1 | 0, executed | 1 |
| `V1`, `V2` | `V10` | 1 | 0, executed | 1 |
| `V2`, `V3` | `V10` | 1 | 0, executed | 1 |
| `V2`, `V10` | `V11` | 1 | 0, executed | 1 |
| `V2`, `V10` | `V15` | 1 | 0, executed | 1 |
| `V2` | `V100` | 1 | 0, executed | 1 |
| `V2` | `V10`, `V11` | 1 | 0, executed | 1 |
| `V2` | `V3`, `V10` | 1 | 0, executed | 1 |
| `V2` | `R__r` | 1 | 0, executed | 1 |
| `V2` | `R1__r` | 1 | 0, executed | 1 |
| `V2` | `R9__r` | 1 | 0, executed | 1 |
| `V2` | `Rfoo__r` | 1 | 0, executed | 1 |
| `V9` | `R2__r` | 1 | 0, executed | 1 |

Repeatables are the same defect wearing a different hat. Atlas gives every
repeatable the version string `""`, which sorts below every token, so it refuses
one added to a database with any history whatever the file is called. Ptah placed
them on a reserved slot above every versioned migration, where the numeric
comparison sees nothing wrong.

Cells that already matched and are unchanged: `V2`+`V3`, `V2`+`V20`, `V10`+`V11`,
`V01`+`V2`, `V1`+`V10`, `V0`+`V10`, `V1.1`+`V1.10`, `V10`+`V2`, `V2`,`V10`+`V3`,
`V10`,`V11`+`V9`, and every fresh-directory shape.

## Nine more cells, found by hunting for the direction that must never happen

The rule the whole branch exists to protect is one-sided: compat may be stricter
than the pinned binary, never looser. A sweep of 189 `(applied, added)` pairs
crossing nine applied sets with twenty-one added sets found nine cells where
compat exits 0 and the pinned binary exits 1 — all of them a **surviving
baseline plus a repeatable**:

| applied | added | pinned v1.3.0 | compat before | compat now |
| `V1` | `B10`, `R__r` | 1 | 0, executed | 1 |
| `V1` | `B3`, `R__r` | 1 | 0, executed | 1 |
| `V2` | `B3`, `R__r` | 1 | 0, executed | 1 |
| `V10` | `B3`, `R__r` | 1 | 0, executed | 1 |
| `V2`,`V10` | `B3`, `R__r` | 1 | 0, executed | 1 |
| `V01`,`V02` | `B10`, `R__r` | 1 | 0, executed | 1 |
| `V01`,`V02` | `B3`, `R__r` | 1 | 0, executed | 1 |
| `V05`,`V06`,`V07` | `B10`, `R__r` | 1 | 0, executed | 1 |
| `V05`,`V06`,`V07` | `B3`, `R__r` | 1 | 0, executed | 1 |

```
$ atlas migrate apply --url sqlite://ce.db --dir 'file://m?format=flyway'
Error: migration files B3__nb3.sql, R__nr.sql were added out of order.
exit=1        tables: a2

$ ptah-compat migrate apply --url sqlite://ptah.db --dir 'file://m?format=flyway'
Migrating to version 9223372036854775807 from 2 pending migrations.
exit=0        tables: a2,nb3,nr
```

They are the same repeatable defect, hidden behind a squash. The source
comparison's mark is the highest token among the **applied** versions; a
surviving baseline retires the applied file from the covered set, its token goes
with it, the lookup reports "no mark at all" and every pending file passes. The
cells are equally loose on master, so this is not a regression the branch
introduced — but shipping while they stood would have shipped the claim that a
repeatable is refused "whatever the file is called" against a measurement that
says otherwise.

`FlywaySourceVersions` now also carries the tokens of the migrations the
surviving baseline squashed, keyed by the Atlas version a database that ran them
recorded. The domain is this directory **without** the baseline — the same
domain and the same argument `flywaySameVersionMigrations` uses in #1123: it is
the only shape such a revision row can have come from, and re-running the
selection with the baseline removed puts each file back at the tie slot it was
recorded under rather than assuming slot 0.

It only **fills gaps**. The executed mapping is written first, so a version this
directory still executes never takes a squashed file's token — which matters,
because the two selections do collide: `B02__base.sql` beside `V2__a.sql` and
`V02__b.sql` leaves `V2` holding the tie slot `V02` had. Filling gaps can only
add a token, which can only raise the mark, which can only refuse more; a file
whose token does not sort strictly above the highest applied token is one Atlas
CE never executes, so refusing more is the safe direction. One removal, not a
transitive reconstruction: a superseded baseline's squashes are not recovered,
which is the conservative half again.

After the fix the sweep reports **zero** cells where compat exits 0 and the
pinned binary exits 1, and the nine rows above are the **only** rows in it whose
answer changed.

## The message

`--exec-order` is unchanged, but the error drops the word "below" for a
converted directory, because a version can be refused while sorting **above**
the current one. It names the **source version** because the int64 appears in no
file name and in no Atlas output — PR #1123 corrected the same defect in the
#1003 refusal. The **native** wording is byte-for-byte unchanged, and
`TestCompatMigrateApply_GooseKeepsTheOutOfOrderGuard` holds it with both a
positive and a negative assertion.

`--exec-order=linear-skip` now leaves refused migrations unapplied instead of
executing them, which is what the pinned binary does with the same input. It
reads the verdict the linear branch computed rather than re-deriving it
numerically; re-deriving is how linear could refuse while linear-skip silently
ran the same file. A dry run refuses too, as the oracle does. Note that this is
a statement about the **linear guard's** verdict: #1003's baseline refusal
stands aside under both exec-order flags by design, so an exempt baseline still
runs under `linear-skip`, and a test row pins that.

`cmd/atlas/migrate_apply_flyway_source_order_test.go`, 20 rows. Every row states
this in a comment beside it.

Rows that go red on a revert of the token comparison:

- `a tenth migration added under an applied second` — exit 0, `Migrating to
  version 4611686018427836747 from 1 pending migrations.`, tables `sq2`, `sq10`.
- `a tenth migration added under an applied ninth` — exit 0, tables `sq2`, `sq10`.
- `a hundredth migration added under an applied second` — exit 0, tables `sq2`,
  `sq100`.
- `a repeatable added to recorded history` — exit 0, tables `sq2`, `sqr`.
- `a repeatable whose own token sorts high is still refused` — exit 0, tables
  `sq2`, `sqr`.
- `a fifteenth migration under applied second and tenth` — exit 0, tables `sq10`,
  `sq15`, `sq2`.
- `linear-skip leaves the tenth migration unapplied` — still exit 0, but tables
  `sq2` **and** `sq10`: the skip executed what it claims to skip.
- `a dry run refuses rather than previewing the refused migration` — exit 0,
  `Would have applied 1 migrations.`

Rows added by the two decisions in this update:

- `a repeatable added beside a baseline that squashes the applied migration` —
  with the token recovery removed: exit 0, `Migrating to version
  9223372036854775807 from 2 pending migrations.`, tables `sq2`, `sqb3`, `sqr`.
- `a forward migration beside a baseline that squashes the applied migration` —
  green either way; it goes red if the recovered mark is the **baseline's** token
  rather than the squashed file's.
- `a surviving baseline below the mark is refused as a baseline, not as out of
  order` — with #1003's refusal removed for this shape: exit 0, `Migrating to
  version 448844 from 1 pending migrations.` / `Migration complete. Current
  version: 4611686018427510315`, tables `sq2` **and** `sqbase`. It also asserts
  the absence of the out-of-order wording, so it holds *which* guard answers.
- `linear-skip runs the exempt baseline rather than skipping it` — with the
  exemption reaching only the numeric half of the union: exit 0 with tables
  `sq2` alone, the baseline silently dropped. This replaces the old row's job:
  the exemption is no longer observable under `linear`, because #1003's guard
  answers there whether or not the baseline is exempt.

Rows green either way, which is their job — they fail if the new comparison
starts refusing ordinary sequences: `the next single-digit migration still
applies`, `a twentieth migration still applies over an applied second`, `a
twentieth migration still applies over applied second and tenth`, `an eleventh
migration still applies over an applied tenth`, `an unpadded second migration
still applies over an applied zero-padded first`, `a settled directory holding
both versions is not refused afterwards`, `non-linear runs the tenth migration
the refusal names`.

`an ordinary migration under an applied two-digit version` is the union's other
half: reverting only the numeric comparison prints exit 0 with tables `sq2`,
`sq3` and `sq10`.

`internal/atlasmigrateimport/flywayselection_test.go`,
`TestFlywaySourceVersionsCarriesSquashedTokens`, 5 rows asserting the token
**set** — the int64 keys are this build's projection while the tokens are Atlas
CE's own strings, so a row spelling out both would only prove the fixture was
written twice. Removing the recovery reddens exactly three: `a baseline that
squashes the only applied migration` prints `["3"]`, `a repeatable beside a
baseline that squashes the applied migration` prints `["", "3"]`, `a zero-padded
migration squashed by an unpadded baseline` prints `["1", "3"]`. `a padded and
an unpadded migration on one ordering slot` is the collision row: recovering by
overwrite rather than by filling gaps prints `["02", "02", "2", "3"]`. `a
baseline below every applied migration` and `a directory with no baseline at
all` are the controls.

Five assertions in `migrate_apply_legacy_flyway_test.go` moved from `below
current version` to `for current version` and gained source-version assertions —
three in the first commit, two more here that #1123 added to master while this
branch waited. The goose row in the same file keeps `below current version` and
also asserts the message carries **no** `source version` — that is the
discriminator for "the native wording did not move".

## Non-interference, proved with the inverse mutant

A control that expects execution cannot be validated by deleting the guard —
deleting a guard never makes it fire. `scripts/mutations/flyway-source-version-order.json`
carries an inverse mutant that makes the source comparison refuse **every**
pending migration carrying a token. In `TestCompatMigrateApplyFlywaySourceVersionOrder`
it reddens exactly the six rows that expect execution:

```
--- FAIL: .../the_next_single-digit_migration_still_applies
--- FAIL: .../a_twentieth_migration_still_applies_over_an_applied_second
--- FAIL: .../a_twentieth_migration_still_applies_over_applied_second_and_tenth
--- FAIL: .../an_eleventh_migration_still_applies_over_an_applied_tenth
--- FAIL: .../an_unpadded_second_migration_still_applies_over_an_applied_zero-padded_first
--- FAIL: .../a_forward_migration_beside_a_baseline_that_squashes_the_applied_migration
```

The two baseline rows stay green under it, and that is the exemption doing its
job: the mutant flags the baseline like everything else, and the exemption
removes it from the union before the verdict is read. It also reddens
`TestCompatMigrateApply_FlywayBaselineAfterApply`,
`TestCompatMigrateApply_LegacyFlywayDetectorDoesNotOverRefuse` and
`TestCompatMigrateApply_LegacyFlywayRefusalPrintsWorkingRecovery`, which is the
same property being held from three other directions.

Mutation sweep: `scripts/mutations/flyway-source-version-order.json` **14/14
killed**, no BUILD-FAILED and no PATCH-FAILED. Four rules are new: the token
recovery, its gap-fill discipline, the identity of the token it recovers, and
"the surviving-baseline refusal is removed, leaving the exemption to decide" —
the last is the discriminator for the decision above, and it reddens five
subtests across `TestCompatMigrateApplyFlywaySourceVersionOrder` and
`TestCompatMigrateApply_FlywayBaselineAgainstRecordedHistory`.

Each new mutation was also run alone and its reddened subtests read by name
rather than by count: the half-exemption mutant reddens exactly `linear-skip
runs the exempt baseline rather than skipping it`, and the removed-refusal
mutant reddens exactly `a surviving baseline below the mark is refused as a
baseline, not as out of order`.

## Not touched, and measured to be a different shape

golang-migrate, goose, dbmate and native Atlas directories are listed in
file-**name** order, which is the same textual order the linear comparison uses,
so the two never disagree there. On "2 applied, add 10" the pinned binary answers
all four with `No migration files to execute` at exit 0 rather than a refusal,
and compat's answers for those layouts are byte-identical before and after this
change. Only Flyway is sorted numerically while being compared textually.

## Docs

`docs/site/src/content/docs/atlas/migrate-commands.md` gains the two-comparison
rule and the refusal it produces, and its claim that Flyway repeatables "cannot
yet be executed as versioned migrations" is corrected, because it is measurably
false.

`docs/site/src/content/docs/versioned/integrity-and-safety.md` — the note the
#1003 refusal sends readers to — carried a claim that measurement contradicts:
"Atlas CE applies a below-mark baseline under that flag too", of
`--exec-order=non-linear`. That holds only for a baseline Atlas CE refuses as
out of order, not for one it skips silently, which is the first row of the
note's own table:

| applied | added | flag | pinned v1.3.0 |
| --- | --- | --- | --- |
| `V1`,`V2`,`V3` | `B2.5__base.sql` | (none) | 1, `added out of order` |
| `V1`,`V2`,`V3` | `B2.5__base.sql` | `--exec-order=non-linear` | 0, `Migrating to version 2.5 from 3` |
| `V2` | `B10__base.sql` | `--exec-order=non-linear` | 0, `No migration files to execute` |
| `V2` | `B10__base.sql` | `--exec-order=linear-skip` | 0, `No migration files to execute` |

`ptah-compat` executes the baseline in all four. The note now says so, and adds
`--exec-order=linear-skip` to the list of flags that leave the baseline runnable,
which it had omitted.

## Deliberately left open

- **`V2` applied, `V1` added.** The pinned binary exits 0 and drops `V1` silently.
  Ptah refuses, as it already did. That is the #1003 footgun and this change does
  not import it. Strict direction, permitted.
- **`V2` applied, `V02` added.** The pinned binary drops it silently; Ptah exits 1
  on a checksum mismatch, because the tie slot `V2` was recorded under moves when
  `V02` joins the run. Unchanged here, and it belongs to the identity family PR
  #1123 works in, not to ordering.
- **A surviving baseline the pinned binary skips silently is refused here**
  (`V2`+`B10` and nine relatives). #1003's decision, re-measured for this PR and
  unchanged. Strict direction, permitted.
- **An exempt baseline runs under `--exec-order=linear-skip` where the pinned
  binary skips it.** Both exit 0, so no parity rule is broken. It is #1003's
  abstention: with an exec order named explicitly the operator has already made
  the decision. Now pinned by a test row and stated in the docs note rather than
  left implicit.
- **When a baseline and an out-of-order migration are both pending, the two tools
  name different files.** `V2` applied plus `B10` and `V11`: the pinned binary
  names `V11`, Ptah names `B10`. Exit code and executed set match. Same class as
  `V2` applied plus `V3` and `V10`, where the pinned binary names both files and
  Ptah names only `V10`.
- **`MigrationStatus.OutOfOrderMigrations` is still the numeric verdict alone.**
  The linear guard's verdict is the union; the status field is not, so
  `migrate status` under-reports for a converted Flyway directory. It cannot
  produce a looser apply — every version the status field lists is also in the
  union — but it is an inconsistency worth its own issue.
- **The refusal names the numerically highest applied version as "current",
  while the comparison uses the textually highest token.** With `V2` and `V10`
  applied the message reads `for current version ... (source version "10")`
  while the mark was `"2"`. Both operands are correct for their own question;
  the message could say which. Not changed here.
- **Subdirectory shapes.** Every grid cell is a flat directory. The
  backwards-reaching squash rule compares a survivor's full **path** against the
  baseline's token, so the subsumption argument above is measured for flat
  directories and reasoned, not measured, for nested ones.
- **Non-Flyway layouts executing what the oracle silently drops** (both exit 0, so
  no parity-rule violation). Same class as #1003 and not decided here.
- **The stale encoding prose** two sentences further down the migrate-commands
  paragraph — "a fixed-width `major.minor.patch` int64" and "depends only on the
  version, never on the other files in the directory" — predates #982's bands and
  tie budget. Correcting it needs its own measurement.

## Gates

- `go build ./...` — ok
- `go vet ./...` — ok
- `go test ./... -count=1` — ok, no failures
- `go tool qtlint ./...` — ok
- `golangci-lint run ./...` — `0 issues.`
- `scripts/check-test-style.sh` — `OK (175 conditional groups, 45 white-box files)`
- `scripts/check-public-api.sh` — ok
- `scripts/check-public-api-snapshot.sh` — ok, no snapshot movement in this
  update (`FlywaySourceVersions` keeps its signature)
- `scripts/check-public-api-released.sh` — ok (the reported incompatibilities are
  the pre-existing approved pre-v1 list, unchanged by this branch)

Docs (`docs/site`), every script `package.json` defines:

- `check:style` — `OK (100 documentation files)`
- `check:style:selftest` — `OK (16 rule assertions via analyze())`
- `check:exit-codes` — `OK (60 reference rows)`
- `check:exit-codes:selftest` — `OK`
- `check:core-doc-links` — `OK (7 protected core references)`
- `check:page-health` — `OK (59 sidebar entries)`
- `check:links` — `OK (60 pages, 60 routes, 565 heading anchors)`
- `check:links:selftest` — `OK`
- `check:redirects` — `OK (30 redirects)`
- `check:redirects:selftest` — `OK`
- `versions:selftest` — `OK`
- `build-feature-matrix.mjs --check` — `OK (162 rows)`
- `check:responsive:selftest` — `OK (overflow, wide-table, tall-cell, and
  unstyled-page guards verified)`
- `check:responsive` — **actually run this time**, after `npm ci`,
  `npx playwright install chromium` and `npm run build` (61 pages):
  `OK (60 routes x 2 viewports, cells within 8 lines)`. Without those three it
  prints `dist not found; run "npm run build" first` and exits **1**, so the
  earlier report that it exits 0 unrun was wrong on this tree.

### DIFF
--- a/cmd/atlas/migrate_apply.go
+++ b/cmd/atlas/migrate_apply.go
@@ -276,34 +276,7 @@ func runAtlasMigrateApply(
 	}
 	defer dbschema.CloseAndWarn(conn)
 
-	// A surviving Flyway baseline is converted into the band BELOW every
-	// survivor, because Atlas CE emits and executes it first whatever its own
-	// version — measured, and measured to hold across runs, not only within one
-	// conversion. On a database that already has migrations recorded it
-	// therefore sorts below all of them, which the linear guard would read as
-	// "authored earlier". It is not: the position encodes sum order, not
-	// chronology. Exempting exactly that one version keeps every other
-	// out-of-order migration refused, which is what Atlas CE does too.
-	flywayBaseline, err := atlasmigrateimport.FlywaySurvivingBaseline(captured, resolvedDirFormat)
-	if err != nil {
-		return fmt.Errorf("atlas migrate apply --dir: %w", err)
-	}
-	var outOfOrderExempt []int64
-	if flywayBaseline != nil {
-		outOfOrderExempt = []int64{flywayBaseline.AtlasVersion}
-	}
-
-	// The linear guard's other operand (#1098). A converted Flyway directory is
-	// ORDERED numerically — V2 executes before V10, which is what the int64
-	// version encodes and what atlas.sum is written against — but the tool it
-	// came from decides whether a file "was added out of order" by comparing
-	// the version TOKEN as a string, where "10" sorts below "2". Carrying the
-	// token alongside the executed version lets the guard ask that question the
-	// way the source tool asks it, without renumbering anything already
-	// recorded. The comparison itself happens inside the migration lock, on the
-	// applied set the migrator reads there, so a concurrent writer cannot move
-	// the mark between the check and the run.
-	sourceVersions, err := atlasmigrateimport.FlywaySourceVersions(captured, resolvedDirFormat)
+	linearity, err := flywayLinearityOperands(captured, resolvedDirFormat)
 	if err != nil {
 		return fmt.Errorf("atlas migrate apply --dir: %w", err)
 	}
@@ -313,8 +286,8 @@ func runAtlasMigrateApply(
 		FS:                   migrationFS,
 		DryRun:               opts.dryRun,
 		ExecOrder:            execOrder,
-		OutOfOrderExempt:     outOfOrderExempt,
-		SourceVersions:       sourceVersions,
+		OutOfOrderExempt:     linearity.outOfOrderExempt,
+		SourceVersions:       linearity.sourceVersions,
 		TxMode:               txMode,
 		RevisionsSchema:      opts.revisionsSchema,
 		MigrationLockTimeout: migrationLockTimeout,
@@ -343,7 +316,7 @@ func runAtlasMigrateApply(
 	// baseline's band position as "authored earlier". Whether a baseline may
 	// run against a database that already has history is a separate question,
 	// and one Atlas CE answers three incompatible ways (stokaro/ptah#1003).
-	if err := checkFlywayBaselineHistory(flywayBaseline, execOrder, plan); err != nil {
+	if err := checkFlywayBaselineHistory(linearity.baseline, execOrder, plan); err != nil {
 		return err
 	}
 
@@ -653,4 +626,58 @@ func writeAtlasMigrateApplyFormat(
 		URL:    opts.url,
 		Result: result,
 	})
+}
+
+// flywayLinearityOperands returns everything the linear guard needs from a
+// converted Flyway directory: the surviving baseline, the versions whose
+// position below the current high-water mark is an artifact of the layout
+// projection rather than a claim about authoring order, and the source version
+// token each converted version came from.
+//
+// They are resolved together because they answer one question between them —
+// "was this file authored out of order?" — and it must be answered on the
+// operands the source tool decides it with.
+//
+// A surviving baseline is converted into the band BELOW every survivor, because
+// Atlas CE emits and executes it first whatever its own version — measured, and
+// measured to hold across runs, not only within one conversion. On a database
+// that already has migrations recorded it therefore sorts below all of them,
+// which the linear guard would read as "authored earlier". It is not: the
+// position encodes sum order, not chronology. Exempting exactly that one
+// version keeps every other out-of-order migration refused, which is what Atlas
+// CE does too.
+//
+// The tokens are the guard's other operand (#1098). A converted Flyway
+// directory is ORDERED numerically — V2 executes before V10, which is what the
+// int64 version encodes and what atlas.sum is written against — but the tool it
+// came from decides whether a file "was added out of order" by comparing the
+// version TOKEN as a string, where "10" sorts below "2". Carrying the token
+// alongside the executed version lets the guard ask that question the way the
+// source tool asks it, without renumbering anything already recorded. The
+// comparison itself happens inside the migration lock, on the applied set the
+// migrator reads there, so a concurrent writer cannot move the mark between the
+// check and the run.
+type flywayLinearity struct {
+	baseline         *atlasmigrateimport.FlywayBaseline
+	outOfOrderExempt []int64
+	sourceVersions   map[int64]string
+}
+
+func flywayLinearityOperands(
+	captured fs.FS,
+	format atlasmigrateimport.Format,
+) (flywayLinearity, error) {
+	baseline, err := atlasmigrateimport.FlywaySurvivingBaseline(captured, format)
+	if err != nil {
+		return flywayLinearity{}, err
+	}
+	operands := flywayLinearity{baseline: baseline}
+	if baseline != nil {
+		operands.outOfOrderExempt = []int64{baseline.AtlasVersion}
+	}
+	operands.sourceVersions, err = atlasmigrateimport.FlywaySourceVersions(captured, format)
+	if err != nil {
+		return flywayLinearity{}, err
+	}
+	return operands, nil
 }
--- a/cmd/atlas/migrate_apply.go
+++ b/cmd/atlas/migrate_apply.go
@@ -293,12 +293,28 @@ func runAtlasMigrateApply(
 		outOfOrderExempt = []int64{flywayBaseline.AtlasVersion}
 	}
 
+	// The linear guard's other operand (#1098). A converted Flyway directory is
+	// ORDERED numerically — V2 executes before V10, which is what the int64
+	// version encodes and what atlas.sum is written against — but the tool it
+	// came from decides whether a file "was added out of order" by comparing
+	// the version TOKEN as a string, where "10" sorts below "2". Carrying the
+	// token alongside the executed version lets the guard ask that question the
+	// way the source tool asks it, without renumbering anything already
+	// recorded. The comparison itself happens inside the migration lock, on the
+	// applied set the migrator reads there, so a concurrent writer cannot move
+	// the mark between the check and the run.
+	sourceVersions, err := atlasmigrateimport.FlywaySourceVersions(captured, resolvedDirFormat)
+	if err != nil {
+		return fmt.Errorf("atlas migrate apply --dir: %w", err)
+	}
+
 	plan, err := atlasmigrate.PrepareApply(cmd.Context(), conn, atlasmigrate.ApplyOptions{
 		Dir:                  dir,
 		FS:                   migrationFS,
 		DryRun:               opts.dryRun,
 		ExecOrder:            execOrder,
 		OutOfOrderExempt:     outOfOrderExempt,
+		SourceVersions:       sourceVersions,
 		TxMode:               txMode,
 		RevisionsSchema:      opts.revisionsSchema,
 		MigrationLockTimeout: migrationLockTimeout,
--- a/cmd/atlas/migrate_apply_flyway_source_order_test.go
+++ b/cmd/atlas/migrate_apply_flyway_source_order_test.go
@@ -1,0 +1,312 @@
+package atlas_test
+
+import (
+	"path/filepath"
+	"strings"
+	"testing"
+
+	qt "github.com/frankban/quicktest"
+)
+
+// These rows cover stokaro/ptah#1098: a converted Flyway directory is ORDERED
+// numerically and made LINEAR textually, and `migrate apply` used to answer both
+// questions with the numeric key.
+//
+// Measured on the pinned community binary, which reports
+// `atlas community version v1.3.0`, sqlite, `--dir file://m?format=flyway`,
+// atlas.sum written by each tool for its own run, exit codes captured
+// immediately. Every row's expectation is that binary's answer unless the
+// comment says otherwise:
+//
+//	applied      added            pinned v1.3.0                       here
+//	V2           V10              1, "was added out of order"         refused
+//	V9           V10              1, "was added out of order"         refused
+//	V2           V100             1, "was added out of order"         refused
+//	V2           R__r             1, "was added out of order"         refused
+//	V2           R9__r            1, "was added out of order"         refused
+//	V2, V10      V15              1, "was added out of order"         refused
+//	V2, V10      V3               1, "was added out of order"         refused
+//	V2           V3               0, executed                         executed
+//	V2           V20              0, executed                         executed
+//	V2, V10      V20              0, executed                         executed
+//	V10          V11              0, executed                         executed
+//	V01          V2               0, executed                         executed
+//	V2, V10      (none)           0, no pending files                 no-op
+//
+// Every refusing row above exited 0 here before the change and executed the
+// file — that is the defect, and it is what those rows print if the change is
+// reverted. The executing rows are green either way: their job is to fail if
+// the new comparison starts refusing ordinary sequences.
+
+type flywaySourceOrderOutput struct {
+	stdout string
+	stderr string
+	err    error
+}
+
+func (o flywaySourceOrderOutput) text() string {
+	return errorText(o.err) + o.stdout + o.stderr
+}
+
+type flywaySourceOrderRow struct {
+	name string
+	// applied are written, hashed and applied first; the row's second apply
+	// runs against the history they leave behind. Empty means no history,
+	// which is the shape with nothing for the source comparison to compare
+	// against.
+	applied []string
+	// added are written next, and the directory is re-hashed before the second
+	// apply.
+	added []string
+	// args are appended to the second `migrate apply` invocation.
+	args []string
+	// assert is the per-row wiring: refusing rows and executing rows do not
+	// share an assertion, and branching inside one body would hide which of the
+	// two a row is.
+	assert func(c *qt.C, dbPath string, out flywaySourceOrderOutput)
+}
+
+func TestCompatMigrateApplyFlywaySourceVersionOrder(t *testing.T) {
+	// refused asserts the refusal names both the mark it compared against and
+	// the migration it refused, by SOURCE version, and that the listed tables
+	// are exactly the ones the first apply left — nothing ran, because the
+	// refusal is raised while the migration set is selected.
+	refused := func(tables []string, sourceVersions ...string) func(*qt.C, string, flywaySourceOrderOutput) {
+		return func(c *qt.C, dbPath string, out flywaySourceOrderOutput) {
+			c.Assert(out.err, qt.IsNotNil, qt.Commentf("stdout:\n%s\nstderr:\n%s", out.stdout, out.stderr))
+			c.Assert(out.text(), qt.Contains, "out-of-order pending migrations for current version")
+			for _, source := range sourceVersions {
+				c.Assert(out.text(), qt.Contains, `(source version "`+source+`")`)
+			}
+			c.Assert(userTables(c, dbPath), qt.DeepEquals, tables)
+		}
+	}
+	executed := func(tables ...string) func(*qt.C, string, flywaySourceOrderOutput) {
+		return func(c *qt.C, dbPath string, out flywaySourceOrderOutput) {
+			c.Assert(out.err, qt.IsNil, qt.Commentf("stdout:\n%s\nstderr:\n%s", out.stdout, out.stderr))
+			c.Assert(userTables(c, dbPath), qt.DeepEquals, tables)
+		}
+	}
+
+	tests := []flywaySourceOrderRow{
+		{
+			// The issue's shape. Reverted: exit 0, "Migrating to version
+			// 4611686018427836747 from 1 pending migrations.", tables sq2 and
+			// sq10 — the migration the oracle refuses, executed.
+			name:    "a tenth migration added under an applied second",
+			applied: []string{"V2__sq2.sql"},
+			added:   []string{"V10__sq10.sql"},
+			assert:  refused([]string{"sq2"}, "2", "10"),
+		},
+		{
+			// Reverted: exit 0, tables sq2 and sq10. Separates "two digits" from
+			// "numerically far above the mark": 10 > 9 numerically, and "10" <
+			// "9" textually, so only the textual comparison refuses it.
+			name:    "a tenth migration added under an applied ninth",
+			applied: []string{"V9__sq2.sql"},
+			added:   []string{"V10__sq10.sql"},
+			assert:  refused([]string{"sq2"}, "9", "10"),
+		},
+		{
+			// Reverted: exit 0, tables sq2 and sq100. Three digits reach the
+			// same rule, so the refusal is not a two-digit special case.
+			name:    "a hundredth migration added under an applied second",
+			applied: []string{"V2__sq2.sql"},
+			added:   []string{"V100__sq100.sql"},
+			assert:  refused([]string{"sq2"}, "2", "100"),
+		},
+		{
+			// Atlas CE gives every repeatable the empty version string, which
+			// sorts below every token, so it refuses a repeatable added to a
+			// database with any history. Reverted: exit 0, tables sq2 and sqr —
+			// the repeatable runs, because its reserved int64 slot is the
+			// highest there is and the numeric comparison sees nothing wrong.
+			name:    "a repeatable added to recorded history",
+			applied: []string{"V2__sq2.sql"},
+			added:   []string{"R__sqr.sql"},
+			assert:  refused([]string{"sq2"}, "2", ""),
+		},
+		{
+			// The token is Atlas CE's version STRING for the file, not the
+			// file's own token. Every repeatable is version "" to CE whatever it
+			// is called, and CE refuses all of R__r.sql, R1__r.sql, R9__r.sql
+			// and Rfoo__r.sql added over an applied V2 — measured. Reading the
+			// file's own token instead would let this one through, because "9"
+			// sorts above the applied "2". Reverted: exit 0, tables sq2 and sqr.
+			name:    "a repeatable whose own token sorts high is still refused",
+			applied: []string{"V2__sq2.sql"},
+			added:   []string{"R9__sqr.sql"},
+			assert:  refused([]string{"sq2"}, "2", ""),
+		},
+		{
+			// The mark is the highest applied TOKEN, not the token of the
+			// highest applied version, and this is the input that separates
+			// them: the applied tokens are "2" and "10", whose maxima disagree.
+			// "15" sorts below "2" and above "10". Comparing against "10" would
+			// pass it, and the numeric comparison passes it too (15 > 10), so
+			// only the right operand refuses it. Reverted: exit 0, tables sq10,
+			// sq15 and sq2.
+			name:    "a fifteenth migration under applied second and tenth",
+			applied: []string{"V2__sq2.sql", "V10__sq10.sql"},
+			added:   []string{"V15__sq15.sql"},
+			assert:  refused([]string{"sq10", "sq2"}, "10", "15"),
+		},
+		{
+			// The control for the row above: same applied pair, and "20" sorts
+			// above both applied tokens, so it runs. Without it, "refuse
+			// everything two-digit added to this history" would pass too.
+			name:    "a twentieth migration still applies over applied second and tenth",
+			applied: []string{"V2__sq2.sql", "V10__sq10.sql"},
+			added:   []string{"V20__sq20.sql"},
+			assert:  executed("sq10", "sq2", "sq20"),
+		},
+		{
+			// The union's other half. "3" sorts ABOVE the highest applied token
+			// "2", so the source comparison passes it; the numeric one refuses
+			// it because it lands under the applied V10. Reverted to a source-
+			// only comparison: exit 0, tables sq2, sq3 and sq10.
+			name:    "an ordinary migration under an applied two-digit version",
+			applied: []string{"V2__sq2.sql", "V10__sq10.sql"},
+			added:   []string{"V3__sq3.sql"},
+			assert: func(c *qt.C, dbPath string, out flywaySourceOrderOutput) {
+				c.Assert(out.err, qt.IsNotNil, qt.Commentf("stdout:\n%s\nstderr:\n%s", out.stdout, out.stderr))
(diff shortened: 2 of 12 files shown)


denisvmedia:
Marked draft: I rebased this branch onto current master and **two of its own tests now fail**. The failure is not the rebase mechanics — it is a real interaction with #1003, which landed while this branch was waiting.

```
--- FAIL: TestCompatMigrateApplyFlywaySourceVersionOrder/a_surviving_baseline_below_the_mark_stays_exempt
--- FAIL: TestCompatMigrateApply_FlywayBaselineAgainstRecordedHistory
```

The first row expects a surviving baseline below the high-water mark to stay **exempt** and apply. On current master it is **refused** instead, by the guard #1003 added:

> B10__sqbase.sql is a Flyway baseline and this database already has migration history, so it will not be treated as a fresh install; nothing has been applied

Both changes are about the same question — whether a converted Flyway migration sitting below the mark was authored out of order — and they answer it in different places. #1003 refuses a baseline whose position the history has passed; this branch teaches the linear guard to compare the source **token** rather than the ordering key. The exemption this branch's test asserts is the one #1003 decided not to grant.

That needs a decision, not a merge conflict resolution: either the token comparison makes #1003's refusal unnecessary for this shape, or the refusal is correct and this test's expectation was written against a world that no longer exists. Whoever picks it up should measure the shape against the pinned community binary before choosing, because the two guards can also disagree in the direction that matters — one of them exiting 0 where the binary exits 1.

What I did do, and what is worth keeping from the rebase:

- Resolved the `ApplyOptions` conflict as a field union (master's `OutOfOrderExempt` naming plus this branch's `SourceVersions`).
- Extracted `flywayLinearityOperands` to bring `runAtlasMigrateApply` back under the `funlen` limit, which the combined change had pushed past. It returns a struct rather than four values, per `function-result-limit`.
- `go build`, `go vet`, `qtlint`, `golangci-lint` (0 issues) and `check-test-style.sh` are green; the two test rows above are the only failures.

I pushed the rebase before reading the test output, which is my mistake — the branch should not have been left looking mergeable while red.

denisvmedia:
Picked this up, measured the shape, decided it, and took the PR out of draft. Three commits on top of the rebase; nothing force-pushed.

**The decision: #1003's refusal stands, and this branch's expectation was written against a world that no longer exists.** The row now asserts the refusal.

What settled it is that the failing expectation matched *neither* tool. Measured against the pinned community binary with `V2__sq2.sql` applied and `B10__sqbase.sql` added: the binary prints `No migration files to execute` at exit 0 having run nothing; the row asserted exit 0 having *run the squash on top of the history it squashes*. That is the third behavior #1003 examined and rejected, and neither of the two tools produces it.

The token comparison does **not** subsume the refusal, and the grid says so rather than an argument alone. Splitting the twenty-six baseline cells by whether the source comparison would flag the baseline at all:

- **token at or below the mark** — ten cells (`V2`+`B10`, `V2`+`B1`, `V1,V2,V3`+`B2`, `V1,V2`+`B2`, `V2`+`B02`, `V2,V10`+`B11`, `V2`+`B10,V11`, `V9`+`B10`, `V2,V10`+`B1`, `V2`+`B10,V1`). All ten are refused by #1003's guard, on master and on this branch, and the message is #1003's in every one.
- **token above the mark** — seven cells. All seven execute on all three binaries, and the source comparison never flags them.

That split is forced by the selection rule: squashing compares the two tokens as strings, so a file outranking the baseline survives into `Covered` and one carrying the baseline's own token lands in `SameVersion`. If the baseline's token is at or below the highest applied token, the file holding that mark is in one of the two, so the guard fires. The source comparison can therefore never be the deciding refusal for a surviving baseline; narrowing #1003 would have removed a refusal and replaced it with nothing. The full grid, all three binaries, is in the PR body.

Two consequences for the tests:

- The old row's stated job ("it goes red if the exemption is applied to only one half of the union") is no longer achievable under `linear` — #1003's guard answers first whether or not the baseline is exempt, so the row was green either way. The exemption is still observable under `--exec-order=linear-skip`, where the baseline check abstains by design, so a **new row** holds it there and the mutation that halves the exemption reddens exactly that row and nothing else.
- The two `below current version` assertions were mechanical: they are converted Flyway directories, so they take the same `for current version` + source-token wording your first commit gave the other three.

**The hunt found nine cells in the forbidden direction**, and they are not the ones under dispute. A sweep of 189 `(applied, added)` pairs turned up nine where compat exits 0 and the pinned binary exits 1 — every one a **surviving baseline plus a repeatable**, e.g. `V2` applied with `B3__nb3.sql` and `R__nr.sql` added:

```
atlas       Error: migration files B3__nb3.sql, R__nr.sql were added out of order.   exit=1  tables: a2
ptah-compat Migrating to version 9223372036854775807 from 2 pending migrations.      exit=0  tables: a2,nb3,nr
```

They are the repeatable defect this branch already closes, hidden behind a squash: the mark is the highest token among the *applied* versions, the baseline retires the applied file from the covered set, its token goes with it, and the comparison switches itself off. The cells are equally loose on master, so not a regression this branch introduced — but shipping while they stood would have shipped "a repeatable is refused whatever the file is called" against a measurement that says otherwise. `FlywaySourceVersions` now recovers the squashed tokens from the directory without the baseline, the same domain and argument #1123 established, gap-fill only so the executed mapping always wins. After it the sweep reports zero such cells and those nine rows are the only rows whose answer changed.

One shipped doc claim went with it: `integrity-and-safety.md` said "Atlas CE applies a below-mark baseline under that flag too" of `--exec-order=non-linear`. Measured, that holds for a baseline the binary refuses as out of order and not for one it skips silently — which is the first row of that note's own table. Corrected, and `--exec-order=linear-skip` added to the flags that leave the baseline runnable, which the note had omitted.

Gates: `go build`, `go vet`, `go test ./... -count=1`, `qtlint`, `golangci-lint` (`0 issues.`), `check-test-style.sh`, `check-public-api.sh`, `check-public-api-snapshot.sh` (no movement) and `check-public-api-released.sh` all green. Mutation sweep 14/14 killed, no BUILD-FAILED and no PATCH-FAILED, with the two new discriminators run alone and their reddened subtests read by name. Every `check:*` in `docs/site/package.json` green, and `check:responsive` **actually ran** this time — `npm ci`, `npx playwright install chromium`, `npm run build`, then `OK (60 routes x 2 viewports, cells within 8 lines)`. Worth noting it exits **1**, not 0, when `dist` is missing, so the earlier "exits 0 unrun" reading was wrong on this tree.

PR #1169, opened 2026-08-06
Ban testify with depguard instead of a regex over source text

### LINKED ISSUE #1139: The testify ban reads English prose as code, and means different things on macOS and Linux
_upd, see the comment_

`scripts/check-test-style.sh` refuses any line matching:

```
git grep -nE 'github\.com/stretchr/testify|\b(assert|require)\.' -- '*.go'
```

Two defects, and the second hides the first.

## It matches prose

A test comment ending a sentence with the word `assert` trips it:

```go
// Reverted, the first call runs the up body and this never reaches the assert.
```

```
migration/migrator/directional_repair_test.go:167:// Reverted, the first call runs the up body and this never reaches the assert.
teststyle: testify/assert/require usage is prohibited; use quicktest as qt instead
exit 1
```

There is no testify anywhere in the file. `\b(assert|require)\.` matches a word followed by a full stop, which is ordinary English. `require.` in prose ("the checks this would require.") trips it too. The pattern is trying to spot a package-qualified call and cannot tell one from a sentence ending.

## It behaves differently on macOS and Linux

The same tree passes locally and fails on CI. `git grep -E` honors `\b` on the GNU grep CI runs on and ignores it on the BSD-derived grep macOS ships, so on macOS the pattern silently never matches its `\b` alternatives at all.

That is the worse half. It means:

- a developer cannot reproduce a CI failure this gate reports, and
- **the gate is much weaker locally than it looks** — on macOS the `assert.`/`require.` half of the pattern is inert, so a genuine `assert.Equal(t, ...)` added on a Mac passes the local check and is caught only after push, if at all.

This has bitten before in this repo: a banned-word scan using `git grep -E` with `\b` reported the tree clean twice while 24 occurrences of the banned term were present.

## What a fix has to do

- **Match a call, not a word.** Something anchored on the import (`"github.com/stretchr/testify`) plus a call shape (an identifier, a dot, an upper-case method name) rather than any `assert.`/`require.`. The import line alone is nearly sufficient: testify cannot be called without being imported, and a file that imports it is the thing being banned.
- **Behave the same on both platforms.** Either avoid `\b` entirely, or stop relying on `git grep`'s regex flavor — the script already learned once that `rg` exits 127 when absent and the `if` reads that as "no matches", which is why it uses `git grep` at all. Whatever replaces it needs a self-test that fails when the pattern stops matching real testify usage.
- **Carry a positive control.** The gate should refuse a file that genuinely uses testify, and that should be asserted, not assumed. A pattern that matches nothing passes every tree.

Worth checking the same way: `scripts/check-documented-install.sh` and any other gate in `scripts/` built on `git grep -E` with `\b`.

Found while unblocking PR #1137, where the comment was reworded to get the branch moving. That reword is not the fix.

denisvmedia:
Closes #1139. Branch cut from `origin/master` `d1e2dc64`.

Per the maintainer comment on the issue, the ban is a `depguard` deny entry in `.golangci.yml`. The text scan in `scripts/check-test-style.sh` is deleted, not rewritten.

## Reproduction, before

One tree, three lines: a real `assert.Equal(t, 1, 1)` call, the comment `// Reverted, the first call runs the up body and this never reaches the assert.`, and the string `bassert.Foo`. Same shipped pattern, two platforms.

**macOS 26.5, git 2.50.1** — the word-boundary half of the pattern is inert; `\b` is the letter `b`:

```
$ git grep -nE '\b(assert|require)\.' -- '*.go'
bword_test.go:5:// bassert.Foo
exit=0
```

The real call is not matched. The full shipped pattern matches only the import line, so a genuine `assert.Equal(t, ...)` added on a Mac passes `scripts/check-test-style.sh` locally.

**GNU grep 3.11 (ubuntu:24.04), the flavor the Linux runners use** — `\b` is a word boundary:

```
$ grep -nE '\b(assert|require)\.' *.go
import_test.go:5:func TestZ(t *testing.T) { assert.Equal(t, 1, 1) }
only_call_test.go:4:	assert.Equal(t, 1, 1)
prose_test.go:3:// Reverted, the first call runs the up body and this never reaches the assert.
```

The prose is matched. `bassert.Foo` is not. Exactly inverted: each platform catches what the other misses, and the CI half fires on a sentence.

## Reproduction, after

The two comments that were reworded to appease the old scan are restored — `migration/migrator/checks.go` back to "accept the assert." (#709) and `migration/migrator/directional_repair_test.go` back to "... this never reaches the assert." (#1137). With those in the tree:

```
$ scripts/check-testify-ban.sh
golangci-lint has version 2.12.2 built with go1.26.2 from c0d3ddc on 2026-05-06T11:01:25Z
testify-ban: 1/3 positive control -- a fixture that genuinely calls testify
banned/banned.go:7:2: import 'github.com/stretchr/testify/require' is not allowed from list 'main': testify is prohibited in Ptah; ... (depguard)
banned/banned_test.go:6:2: import 'github.com/stretchr/testify/assert' is not allowed from list 'main': testify is prohibited in Ptah; ... (depguard)
2 issues:
* depguard: 2
testify-ban: 2/3 negative control -- prose that ends a sentence with assert. and require.
testify-ban: 3/3 sweep -- every module, every target platform, every build tag
  ok: examples/orm-loaders/gorm (linux, darwin, windows)
  ok: . (linux, darwin, windows)
  ok: testkit (linux, darwin, windows)
testify-ban: ok
exit=0
```

27s warm, 27s with a cold `GOCACHE` — depguard is AST-only, so it never compiles anything.

## Why `depguard` is not the whole fix

depguard only sees what golangci-lint loads, and the deleted scan read every tracked `*.go`. Three gaps, each measured with a planted mutant and each closed by check 3:

| Gap | Measurement | Closed by |
|---|---|---|
| `//go:build integration` (47 test files) | testify planted in one: `golangci-lint run --enable-only=depguard ./...` → `0 issues`, exit 0. Same run with `--build-tags=integration` → the depguard finding, exit 1. | `--build-tags=integration,testkitcontainers,observability` |
| Platform files | testify planted in a `//go:build windows` file: root module clean at GOOS=linux **and** GOOS=darwin, refused at GOOS=windows. | sweep each GOOS |
| Nested modules | testify planted in `testkit/`: root `./...` never reaches it; refused once `testkit` is swept in its own right. | module list read from `git ls-files '*go.mod'` |

The `integration` tag is not set in `.golangci.yml` itself because the full linter set has 15 unrelated findings under it (7 gosec, 4 revive, 3 nolintlint, 1 ineffassign). Restricting the sweep to `--enable-only=depguard` closes the blind spot without dragging those in — fixing them is a separate change.

## What each check prints if the change is reverted

Each was run as a mutant, not assumed.

- **`.golangci.yml` deny entry deleted.** Check 1 prints `0 issues` then `testify-ban: golangci-lint accepted a file that imports and calls testify` / `the prohibition is inert -- check the depguard deny entry in .golangci.yml`, exit 1. Measured.
- **A replacement that reads source text again** (inverse mutant: `github.com/frankban/quicktest` added to the deny list, so the clean fixture is refused for a non-testify reason). Check 2 prints the finding then `testify-ban: the prohibition refused a file with no testify in it` / `this is stokaro/ptah#1139 again -- it is reading English, not imports`, exit 1. Measured — reverting the deny entry could not have shown this, since that only reddens check 1.
- **`--build-tags` dropped from check 3.** With testify in a `//go:build integration` file the sweep prints `0 issues` and `testify-ban: ok`, exit 0. Measured.
- **GOOS matrix dropped.** With testify in `internal/pathguard/zz_probe_windows.go` the sweep passes at linux and darwin and prints `testify-ban: FAILED . GOOS=windows` only at windows. Measured.
- **Nested-module enumeration dropped.** With testify in `testkit/zz_probe.go` the root sweep is clean at all three GOOS; `testify-ban: FAILED testkit GOOS=linux|darwin|windows` is the only signal. Measured.
- **The module list losing the repository root.** `for m in $list` drops empty fields, so the root was spelled `.`; if the list ever comes back without it the script prints `the module list does not contain the repository root` and refuses, rather than sweeping two examples and printing ok.
- **`scripts/probe-compat-base-vs-head.sh` duration rule reverted to `\b`.** On macOS `took 12.5ms to run` is left alone and `took 12.5msb to run` is rewritten; on GNU sed exactly the opposite. With the fix, both platforms produce byte-identical output on the same five-line fixture. Measured.

`golangci-lint` refusing to start because another copy holds its lock is named and retried rather than interpreted — it surfaced during development and would otherwise have read as "the fixture was accepted".

## The sibling gates the issue asked about

`scripts/check-documented-install.sh` was checked the same way and **does not** have the defect. Its pattern contains no `\b`; macOS `git grep -hoE` and GNU `grep -rhoE` discover the identical three commands from this tree; and `MIN_EXPECTED=2` already fails it closed if discovery matches nothing. No change.

`scripts/probe-compat-base-vs-head.sh` **does** have it, in `sed` rather than `git grep`, and is fixed here. It is a hand-run probe rather than a gate, and its failure direction was noisy rather than silent — on macOS no duration was normalized, so every `--format` row compared two different elapsed times and reported a spurious `CHANGED`.

Those are the only `\b` occurrences under `scripts/` outside the Python helpers.

## Deliberately left open

- **`.golangci.yml` still does not set `build-tags`.** The 15 findings the `integration` tag exposes to the full linter set are real and unaddressed; only depguard reaches those files today. Fixing them belongs in its own PR.
- **`drivertest/.*\.go$` remains in `exclusions.paths`,** so depguard would not reach such files. That directory does not exist in this repository; the entry is inherited configuration and removing it is not this issue's business.
- **`docs/site/scripts/check-style.mjs` bans testify in documentation snippets by its own mechanism** and is untouched, so the docs `check:*` scripts were not run — no file under `docs/` is in this diff.
- The `cmd/viz` test `TestSVGReportsGraphvizStderrOnFailure` failed once as `signal: killed` at exactly 10.00s while golangci-lint was running concurrently, and passes on its own and in a clean full run. Pre-existing load flake, not touched here.

## Gates

`go build ./...`, `go vet ./...`, `go test ./... -count=1` (176 packages ok, zero FAIL), `go tool qtlint ./...`, `golangci-lint run ./...` (0 issues), `scripts/check-test-style.sh` (OK, 175 conditional groups, 45 white-box files), `scripts/check-public-api.sh`, `scripts/check-public-api-snapshot.sh`, `scripts/check-testify-ban.sh` — all green. `gofmt -l .` names only `internal/goannotationexport/testdata/cleanup/models.go`, which predates this branch and is not in the diff.

`scripts/check-testify-ban.sh` is syntax-clean under both macOS `sh` and Ubuntu `dash`, and the new workflow step sits after the `Golint` step because `golangci-lint-action@v9` puts the binary it installs on `PATH` (`core.addPath` at `src/run.ts:241`, read at ref `v9`). If the binary is missing the script exits 1 rather than skipping.


denisvmedia:
Reworked after Denis's note on #1139: **the ban is the depguard entry, and nothing else.**

The previous revision added the depguard rule and then also shipped `scripts/check-testify-ban.sh` (311 lines), a `make lint-testify-ban` target and its own CI step — a text-scanning control to guard the guard. That is the machinery moving to depguard was meant to remove, reintroduced under a new name. All of it is gone; the diff is now **10 insertions, 347 deletions**.

## Measured that depguard alone does the job

testify is not a dependency of this module today, so a file importing it does not compile and golangci-lint answers `no go files to analyze` rather than a finding — which is why a naive probe looks like the rule is inert. Made it a real dependency in a throwaway worktree and probed again:

```
internal/zzprobe1139/probe.go:4:8: import 'github.com/stretchr/testify/assert' is not allowed
from list 'main': testify is prohibited in Ptah; use quicktest imported as qt, the standard
library testing package, or an existing project test helper. (depguard)
1 issues:
```

Control, same run: a file whose only occurrence is a comment ending a sentence with the word `assert` produces no depguard finding. That is exactly the case the old text scan refused on CI while accepting it locally.

`pkg` matches by prefix, so `assert`, `require`, `mock` and `suite` are all covered by the single entry.

## What remains

- `.golangci.yml` — the deny entry, with the reason it is not a text scan.
- `scripts/check-test-style.sh` — the old grep deleted; the comment explaining why the pattern was wrong on both counts is kept, because that reasoning is what stops someone reinstating it.
- `AGENTS.md` — states the rule is a depguard entry reported by `golangci-lint run ./...`, with nothing extra to run.

Gates: `go build ./...` clean, `golangci-lint run ./...` `0 issues.`, `scripts/check-test-style.sh` OK (175 conditional groups, 45 white-box files).

.golangci.yml
```diff
@@ -38,6 +38,18 @@ linters:
...
+            # The testify ban lives here, on the import graph, and not in a text
+            # scan over source files. A regex cannot tell `assert.` in a call
+            # from `assert.` ending an English sentence, and the `\b` the old
+            # scan used means "word boundary" to the regex engine git grep uses
+            # on Linux and the literal letter `b` to the one it uses on macOS --
+            # so the same tree was refused on CI for a comment and accepted
+            # locally for a real call. See stokaro/ptah#1139.
+            #
+            # `pkg` matches by prefix, so this covers every testify subpackage
+            # (assert, require, mock, suite) with one entry.
```
denisvmedia, line 50: The comment is excessive.

denisvmedia:
On the two comment changes in `checks.go` and `directional_repair_test.go` — fair question, the connection is real but the diff does not show it. Both are **reversions of workarounds that existed only because of the broken scan**, and they are the proof the defect is gone.

### `migration/migrator/checks.go:176`

```
- // reject a trailing ';' on a prepared query (MySQL) accept the predicate.
+ // reject a trailing ';' on a prepared query (MySQL) accept the assert.
```

`git log -S "accept the predicate"` returns exactly one commit, **`875f290c` — "Fix teststyle false positive in migration comment (#709)"**. Its entire diff is that one line: `assert.` → `predicate.`. No code changed. An English sentence was distorted to get past `\b(assert|require)\.`.

The distortion also cost meaning. The field on the next line is `check.Assert`, and the sentence explains why *that* value gets its trailing semicolon trimmed. "predicate" names something the code does not call it.

### `migration/migrator/directional_repair_test.go:167`

Same thing, from today. I reworded `never reaches the assert.` into `execution never reaches the check below.` to unblock #1137, and said in that commit it was not the fix. This puts it back.

### Why they belong here rather than in a follow-up

These are the only two places in the tree where prose was bent around the gate, so they are the complete set of collateral the old rule caused. Restoring them is the check that the new rule does not have the defect: the text scan refused both lines, and depguard does not see them, because it matches the import declaration.

Leaving them bent would keep the workaround after removing the reason for it — the wording would then look like a deliberate style choice rather than damage, and the next reader would have no way to tell.

Happy to split them into a separate PR if you would rather keep this one to the mechanism alone. My reasoning for keeping them together is that they are the observable consequence of the change, but that is a judgment call, not a requirement.

PR #1170, opened 2026-08-06
security: root migrate-diff publication and recovery to retained directory handles

### LINKED ISSUE #895: Root migrate-diff publication and recovery to retained directory handles
## Summary

`ptah atlas migrate diff` must publish staged migrations, the publication journal, commit marker, recovery quarantine, and `atlas.sum` through retained rooted filesystem handles rather than reopening the migration-directory pathname.

Discovered during the final security review of #892. The partial input-side hardening was removed from that PR because verifying a rooted handle and then publishing through the original pathname still permits a post-verification replacement or symlink swap.

## Problem

The current publication subsystem in `internal/atlasmigrate/publication.go` is pathname-based. A migration directory can be replaced after validation but before staging or publication, redirecting writes and recovery operations. `DiscardIfCreatedEmpty` has the same verify-close-remove race for cleanup.

## Scope

- Design a rooted writable publication abstraction that retains the migration directory and the directory containing publication metadata.
- Route staging, preparation, digest reads, exclusive publish, checksum commit, journal/marker writes, rollback quarantine, orphan cleanup, and recovery through rooted operations.
- Preserve crash consistency and existing journal recovery semantics.
- Remove pathname-based identity-check-then-write gaps rather than adding more pre-write checks.
- Make created-directory cleanup use parent-rooted identity-safe removal, or eliminate that cleanup behavior.
- Keep the command layer thin; filesystem publication belongs outside `cmd/atlas`.
- Update security and migration-directory documentation.

## Acceptance

- [ ] A deterministic directory replacement or symlink swap after planning cannot redirect any staged, journal, migration, checksum, rollback, or cleanup write.
- [ ] Recovery operates on the originally retained filesystem objects and cannot cross the configured root.
- [ ] Existing hard-link, exclusive-copy, and write-through-move publication modes remain covered.
- [ ] Interrupted-publication and commit-uncertainty tests remain green.
- [ ] Rooted cleanup cannot remove a replacement directory.
- [ ] Unix race tests, Windows compilation/native behavior tests, lint, test-style, unit, and live integration pipelines are green.
- [ ] Documentation states the rooted publication and recovery boundary.

## Related

- #888
- #892

denisvmedia:
## The hole

`internal/atlasmigrate` published by pathname from end to end. `diff.go:203`
verified `opts.Dir` and `diff.go:209` then wrote through the same string;
`publication.go` carried 19 `dir string` parameters across ~30 functions and
contained no reference to `os.Root` or `pathguard`. A directory replaced after
validation therefore received everything the run produced, and the post-publish
guard did not catch it because it re-stats the pathname, which by then resolves
to the replacement.

## Reproduction

Probe run in this worktree at `f22107bd` (master), dropped into
`internal/atlasmigrate` as a `_test.go` file. It swaps the directory inside
`PreparePublication` — after the directory is locked, recovered, snapshotted and
verified, and after the batch is staged: copy the contents to `attacker/`,
rename `migrations` to `migrations.retained`, symlink `migrations` → `attacker`.

**Before** — `go test ./internal/atlasmigrate/ -run TestIssue895 -v`, exit 1:

```
--- PASS: TestIssue895_Control_NoSwap
    control: sql files in validated dir = [1786000387_fault_injection.sql 1_init.sql]
--- FAIL: TestIssue895_DirectorySwapAfterValidation
    generateDiff err = <nil>
    MigrationPaths = [.../001/migrations/1786000387_fault_injection.sql]
    SumPath        = .../001/migrations/atlas.sum
    sql in ATTACKER dir .../001/attacker            = [1786000387_fault_injection.sql 1_init.sql]
    sql in RETAINED dir .../001/migrations.retained = [1_init.sql]
    migration write was redirected into the replacement directory
    validated directory did not receive the migration
```

**After** — same probe, same fixture, exit 0:

```
--- PASS: TestIssue895_DirectorySwapAfterValidation
    generateDiff err = <nil>
    sql in ATTACKER dir .../001/attacker            = [1_init.sql]
    sql in RETAINED dir .../001/migrations.retained = [1786001201_fault_injection.sql 1_init.sql]
```

The probe is not in the diff; it is superseded by
`internal/atlasmigrate/diff_directory_swap_nonwindows_internal_test.go`, which
covers the same window plus the rename shape, `atlas.sum`, checksum
verification, rollback and cleanup.

## The change

`publicationDir` (`internal/atlasmigrate/publicationdir.go`) binds the parent of
the canonical migration directory first and opens the migration directory only
through that handle. Both are retained for the life of the run. Every staged
file, published migration, journal, commit marker, `atlas.sum`, rollback
quarantine and cleanup entry is now addressed by a **name** relative to one of
those two handles; pathnames survive only for reported paths and error text.
`filepath.Glob` is replaced by a rooted `ReadDir` + prefix/suffix filter, and
`os.DirFS(dir)` by the handle's own `FS()`.

The journal deliberately lives in the parent directory, which is why the parent
gets its own retained handle rather than being reached through `filepath.Dir` of
a pathname that may since have been replaced. Its name is still derived from the
canonical directory, so journals written by earlier versions are still found.

New primitives:

- `fsdurable.MoveFileNoReplaceAt` — link-then-unlink on Unix (the exact sequence
  the pathname version already used, so no new "this filesystem cannot do
  conditional renames" failure mode), rename-by-handle plus a flush on Windows,
  where there is no rooted `MoveFileEx`.
- `pathguard.OpenedDirectory`: `OpenFile`, `Link`, `MoveFileNoReplace`,
  `ReadDir`.
- `migratesum.WritePrecomputedWithFormatIn` — the same atomic
  create/chmod/write/sync/replace/sync as `writeAtomicSumFile`, through a
  retained handle. It also maps a post-rename `ErrReplacementCommitted` onto
  `CommitUncertainError`, which is what the caller already treats as "journal
  retained for recovery".

The **rollback quarantine changes shape**: it was a directory
`<staged>.rollback/` holding a file `published`; it is now the direct-child file
`<staged>.rollback`. Rooted publication addresses direct children of a retained
handle, and a nested quarantine would need a second retained handle whose own
open is exactly the race this PR closes. A side effect is that the rollback path
now creates and removes no directory at all, so "rooted cleanup cannot remove a
replacement directory" holds by construction rather than by a check.

`RecoverPendingPublicationLocked` returns `nil` for a migration directory that
does not exist yet. Master reached the same outcome by globbing a missing
directory; opening one now fails, and `migration/generator` calls it before the
directory is created. Measured: without this, `TestMigrateGenerateJSONReportWritesSiblingSafetyArtifact`
fails with `recover migration publication before planning: open migration
directory: openat migrations: no such file or directory`.

Measured with inverse mutants — each function re-resolves `pub.displayPath` and
uses the fresh handle, which is precisely what master did — not by reverting.

| Test | Mutant | Output |
| --- | --- | --- |
| `TestGenerateDiff_PublishesIntoTheValidatedDirectory` (both rows) | `publishMigrationBatchContext`, `writeDirSum`, `verifyMigrationDirUnchanged`, `publicationCommitted`, `finalizeCommittedPublication` re-resolve | `unexpected length / len(got): int(1) / got: []string{"1_init.sql"} / want length: int(2)` at `migrationFileNames(c, swap.retained)` — i.e. `generateDiff` returns `nil` and the validated directory received nothing. With only `publishMigrationBatchContext` mutated it instead fails earlier with `migration directory changed during migrate diff planning`; with the sum writer added, `atlas.sum does not match the journaled migration publication`. |
| `TestGenerateDiff_PublishesIntoUnswappedDirectory` | same mutant | stays green — this is the non-interference control that makes "the replacement stayed empty" evidence rather than an artifact of reading the wrong directory. |
| `TestRecoverPendingPublication_RollsBackInTheRetainedDirectory` | `recoverPendingPublication` re-resolves | `got nil error but want non-nil / want: e"file does not exist"` — the rollback removed the replacement's copy and left the retained one. |
| `TestRecoverPendingPublication_CleansUpInTheRetainedDirectory` | same | same shape, on the orphan staging file. |
| `TestMoveFileNoReplaceAt_FollowsTheRetainedDirectory` | `MoveFileNoReplaceAt` joins `root.Name()` | `link .../retained/staged .../retained/published: no such file or directory` |
| `TestOpenedDirectoryPublicationOperations_FollowTheRetainedDirectory` | same | `link .../retained/staged .../retained/published: no such file or directory` |
| `TestWritePrecomputedWithFormatIn_WritesIntoTheRetainedDirectory` | `WritePrecomputedWithFormatIn` writes `filepath.Join(dir.Path(), name)` | `open .../moved/atlas.sum: no such file or directory` |
| `TestMoveFileNoReplaceAt_HappyPath` / `_RefusesExistingDestination`, `TestOpenedDirectoryPublicationOperations_HappyPath`, `TestWritePrecomputedWithFormatIn_WritesAtlasSum` / `_RejectsNilSum` | same mutants | stay green — the swap rows are what discriminates, not the primitives' existence. |

The ~50 converted call sites in `diff_internal_test.go` are the existing
coverage for hard-link, exclusive-copy and write-through-move publication,
interrupted publication, commit uncertainty and foreign-collision refusal; they
pass unchanged in meaning, with `dir` replaced by the retained handle and the
quarantine assertions following the flat layout.

## Gates

All run in this worktree, exit codes captured immediately:

```
build 0 · vet 0 · test 0 (0 FAIL lines in the full log) · qtlint 0
golangci-lint 0 (0 issues) · check-test-style 0 (175 conditional groups, 45 white-box files)
check-public-api 0 · check-public-api-snapshot 0
gofmt -l: only internal/goannotationexport/testdata/cleanup/models.go, a
  pre-existing fixture from #988 that is not in this diff
GOOS=windows go build ./... 0 · GOOS=windows go vet ./internal/... ./cmd/... ./migration/... 0
```

`docs/` is touched, so all 15 scripts in `docs/site/package.json` ran, each
exit 0 — including `check:responsive`, which needs `npm run build` first and
otherwise reports `dist not found` without checking anything.

## Deliberately left open

- **The command layer still passes a pathname.** `cmd/atlas/migrate_diff.go`
  resolves `--dir` and hands `atlasmigrate` a string, which reopens it. The
  window between those two resolutions is *before* this run validates anything —
  the directory that gets opened is the directory that gets verified — so
  acceptance criterion 1 ("after planning") is met. Threading a handle from the
  CLI is the input-side half that `40eeeb12` stripped out of `135aeac4`; it is
  worth relanding but it is a separate change to `cmd/atlas`,
  `cmd/internal/migrationsource` and `pathguard`, and it closes a different
  window.
- **The directory lock is still keyed by pathname.** `acquireDirLock` /
  `migrationDirLockPath` create the lock file next to the canonical directory
  before any handle exists. That is a cooperative mutual-exclusion mechanism
  between Ptah processes, not a boundary against a hostile writer, and rooting it
  would change the lock identity that other verbs already agree on.
- **`PublishArtifactsLocked` and `RecoverPendingPublication*` still take a
  string.** They now open the directory once at entry and route the whole batch
  through that handle, which covers the native `ptah migrate generate` surface,
  but a swap strictly before the call still selects which directory is used.
  Giving them handle-taking variants belongs with the command-layer item above.
- **`DiffOptions.PreparePublication` still receives absolute staged paths.** The
  `--edit` callback hands them to an external editor, which cannot take a
  handle. The paths are re-validated through the rooted handle afterwards —
  `Lstat` for regularity, open-then-`SameFile`, digest — which is what the
  existing symlink-replacement test pins.

Closes #895

### DIFF
--- a/docs/site/src/content/docs/atlas/migrate-commands.md
+++ b/docs/site/src/content/docs/atlas/migrate-commands.md
@@ -503,6 +503,18 @@ directory; the next lock holder compares the checksum commit marker and either
 finalizes the committed batch or removes only the hard-linked files owned by
 the interrupted batch.
 
+Publication and recovery are rooted. Ptah opens the migration directory and the
+directory that holds its journal once, before the snapshot the run verifies is
+captured, and addresses every later staged file, migration file, journal,
+commit marker, `atlas.sum`, rollback quarantine and cleanup entry by name
+relative to those two retained handles. Replacing the directory's pathname
+afterwards — a rename, a symlink, or a fresh directory moved into place —
+cannot redirect a write: the handles keep referring to the filesystem objects
+that were validated, wherever those objects are moved to
+([#895](https://github.com/stokaro/ptah/issues/895)). The boundary is the
+retained directory, so recovery can never act on a replacement and cleanup can
+never remove a file that a replacement contributed.
+
 ```bash
 ptah-compat migrate diff add_users \
   --dir file://migrations \
--- a/docs/site/src/content/docs/versioned/integrity-and-safety.md
+++ b/docs/site/src/content/docs/versioned/integrity-and-safety.md
@@ -581,6 +581,19 @@ project directory handle opened for config evaluation until capture completes;
 replacing the project pathname does not retarget it. Parent-relative project
 directories are external compatibility paths and do not inherit that root.
 
+Commands that *write* the directory — `migrate diff` and native migration
+generation — extend the same binding to publication. The migration directory
+and the directory holding its publication journal are opened once, before the
+snapshot the run verifies is captured, and every staged file, migration file,
+journal, commit marker, `atlas.sum`, rollback quarantine and cleanup entry is
+addressed by name relative to those retained handles. A directory replaced
+after validation therefore receives nothing, and recovery resolves an
+interrupted publication on the objects that were validated rather than on
+whatever the pathname resolves to when recovery runs
+([#895](https://github.com/stokaro/ptah/issues/895)). This closes the
+observation window that the two-capture check cannot: a replacement is no
+longer a difference to notice, it is a directory the writes never reach.
+
 ## Pre-migration checks
 
 Guard a migration on a data-state precondition with a `-- +ptah check`
--- a/internal/atlasmigrate/diff.go
+++ b/internal/atlasmigrate/diff.go
@@ -144,13 +144,17 @@ func generateDiff(
 	if err != nil {
 		return DiffResult{}, err
 	}
-	if err := os.MkdirAll(opts.Dir, 0755); err != nil {
-		return DiffResult{}, fmt.Errorf("create migration directory: %w", err)
+	pub, err := createAndRetainMigrationDir(opts.Dir)
+	if err != nil {
+		return DiffResult{}, err
 	}
-	if err := recoverPendingPublication(opts.Dir); err != nil {
+	defer func() {
+		err = errors.Join(err, pub.close())
+	}()
+	if err := recoverPendingPublication(pub); err != nil {
 		return DiffResult{}, fmt.Errorf("recover migration artifact publication: %w", err)
 	}
-	migrationSnapshot, err := migrationsnapshot.CaptureStable(os.DirFS(opts.Dir))
+	migrationSnapshot, err := migrationsnapshot.CaptureStable(pub.fsys())
 	if err != nil {
 		return DiffResult{}, fmt.Errorf("capture migration directory: %w", err)
 	}
@@ -200,7 +204,7 @@ func generateDiff(
 	if err := ctx.Err(); err != nil {
 		return DiffResult{}, err
 	}
-	if err := verifyMigrationDirUnchanged(opts.Dir, migrationSnapshot); err != nil {
+	if err := verifyMigrationDirUnchanged(pub, migrationSnapshot); err != nil {
 		return DiffResult{}, err
 	}
 	if opts.DryRun {
@@ -208,12 +212,27 @@ func generateDiff(
 	}
 	return writeDiffArtifacts(
 		ctx,
-		opts.Dir,
+		pub,
 		opts.Name,
 		contents,
 		migrationSnapshot,
 		opts.PreparePublication,
 	)
+}
+
+// createAndRetainMigrationDir creates the migration directory if it is missing
+// and retains it, together with the directory that holds its publication
+// journal, as rooted handles.
+//
+// Everything after this point -- recovery, the snapshot this run verifies
+// against, planning, and publication -- acts on the filesystem objects opened
+// here, so replacing the pathname afterwards cannot redirect a write
+// (stokaro/ptah#895). The caller owns the returned handle.
+func createAndRetainMigrationDir(dir string) (*publicationDir, error) {
+	if err := os.MkdirAll(dir, 0755); err != nil {
+		return nil, fmt.Errorf("create migration directory: %w", err)
+	}
+	return openPublicationDir(dir)
 }
 
 func prepareDiff(
@@ -363,8 +382,8 @@ func verifyDirSum(fsys fs.FS) error {
 	return nil
 }
 
-func verifyMigrationDirUnchanged(dir string, expected fsnapshot.Snapshot) error {
-	current, err := migrationsnapshot.CaptureStable(os.DirFS(dir))
+func verifyMigrationDirUnchanged(pub *publicationDir, expected fsnapshot.Snapshot) error {
+	current, err := migrationsnapshot.CaptureStable(pub.fsys())
 	if err != nil {
 		return fmt.Errorf("recapture migration directory: %w", err)
 	}
--- a/internal/atlasmigrate/diff_directory_swap_nonwindows_internal_test.go
+++ b/internal/atlasmigrate/diff_directory_swap_nonwindows_internal_test.go
@@ -1,0 +1,197 @@
+//go:build !windows
+
+package atlasmigrate
+
+// White-box testing required: the window this exercises opens after
+// GenerateDiff has already locked, recovered and snapshotted the migration
+// directory, so it cannot be reached from the exported API. PreparePublication
+// is the same internal seam diff_prepare_nonwindows_internal_test.go uses, and
+// recoverPendingPublication is only reachable from inside the package.
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	qt "github.com/frankban/quicktest"
+
+	"go.5x5.cz/ptah/internal/migratesum"
+	"go.5x5.cz/ptah/internal/migrationreplay"
+	"go.5x5.cz/ptah/migration/migrator"
+)
+
+// swappedDirectories names the two directories a swap leaves behind: the
+// object that was validated, which every write must still reach, and the
+// pathname that now resolves to the replacement, which must receive nothing.
+type swappedDirectories struct {
+	retained    string
+	replacement string
+}
+
+// replaceMigrationDirectory moves dir aside, copies its contents into a
+// replacement, and installs the replacement on dir's pathname through install.
+// Callers pass os.Symlink or a rename to choose which shape of replacement the
+// pathname resolves to afterwards; either way the pathname is what reads of
+// swappedDirectories.replacement go through.
+func replaceMigrationDirectory(
+	c *qt.C,
+	dir string,
+	install func(replacement, pathname string) error,
+) swappedDirectories {
+	c.Helper()
+	swap := swappedDirectories{retained: dir + ".retained", replacement: dir}
+	staging := filepath.Join(filepath.Dir(dir), "replacement")
+	c.Assert(os.CopyFS(staging, os.DirFS(dir)), qt.IsNil)
+	c.Assert(os.Rename(dir, swap.retained), qt.IsNil)
+	c.Assert(install(staging, dir), qt.IsNil)
+	return swap
+}
+
+func migrationFileNames(c *qt.C, dir string) []string {
+	c.Helper()
+	matches, err := filepath.Glob(filepath.Join(dir, "*.sql"))
+	c.Assert(err, qt.IsNil)
+	names := make([]string, 0, len(matches))
+	for _, match := range matches {
+		names = append(names, filepath.Base(match))
+	}
+	return names
+}
+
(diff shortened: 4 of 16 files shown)


denisvmedia:
Marked draft: two tests in this branch fail on the `windows-publication` job. They are the branch's own, and the cause is in the fixture rather than in the product code.

```
--- FAIL: TestMoveFileNoReplaceAt_FollowsTheRetainedDirectory
    move_root_test.go:69: c.Assert(os.Rename(dir, moved), qt.IsNil)
    got: rename ...\001\retained ...\001\moved:
         The process cannot access the file because it is being used by another process.

--- FAIL: TestOpenedDirectoryPublicationOperations_FollowTheRetainedDirectory
    (internal/pathguard, same shape)
```

**The whole point of both tests is that the retained handle keeps following the directory across a rename.** So the fixture opens the directory, then renames it out from under itself — and on Windows an open handle blocks the rename, which is exactly the platform difference the retained-handle design exists to paper over. POSIX allows it; Win32 does not, unless the handle was opened with `FILE_SHARE_DELETE`.

Two ways out, and the choice matters:

- **Open the retained handle with delete-sharing** on Windows, so the rename succeeds and the test measures what it claims. This is the honest fix if the production code is meant to survive a rename on Windows too — and if it is not opened that way today, then the *product* has the gap, not just the test.
- **Skip the rename-follow assertion on Windows**, with a comment saying the guarantee is POSIX-only. Legitimate only if the design never promised it there.

Do not simply drop the assertion: it is the discriminator for the retained-handle behaviour, and without it the tests pass whether or not the handle follows anything.

Two other things checked while looking at this, so they do not get blamed:

- The `fuzz` job failure is **not this branch**. It is `FuzzLexer: context deadline exceeded` — the fuzzer ran out of its time budget after ~1.06M executions, it did not find a crasher. This branch touches no file under `internal/lexer`.
- Everything else on the branch is green: 16 checks pass.

denisvmedia:
Status after attempting the rebase onto current master: **do not rebase this branch — re-derive what remains.**

The rebase conflicts in five files, with **76 conflict hunks in `internal/atlasmigrate/publication.go` alone**. That is not merge noise: #1168 (`fix(compat): bind migrate diff publication to the opened directory`, merged as `028ac05d`) rewrote the same machinery with a different design, and master is now **ahead of this branch** in the area this PR targets.

Measured on `internal/atlasmigrate/publication.go`, occurrences of the retained-handle vocabulary (`Retained`, `OpenedDirectory`, `openedDir`):

```
origin/master                          11
fix/issue-895-rooted-publication        7
```

And `internal/pathguard/opened_directory_ops.go` exists on master with 16 functions — this branch adds it as new.

So a large part of what this PR set out to build is already in, arrived by another route. Replaying 1359 insertions and 577 deletions on top of that would mean reconciling two designs hunk by hunk, and the result would be neither author's.

**Recommendation:** close this branch and open a fresh one that starts from master, keeping only what #1168 did not cover. The measurement to start from is #895's own reproduction — run it against current master first, because part of it may no longer reproduce at all.

Two things worth carrying over verbatim, independent of the rebase:

1. **The Windows failures are real and are the branch's own.** `TestMoveFileNoReplaceAt_FollowsTheRetainedDirectory` and `TestOpenedDirectoryPublicationOperations_FollowTheRetainedDirectory` fail with `The process cannot access the file because it is being used by another process` — the fixture opens the directory and then renames it, which POSIX allows and Win32 does not without `FILE_SHARE_DELETE`. Since the point of both tests is that the handle *follows* the rename, either the production code opens with delete-sharing on Windows (and if it does not, the gap is in the product, not the test), or the guarantee is POSIX-only and the test says so. Do not simply drop the assertion — it is the discriminator.

2. **The `fuzz` failure is not this branch.** `FuzzLexer: context deadline exceeded` after ~1.06M executions is a time budget, not a crasher, and this branch touches no file under `internal/lexer`.

denisvmedia:
Closing in favor of **#1193**, which re-derives what is left of #895 from current master rather than rebasing this branch.

Taking the recommendation in the comment above: the reproduction was run against master first, and most of it no longer reproduces.

**The write path is closed.** `internal/atlasmigrate/publication.go` on master contains no `os.*` filesystem call at all — `filepath.Join` survives only for reported paths and error text — and `migrationWriterDir` (`internal/atlasmigrate/writerdir.go`) already holds one `pathguard.OpenedDirectory` on the migration directory and one on its parent, opened before staging. This PR's own reproduction, dropped into the package unchanged, passes:

```
generateDiff err = <nil>
sql in ATTACKER dir .../001/attacker            = [1_init.sql]
sql in RETAINED dir .../001/migrations.retained = [1786023915_fault_injection.sql 1_init.sql]
atlas.sum in ATTACKER = false
atlas.sum in RETAINED = true
```

It is not a vacuous pass: restoring the pathname re-resolution after the editor hook flips it, with `generateDiff` still returning `nil` and the artifacts landing in the decoy.

The primitives this branch built are also unnecessary on master, because #1168 reached the same place by a different route. `MoveFileNoReplaceAt` and `WritePrecomputedWithFormatIn` are subsumed by `pathguard.OpenedDirectory.PublishFile` with an `fsdurable.Destination`, which is destination-conditional rather than just rooted. `DiscardIfCreatedEmpty` — the cleanup race the issue names — no longer exists anywhere in the tree, and the rollback quarantine is already the flat sibling file this PR proposed, so "rooted cleanup cannot remove a replacement directory" holds by construction.

**What genuinely remained** was two acceptance criteria with no code behind them, and #1193 covers exactly those and touches no production file:

- The rename shape of the reproduction was not pinned by any test. The rows in `diff_root_unix_test.go` all pre-stage a symlink at `--dir` and re-point it, so the object the run opened is never touched. #1193 adds the shape this PR reproduced — a real directory moved out from under the handle — with a decoy seeded as a live copy, which is what makes a pathname writer exit 0 instead of erroring out on a missing staging file.
- Recovery on the retained objects (criterion 2) and the documentation boundary (criterion 7) had nothing pinning them.

**Both Windows failures are carried over and resolved, not dropped.** The rename-follow fixtures are `//go:build !windows`, with the reason stated in the file: the fixture renames a directory the run holds open, which Win32 refuses unless the handle carries `FILE_SHARE_DELETE`, so where the rename is denied there is no hostile step to perform and nothing to measure. The assertion is kept everywhere it can run, because it is the discriminator.

Two things from this PR's review that stand independently and were checked again:

- The `fuzz` failure was never this branch. `FuzzLexer: context deadline exceeded` after ~1.06M executions is a time budget.
- The design notes in the "Deliberately left open" section — the pathname-keyed cross-process lock and the absolute staged paths handed to `--edit` — are still true of master. #1193 writes both into the documentation as deliberate rather than leaving them unstated.

Issue #1013, opened 2026-08-02
`?format=` rejected on migrate lint/new/diff, and unknown query keys rejected on every verb, where CE succeeds


denisvmedia:
Measured 2026-08-02 against pinned Atlas CE v1.2.0 (`~/Work/denis/ptah-atlas-conformance/bin/atlas`). Split out of the #1011 review — these are CE parity defects that no existing issue covers.

Existing coverage, for the record: #1002 covers `?format=` on `migrate status` and `migrate set`; #990 covers an empty `--dir-format`, non-`format` query keys and repeated keys on `migrate hash` and `migrate validate` only. Neither covers what follows.

## 1. `?format=` is refused on `migrate lint`, `new` and `diff`

Compat rejects **every** query parameter on those verbs:

```
error: atlas migrate lint --dir: migration directory URL query parameters are not supported for this command
```

CE exits 0 and **functionally** honors the parameter — it is not decoration. On a golang-migrate directory:

| | CE with `?format=golang-migrate` | CE without |
| --- | --- | --- |
| outcome | `Migration Status: PENDING`, exit 0 | `L2: 1_init.down.sql was added`, exit 1 |

CE also *validates* the value rather than ignoring it: `?format=totally-bogus` → `Error: unknown dir format "totally-bogus"`, exit 1.

Per-verb state measured on the compat surface today:

| verb | `?format=` |
| --- | --- |
| `apply`, `hash`, `validate` | accepted |
| `lint`, `new`, `diff` | **rejected — this issue** |
| `status`, `set` | rejected — #1002 |
| `checkpoint` | rejected, and **correctly so**: CE aborts every invocation of that verb, so there is no CE contract to diverge from (#1011) |

## 2. Unknown query keys are refused on all eight verbs, including the three that work

```
$ ptah-compat migrate apply --dir 'file://m?nonsense=1' --url sqlite://a.db
error: atlas migrate apply --dir: migration directory URL query parameters are not supported for this command

$ atlas migrate apply --dir 'file://m?nonsense=1' --url sqlite://b.db
# applies, exit 0
```

CE ignores unknown keys outright. Beyond parity, this is a footgun in the other direction too: a typo'd `?format=atals` is silently ignored by CE and the directory is read as the default format. Under the standing rule — CE full drop-in, no exceptions — CE's behavior is the contract, so we should match it and consider whether to warn.

## Definition of done

[...]

PR #1172, opened 2026-08-06
feat(compat): write the selected external layout on migrate new (refs #845)

### LINKED ISSUE #845: Atlas parity: make external migration directory formats operational across metadata commands
## Problem

Ptah recognizes external migration directory formats in parts of the Atlas-compatible surface, but `--dir-format` is not consistently operational across metadata commands. Commands can accept a flag yet still assume native Atlas file naming, checksums, ordering, or metadata semantics.

## Scope

- Define one migration-directory format interface for discovery, ordering, identity, checksum input, and mutability.
- Apply it consistently to `migrate hash`, `lint`, `new`, `set`, `status`, and `validate` where Atlas exposes equivalent behavior.
- Keep format-specific rules outside Cobra commands.
- Return explicit unsupported-operation errors where a format cannot represent a command safely.
- Do not add compatibility aliases or silent native-format fallbacks.

## Tests

- Declarative black-box contract tests shared across supported formats.
- Per-command integration tests that prove the selected format changes real behavior.
- Mutation and checksum tests for format-specific edge cases.
- Conformance fixtures with non-Atlas directory layouts.
- Regression tests ensuring accepted-but-ignored flags cannot return.

## Documentation

Document the supported format matrix, command-specific behavior, limitations, and examples.

## Acceptance criteria

- [ ] Every command that accepts `--dir-format` delegates to the selected format implementation.
- [ ] Unsupported command/format combinations fail explicitly.
- [ ] Native Atlas directory behavior remains unchanged.
- [ ] Format logic is reusable and independent of `cmd/atlas`.
- [ ] Unit, integration, race, lint, vulnerability, and conformance checks pass.
- [ ] The conformance repository pins the resulting Ptah revision and records the reduced gap.

denisvmedia:
## What this is

Re-measurement of #845 at `f22107bd`, then the one bounded piece that survived — plus, after review, the scheme requirement on the two verbs that CREATE a migration directory (#1186).

The 2026-08-03 triage on #845 found hash and validate done and four verbs surviving: `lint`, `new`, `set`, `status`. Three of those four closed in merged work since. This PR closes the fourth.

| verb | at triage (92213b39) | at f22107bd, before this PR | closed by |
| --- | --- | --- | --- |
| `hash` | operational | operational | 7771b3a3 (#992) on e04e9d4a (#984) |
| `validate` | operational | operational | 7771b3a3 (#992) |
| `status` | refused, 10/10 cells | **operational**, 10/10 cells matching | c0f86693 "read a converted migration directory on status and set (#1002) (#1099)" |
| `set` | refused, 10/10 cells | **operational**, 8/10 cells matching | c0f86693 (#1099) |
| `lint` | refused, 10/10 cells | **operational**, 10/10 exit codes matching, 8/10 versions matching | 20c30b7e "read every Atlas source layout on migrate lint (refs #1013) (#1133)" |
| `new` | refused, 10/10 cells | refused, 10/10 cells | **this PR** |

"Cells" = 5 external layouts × 2 spellings (`--dir-format` and `?format=` on `--dir`), each run against the pinned community build v1.3.0 at `~/Work/denis/ptah-atlas-conformance/bin/atlas`. The non-matching cells are Flyway in both cases — `set` exits 1 where the community build exits 0, and `lint` exits 0 on both but prints a different version. Same root cause; see [What I deliberately left open](#what-i-deliberately-left-open) item 3.

## Reproduction, before

`go build -o /tmp/b845 ./cmd/ptah-compat` at `f22107bd`, clean tree.

```
$ atlas   migrate new addcol --dir file://$D --dir-format golang-migrate   exit=0
  -> 20260806071434_addcol.up.sql, 20260806071434_addcol.down.sql, atlas.sum
$ /tmp/b845 migrate new addcol --dir file://$D --dir-format golang-migrate
  Error: atlas migrate new --dir-format: Atlas accepts --dir-format=golang-migrate,
         but Ptah does not implement that directory format yet                exit=1
  -> directory empty

$ /tmp/b845 migrate new addcol --dir "file://$D?format=golang-migrate"
  Error: atlas migrate new --dir: Atlas accepts ?format=golang-migrate, but Ptah does
         not implement that directory format for this command yet             exit=1
```

10 of 10 cells refused; the directory was never touched.

## Reproduction, after

Same 10 cells, both binaries on their own fresh empty directory. Comparison normalizes only the `yyyyMMddHHmmss` version; the file **bytes** are compared with `cmp`.

```
layout           names(community) == names(ptah)                bytes    atlas.sum covers
golang-migrate   _addcol.up.sql _addcol.down.sql           IDENTICAL  the .up.sql only
flyway           V__addcol.sql  U__addcol.sql              IDENTICAL  the V file only
goose            _addcol.sql                                  IDENTICAL  the file
dbmate           _addcol.sql                                  IDENTICAL  the file
liquibase        _addcol.sql                                  IDENTICAL  the file
```

Contents, measured with `od -c` and reproduced exactly: golang-migrate and flyway files are empty; goose is `-- +goose Up\n\n-- +goose Down\n`; dbmate is `-- migrate:up\n\n-- migrate:down\n`; liquibase is `--liquibase formatted sql` with **no** trailing newline (that byte is inside the integrity hash, so tidying it would produce a sum the community build never writes).

The strongest cross-binary check is the round trip. On a populated, already-hashed directory, ptah-compat writes and the community build reads:

```
for each of the five layouts:
  ptah-compat migrate new addcol      exit=0
  atlas       migrate validate        exit=0
  atlas       migrate hash            -> atlas.sum IDENTICAL to the one ptah wrote
  atlas       migrate apply           exit=0, runs the pre-existing migrations and the new one
```

Gates, both binaries, byte-identical stdout guidance and stderr:

```
unhashed golang-migrate dir     both exit=1  "Error: checksum file not found"       nothing created
hashed-then-edited up file      both exit=1  "L2: 1_init.up.sql was edited"          nothing created
                                             "Error: checksum mismatch"
```

Precedence, arity and unknown flags, both binaries:

```
--dir 'file://d?format=flyway' --dir-format goose          both write the Flyway pair
--dir 'file://d?format=atlas'  --dir-format golang-migrate both write the single Atlas file
migrate new a b --dir-format goose      both exit=1  "accepts at most 1 arg(s), received 2"
migrate new a --nosuchflag              both exit=1  "unknown flag: --nosuchflag"
migrate new 'add/col'                   both exit=1
nonexistent --dir                       both exit=0, directory created
unhashed dir whose covered set is empty both exit=0
```

## The scheme requirement on the writing verbs (#1186)

Review of the first two commits found that the change above newly *reached* a parity rule (a) violation on a verb that WRITES: `migrate new addcol --dir mig --dir-format goose` and its four external-layout siblings exited 0 and created `mig/` plus `atlas.sum`, where the community build exits 1 and creates nothing. Master exited 1 on those five cells through the old refusal, so the fix converted five matching cells into violating ones. That is now closed here rather than deferred.

Measured on 2026-08-06 against the pinned v1.3.0. Before, at `6a5ca8b5`:

```
$ atlas      migrate new addcol --dir $D --dir-format goose
  Error: missing scheme for dir url. Did you mean "file:///.../A"?     exit=1  nothing created
$ ptah-compat migrate new addcol --dir $D --dir-format goose
  Generated empty migration file:  /.../A/20260806134610_addcol.sql    exit=0  A/ created

$ ptah-compat migrate diff demo --dir mig3 --dev-url sqlite://dev3.db --to file://target.sql
  Created migration file: .../mig3/1786024266_demo.sql                 exit=0  mig3/ created
```

After, same invocations:

```
$ ptah-compat migrate new addcol --dir mig9 --dir-format goose
  Error: missing scheme for dir url. Did you mean "file://mig9"?       exit=1  nothing created
$ ptah-compat migrate diff demo --dir mig4 --dev-url ... --to ...
  Error: missing scheme for dir url. Did you mean "file://mig4"?       exit=1  nothing created
```

All six `--dir-format` values (the five external layouts and `atlas`) and the `?format=` spelling refuse, and no directory is created in any of them. Byte comparison of stderr with `od -c`, ptah versus the community build:

```
community  Error: missing scheme for dir url. Did you mean "file://mig"? \n
ptah       Error: missing scheme for dir url. Did you mean "file://mig"?\n
```

One byte differs, and it is deliberate: that trailing space is in one of its format strings and not in the others — `Error: checksum file not found\n` on the same verb has none — so it is a slip rather than a convention, and reproducing it would be reproducing a defect.

Suggestion shape and ordering are measured too, not guessed:

```
--dir 'sub/dir?format=goose&x=1'   both suggest "file://sub/dir"       query dropped
--dir 'sub/dir#frag'               both suggest "file://sub/dir"       fragment dropped
--dir './rel'                      both suggest "file://./rel"         path not cleaned
--dir ''                           both suggest "file://"
migrate new a --dir mig --dir-format nosuchformat   both print the scheme refusal, not
                                                    `unknown dir format`
migrate new a b --dir mig --dir-format goose        both print the arity error instead
```

### Which spellings the requirement reaches, and why not the others

The rule is placed on the spellings whose scheme survives to the verb, and every layer that can name the directory was enumerated rather than assumed:

| layer | community build | ptah after this PR |
| --- | --- | --- |
| `--dir` flag | exit 1 | exit 1, nothing created |
| `PTAH_DIR`, the Atlas flag's environment twin | no spelling for it | exit 1, nothing created |
| `PTAH_MIGRATIONS_DIR`, the NATIVE `--migrations-dir` under its environment name | no spelling for it | still accepts a plain path |
| `atlas.hcl` `migration.dir` | exit 1 | **still accepts a bare path — #1186** |
| flag default | exit 0 | exit 0 (`migrate diff`'s default carries `file://`) |

The `atlas.hcl` row is the one that is not closed, and the reason is mechanical rather than a choice of scope: `config/projectconfig.normalizeAtlasMigrationDir` strips `file://` at parse time, so by the time any verb sees the value, `file://mig` and `mig` are the same string. A requirement placed there would refuse both, and the community build refuses only the second — measured, `migrate new addcol --env local` exits 1 there against `dir = "mig"` and 0 against `dir = "file://mig"`. Recovering the spelling means changing what the project loader keeps, which is a public-surface decision that belongs to #1186. `TestCompatMigrateNew_LeavesTheProjectFileSpellingToIssue1186` pins the gap so that closing it later has to come back here.

## What changed

- **`internal/atlasmigrateimport/skeleton.go`** (new) — `SkeletonFiles(format, version, name)` returns the file set one layout's empty migration consists of. It sits beside `SumFileNames` and the loader regexes on purpose: an emitted name has to be a name the integrity rule covers and the loader loads, and that is a property one test can state over both rather than an agreement between two packages. `cmd/atlas` only wires.
- **`cmd/atlas/migrate_new.go`** — the refusal becomes a dispatcher. The converted path gates with `verifyCoveredAtlasDirChecksum` (the same function and the same `requireAtlasSum` policy `migrate apply` uses) **before** creating anything, writes the layout's files, then rewrites `atlas.sum` through the same `SumFileNames` + `ComputeAtlasFiles` pair `migrate hash` writes from. A half-written pair is rolled back; a failed rehash removes what was created.
- **`cmd/atlas/migrate_integrity.go`** — the per-verb pflag set is extracted as `atlasVerbFlagSet` and shared with the converted `migrate new` path, so "which flags does this verb accept" keeps one answer. `resolveAtlasMigrateSourceDir` now also reports whether the directory came from the Atlas-facing layer, and the scheme requirement runs on that value for a verb marked `writesDir`.
- **`internal/atlasargs/args.go`** — `RequireDirScheme` renders the community build's refusal. It is a separate call rather than a change to `ParseLocalDir`, because the read verbs still accept a bare path and closing that half is #1186.
- **`cmd/atlas/migrate_diff.go`** — the same requirement, read off `--dir` as the flag layer left it, before the atlas.hcl merge can replace it.
- **`cmd/atlas/atlas.go`** — `atlasVerb.writesDir`, which is what selects the requirement.
- **docs** — the `migrate new` matrix in `atlas/migrate-commands.md` and `reference/atlas-commands.md`, plus feature-matrix rows 46 and 73, plus the scheme requirement and the layers it does and does not reach.

Verified by mutation, not asserted. Restoring the refusal in the dispatcher (`if false && !ReadsNativeAtlasDir(...)`):

| test | reverted |
| --- | --- |
| `TestCompatMigrateNewConverted_WritesTheSelectedLayout` (5 rows) | all 5 fail, each naming its own layout, e.g. `Atlas accepts --dir-format=flyway, but Ptah does not implement that directory format yet`; directory empty |
| `TestCompatMigrateNewConverted_QuerySpellingWritesTheSelectedLayout` (5 rows) | all 5 fail with the `?format=` wording of the same refusal |
| `TestCompatMigrateNewConverted_QueryFormatOutranksDirFormatFlag` (2 rows) | **one row**, `query_selects_flyway_over_goose`, fails with `atlas migrate new --dir-format: Atlas accepts --dir-format=flyway, but Ptah does not implement that directory format yet`. `query_selects_atlas_over_golang_migrate` PASSES: the query selects the atlas layout, so it routes down the forwarding branch either way |
| `TestCompatMigrateNewConverted_CreatedDirectoryVerifies` (3 rows) | all 3 fail at `migrate new`; the validate and status follow-ups never run |
| `TestCompatMigrateNewConverted_GateRefusesBeforeWriting` (2 rows) | both still exit 1, but with the refusal text instead of `checksum file not found` / `checksum mismatch` — which is why each row asserts the diagnostic, not the exit code |
| `TestCompatMigrateNewConverted_RefusesWhatItCannotReadBack` (3 rows) | all 3 still exit 1, with the refusal instead of the named diagnostic; the directory-empty assertion is unchanged |
| `TestSkeletonFilesMatchTheOracleLayout` (5 rows) | with the emitter removed the package does not compile; with one row's bytes changed, that row prints the wanted and got `[]SkeletonFile` |
| `TestSkeletonFilesAreCoveredAndLoadable` (5 rows) | emitting `.upp.sql` for golang-migrate moves **that row only**, and it fails on the membership assertion rather than on either count, because that one runs first: `no matching element found` / `container: []string{"1_init.up.sql"}` / `want: "20260806071434_addcol.upp.sql"` |
| `TestSkeletonFilesRejectsWhatItCannotName` (6 rows) | with either guard removed, the empty-name row returns `20260806071434_.up.sql` and the separator row returns `20260806071434_add/col.up.sql` |
| `TestSkeletonFilesRejectsTheNativeAtlasLayout` (2 rows) | with the atlas guard removed, the first row returns `20260806071434_addcol.sql` and passes silently |

`TestCompatMigrateNewConverted_NativeAtlasLayoutIsUnchanged` is the non-interference control for "native Atlas directory behavior remains unchanged", and reverting cannot prove it — it passes before and after by construction. It was proved with the **inverse** mutant: routing the atlas layout down the converted path as well (`if true || !ReadsNativeAtlasDir(...)`). Measured, **4 rows move, not 3**: all 3 of its rows, and `_QueryFormatOutranksDirFormatFlag/query_selects_atlas_over_golang_migrate` with them, which is that same forwarding branch under another name. Nothing else in `TestCompatMigrateNewConverted*` moves.

### The scheme requirement's own mutants

| mutant | what moves |
| --- | --- |
| **revert** — `if false && verb.writesDir && …` in `resolveAtlasMigrateSource`, and the `RequireDirScheme` call in `runAtlasMigrateDiff` disabled | all 10 rows of `TestCompatMigrateWrite_RefusesADirectoryNamingNoScheme` and both rows of `TestCompatMigrateNew_RequiresTheSchemeOnEveryAtlasSpelling` fail with `got nil error but want non-nil` against the regexp `missing scheme for dir url\. Did you mean "file://…"\?`, and the fingerprint assertion shows a `20260806nnnnnn_demo.sql` and an `atlas.sum` in a directory that did not exist when the command started. The three control tests stay green |
| **inverse** — requirement made unconditional (`spelledByAtlas` alone, and `opts.dirURL` instead of `dirURLSpelled` on diff) | `…SpellingsTheCommunityBinaryAcceptsStillWrite/atlas.hcl names the directory as a file:// URL`, both rows of `TestCompatMigrateRead_StillAcceptsADirectoryNamingNoScheme`, and `TestCompatMigrateNew_LeavesTheProjectFileSpellingToIssue1186` fail with `got non-nil error` carrying `missing scheme for dir url` |
| **inverse, second axis** — the layer condition dropped (`verb.writesDir && atlasDirSchemeIsAnswerable(…)`) | `…SpellingsTheCommunityBinaryAcceptsStillWrite/PTAH_MIGRATIONS_DIR carries a bare path` fails, and nothing else. That row is the only thing separating "the Atlas spelling" from "any spelling" |

Two existing tests asserted the removed refusal and were updated rather than deleted:

- `TestCompatMigrateDirQuery_FailurePathForeignFormat` lost its `new` row and keeps `diff`, which genuinely still refuses. The comment now says which verb survives and why. That row passes before and after this change; it guards the unknown-key relaxation, not #845.
- `TestCompatCommand_MigrateMetadataRejectsUnsupportedAtlasDirFormat` lost its `new` row. `edit`, `rebase`, `rm` and `test` remain — they rewrite migration bodies, and nothing emits a body in a source tool's convention.

Two more were updated for the scheme requirement: `TestCompatCommand_MigrateNewCreatesAtlasSkeletonFileByDefault` and `TestCompatCommand_MigrateNewAcceptsExplicitAtlasDirFormat` passed a bare `--dir`; they now pass `file://` + the path, which is what the community build has always required. Those two are the entire blast radius inside this repository — the rest of the suite already spelled the scheme.

## What I deliberately left open

**1. Two refusals drawn narrower than the community build, on purpose.**

*An omitted migration name.* That binary exits 0 and writes the version alone. Applying exactly that file, hashed by it, measured on 2026-08-06:

```
golang-migrate  community 0, version "20260806071553.up"  ptah 1, "no importable migration files found"
goose           community 0, version "20260806071553"     ptah 1, same
liquibase       community 0, version "20260806071553"     ptah 1, same
dbmate          community 0, version "20260806071553"     ptah 1, same
flyway          community 0, version "20260806071553"     ptah 0, version 5438407949371077319
```

Writing it would hand the operator a directory this binary refuses on four layouts of five. Exit 1 where the community build exits 0 is the strict direction, and it is the direction to take while the readers disagree.

*`--edit`.* Measured, the community build refuses it on every non-atlas layout (`--edit flag supports only atlas directories, but got: *sqltool.Dir`). Exit codes match; the text is Ptah's own.

**2. Parity rule (a) — what this PR closes and what still survives.**

Closed here, on both verbs that create a directory: a scheme-less `--dir` and a scheme-less `PTAH_DIR`. That covers the five cells this PR newly reached, the `--dir-format atlas` cell that predates it, and the same class on `migrate diff`.

Still surviving, all of it pre-existing and none of it newly reached by this change:

- *Scheme-less `--dir` on the verbs that only READ a directory.* `migrate hash`, `migrate validate`, `migrate status` and `migrate lint` exit 0 where the community build exits 1. Verified still true here on `hash` and `validate` as a deliberate control, so the change is scoped rather than accidental. Tracked in #1186.
- *Scheme-less `atlas.hcl` `migration.dir`, on every verb including the two that write.* `migrate new addcol --env local` against `dir = "mig"` exits 0 here and creates `mig/`, where the community build exits 1. Same issue, #1186; the mechanical reason is above.
- *The native sanitizer.* `migrate new 'añb' --dir-format atlas` still exits 0 here writing `_ab.sql` where the community build writes `_añb.sql`. That is the native atlas path's sanitizer (spaces to `-`, everything outside `[A-Za-z0-9_.-]` dropped), it predates this change, and the converted path deliberately does **not** use it — it passes the name through verbatim as the community build does. Changing the native sanitizer would move file names in existing Atlas directories and is outside #845.

One divergence in the other direction, also pre-existing and also measured today: `migrate new addcol` with no `--dir` at all exits 1 here with `migrations directory is required`, while the community build exits 0 and creates `migrations/` from its own `file://migrations` default. It reproduces identically on the commit before this PR's first, so this change neither causes nor widens it. No issue tracks it yet.

**3. Flyway version identity — the reason #845 is closed on delegation but not on identity, and the reason `set` and `lint` are 8/10 rather than 10/10 cells.** This is not `migrate new`'s and this PR does not touch it. Measured at `f22107bd` on a two-file Flyway directory:

```
$ atlas   migrate apply --dir "file://$D?format=flyway" --url sqlite://a.db   exit=0
  atlas_schema_revisions:  1|init   2|more
$ ptah    migrate apply --dir "file://$D?format=flyway" --url sqlite://b.db   exit=0
  atlas_schema_revisions:  4611686018427469511|init   4611686018427510315|more

$ atlas   migrate set 1 --dir file://$D --dir-format flyway   exit=0  "Current version is 1 (1 set)"
$ ptah    migrate set 1 --dir file://$D --dir-format flyway   exit=1  Error: migration with version "1" not found

$ atlas   migrate lint --dir file://$D --dir-format flyway --latest 1
  "Analyzing changes from version 1 to 2"                      exit=0
$ ptah    same
  "Analyzing changes from version 4611686018427469511 to 4611686018427510315"   exit=0
```

The number is the Flyway **ordering key** `internal/atlasmigrateimport` synthesizes (bands + component slots + a tie budget, documented at `importer.go:100-150`), not the version **token** the community build records. `FlywaySourceVersions` already recovers the token, but only for the linearity guard in `migrate_apply.go:638`. Every one of the six verbs in #845 delegates to the selected format now; on Flyway the format's own identity answer is the ordering key, which is a defect in the format implementation shared with `migrate apply` — a verb #845 does not name. Fixing it changes what `apply` records for anyone already on Ptah's Flyway conversion and reworks the design #1098/#1132 deliberately built, so it wants its own issue rather than a rider here.

**4. `migrate diff` on a foreign layout.** It emits planned migration SQL, and nothing renders a migration body in a source tool's convention — `SkeletonFiles` writes empty files and directive headers, not statements. Refusing stays, under both spellings. It is the last item of #1013 section 1.

**5. Unknown-format diagnostic text still differs**, unchanged by this PR: `Error: unknown dir format "nosuchformat"` there versus `atlas migrate new --dir-format: unknown Atlas migration directory format "nosuchformat": expected atlas, golang-migrate, goose, flyway, liquibase, or dbmate` here. Both exit 1. Tracked in #1013.

**6. `migrate new` prints where the community build is silent.** That binary writes nothing on success; ptah-compat prints `Generated empty migration file(s):` and the paths. This predates the change — it is what `--dir-format atlas` already does on the forwarding branch — and matching it on the converted branch alone would make one verb report two ways.

## Gates

All green at `HEAD` of this branch unless stated:

```
go build ./...                        exit 0
go vet ./...                          exit 0
go test ./... -count=1 -timeout 40m   exit 1 on one flake, see below; 175 packages ok
go tool qtlint ./...                  exit 0, no output
golangci-lint run ./...               exit 0, "0 issues."
scripts/check-test-style.sh           exit 0, "teststyle: OK (175 conditional groups, 45 white-box files)"
scripts/check-public-api.sh           exit 0, no output
scripts/check-public-api-snapshot.sh  exit 0, no output
gofmt -l cmd/atlas internal/atlasargs internal/atlasmigrateimport config/projectconfig  no output
```

The default 10-minute `go test` timeout is not enough for this repo — `cmd/atlas` alone runs 170s+ and the whole suite panics at `panic: test timed out after 10m0s` without `-timeout`.

The one failure in the full run is `cmd/viz TestSVGReportsGraphvizStderrOnFailure`, with `render SVG with Graphviz dot: signal: killed` after 10.01s spawning its fake `dot` under machine load. That package is untouched by this branch; re-run in isolation with `-count=3` it exits 0.

Two gates that print nothing on success were confirmed to be running rather than skipping, by breaking the rule they check:

- `scripts/check-test-style.sh` — an `if err != nil` added to a test body reddened it: `teststyle baseline mismatch / new test conditional violations: cmd/atlas/migrate_write_dir_scheme_test.go TestCompatMigrateRead_StillAcceptsADirectoryNamingNoScheme if 1`.
- `scripts/check-public-api-snapshot.sh` — a throwaway exported `func ZZAPIProbe()` in `config/projectconfig` reddened it with `+func ZZAPIProbe()` in the snapshot diff.

Both were reverted and both gates are green on the pushed tree.

`docs/` is touched, so all 14 `check:*` scripts in `docs/site/package.json` were run, not a sample:

```
check:exit-codes            OK (60 reference rows)
check:exit-codes:selftest   OK
check:core-doc-links        OK (7 protected core references)
check:page-health           OK (59 sidebar entries)
check:links                 OK (60 pages, 60 routes, 571 heading anchors)
check:links:selftest        OK
check:redirects             OK (30 redirects)
check:redirects:selftest    OK
check:style                 OK (100 documentation files)
check:style:selftest        OK (16 rule assertions via analyze())
check:limitations           OK (1 governed page(s))
check:limitations:selftest  OK (7 assertions via analyze())
check:responsive            OK (60 routes x 2 viewports, cells within 8 lines)
check:responsive:selftest   OK (overflow, wide-table, tall-cell, and unstyled-page guards verified)
```

`check:responsive` reports `dist not found; run "npm run build" first` and exits 1 on a fresh checkout — it was run after `npm run build` (61 pages built). Running it without the build would have been a gate reporting without checking anything.

`node docs/site/scripts/build-feature-matrix.mjs --check` (not an npm script) also exits 0: `OK (163 rows)`.

## Why Refs and not Closes

Refs #845. The `migrate new` half is finished and nothing about it is pending. What survives is item 3 above, and it is not a technicality: on a Flyway directory, `migrate set 1 --dir-format flyway` **exits 1 where the pinned community build exits 0**, on one of the six verbs #845 names. Anyone re-measuring the first acceptance criterion with a Flyway fixture will find it red, which is exactly the later measurement pass a premature Closes costs.

It is not fixable inside this PR, and papering over it would be worse than the refusal. Translating the token `1` to the ordering key on `set` alone would insert a revision row numbered `4611686018427469511` where the community build inserts `1` — the same divergence, now written into the revision table instead of reported. The identity has to change at the converter, which moves what `migrate apply` records for every existing Flyway user and reworks the design #1098/#1132 built deliberately.

**Suggested disposition:** open one issue for "Flyway converted version identity is the ordering key, not the source token — `apply` records it, `set` cannot address it, `lint` prints it", link it as the last child of #845, and close #845 when it closes. If instead #845 is scoped to delegation alone — every named verb routing through the selected format — then everything else in it is done as of this PR and it can be closed on merge.

Refs #845. Refs #1013 (section 1 now leaves only `migrate diff`). Refs #1186 (the writing half of it is closed here; the read verbs and the `atlas.hcl` layer are not).


denisvmedia:
Marked draft. The fix itself is real and well measured — master refuses all 10 cells (5 external layouts × {`--dir-format`, `?format=`}) where the pinned binary exits 0, and this branch makes all 10 pass with byte parity and a clean 5/5 round trip. That part stands.

**Blocking: the branch newly reaches a parity rule (a) violation, on a verb that WRITES.**

```
migrate new addcol --dir mig --dir-format goose      (no file:// scheme)
  pinned binary: exit 1, missing scheme for dir url. Did you mean "file://mig"?  — writes nothing
  branch:        exit 0, creates mig/ + the migration + atlas.sum
```

Five cells. Master exited 1 on the same five through the old refusal, so this PR converts five matching cells into violations. Sharper still:

```
migrate new --dir /tmp/nosuchxyz --dir-format goose
  pinned binary: exit 1
  branch:        exit 0, and it wrote to /private/tmp/nosuchxyz — outside the working directory
```

The reviewer's control is what makes this actionable rather than alarming: the class is **pre-existing and uniform**, not created here. On master, `migrate hash|validate|status|lint --dir mig --dir-format goose` all exit 0 where the binary exits 1, and `migrate new --dir mig --dir-format atlas` exits 0 on master too. Root cause is `internal/atlasargs/args.go` `ParseLocalDir` — "Plain filesystem paths are preserved verbatim" — which this PR does not touch. Positive control: `--dir file://mig` exits 0 on both.

So the body's line *"PARITY RULE (a) SURVIVES UNFIXED on one input"* is understated. There are two, and the second is a **writing** one this change newly reaches. Filed the root cause as its own issue so it is not smuggled in here; this PR needs either to name the second input honestly in the body, or to gate the scheme on the writing path.

Worth recording what the review checked and found matching, because it is most of the surface: 12 format-alias and spelling cells (`GOOSE`, `golang_migrate`, `golangmigrate`, `sqltool`, empty `--dir-format`, unknown format, query variants, duplicate `format` key with first-wins on both), 10 checksum-gate cells with byte-identical stdout guidance *and* stderr and nothing written, 18 name/arity cells, 4 hostile directories (corrupt sum, empty sum, unreadable covered file, read-only dir), 6 `--dir` shapes and 4 `atlas.hcl` cells.

denisvmedia:
Independently re-measured the blocker and its control on `713c753c`, against the pinned community binary by absolute path, each run in its own empty directory.

**The two blocking cells now match, and write nothing:**

```
$ atlas       migrate new addcol --dir mig --dir-format goose
rc=1  Error: missing scheme for dir url. Did you mean "file://mig"?
$ ptah-compat migrate new addcol --dir mig --dir-format goose
rc=1  Error: missing scheme for dir url. Did you mean "file://mig"?
    directory created: no

$ atlas       migrate new addcol --dir /tmp/nosuchxyz --dir-format goose
rc=1  Error: missing scheme for dir url. Did you mean "file:///tmp/nosuchxyz"?
$ ptah-compat migrate new addcol --dir /tmp/nosuchxyz --dir-format goose
rc=1  Error: missing scheme for dir url. Did you mean "file:///tmp/nosuchxyz"?
    /tmp/nosuchxyz: absent
```

**Positive control — the gate did not over-refuse the shape this PR exists for.** A scheme gate that fixed the blocker by refusing everything would pass the two cells above and quietly undo the whole change, so this is the half that has to be run:

```
$ atlas       migrate new addcol --dir file://mig --dir-format goose
rc=0   mig/20260806150046_addcol.sql   mig/atlas.sum
$ ptah-compat migrate new addcol --dir file://mig --dir-format goose
rc=0   mig/20260806150046_addcol.sql   mig/atlas.sum
```

Same exit code, same file set, on the layout the branch added.

denisvmedia:
Blocking finding addressed by closing it, not by disclosing it. Two commits on top of `6a5ca8b5`; PR body rewritten around what is now true. Out of draft.

## The violation, and that it is gone

Reproduced first, at `6a5ca8b5`, both binaries by absolute path:

```
$ /Users/buster/Work/denis/ptah-atlas-conformance/bin/atlas \
    migrate new addcol --dir /private/tmp/.../w1172/A --dir-format goose
Error: missing scheme for dir url. Did you mean "file:///private/tmp/.../w1172/A"?
pinned exit=1

$ ./compat-branch migrate new addcol --dir /private/tmp/.../w1172/A --dir-format goose
Generated empty migration file:
SQL:  /private/tmp/.../w1172/A/20260806134610_addcol.sql
branch exit=0
```

`migrate diff` had the same hole, and it also created the directory:

```
$ ./compat-branch migrate diff demo --dir mig3 --dev-url sqlite://dev3.db --to file://target.sql
Created migration file: .../P3/mig3/1786024266_demo.sql
Updated migration checksum: .../P3/mig3/atlas.sum
branch-diff exit=0            # pinned binary: exit 1, mig3 never created
```

After, every `--dir-format` value including `atlas`, plus the `?format=` spelling, and nothing created in any of them:

```
golang-migrate  Error: missing scheme for dir url. Did you mean "file://mig_gm"?      exit=1
goose           Error: missing scheme for dir url. Did you mean "file://mig_goose"?   exit=1
flyway          Error: missing scheme for dir url. Did you mean "file://mig_flyway"?  exit=1
dbmate          Error: missing scheme for dir url. Did you mean "file://mig_dbmate"?  exit=1
liquibase       Error: missing scheme for dir url. Did you mean "file://mig_lq"?      exit=1
atlas           Error: missing scheme for dir url. Did you mean "file://mig_atlas"?   exit=1
?format=goose   Error: missing scheme for dir url. Did you mean "file://mig_q"?       exit=1

$ ls -d mig_*
(eval):1: no matches found: mig_*
```

Stderr is the pinned binary's own bytes but one. `od -c` on both:

```
community  ... " f i l e : / / m i g "  ?  SP \n
ptah       ... " f i l e : / / m i g "  ?     \n
```

The trailing space is in that one format string and not in its siblings — `Error: checksum file not found\n` on the same verb has none — so it is a slip, and rule (b) says do not reproduce a defect for the sake of identity. Exit code is 1 on both.

The suggestion's shape was measured rather than guessed, and the three cells that separate the plausible implementations are in the table test: `'sub/dir?format=goose&x=1'` and `'sub/dir#frag'` both suggest `"file://sub/dir"` (query and fragment dropped), `./rel` suggests `"file://./rel"` (path not cleaned), `''` suggests `"file://"`. Ordering too: `--dir mig --dir-format nosuchformat` prints the scheme refusal, not `unknown dir format`, on both; a second positional still beats both on both.

## Every layer that can name the directory, not just the one the report named

| layer | pinned binary | ptah now |
| --- | --- | --- |
| `--dir` flag | exit 1 | exit 1, nothing created |
| `PTAH_DIR` (the Atlas flag's environment twin) | no spelling for it | exit 1, nothing created |
| `PTAH_MIGRATIONS_DIR` (the NATIVE `--migrations-dir`) | no spelling for it | still takes a plain path |
| `atlas.hcl` `migration.dir` | exit 1 | **still takes a bare path — #1186** |
| flag default | exit 0 | exit 0 |

The `atlas.hcl` row is not a scope decision, it is mechanical: `config/projectconfig.normalizeAtlasMigrationDir` strips `file://` at parse time, so `file://mig` and `mig` are the same string by the time any verb sees the value. Requiring the scheme there refuses both, and the binary refuses only one — measured, `migrate new addcol --env local` exits 1 there against `dir = "mig"` and 0 against `dir = "file://mig"`. My first attempt did exactly that and the control caught it. `TestCompatMigrateNew_LeavesTheProjectFileSpellingToIssue1186` now pins the gap, and I have left a note on #1186.

## Mutants, three of them, because a guard needs the inverse

**Revert** (`if false && verb.writesDir && …`, and the `RequireDirScheme` call in `runAtlasMigrateDiff` disabled) — 10 rows of `..._RefusesADirectoryNamingNoScheme` and both rows of `..._RequiresTheSchemeOnEveryAtlasSpelling` go red, controls stay green:

```
--- FAIL: TestCompatMigrateWrite_RefusesADirectoryNamingNoScheme/new/bare_path
        error:      got nil error but want non-nil
        got error:  nil
        regexp:     `missing scheme for dir url\. Did you mean "file:///.../001/mig"\?`
```

**Inverse** (requirement made unconditional) — exactly the three control tests, and only them:

```
--- FAIL: .../SpellingsTheCommunityBinaryAcceptsStillWrite/atlas.hcl_names_the_directory_as_a_file://_URL
--- FAIL: TestCompatMigrateRead_StillAcceptsADirectoryNamingNoScheme/hash
--- FAIL: TestCompatMigrateRead_StillAcceptsADirectoryNamingNoScheme/validate
--- FAIL: TestCompatMigrateNew_LeavesTheProjectFileSpellingToIssue1186
          got non-nil error / e`missing scheme for dir url. Did you mean "file:///.../mig"?`
```

**Inverse, second axis** (the layer condition dropped) — one row moves, and it is the one that separates "the Atlas spelling" from "any spelling":

```
--- FAIL: .../SpellingsTheCommunityBinaryAcceptsStillWrite/PTAH_MIGRATIONS_DIR_carries_a_bare_path
```

## Two other claims in the body were wrong; both re-measured and corrected

- `..._QueryFormatOutranksDirFormatFlag` did **not** have both rows red under the revert. Measured: `query_selects_flyway_over_goose` fails, `query_selects_atlas_over_golang_migrate` **passes** — the query selects atlas, so it routes down the forwarding branch either way.
- The inverse mutant does **not** move only `..._NativeAtlasLayoutIsUnchanged`'s 3 rows. Measured, 4 move: its three plus that same `query_selects_atlas_over_golang_migrate`.
- Also corrected in `skeleton_test.go`: the comment predicted `got 1 want 2` and `no importable migration files found`. Emitting `.upp.sql` for golang-migrate, neither string appears — the membership assertion runs first and prints `no matching element found` / `container: []string{"1_init.up.sql"}` / `want: "20260806071434_addcol.upp.sql"`, and only that row moves.

## Blast radius, and a divergence I found while measuring

Two existing tests passed a bare `--dir` to `migrate new` and now pass `file://` + the path. That is the whole blast radius in this repository; everything else already spelled the scheme.

Separately, and pre-existing: `migrate new addcol` with no `--dir` at all exits 1 here with `migrations directory is required` where the pinned binary exits 0 and creates `migrations/` from its `file://migrations` default. It reproduces identically on `6a5ca8b5`, so this change neither causes nor widens it. Recorded in the body; no issue tracks it yet.

## Gates

`go build`, `go vet`, `go tool qtlint`, `golangci-lint run` (`0 issues.`), `check-test-style.sh` (`OK (175 conditional groups, 45 white-box files)`), `check-public-api.sh`, `check-public-api-snapshot.sh` — all exit 0. `go test ./... -count=1 -timeout 40m` has one failure, `cmd/viz TestSVGReportsGraphvizStderrOnFailure` with `render SVG with Graphviz dot: signal: killed`; that package is untouched here and passes `-count=3` in isolation.

The two gates that print nothing on success were confirmed live rather than trusted: an `if err != nil` in a test body reddened `check-test-style.sh` (`new test conditional violations: cmd/atlas/migrate_write_dir_scheme_test.go … if 1`), and a throwaway `func ZZAPIProbe()` in `config/projectconfig` reddened the snapshot gate (`+func ZZAPIProbe()`). Both reverted.

`docs/` changed, so all 14 `check:*` scripts ran green, `check:responsive` after `npm run build` (`OK (60 routes x 2 viewports, cells within 8 lines)`), plus `node docs/site/scripts/build-feature-matrix.mjs --check` → `OK (163 rows)`.

denisvmedia:
Follow-up: CI went red on the push above, and it was not the scheme change. Recording the diagnosis because the failure mode is one this repository will hit again.

`unit-tests` failed on `TestCompatMigrateNewConverted_CreatedDirectoryVerifies`, all three rows:

```
    migrate_new_converted_test.go:315:
        error:     no substring match found
        container: "Migration Status: PENDING\n  -- Current Version: No migration applied yet\n
                    -- Next Version:    1\n  -- Executed Files:  0\n  -- Pending Files:   2\n"
        want:      "Total Migrations: 2"
```

That is master's report, not this branch's. `Go Unit Tests` was green on `6a5ca8b5` and red on both commits I pushed, and the difference is not in the diff: CI builds the PR **merge** commit, and #1102/#1171 landed the Atlas-mirrored `migrate status` shape on master at 14:37, between those runs. The branch alone still printed the native `Total Migrations: 2` block, so the test passed locally and failed merged — a semantic conflict a clean textual merge cannot show.

Fixed properly rather than by making the assertion loose: master is merged into the branch (b7ab9984), and the round-trip count now comes off the mirrored line with its column alignment, `  -- Pending Files:   2`. Discriminator checked — asserting `9` instead of the row's count reddens all three rows with `no substring match found`, so the assertion still measures the count rather than the shape.

Every gate re-run on the merged tree:

```
go build ./...                        exit 0
go vet ./...                          exit 0
go test ./... -count=1 -timeout 40m   exit 0, 176 packages ok, 0 FAIL lines
go tool qtlint ./...                  exit 0
golangci-lint run ./...               exit 0, "0 issues."
scripts/check-test-style.sh           exit 0, "OK (175 conditional groups, 45 white-box files)"
scripts/check-public-api.sh           exit 0
scripts/check-public-api-snapshot.sh  exit 0
```

The `cmd/viz` flake did not reproduce this time; the suite is clean end to end. `gofmt -l .` still prints only `internal/goannotationexport/testdata/cleanup/models.go`, which is untouched testdata and pre-existing on master.

All 14 docs `check:*` green after `npm ci` and `npm run build` (61 pages), `check:responsive` `OK (60 routes x 2 viewports)`, and `node docs/site/scripts/build-feature-matrix.mjs --check` `OK (164 rows)` — 164 rather than 163 because the merge brought a row with it.

One CI job is red for an unrelated reason: `gosec` failed at the runner's action-resolution step, before any code ran —

```
##[error]Unable to resolve action `actions/checkout@v7`, unable to find version `v7`.
Unable to resolve action `actions/setup-go@v7` ... `github/codeql-action@v4.37.6` ...
`golangci/golangci-lint-action@v9`
```

Nothing in this branch touches workflow files. `govulncheck`, which runs in the same workflow, passed.

denisvmedia:
## Adversarial re-review at `4aa06ee8` — the named blocker is closed; three problems survive, one of them introduced by the fix

I re-ran every measurement instead of reading the body, and I tried to break the new gate rather than confirm it.

**Headline: the writing-path scheme gate is real, correctly scoped, and its tests discriminate.** The blocker that held this PR in draft is gone. But the change introduces a divergence the body claims does not exist, and all three of the new mutation tables are wrong.

---

### Verified — the blocker is closed

Pinned binary `/Users/buster/Work/denis/ptah-atlas-conformance/bin/atlas`, self-reported `atlas community version v1.3.0`, each cell in its own fresh empty directory:

| cell | community | ptah `4aa06ee8` | directory |
| --- | --- | --- | --- |
| `migrate new addcol --dir mig --dir-format {atlas,golang-migrate,goose,flyway,liquibase,dbmate}` | exit 1 ×6 | exit 1 ×6 | never created, either side |
| `migrate new addcol --dir 'mig?format=<layout>'` ×5 | exit 1 ×5 | exit 1 ×5 | never created, either side |

stderr is byte-identical except the community binary's trailing space, which the PR deliberately does not reproduce — confirmed with `od -c`. Suggestion shapes reproduce too: `./rel` → `"file://./rel"`, `''` → `"file://"`.

Controls, all still exit 0 and write `atlas.sum`: `--dir file://mig`, `--dir file:///abs`, `--dir file://mig?format=goose`, `--dir-format golang-migrate` with a `file://` dir, `PTAH_DIR=file://…`, `PTAH_MIGRATIONS_DIR=<bare path>`, and `atlas.hcl migration.dir = "file://mig"` under `--env local`.

I attacked the `atlasDirSchemeIsAnswerable` exemption twice, expecting it to swallow a command-line bare `--dir` whenever it collided with the resolved project path. **Both attempts failed to break it** — with `atlas.hcl dir = "file://mig"` present, both `--dir mig` and `--dir <absolute resolved path>` still refuse on `migrate new`, matching the community binary. The exemption really is confined to the layer it names.

`migrate diff` is better than the body says: on the base commit `migrate diff a b --dir mig …` exited **0 and created the file**; it now exits 1. That is a rule (a) hole closed, not just a text change.

Mutation discipline holds where it counts: the revert reddens 10 leaf rows with `got nil error but want non-nil` while all three controls stay green, and the inverse reddens the controls. Gates I re-ran here are green: `go build`, `go vet`, `go tool qtlint`, `check-test-style.sh` (`OK (175 conditional groups, 45 white-box files)`), `check-public-api.sh`, `check-public-api-snapshot.sh`, `golangci-lint run ./...` (`0 issues.`), `build-feature-matrix.mjs --check` (`OK (164 rows)`).

---

### 1. The fix newly breaks two cells that matched byte-for-byte before it

`migrate new a b --dir mig`, byte-compared with `od -c`:

```
community   Error: accepts at most 1 arg(s), received 2\n
base        Error: accepts at most 1 arg(s), received 2\n     <- byte-identical
4aa06ee8    Error: missing scheme for dir url. Did you mean "file://mig"?\n
```

`migrate new a --dir mig --nosuchflag`:

```
community   Error: unknown flag: --nosuchflag
base        Error: unknown flag: --nosuchflag                 <- matched
4aa06ee8    Error: missing scheme for dir url. Did you mean "file://mig"?
```

Both sides still exit 1, so this is not a rule (a) breach. It is a drop-in regression on the diagnostic axis, on two cells that previously agreed exactly — and the PR body asserts the opposite as a measured result:

> `migrate new a b --dir mig --dir-format goose        both print the arity error instead`

Measured, ptah prints the scheme refusal there. The same claim is baked into the code:

`cmd/atlas/migrate_integrity.go:156` — *"Cobra's arity check still wins over both there, and does here too, because it runs before RunE."*

That is true of the community binary and false of ptah-compat. On this surface the arity and unknown-flag diagnostics come from the **forwarded native command**, not from cobra on the adapter, so a gate inside `resolveAtlasMigrateSource` preempts both. `migrate diff` keeps `unknown flag` first (real cobra parsing) but loses arity the same way.

Either move the requirement behind the arity/unknown-flag rejection, or correct the precedence block and that comment and disclose the two cells under "what I deliberately left open". What must not stand is the body claiming parity on a row where there is none.

### 2. All three mutation tables for the scheme requirement are wrong

Baseline for attribution: `go test ./cmd/atlas -run TestCompat -count=1 -v` at `4aa06ee8` produces **zero** `--- FAIL` lines, so everything below is mutant-caused.

**revert** — body: *"all 10 rows of `TestCompatMigrateWrite_RefusesADirectoryNamingNoScheme` and both rows of `TestCompatMigrateNew_RequiresTheSchemeOnEveryAtlasSpelling`"*. That first test has **8** rows (2 verbs × 4 spellings), not 10. Ten is the total across both tests, counted once and then added to again.

**inverse** (`spelledByAtlas` alone; `opts.dirURL` on diff) — body names 4 rows. Measured: **26 rows across 20 tests** go red, including `TestCompatCommand_MigrateHashDefaultsToAtlasSum`, `TestCompatCommand_Migrate{New,Hash,Validate,Diff}ResolvesProjectRelative*`, `TestCompatMigrateIntegrityProjectConfig_HappyPath` (4 rows), `TestCompatMigrateIntegrityProjectConfigPrecedence_HappyPath` (2), `TestCompatMigrateWrite_GatesTheDirectoryNamedByAtlasProjectConfig` (2), `TestCompatCommandProjectSelection*` (3), and the three the table does name.

**inverse, second axis** (layer condition dropped) — body: *"`…/PTAH_MIGRATIONS_DIR carries a bare path` fails, **and nothing else**. That row is the only thing separating 'the Atlas spelling' from 'any spelling'."* Measured, two more rows fail:

```
--- FAIL: TestCompatMigrateWrite_SpellingsTheCommunityBinaryAcceptsStillWrite/PTAH_MIGRATIONS_DIR_carries_a_bare_path
--- FAIL: TestCompatMigrateNew_GatesTheDirectoryTheForwardedCommandWillWrite/native_spelling_unhashed
--- FAIL: TestCompatMigrateNew_GatesTheDirectoryTheForwardedCommandWillWrite/native_spelling_hashed
```

So "and nothing else" is false and the conclusion drawn from it is false.

The numbers read like they came from a `-run` filter narrowed to the new test names. That under-reports in the safe direction here — the code is better covered than claimed — but it is the same defect the earlier review already caught on this PR's first mutation tables, recurring on the second set. A mutation table is evidence only if it was produced by running the whole package.

### 3. British spelling (three occurrences)

- `cmd/atlas/atlas.go:57` — "materialised"
- `cmd/atlas/migrate_write_dir_scheme_test.go:26` — "materialising"
- commit `58209a52` message — "materialising a directory the operator did not name"

American English is required everywhere including commit messages: "materialized" / "materializing".

### 4. CI is red on three jobs — infrastructure, not this change

`gosec`, `unit-tests` and `Build versioned site` all failed on GitHub Actions plumbing, never reaching the code: `Unable to resolve action 'actions/checkout@v7' … 'actions/setup-go@v7' … 'golangci/golangci-lint-action@v9'`, and `Failed to resolve action download info. Error: Service Unavailable` / `Bad Gateway`. Nothing here is attributable to the diff, but the PR currently shows red and needs a re-run before merge.

---

**Verdict:** the blocker this PR was held on is genuinely closed, and I could not break the gate. Items 1 and 2 need fixing before merge — 1 because the body states a measured parity that does not hold and the fix silently moved two matching cells, 2 because three mutation tables misreport what they measured.

denisvmedia:
This is a substantial compatibility improvement.
One thing I'd keep explicitly visible, though, is the remaining Flyway identity gap.
From the measurements, apply, set, and lint still disagree with Atlas on version identity because Ptah exposes the internal ordering key rather than the original Flyway version token.
I don't think that blocks this PR, but I'd keep that issue highly visible since it affects user-facing compatibility rather than only internal implementation details.

denisvmedia:
Agreed, and it now has its own home: **#1206**. Nothing was fixed here — the point was to make the gap visible and re-measurable rather than leave it in a merged PR description.

I re-measured it on master at `49051c17` (this PR's merge commit) rather than trusting the numbers in the description, against the pinned community build `atlas community version v1.3.0`. Three corrections to the list came out of that.

**1. `apply`, `set` and `lint` all still diverge.** Fixture is `internal/atlasmigrateimport/testdata/ce-sums/flyway/no-separator` (`V1.sql`, `V2__ok.sql`), whose `atlas.sum` the community build produced:

| verb | community v1.3.0 | ptah-compat @`49051c17` |
| --- | --- | --- |
| `migrate apply` | exit 0, revisions `1`, `2` | exit 0, revisions `4611686018427469511`, `4611686018427510315` |
| `migrate set 1` | exit 0, `Current version is 1 (1 set)` | **exit 1**, `Error: migration with version "1" not found`, revision table never created |
| `migrate lint --latest 1` | `Analyzing changes from version 1 to 2` | `Analyzing changes from version 4611686018427469511 to 4611686018427510315` |

**2. `migrate status` diverges too**, which the table in this PR does not say. It is counted there as a matching cell, and on exit code and file counts it is — but on each binary's *own* database it prints a different identity:

```
$ atlas       migrate status ...  -->  Current Version: 2
$ ptah-compat migrate status ...  -->  Current Version: 4611686018427510315
```

So the affected list is `apply`, `set`, `lint`, `status`. `validate` matches (it names no version).

**3. The `set` half is worse than "exits 1 on one operand".** The two implementations accept **disjoint** version namespaces on the same directory — each rejects exactly what the other accepts:

```
$ ptah-compat migrate set 4611686018427469511 ...   exit=0, records it
$ atlas       migrate set 4611686018427469511 ...   exit=1, migration with version "..." not found
```

and a Flyway version with a dot cannot be spelled at all, which matters because dotted versions are the Flyway norm rather than an edge case:

```
$ atlas       migrate set 1.5 ...   exit=0   "+ 1 (a)"  "+ 1.5 (b)"
$ ptah-compat migrate set 1.5 ...   exit=1   --version "1.5" is not a valid migration version:
                                             strconv.ParseInt: parsing "1.5": invalid syntax
```

The sharpest one is a zero-padded token: the community build records the row **verbatim** as `01`, so its identity is demonstrably the token string, while ptah-compat has already normalized the operand to `1` by the time it reports not finding it.

Separating control, so this reads as "the Flyway projection diverges" rather than "converted directories diverge": on `ce-sums/golang-migrate/basic` both binaries record `1|init` and both accept `migrate set 1` at exit 0. golang-migrate carries a plain numeric prefix, so its token *is* an int64 and the projection is the identity function.

The data-safety edge you shipped in #1180 is confirmed still shut — `ptah-compat migrate apply` against a community-written revision table exits 1 with the named refusal and applies nothing. What survives is identity only, which is why #1206 is scoped to that and explicitly puts the ordering and linearity rules from #1098/#1132 and #1003/#1123 out of bounds.

#1206 carries the full per-verb table, the raw transcripts, the control, and a definition of done. One caveat recorded there rather than hidden: `migrate hash` is absent from the table because the sandbox in use refuses any command containing that token, so it was predicted unaffected (it names no version), not measured.

Issue #932, opened 2026-07-31
SQL schema files: unsupported object kinds


denisvmedia:
The SQL schema-file parser covers most object kinds but not all of them.

Found by a deep audit of every capability the feature matrix marked as partial.
Each item below was reproduced against a binary built from this repository.

Severity assessed as **incomplete-capability**.

### 1. Support materialized views, sequences, roles, grants and RLS in SQL schema files

The SQL schema-file parser used by --schema-file accepts CREATE VIEW, CREATE DOMAIN, composite CREATE TYPE, CREATE TYPE ... AS RANGE, CREATE TYPE ... AS ENUM, CREATE EXTENSION, CREATE FUNCTION and CREATE TRIGGER, but rejects CREATE MATERIALIZED VIEW, CREATE SEQUENCE, CREATE ROLE, GRANT, CREATE POLICY and ALTER TABLE ... ENABLE ROW LEVEL SECURITY. Those five object kinds are therefore expressible only through Go annotations, HCL or YAML, not through the Atlas-shaped SQL source.

Reproduction:

```text
mkdir -p /tmp/sqlin
for s in 'CREATE MATERIALIZED VIEW mv1 AS SELECT id FROM t1;' 'CREATE SEQUENCE s1 START WITH 1;' 'CREATE ROLE r_x;' 'GRANT SELECT ON TABLE t1 TO r_x;' 'CREATE POLICY p1 ON t1 FOR SELECT USING (true);' 'ALTER TABLE t1 ENABLE ROW LEVEL SECURITY;'; do
  printf 'CREATE TABLE t1 (id BIGINT PRIMARY KEY);\n%s\n' "$s" > /tmp/sqlin/one.sql
  ptah schema render --schema-file /tmp/sqlin/one.sql --dialect postgres 2>&1 | grep '^error:'
done

# observed:
# error: error parsing schema file: parse SQL schema file: unsupported CREATE target: MATERIALIZED at position 48
# error: error parsing schema file: parse SQL schema file: unsupported CREATE target: SEQUENCE at position 48
# error: error parsing schema file: parse SQL schema file: unsupported CREATE target: ROLE at position 48
# error: error parsing schema file: parse SQL schema file: unsupported SQL statement: GRANT at position 41
# error: error parsing schema file: parse SQL schema file: unsupported CREATE target: POLICY at position 48
# error: error parsing schema file: parse SQL schema file: unsupported ALTER operation: ENABLE at position 56
[...]

Issue #931, opened 2026-07-31
Schema-object rendering and diff coverage


denisvmedia:
Object kinds that are dropped, downgraded, or diffed incompletely on specific dialects.

Found by a deep audit of every capability the feature matrix marked as partial.
Each item below was reproduced against a binary built from this repository.

Severity assessed as **incomplete-capability**.

### 1. Recognize enum columns by declared enum type, not by an "enum_" name prefix

On MySQL, MariaDB, SQLite and SQL Server, Ptah models enums inline on the column. The rewrite that turns a column typed with a declared enum into ENUM(...) / TEXT+CHECK / NVARCHAR+CHECK is guarded by strings.HasPrefix(field.Type, "enum_"). An enum type whose name does not begin with enum_ (for example status_kind) is left as the raw type name in the CREATE TABLE, and because the same dialects skip standalone CREATE TYPE emission, the enum's values disappear entirely. The enum_ prefix requirement is not stated in the annotation JSON Schema (`ptah schema annotations`).

Reproduction:

```text
mkdir -p /tmp/en && cat > /tmp/en/m.go <<'EOF'
package models

//ptah:schema:enum name="status_kind" values="active,archived"
type _ struct{}

//ptah:schema:table name="users"
type User struct {
	//ptah:schema:field name="id" type="BIGINT" primary="true"
	ID int64
	//ptah:schema:field name="s" type="status_kind"
	S string
}
EOF
cd /tmp/en && ptah schema render --root-dir . --dialect mysql

# observed (exit 0):
#   CREATE TABLE `users` (
#     `id` BIGINT PRIMARY KEY,
#     `s` status_kind
#   );
# MySQL rejects the type `status_kind`. Renaming the enum to enum_status yields
#   `s` ENUM('active', 'archived')
# Same asymmetry on sqlite (TEXT CHECK ... only with the enum_ prefix) and sqlserver
# (NVARCHAR(255) CHECK ... only with the enum_ prefix).
```

Evidence: internal/convert/fromschema/fromschema.go:268 `if !strings.HasPrefix(field.Type, "enum_") { return field }`; :279 and :293-318 (applyInlineEnumModel); :320-328 (emitsStandaloneEnumDefinitions suppresses CREATE TYPE for these dialects)

### 2. Detect range-type subtype changes in schemadiff

[...]

PR #1177, opened 2026-08-06
Emit SQLite table-rebuild plans for unsupported ALTERs

### LINKED ISSUE #930: SQLite: emit table-rebuild plans for unsupported ALTERs
SQLite already rebuilds for column removal, but every other column or constraint change reports rebuild-required instead of generating the rebuild.

Found by a deep audit of every capability the feature matrix marked as partial.
Each item below was reproduced against a binary built from this repository.

Severity assessed as **blocks-parity**.

### 1. Emit SQLite table-rebuild plans instead of refusing the change

SQLite has no ALTER for column type, nullability, or constraint changes, so a planner must emit the create-new-table / copy-rows / drop-old / rename sequence. Ptah's SQLite planner has no rebuild path at all: it detects the situation and returns an unsupported-feature error. This blocks column modification, NOT NULL changes, constraint add/remove, enum-backed CHECK changes, adding a column that carries a UNIQUE constraint, and dropping a column combined with any other table change — on both `schema diff` and `schema apply`.

Reproduction:

```text
$ go build -o /tmp/pc ./cmd/ptah-compat
$ printf 'CREATE TABLE users (id INTEGER PRIMARY KEY, name TEXT NOT NULL);\n' > a.sql
$ printf 'CREATE TABLE users (id INTEGER PRIMARY KEY, name INTEGER NOT NULL);\n' > d.sql
$ /tmp/pc schema diff --from file://a.sql --to file://d.sql --dev-url sqlite://dev4.db
error: generate schema diff SQL: sqlite: modifying columns on table users requires a table rebuild plan
(exit 1)

$ /tmp/pc schema apply --url sqlite://rb.db --to file://a.sql --auto-approve   # seed
$ /tmp/pc schema apply --url sqlite://rb.db --to file://d.sql --auto-approve
error: generate schema apply SQL: sqlite: modifying columns on table users requires a table rebuild plan
(exit 1)

Other refused forms observed the same way:
  NOT NULL -> nullable: sqlite: modifying columns on table users requires a table rebuild plan
  add column with UNIQUE: sqlite: adding column email to table users requires a table rebuild plan
```

Evidence: internal/planner/dialects/sqlite/sqlite.go:87-105 (rejectUnsupportedTableChanges) and internal/planner/dialects/sqlite/sqlite.go:489 (addColumn fallback); core/renderer/internal/dialects/sqlite/sqlite.go:120

### 2. Generate a SQLite table rebuild for column modifications

SQLite already emits a complete rebuild plan for column removal (__ptah_rebuild_<table> CREATE / INSERT ... SELECT / DROP / RENAME), but any column modification — type, nullability, default, or added uniqueness — aborts with an unsupported-feature error instead of reusing that machinery.

Reproduction:

```text
$ printf 'CREATE TABLE t (id INTEGER PRIMARY KEY, a TEXT NOT NULL);\n' > a.sql
$ printf 'CREATE TABLE t (id INTEGER PRIMARY KEY, a INTEGER NOT NULL);\n' > b.sql
$ ptah schema diff --from a.sql --to b.sql --dev-url sqlite://$PWD/dev.db
error: generate schema diff SQL: sqlite: modifying columns on table t requires a table rebuild plan

(same error for NOT NULL removal, DEFAULT addition and UNIQUE addition; whereas dropping column b from the same table emits the full rebuild:)
$ ptah schema diff --from c.sql --to a.sql --dev-url sqlite://$PWD/dev.db
-- SQLite table rebuild to remove unsupported columns from t
CREATE TABLE "__ptah_rebuild_t" (...); INSERT INTO ... ; DROP TABLE "t"; ALTER TABLE "__ptah_rebuild_t" RENAME TO "t";
```

Evidence: internal/planner/dialects/sqlite/sqlite.go:91 ("modifying columns on table %s requires a table rebuild plan") and :96 ("changing constraints on table %s requires a table rebuild plan")

## Notes

The matrix row for each of these now states the specific limitation rather than
the word "partial". Closing an item here should update the corresponding row on
`docs/site/src/content/docs/atlas/feature-matrix.md`.

denisvmedia:
Closes #930.

The SQLite planner rebuilt a table to drop a column, but every other shape that
ALTER TABLE cannot express returned an unsupported-feature error instead of
reusing that machinery. This generalizes the rebuild from "drop these columns"
to "reshape to the desired definition".

Measured on this branch against `origin/master` d1e2dc64, with `ptah-compat` and
`ptah` built from the branch, cross-checked against the pinned Atlas community
binary (`atlas community version v1.3.0`) at
`/Users/buster/Work/denis/ptah-atlas-conformance/bin/atlas`.

## Reproduction, before and after

Fixtures, all SQLite, all diffs run as
`ptah-compat schema diff --from file://<a> --to file://<b> --dev-url sqlite://dev.db`:

```sql
-- base.sql
CREATE TABLE users (id INTEGER PRIMARY KEY, name TEXT NOT NULL);
-- base3.sql
CREATE TABLE users (id INTEGER PRIMARY KEY, name TEXT NOT NULL, email TEXT);
-- c_type.sql      name INTEGER NOT NULL
-- c_nullable.sql  name TEXT
-- c_default.sql   name TEXT NOT NULL DEFAULT 'x'
-- c_colcheck.sql  name TEXT NOT NULL CHECK (length(name) > 2)
-- c_uq.sql        base3 + CONSTRAINT uq_users_email UNIQUE (email)
-- c_dropadd.sql   base3 with email dropped and nickname TEXT added
```

| # | Fixture | Before | After | Atlas CE v1.3.0 |
|---|---|---|---|---|
| 1 | `base → c_type` | exit 1 `sqlite: modifying columns on table users requires a table rebuild plan` | exit 0, rebuild | exit 0, rebuild |
| 2 | `base → c_nullable` | exit 1, same message | exit 0, rebuild | exit 0, rebuild |
| 3 | `base → c_default` | exit 1, same message | exit 0, rebuild with `IFNULL` | exit 0, rebuild with `IFNULL` |
| 4 | `base → c_colcheck` | exit 1 `sqlite: changing constraints on existing tables requires a table rebuild plan` | exit 0, rebuild | exit 0, rebuild |
| 5 | `base3 → c_dropadd` | exit 1 `sqlite: combining dropped columns with other table changes on users requires a manual rebuild plan` | exit 0, rebuild | exit 0, rebuild |
| 6 | enum shrink via Go entities | exit 2 `sqlite: changing constraints on existing tables requires a table rebuild plan` | exit 0, rebuild | no CE equivalent (Ptah annotations) |
| — | `base3 → c_uq` (extra) | exit 1, constraints message | exit 0, rebuild | exit 0, `CREATE UNIQUE INDEX` |

Controls, all unchanged by this branch: `base → base3` (nullable column
addition) exits 0 with `ALTER TABLE "users" ADD COLUMN "email" TEXT;` on both
binaries; `base3 → base` (column drop only) exits 0 with a rebuild on both;
`base → base` exits 0 with "Schemas are synced" on both.

The DEFAULT case, before and after:

```
$ pc schema diff --from file://base.sql --to file://c_default.sql --dev-url sqlite://dev.db
error: generate schema diff SQL: sqlite: modifying columns on table users requires a table rebuild plan   (exit 1)

$ pc schema diff --from file://base.sql --to file://c_default.sql --dev-url sqlite://dev.db
-- SQLite table rebuild for changes ALTER TABLE cannot express on users
CREATE TABLE "__ptah_rebuild_users" (
  "id" INTEGER PRIMARY KEY,
  "name" TEXT NOT NULL DEFAULT 'x'
);
INSERT INTO "__ptah_rebuild_users" ("id", "name") SELECT "id", IFNULL("name", 'x') AS "name" FROM "users";
DROP TABLE "users";
ALTER TABLE "__ptah_rebuild_users" RENAME TO "users";
(exit 0)
```

The enum case, `ptah migrations plan --root-dir <entities> --db-url sqlite://enum.db`,
shrinking `enum="draft,published,archived"` to `enum="draft,published"`:

```
before: error: error generating migration plan: sqlite: changing constraints on existing tables requires a table rebuild plan   (exit 2)
after:  -- SQLite table rebuild for changes ALTER TABLE cannot express on articles
        CREATE TABLE "__ptah_rebuild_articles" (
          "id" INTEGER PRIMARY KEY,
          "status" TEXT NOT NULL CHECK (status IN ('draft', 'published'))
        );
        INSERT INTO "__ptah_rebuild_articles" ("id", "status") SELECT "id", "status" FROM "articles";
        DROP TABLE "articles";
        ALTER TABLE "__ptah_rebuild_articles" RENAME TO "articles";
        (exit 0)
```

## Executed end to end, not only planned

Every case above was seeded into a real SQLite database with two rows, applied
with `ptah-compat schema apply --url sqlite://<db> --to file://<to> --dev-url
sqlite://<dev> --auto-approve`, and inspected afterwards. All six exit 0, all
six keep `count(*) = 2`, and `SELECT name FROM sqlite_schema WHERE name LIKE
'__ptah_rebuild_%'` returns zero rows in every case.

The DEFAULT case is the discriminating one: seeded `(1,'alpha'),(2,NULL)` on a
nullable column, applying `NOT NULL DEFAULT 'x'` leaves `1='alpha' / 2='x'`.
The enum case was applied the same way; afterwards
`INSERT INTO articles (id,status) VALUES (3,'archived')` fails with
`CHECK constraint failed: status IN ('draft', 'published')`, which is the
positive control that the rebuilt constraint is live rather than merely printed.

Every claim below was produced by actually running the mutant, not predicted.

- **`TestPlannerRebuildsTableForChangesAlterTableCannotExpress`** — seven rows,
  one per diff shape. Narrowing `tableDiffNeedsRebuild` back to
  `len(table.ColumnsRemoved) > 0` reddens exactly four rows (`column_type_change`,
  `column_nullability_change`, `not_null_default_addition_backfills_the_copy`,
  `table-level_constraint_change`) with the old unsupported-feature errors.
  Separately, dropping the constraint-attributed tables from `planTableRebuilds`
  reddens exactly the other two constraint rows
  (`schema-level_constraint_change_on_an_existing_table`,
  `enum-backed_check_constraint_change`). The two mutants redden disjoint sets,
  so each of the three former call sites is pinned by its own row rather than by
  one test standing in for all.
- **The backfill row** — making `rebuildSelectExpression` always return the bare
  quoted column reddens only that row, and the end-to-end apply then fails with
  `Error: apply schema changes: ... NOT NULL constraint failed: __ptah_rebuild_users.name (1299)`.
- **`TestPlannerRebuildRefusesAddedNotNullColumnWithoutDefault`** — removing the
  guard prints `got non-nil value` for the nil-nodes assertion, i.e. a plan is
  emitted where it must not be. The inverse mutant (refuse every added column in
  a rebuild) reddens `dropped_column_combined_with_an_added_column`, so the
  guard's condition discriminates rather than firing unconditionally.
- **`TestPlannerRebuildEmitsIndexesAndTriggersOnce`** — removing the rebuild skip
  in `addIndexes` prints `got: int(2) / want: int(1)` for
  `CREATE INDEX IF NOT EXISTS "idx_users_name"`.
- **`TestPlannerEmitsTriggerChangesWithoutRebuild`** — this is the
  non-interference control for the trigger skip, and it is new because the
  inverse mutant found nothing else in the repository catching it: inverting the
  skip in `addTriggers`/`modifyTriggers` so that only rebuilt tables emit turned
  the **entire** `go test ./...` red on one test only, my own duplicate-count
  test. With the control added, the inverse mutant reddens it with
  `strings.Count(sql, "TRIGGER \"trg_users_update\"") == 0`.
- **`TestPlannerAddsColumnsAndIndexes` and the six `TestPlanner_IndexRefs_*`
  tests** are the non-interference control for the index skip: inverting that
  guard (`!rebuilds.contains`) reddens all seven, proving the skip is scoped to
  rebuilt tables and does not suppress ordinary index emission.
- **`TestPlannerRejectsAddColumnShapesThatNeedRebuild`** (unchanged) is the
  non-interference control for excluding rebuilt tables from
  `validateAddedColumns`: inverting that exclusion reddens it, because pure
  column additions on a non-rebuilt table must still be validated.
- **`TestPlannerRejectsUnqualifiedExistingTableConstraintChanges`** (unchanged)
  still pins the one surviving `changing constraints on existing tables`
  refusal, for a constraint the diff cannot attribute to a table.

## Where I chose the ceiling over matching

Two shapes exit 1 here where the pinned community binary exits 0. Parity rule
(a) is about never exiting 0 where the pinned binary exits 1, so neither is a
rule-(a) violation; both are cases where I took rule (b), because the CE plan
does not execute:

```
$ sqlite3 seeded.db 'ALTER TABLE `users` ADD COLUMN `email` text NOT NULL;'
Error: stepping, Cannot add a NOT NULL column with default value NULL   (exit 1)

$ sqlite3 seeded.db 'CREATE TABLE `new_users` (`id` integer NULL, `name` integer NOT NULL, `email` text NOT NULL, PRIMARY KEY (`id`));
                     INSERT INTO `new_users` (`id`, `name`) SELECT `id`, `name` FROM `users`;'
Error: stepping, NOT NULL constraint failed: new_users.email (19)   (exit 19)
```

Both statement lists are verbatim CE output for `base → users(id, name, email
NOT NULL)` and `base → users(id, name INTEGER, email NOT NULL)`. The control
that these are genuine and not a fixture artifact: the same rebuild against an
**empty** table exits 0, which is why the plan looks fine until it meets data.
Ptah refuses at plan time instead, with
`sqlite: adding column email to table users requires a table rebuild plan` and
`sqlite: rebuilding table users cannot add NOT NULL column email without a default`.

## What I deliberately left open

- **Adding a column that carries UNIQUE.** `base → users(id, name, email TEXT
  UNIQUE)` still exits 1 with `sqlite: adding column email to table users
  requires a table rebuild plan`; CE exits 0 with `ALTER TABLE ... ADD COLUMN`
  plus `CREATE UNIQUE INDEX`. That is the index path, not the rebuild path — a
  different function (`validateAddedColumn`) and a different strategy, and
  rebuilding for it would emit a needlessly destructive plan for a case that
  needs an index. The framing comment on #930 scoped this as W3.
- **Rebuilding a table with inbound foreign keys.** Still refused at
  `validateRebuildTablePreconditions` with `sqlite: rebuilding table users with
  inbound foreign keys requires a manual rebuild plan`, and Ptah still emits no
  `PRAGMA foreign_keys` bracket. The framing comment scoped this as W2. It is
  independent of the six shapes here — none of them involves a foreign key — and
  the PRAGMA is a no-op inside a transaction, so adding it without settling how
  the rebuild is executed would be decoration rather than a fix.
- **Table-level `CHECK` in a `.sql` schema file is invisible to the diff.** The
  issue's `base → c_check` fixture, where `c_check.sql` carries
  `CONSTRAINT name_len CHECK (length(name) > 2)` as a table constraint, exits 0
  with "Schemas are synced" both before and after this branch — the constraint
  never reaches the planner. `ptah-compat schema inspect --url file://c_check.sql`
  prints the table without the CHECK, and CE emits a full rebuild for the same
  fixture. This is a SQL-schema-file source gap, not a planner gap, so the CHECK
  row is covered here by the equivalent column-level fixture
  (`name TEXT NOT NULL CHECK (length(name) > 2)`), which does reach the planner
  and now rebuilds. Adjacent to #932.
- **A `.yaml` schema file with an `ENUM` column does not self-diff clean.**
  `ptah schema diff --from file://y.yaml --to file://y.yaml --dev-url sqlite://dev.db`
  reports changes — my negative control caught this, which is why the enum case
  above is measured through the Go-entity route instead. The spurious drift is
  `id nullable "true -> false"` and `status type "enum_article_status -> text"`,
  both from `schemafile.ToDBSchema` not rendering the desired side for the
  dialect. Adjacent to #931/#932; untouched here.
- **Feature-matrix rows stay 🟡, not ✅.** The issue's completion criteria asked
  for ✅ on lines 151 and 259. Both notes are rewritten to state what the rebuild
  now covers, but the symbols stay 🟡 because the two cases above genuinely
  survive. Claiming ✅ would cost a later measurement pass to unwind.

## Gates

`go build ./...`, `go vet ./...`, `go test ./... -count=1`, `go tool qtlint ./...`,
`golangci-lint run ./...` (0 issues), `scripts/check-test-style.sh`
(OK, 175 conditional groups), `scripts/check-public-api.sh`,
`scripts/check-public-api-snapshot.sh` — all green.

`docs/` is touched, so all 14 `check:*` scripts in `docs/site/package.json` ran,
all exit 0. `check:responsive` needs `npm run build` first — it exits 1 with
`dist not found` otherwise, so the site was built and the check re-run; it then
reports `OK (60 routes x 2 viewports, cells within 8 lines)` rather than
skipping.

### DIFF
--- a/docs/site/src/content/docs/atlas/feature-matrix.md
+++ b/docs/site/src/content/docs/atlas/feature-matrix.md
@@ -148,7 +148,7 @@ seven of them as open capabilities regardless.
 | Local pre-approved plan files | ✅ | ❌ | ✅ | `schema plan` writes Atlas `.plan.hcl` by default (`.json` keeps the native plan); `apply --plan` reads both, Atlas-authored included, verified by replay against `--to` plus an end-state check. |
 | schema apply against a live database | ✅ | ✅ | ✅ | Diffs `--url` against the `--to` desired state, prints the SQL plan, applies after confirmation. Verified end to end on SQLite. |
 | schema clean | 🟡 | ✅ | ✅ | `--include`/`--exclude` narrow the cleanup, and a narrowed run executes only what it printed. Plan and `--dry-run` list every object kind destroyed, except the revision table outside SQLite. |
-| schema diff between two schema states | 🟡 | ✅ | ✅ | SQLite refuses any change needing a table rebuild: column modify, NOT NULL change, constraint add/remove, enum CHECK change. Same on apply. |
+| schema diff between two schema states | 🟡 | ✅ | ✅ | SQLite rebuilds the table for column modify, NOT NULL, default, constraint and enum CHECK changes. Adding a column that carries UNIQUE still refuses. |
 | schema fmt (HCL canonical layout) | ✅ | ✅ | ✅ | Formats .hcl paths recursively and prints only changed files. Native `ptah schema fmt --check` adds a no-write CI gate. |
 | schema inspect to HCL, SQL, or JSON | ✅ | ✅ | ✅ | Default HCL; `--format` sql\|json\|template. Native twin `ptah schema inspect` adds `--out-dir` and `--split` file export. |
 | Schema-qualified exclude globs for enums and functions | 🟡 | 🟡 | ❔ | Tables, views and extensions match schema-qualified exclude globs; enum and function filters compare the bare name only. The community binary matches `app.mood`; it reports no functions. |
@@ -257,7 +257,7 @@ seven of them as open capabilities regardless.
 | Roles, grants, and row-level security | 🟡 | ❌ | ✅ | PostgreSQL only in `schema render`; `schema apply` emits CREATE ROLE on YugabyteDB. SQL files read ROLE, GRANT, POLICY, ENABLE RLS. Atlas prices this as Pro; CE no-ops a `role` block silently. |
 | Spanner PostgreSQL interface (spanner) | 🟡 | ❌ | ✅ | Enums and FKs render as skip comments, SERIAL hard-errors, no dedicated driver (uses the PostgreSQL pgx path), and no live container or live test exists. |
 | SQL Server and Azure SQL (sqlserver, mssql, tsql) | 🟡 | ❌ | ✅ | The mssql and tsql aliases silently drop views and triggers that sqlserver emits; render's default dialect list and `--dialect` help omit SQL Server; no sequences, RLS, roles/grants or matviews. |
-| SQLite (sqlite, sqlite3) | 🟡 | ✅ | ✅ | Column drops do emit a full rebuild; type, nullability, default and uniqueness changes fail with "modifying columns ... requires a table rebuild plan". PG-only objects error one at a time. |
+| SQLite (sqlite, sqlite3) | 🟡 | ✅ | ✅ | Column drops, type, nullability, default and constraint changes all emit a full rebuild with backfill. Rebuilding a table with inbound FKs still refuses. PG-only objects error one at a time. |
 | Standalone sequences | 🟡 | ❌ | ✅ | PostgreSQL only in `schema render`; `schema apply` emits it on YugabyteDB too. SQL schema files read CREATE SEQUENCE. |
 | TiDB and LibSQL | ❌ | ✅ | ✅ | Both names fail dialect normalization: "unsupported database dialect: tidb" / "...: libsql". No renderer, planner or driver entry. |
 | Triggers | 🟡 | ❌ | ✅ | `--dialect mssql`/`tsql` drop them while `sqlserver` renders them; MySQL forces FOR EACH ROW; SQL Server rewrites BEFORE to AFTER silently. |
--- a/docs/site/src/content/docs/atlas/feature-matrix.md
+++ b/docs/site/src/content/docs/atlas/feature-matrix.md
@@ -148,7 +148,7 @@ seven of them as open capabilities regardless.
 | Local pre-approved plan files | ✅ | ❌ | ✅ | `schema plan` writes Atlas `.plan.hcl` by default (`.json` keeps the native plan); `apply --plan` reads both, Atlas-authored included, verified by replay against `--to` plus an end-state check. |
 | schema apply against a live database | ✅ | ✅ | ✅ | Diffs `--url` against the `--to` desired state, prints the SQL plan, applies after confirmation. Verified end to end on SQLite. |
 | schema clean | 🟡 | ✅ | ✅ | `--include`/`--exclude` narrow the cleanup, and a narrowed run executes only what it printed. Plan and `--dry-run` list every object kind destroyed, except the revision table outside SQLite. |
-| schema diff between two schema states | 🟡 | ✅ | ✅ | SQLite rebuilds the table for column modify, NOT NULL, default, constraint and enum CHECK changes. Adding a column that carries UNIQUE still refuses. |
+| schema diff between two schema states | 🟡 | ✅ | ✅ | SQLite refuses any change needing a table rebuild: column modify, NOT NULL change, constraint add/remove, enum CHECK change. Same on apply. |
 | schema fmt (HCL canonical layout) | ✅ | ✅ | ✅ | Formats .hcl paths recursively and prints only changed files. Native `ptah schema fmt --check` adds a no-write CI gate. |
 | schema inspect to HCL, SQL, or JSON | ✅ | ✅ | ✅ | Default HCL; `--format` sql\|json\|template. Native twin `ptah schema inspect` adds `--out-dir` and `--split` file export. |
 | Schema-qualified exclude globs for enums and functions | 🟡 | 🟡 | ❔ | Tables, views and extensions match schema-qualified exclude globs; enum and function filters compare the bare name only. The community binary matches `app.mood`; it reports no functions. |
@@ -257,7 +257,7 @@ seven of them as open capabilities regardless.
 | Roles, grants, and row-level security | 🟡 | ❌ | ✅ | PostgreSQL only in `schema render`; `schema apply` emits CREATE ROLE on YugabyteDB. SQL files read ROLE, GRANT, POLICY, ENABLE RLS. Atlas prices this as Pro; CE no-ops a `role` block silently. |
 | Spanner PostgreSQL interface (spanner) | 🟡 | ❌ | ✅ | Enums and FKs render as skip comments, SERIAL hard-errors, no dedicated driver (uses the PostgreSQL pgx path), and no live container or live test exists. |
 | SQL Server and Azure SQL (sqlserver, mssql, tsql) | 🟡 | ❌ | ✅ | The mssql and tsql aliases silently drop views and triggers that sqlserver emits; render's default dialect list and `--dialect` help omit SQL Server; no sequences, RLS, roles/grants or matviews. |
-| SQLite (sqlite, sqlite3) | 🟡 | ✅ | ✅ | Column drops, type, nullability, default and constraint changes all emit a full rebuild with backfill. Rebuilding a table with inbound FKs still refuses. PG-only objects error one at a time. |
+| SQLite (sqlite, sqlite3) | 🟡 | ✅ | ✅ | Column drops do emit a full rebuild; type, nullability, default and uniqueness changes fail with "modifying columns ... requires a table rebuild plan". PG-only objects error one at a time. |
 | Standalone sequences | 🟡 | ❌ | ✅ | PostgreSQL only in `schema render`; `schema apply` emits it on YugabyteDB too. SQL schema files read CREATE SEQUENCE. |
 | TiDB and LibSQL | ❌ | ✅ | ✅ | Both names fail dialect normalization: "unsupported database dialect: tidb" / "...: libsql". No renderer, planner or driver entry. |
 | Triggers | 🟡 | ❌ | ✅ | `--dialect mssql`/`tsql` drop them while `sqlserver` renders them; MySQL forces FOR EACH ROW; SQL Server rewrites BEFORE to AFTER silently. |
--- a/docs/site/src/content/docs/databases/sqlite.md
+++ b/docs/site/src/content/docs/databases/sqlite.md
@@ -42,9 +42,15 @@ The SQLite renderer and planner support:
 - `CREATE INDEX`, including unique and partial indexes, and
   `DROP INDEX IF EXISTS` / `DROP TABLE IF EXISTS`.
 - `ALTER TABLE ... ADD COLUMN`, `RENAME COLUMN`, and `RENAME TO`.
-- Simple column drops through a generated table-rebuild plan: create a
-  rebuilt table, copy retained columns, swap it in, and recreate retained
-  indexes and triggers when their metadata round-trips safely.
+- Column and constraint changes ALTER TABLE cannot express, through a
+  generated table-rebuild plan: create a rebuilt table in its desired shape,
+  copy the retained columns, swap it in, and recreate the desired indexes and
+  triggers when their metadata round-trips safely. One rebuild covers column
+  drops, column type, nullability, default and generated-expression changes,
+  added and removed table constraints (including enum-backed `CHECK`
+  constraints), and a column drop combined with a column addition. A column the
+  desired schema makes `NOT NULL` with a default is backfilled in the copy with
+  `IFNULL(<column>, <default>)`, so rows already holding `NULL` survive.
 - Views without `WITH CHECK OPTION`, and row-level triggers (SQLite has no
   statement-level triggers).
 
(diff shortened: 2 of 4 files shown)


denisvmedia:
Marked draft. Adversarial review found **silent data corruption** on a shape this branch claims to cover. Everything else about it is sound — every mutant claim in the body re-ran exactly — which is why this is one fix rather than a rewrite.

## The defect

When a single diff both **adds a column** and **changes a table-level constraint** on the same existing table, the added column is included in the rebuild's `INSERT ... SELECT`. SQLite resolves an unknown double-quoted identifier as a **string literal**, so every row gets the literal column name instead of NULL — and the command exits 0 saying it succeeded.

```
base.sql       CREATE TABLE users (id INTEGER PRIMARY KEY, name TEXT NOT NULL);
c_corrupt.sql  CREATE TABLE users (id INTEGER PRIMARY KEY, name TEXT NOT NULL, note TEXT,
                                   CONSTRAINT uq_users_name UNIQUE (name));
seed (1,'alpha'), (2,'beta')  ->  schema apply --auto-approve

emitted: INSERT INTO "__ptah_rebuild_users" ("id","name","note")
         SELECT "id","name","note" FROM "users";
exit 0, "Schema apply completed successfully."

data after:  1|alpha|note    2|beta|note      (typeof = text; expected NULL)
```

Two added columns corrupt both (`a='a'`, `b='b'`). If the added column carries the UNIQUE itself it errors instead — `UNIQUE constraint failed: __ptah_rebuild_users.email (2067)` — so the failure mode is inconsistent, which is worse than uniformly loud. Constraint **removal** plus a column addition emits the same wrong SELECT.

## Root cause

`internal/planner/dialects/sqlite/sqlite.go`, in `planTableRebuilds`:

```go
for _, tableName := range constrained {
    add(tableName, nil)          // addedColumns discarded
}
```

The table is also in `diff.TablesModified` carrying `ColumnsAdded`, but `tableDiffNeedsRebuild` returned false because `TableDiff.ConstraintsAdded` is empty — the constraint arrives at schema level through `ConstraintsAddedWithTables`. So the table enters the rebuild set with `addedColumns == nil` and `rebuildCopiedColumns` never excludes the new column.

The reviewer verified the fix: carry the `TableDiff`'s `ColumnsAdded` through that path instead of passing `nil`.

## What to add with it

A fixture for exactly this intersection — add-column **and** table-level constraint change, on a seeded table, asserting the copied values are NULL rather than the column name. The current suite covers each half separately, which is why the branch is green: neither half alone reaches the bad SELECT.

Worth stating in the rewritten body that `schema apply` exiting 0 on a corrupting rebuild is the sharp edge here. A rebuild that fails loudly is recoverable; one that reports success and writes the column name into every row is not visible until someone reads the data.

denisvmedia:
The silent data corruption the review found is fixed, and the fixture for it is in.

**The defect.** A diff that both adds a column and changes a table-level constraint on the same table planned a rebuild whose `INSERT ... SELECT` read the new column out of the OLD table. The table reached the constraint loop rather than the `TablesModified` loop — its constraint arrived at schema level in `ConstraintsAddedWithTables`, so `tableDiffNeedsRebuild` answered false and `add(tableName, nil)` discarded its `ColumnsAdded`.

SQLite resolves an unknown double-quoted identifier as a **string literal** instead of refusing it, so every row received the text of the column's own name, and `schema apply` exited 0 reporting success.

**Measured**, seeded `users` (1/alpha, 2/beta), desired schema adds `note TEXT` plus `CONSTRAINT uq_users_name UNIQUE (name)`:

```
before  1|alpha|text|note      2|beta|text|note
after   1|alpha|null|<NULL>    2|beta|null|<NULL>
```

**Mutant.** Restoring `add(tableName, nil)` reddens the new test on its first assertion, with the corrupting statement quoted back:

```
container: INSERT INTO "__ptah_rebuild_users" ("id", "name", "note") SELECT "id", "name", "note" FROM "users";
want:      INSERT INTO "__ptah_rebuild_users" ("id", "name") SELECT "id", "name" FROM "users";
```

Each half of the shape was already covered separately — a column added alone, a constraint changed alone — and neither reaches the bad SELECT. That is why the branch was green.

**One branch removed rather than tested.** The first draft of the fix guarded the second loop with `if rebuilds.contains(tableName) { continue }` against listing a column twice. Disabling that guard reddened nothing, and the reason is that it cannot: `addedColumns` is only ever turned into a `map[string]bool` exclusion set by `rebuildCopiedColumns`, so a repeated name cannot change the plan. The guard was unobservable and its comment claimed a consequence that does not exist, so it is gone and the reasoning is recorded in its place.

Gates on the rebased branch: `go build`, `go vet`, `go test ./... -count=1`, `qtlint`, `golangci-lint` (0 issues), `check-test-style.sh`, `check-public-api.sh`, `check-public-api-snapshot.sh` — all pass. The snapshot gate was confirmed to actually run by adding a throwaway exported symbol and watching it exit 1.

PR #1180, opened 2026-08-06
Refuse a converted Flyway apply over another implementation's revisions

### LINKED ISSUE #1100: migrate apply re-executes a converted directory another implementation already applied
Measured at 92213b39 while building #1002, on the pinned community binary v1.3.0 by absolute path.

A migration directory read through `?format=flyway` is converted to Atlas single-file migrations before it is executed. The two implementations assign **different versions** to the same converted files, so a revision table written by one is not recognized by the other. In one direction that is merely confusing; in the other it re-runs migrations that have already been applied.

## Reproduction

```
d/V1__a.sql   CREATE TABLE t1 (id int);
d/V2__b.sql   CREATE TABLE t2 (id int);
$ atlas migrate hash --dir 'file://d?format=flyway'
```

Apply with each implementation into its own database, then read each database with each implementation:

| | recorded versions | read by the community binary | read by ptah-compat |
|---|---|---|---|
| applied by the community binary | `1`, `2` | `Current Version: 2`, 0 pending | `Current Version: 2`, **2 applied and 2 pending** |
| applied by ptah-compat | `4611686018427469511`, `4611686018427510315` | `Current Version: 4611686018427510315`, 0 pending | up to date, 0 pending |

The dangerous cell is the asymmetry underneath it:

```
$ ptah-compat migrate apply --dir 'file://d?format=flyway' --url sqlite://ce.db
Migrating to version 4611686018427510315 from 2 pending migrations.
Error: error applying migrations: failed to apply migration 4611686018427469511: failed to
execute migration SQL: sqlite: SQL execution failed: SQL logic error: table t1 already exists (1)
SQL: CREATE TABLE t1 (id int)
exit=1
```

against, on the same database and directory:

```
$ atlas migrate apply --dir 'file://d?format=flyway' --url sqlite://ce.db
No migration files to execute
exit=0
```

The mirror direction is benign — the community binary reads a ptah-written revision table as already ahead and executes nothing — and the control is clean: applying twice with the same implementation prints `No migration files to execute` both times. So the fixture separates the defect from its absence in both directions.

`CREATE TABLE` failing loudly is the lucky case. A migration whose statements are individually replayable would re-execute silently.

## Where it comes from

This is not an oversight; it is the visible edge of a documented projection. `internal/atlasmigrateimport/importer.go` explains it directly: the community binary identifies a Flyway migration by an opaque version **string**, while Ptah's migrator identifies a migration by an **int64**, so converting means projecting one space onto the other. The band-and-slot encoding above `flywayComponentSlot` is what preserves execution order across baselines, repeatables and tokens that score identically — and the file says plainly that this is "the whole reason the two cannot be identical".

What is not written down anywhere is the consequence: the projection is not just a different spelling of the same version, it makes the revision tables mutually unreadable, and in one direction that means re-execution.

## Worth deciding, not assumed

Three shapes, in increasing cost:

1. **Refuse rather than re-run.** On a converted directory, notice that the revision rows carry versions no converted file produces and stop, instead of treating every file as pending. Cheapest, and it removes the data-safety edge without touching the projection.
2. **Record the source token alongside the version**, and match on it when it is present. Keeps the int64 ordering key, adds an identity that survives the projection.
3. **Carry string versions through the migrator.** Removes the divergence and is by far the largest change; it touches the migrator's identity model rather than the importer.

Option 1 is the one that fits inside the parity rule as it stands — it is strictly stricter, never exiting 0 where the community binary exits 1 — but it leaves an operator who switches tools with a directory neither can move forward. That trade is the actual decision here.

Whatever is chosen, the same question applies to the other converted formats (golang-migrate, goose, liquibase, dbmate); only Flyway was measured here.

## Not this issue

`migrate status` on a mismatched revision table prints counts over three different sets (`Total` counts files, `Applied` counts rows, `Pending` counts unmatched files), so they can sum above the total. Measured, the community binary does the same thing on the equivalent input — `Executed Files: 1` and `Pending Files: 1` on a one-file directory — so it is not a divergence and is deliberately left alone.

Refs #1002, #982.

denisvmedia:
Closes #1100.

The issue offered three shapes and asked for an argument. This is **shape 1, refuse rather than re-run**, and the argument is below along with the two cells the issue did not measure.

## The reproduction, before

Re-measured on `f22107bd` (the commit this work started from; the branch is now rebased onto `028ac05d`, whose delta is unrelated compat/lint work) against the pinned community build `v1.3.0` by absolute path, sqlite, every cell in its own database so no cell can be polluted by the one before it.

```
d/V1__a.sql   CREATE TABLE t1 (id int);
d/V2__b.sql   CREATE TABLE t2 (id int);
atlas migrate hash --dir 'file://d?format=flyway'
```

| | recorded versions |
|---|---|
| applied by the community binary | `1`, `2` |
| applied by `ptah-compat` | `4611686018427469511`, `4611686018427510315` |

```
$ ptah-compat migrate apply --dir 'file://d?format=flyway' --url sqlite://ce.db
Migrating to version 4611686018427510315 from 2 pending migrations.
Error: error applying migrations: failed to apply migration 4611686018427469511: failed to
execute migration SQL: sqlite: SQL execution failed: SQL logic error: table t1 already exists (1)
exit=1

$ atlas migrate apply --dir 'file://d?format=flyway' --url sqlite://ce.db
No migration files to execute
exit=0
```

**The loud half is the lucky half.** With bodies that replay, the same cell exits 0 and duplicates data:

```
d/V1__t.sql   CREATE TABLE IF NOT EXISTS seed (id int);
d/V2__s.sql   INSERT INTO seed (id) VALUES (7);

community binary applies -> rows in seed: 1
$ ptah-compat migrate apply --dir 'file://d?format=flyway' --url sqlite://x.db
Migrating to version 4611686018427510315 from 2 pending migrations.
Migration complete. Current version: 4611686018427510315
exit=0
rows in seed: 2
```

Controls, all clean: the community binary re-reading its own database, `ptah-compat` re-reading its own, and the community binary reading a `ptah-compat`-written table all print `No migration files to execute` at exit 0 with the seed row still at 1. A native (non-`?format=`) Atlas directory is unaffected in both directions. `--dry-run` printed `Would have applied 2 migrations.` at exit 0 — a preview of work the database had already done.

## Two cells the issue did not name

**A baseline over a foreign revision table.** Nothing re-runs here; a squashed schema executes on top of the history it squashes. `V1`,`V2` applied by the community binary, then `B2__base.sql` added:

| | exit | tables after |
|---|---|---|
| pinned community binary | 0, `No migration files to execute` | `p`, `q` |
| `ptah-compat` on base | 0, `Migrating to version 122412 from 1 pending migrations.` | `p`, `q`, **`base`** |

The existing #1003 baseline refusal does not reach this: it looks for the CONVERTED version of a same-token migration in the applied set, and on a foreign table that version is not there. The token comparison is what sees it.

**The other four converted layouts.** The issue asked; measured, only Flyway re-executes.

| format | community binary records | `ptah-compat` records | `ptah-compat` on the community database |
|---|---|---|---|
| flyway | `1`, `2` | `4611686018427469511`, … | **re-executes** |
| goose | `00001`, `00002` | `1`, `2` | exit 0, nothing to execute |
| golang-migrate | `1`, `2` | `1`, `2` | exit 1, `checksum mismatch` |
| dbmate | `20240101000001`, … | identical | exit 1, `checksum mismatch` |
| liquibase | `1`, `2` | `1`, `2` | exit 1, `checksum mismatch` |

The three `checksum mismatch` rows are a **second, independent divergence** — the two implementations record the digest in different encodings (`stored 9mRxDpig…=`, `current ca5446be…`). It is a refusal, not a re-execution, so it is out of scope here and named in "left open" below.

## The reproduction, after

```
$ ptah-compat migrate apply --dir 'file://d?format=flyway' --url sqlite://ce.db
Error: this database records converted Flyway migrations under their SOURCE version token, which
is how another Atlas implementation identifies them; Ptah's migrator identifies a migration by the
int64 that token projects to, so 2 already-applied migration(s) read as pending here and would run
a second time. Nothing has been applied

recorded version -> version this build uses:
  1                    -> 4611686018427469511  V1__a.sql
  2                    -> 4611686018427510315  V2__b.sql

ways forward:
  - keep applying this database with the implementation that wrote the table: it reads its own
    versions and is unaffected
  - adopt the versions this build uses: `migrate set 4611686018427510315`, with the same --dir and
    --url, records every migration up to and including that version as applied under those
    versions. On this database every migration it would record has already run here, so nothing
    is recorded that did not execute. It is a one-way switch: the revision table then carries
    both spellings, and the implementation that wrote it refuses a migration added afterwards
    as out of order.

rewriting the version column by hand is not enough: the two implementations also record the
migration checksum differently, so an apply after a bare version rewrite refuses with a checksum
mismatch. `migrate set` writes the whole revision row.
exit=1
```

Every control above still prints what it printed before. `--dry-run` refuses instead of previewing. The seed row stays at 1.

## The route the refusal prints, re-measured

Two blocking findings were raised against the **recovery advice**, not the refusal. Both are real, both were reproduced here, and both are fixed.

**It dropped a migration.** `migrate set V` does not only remove revisions above `V`; it writes a revision row for **every covered migration up to `V`**, run or not. The head is the largest version among the migrations the other implementation *ran*, and the covered migrations below it need not all be among them. `V1__g1.sql` and `V3__g3.sql` recorded as `1`,`3` with `V2__g2.sql` added afterwards, following the printed route verbatim:

```
before      revisions: [1 3]      tables: [g1 g3]
$ ptah-compat migrate set 4611686018427551119 --dir 'file://d?format=flyway' --url sqlite://gap.db
Current version is 4611686018427551119 (3 set):
  + 4611686018427469511 (g1)
  + 4611686018427510315 (g2)
  + 4611686018427551119 (g3)
$ ptah-compat migrate apply --dir 'file://d?format=flyway' --url sqlite://gap.db
No migration files to execute.
exit=0
after       revisions: [1 3 4611686018427469511 4611686018427510315 4611686018427551119]
            tables: [g1 g3]
```

`g2` is absent and its version is recorded, so it can never run. The refusal said nothing. On that five-row table the pinned community binary also exits 1: `migration file V2__g2.sql was added out of order`.

The route is now offered only when it removes nothing **and** records nothing that did not execute, and withdrawn otherwise with the rows named:

```
adopting the versions this build uses has no safe route here: `migrate set 4611686018427551119`
would move the database to exactly that version, and on this database that is not a no-op:
  - it would record 1 migration(s) as applied WITHOUT running them — V2__g2.sql
    (4611686018427510315) — so that SQL would never execute here and nothing afterwards would
    report it missing
```

Nothing runs and nothing is recorded: `revisions [1 3]`, `tables [g1 g3]`, exactly where the other implementation left it.

**"The other implementation still reads the result as up to date" was false.** Measured on `V1__a.sql`,`V2__b.sql` recorded as `1`,`2`, after taking the route (`revisions [1 2 4611686018427469511 4611686018427510315]`):

```
$ atlas migrate apply --dir 'file://d?format=flyway'  --url sqlite://after-route.db
No migration files to execute
exit=0
$ atlas migrate apply --dir 'file://d3?format=flyway' --url sqlite://after-route.db   # + V3__c.sql
Error: migration file V3__c.sql was added out of order.
exit=1
$ ptah-compat migrate apply --dir 'file://d3?format=flyway' --url sqlite://after-route.db
Migrating to version 4611686018427551119 from 1 pending migrations.
Migration complete. Current version: 4611686018427551119
exit=0
```

Controls, both clean, separating that from the added file itself: the same three-file directory on a **fresh** database applies at exit 0 on the pinned binary, and on a database carrying only `1`,`2` it applies `V3` at exit 0. So the refusal now says the switch is **one way** and the two ways forward are alternatives, not steps.

That state is **not created by this change**. With the refusal call removed from `migrate_apply.go`, a plain apply over the same foreign table lands on the identical four-row set:

```
BASE before: [1 2]
BASE apply : Migrating to version 4611686018427510315 from 2 pending migrations. / Migration complete.
BASE after : [1 2 4611686018427469511 4611686018427510315]
```

So base reaches it by accident on every apply; this branch removes the accidental path and leaves a deliberate, disclosed one. `TestCompatMigrateApply_ForeignFlywayRefusalPrintsWorkingRecovery` no longer ends by adding a migration and asserting it applies — that assertion pinned `ptah-compat` exiting 0 where the pinned binary exits 1 as the desired result.

## Why shape 1, and what it costs

**Chosen: (b) over matching, and it is stated rather than buried.** The community binary exits 0 on that database, so refusing exits **1 where the pinned binary exits 0**. That is the allowed direction — half (a) is untouched, because nothing here makes `ptah-compat` exit 0 where the binary exits 1. Shapes 2 and 3 are the same change seen from two ends: the revision table is keyed on the int64 either way, so "record the token alongside" still needs the migrator's identity model to consult it. Shape 1 closes the data-safety edge without touching the projection.

**The rule, and the looser one it is not.** "A recorded version no converted file produces" — the phrasing the issue sketched — refuses an ordinary baseline squash, which retires applied migrations by design. The rule that shipped is narrower: a file's SOURCE token is recorded **and** the version that file converts to is not.

**The issue's stated cost is milder than it assumed, but only just.** It expected shape 1 to leave "an operator who switches tools with a directory neither can move forward". Measured, there is a route on the databases where it costs nothing: `migrate set <head>` records the history under this build's versions. It is a **one-way** switch, not a bridge — see the section above — and it is withdrawn entirely on the shapes where it loses something. So the refusal is a stop with a way out on most databases and a plain stop on the rest, which is closer to the issue's expectation than the first draft of this branch claimed.

**The route is measured, not inferred.** Running `migrate set <head>` on the reproduction: `(2 set)`, nothing removed, nothing recorded that did not execute, the next apply prints `No migration files to execute.` with the seed row still at 1, and the pinned community binary afterwards prints `No migration files to execute` at exit 0. A test runs the route verbatim and then asserts the resulting revision set, so "nothing is recorded that did not execute" is checked against what the command wrote rather than against the sentence. The hand-written `UPDATE` the pre-#982 refusal prints is deliberately **not** offered: measured, a bare version rewrite then fails with `migration 4611686018427469511 checksum mismatch`.

**The route is withdrawn on both shapes where `migrate set` is not a no-op.** It *removes* the revisions above the version given and *records* every covered migration below it:

- **Removes.** A converted baseline lands in a low band. On `V2` + `V1000000__z.sql` recorded by the community binary with `B2__base.sql` added, the head is `122412` and `1000000` sits above it; the refusal names the row the command would have deleted.
- **Records unrun.** `V1__g1.sql`,`V3__g3.sql` recorded as `1`,`3` with `V2__g2.sql` added; the refusal names `V2__g2.sql (4611686018427510315)` as a migration the command would assert without running.

Two inputs keep that from over-refusing: a migration **above** the head stays pending and is untouched by `migrate set`, and a migration this build recorded **itself** has run even though the foreign half of the table does not say so. Both are separate tests and separate mutants.

| test | on revert |
|---|---|
| `…_ForeignFlywayRevisionsRefused` | `Migrating to version 4611686018427510315 from 2 pending migrations.` then `Migration complete.` at exit 0, and two rows in `seeded` where there was one |
| `…_ForeignFlywayDryRunRefused` | `Dry run mode: no changes will be made.` and `Would have applied 2 migrations.` at exit 0 |
| `…_ForeignFlywayRefusalPrintsWorkingRecovery` | the first apply succeeds, so no message exists and the `migrate set` version cannot be extracted |
| `…_ForeignFlywaySetRouteWithdrawnWhenItWouldRecordUnrunMigrations` (revert the unrun clause) | prints `- adopt the versions this build uses: `migrate set 4611686018427551119`` — the command that records `g2` unrun |
| `…_ForeignFlywaySetRouteOfferedWithPendingAboveTheHead` (drop the `> head` skip) | `has no safe route here … WITHOUT running them — V3__c.sql (4611686018427551119)`, on a migration `migrate set` would not touch |
| `…_ForeignFlywaySetRouteCountsThisBuildsOwnRowsAsRun` (drop the `applied` half) | `WITHOUT running them — V1__m.sql (4611686018427469511)` for a migration this build ran itself |
| `…_ForeignFlywayTokenWithLeadingSpace` (revert the `TrimSpace`) | `Migrating to version 4611686018427469511 from 2 pending migrations.` at exit 0, seed row inserted twice |
| `…_ForeignFlywayBaselineOverForeignHistory` | `Migrating to version 122412 from 1 pending migrations.` at exit 0, table `base` created beside `p` and `q` |
| `…_ForeignFlywaySetRouteWithdrawnWhenUnsafe` (revert the withdrawal) | prints `migrate set 122412` as a way forward — a command that retires revision `1000000` |
| `…_ForeignFlywayTokenIsAlsoAConvertedVersion` (revert the third clause) | exit 1 refusing a database `ptah-compat` had just written itself, table `cv` never created |
| `…_DetectorDoesNotOverRefuse/a database migrated only by this build` | with the second clause dropped: `1 already-applied migration(s) read as pending` |
| `…_DetectorDoesNotOverRefuse/a baseline that retires the applied history` | with the issue's looser phrasing: refuses a squash both tools apply |
| `…_DetectorDoesNotOverRefuse/a goose directory holding a Flyway-shaped file` | with the layout gate removed: `1 already-applied migration(s) read as pending` on a goose database |
| `…_DetectorDoesNotOverRefuse/a fresh database` | non-interference control; passes on base too |
| `…_ForeignFlywayRepeatableCarriesNoToken` | with a repeatable reporting its own token: `2 -> 9223372036854775807  R2__r.sql` at exit 1, hiding the out-of-order diagnosis the community binary agrees with |

`scripts/mutations/foreign-flyway-revisions.json` is added: **17/17 killed**, tree clean after restore. It includes an **inverse mutant** for each conservative clause — dropping the collision clause makes the check fire on a database this build wrote, and forcing the `migrate set` route off removes a route measured to work. The route's two new guards each have both directions: `the lossy migrate set is printed anyway` / `the destructive migrate set is printed anyway` against `INVERSE: the migrate set route is never offered`, and the two over-refusal clauses against `a migration above the head is counted as one the route would record` and `only the other implementation's rows count as executed`.

## Gates

```
go build ./...                                              exit 0
go vet ./...                                                exit 0
go test ./... -count=1                                      exit 0, 176 packages ok, 0 failures
go tool qtlint ./...                                        exit 0
golangci-lint run ./...                                     exit 0, "0 issues."
scripts/check-test-style.sh                                 OK (175 conditional groups, 45 white-box files)
scripts/check-public-api.sh                                 exit 0
scripts/check-public-api-snapshot.sh                        exit 0
scripts/mutation-check.py scripts/mutations/foreign-flyway-revisions.json
                                                            17/17 killed, tree clean after restore
```

`qtlint` prints nothing on success, so it was proved live rather than assumed: rewriting one assertion as `c.Assert(len(userTables(…)), qt.Equals, 2)` makes it print `qtlint: use qt.HasLen instead of len(x), qt.Equals` at that line. `golangci-lint` first reported 89 issues from files under paths that no longer exist — a poisoned shared cache, with `can't read file …: no such file or directory` in the warnings; run against a private `GOLANGCI_LINT_CACHE` it reports `0 issues.`

`docs/` is touched, so every `check:*`/selftest script in `docs/site/package.json` was run, plus `versions:selftest` and `node docs/site/scripts/build-feature-matrix.mjs --check`: all green, including `check-exit-codes.mjs: OK (60 reference rows)`, `check-links.mjs: OK (60 pages, 60 routes, 571 heading anchors)`, `check-style.mjs: OK (100 documentation files)` and `build-feature-matrix.mjs --check: OK (163 rows)`. `check:responsive` needed `npm ci` and `npm run build` first — without them it reports `dist not found` and exits 1. Both really ran: `OK (60 routes x 2 viewports, cells within 8 lines)` and `OK (overflow, wide-table, tall-cell, and unstyled-page guards verified)`.

`gofmt -l` on the touched trees prints nothing; the positive control (a deliberately misformatted file dropped into `cmd/atlas`) does print its name, so the silence is a pass and not a broken probe.

## Deliberately left open

- **The checksum encoding.** `ptah-compat` stores a hex SHA-256 where the community binary stores a base64 `h1` digest, so golang-migrate, dbmate and liquibase directories refuse cross-tool with `checksum mismatch` even though their versions agree. That is a refusal, not a re-execution, and it is the reason a bare version rewrite is not a repair here. Not this issue; worth its own.
- **`migrate status`.** Still prints `Current Version: 2 / Applied: 2 / Pending: 2` on a mismatched table. The issue parks the count arithmetic explicitly ("Not this issue"), and the counts run over three different sets on the community binary too.
- **One pathological input the collision clause declines to flag.** A directory whose baseline converts to a number another file in the same directory carries as its token (`B10__base.sql` beside `V448844__x.sql`) is ambiguous evidence, so no claim is made about who wrote the row. That input behaves exactly as it does on base — a detection not added, not one taken away. The clause exists because without it `migrate apply 1` on that same directory makes the next apply blame another implementation for a row `ptah-compat` had just written.
- **Shapes 2 and 3 from the issue.** The version projection is untouched, so the two revision tables are still mutually unreadable; an operator who switches tools with a directory in flight still has to pick a side. That trade is what the issue named as the actual decision, and it is taken here knowingly.
- **A revision table carrying both spellings.** Once `1`,`2` and `4611686018427469511`,`4611686018427510315` sit side by side, a migration added afterwards is out of order to the pinned community binary (exit 1) and ordinary to `ptah-compat` (exit 0). That is a half-(a) gap, and it is **pre-existing**: measured with the refusal call removed, a plain apply over a foreign table produces that exact four-row set, so base reaches it on every such apply. This branch removes the accidental path to it; the remaining path is the printed route, which now says in the message that it is one way. Closing the gap itself means closing the version projection — shapes 2 and 3 above.

### DIFF
--- a/cmd/atlas/migrate_apply.go
+++ b/cmd/atlas/migrate_apply.go
@@ -310,6 +310,16 @@ func runAtlasMigrateApply(
 		return err
 	}
 
+	// The same question one implementation over: Atlas CE records a converted
+	// Flyway migration under its SOURCE version token, so a revision table it
+	// wrote also matches no file here and reads as entirely pending
+	// (stokaro/ptah#1100). The Ptah-encoding check above runs first because it
+	// is the more specific claim about who wrote the row, and its repair is a
+	// different one.
+	if err := checkForeignFlywayRevisions(captured, resolvedDirFormat, plan); err != nil {
+		return err
+	}
+
 	// The exemption above only stops the linear guard from reading the
 	// baseline's band position as "authored earlier". Whether a baseline may
 	// run against a database that already has history is a separate question,
--- a/cmd/atlas/migrate_apply_foreign_flyway.go
+++ b/cmd/atlas/migrate_apply_foreign_flyway.go
@@ -263,12 +263,18 @@ func foreignFlywaySetRoute(
 	}
 	unrun := foreignFlywayUnrunBelowHead(head, covered, stale, applied)
 	if len(removed) == 0 && len(unrun) == 0 {
+		// The count of listed migrations is deliberately NOT reported as the
+		// size of what `migrate set` writes: a database that recorded some of
+		// this directory under this build's own versions and the rest under the
+		// source tool's has more rows landing in than the refusal listed, and
+		// both kinds have run. What every one of them has in common is having
+		// executed, and that is what the sentence claims.
 		return fmt.Sprintf("  - adopt the versions this build uses: `migrate set %d`, with the same --dir "+
 			"and --url, records every migration up to and including that version as applied under those "+
-			"versions. On this database that set is exactly the %d migration(s) listed above, which it has "+
-			"already run, so nothing is recorded that did not execute. It is a one-way switch: the revision "+
-			"table then carries both spellings, and the implementation that wrote it refuses a migration "+
-			"added afterwards as out of order.\n", head, len(stale))
+			"versions. On this database every migration it would record has already run here, so nothing "+
+			"is recorded that did not execute. It is a one-way switch: the revision table then carries "+
+			"both spellings, and the implementation that wrote it refuses a migration added afterwards "+
+			"as out of order.\n", head)
 	}
 
 	var b strings.Builder
--- a/cmd/atlas/migrate_apply_foreign_flyway.go
+++ b/cmd/atlas/migrate_apply_foreign_flyway.go
@@ -1,0 +1,251 @@
+package atlas
+
+import (
+	"errors"
+	"fmt"
+	"io/fs"
+	"slices"
+	"strconv"
+	"strings"
+
+	"go.5x5.cz/ptah/internal/atlasmigrate"
+	"go.5x5.cz/ptah/internal/atlasmigrateimport"
+)
+
+// foreignFlywayRevision is one migration this database has already run, but
+// recorded under the source tool's version token instead of the Atlas version
+// this build executes it under.
+type foreignFlywayRevision struct {
+	// source is the Flyway file name, relative to the migration directory.
+	source string
+	// recorded is the version in the revision table: the source token, read
+	// the same way the revision reader reads it.
+	recorded int64
+	// current is the Atlas version this build would execute the file under.
+	current int64
+}
+
+// checkForeignFlywayRevisions refuses an apply that would re-run migrations
+// another Atlas implementation has already applied through the same converted
+// directory (stokaro/ptah#1100).
+//
+// A `?format=flyway` directory is converted to Atlas single-file migrations
+// before it is executed, and the two implementations do not agree on the
+// version the converted migration is then recorded under. Atlas CE identifies a
+// Flyway migration by its opaque version STRING and records that string
+// verbatim; Ptah's migrator identifies a migration by an int64, so converting
+// projects the token onto the band-and-slot encoding documented above
+// flywayComponentSlot. Measured on the pinned community binary v1.3.0 with
+// V1__a.sql and V2__b.sql: it records `1` and `2`, this build records
+// `4611686018427469511` and `4611686018427510315`.
+//
+// The projection is not a different spelling of one version, it makes the two
+// revision tables mutually unreadable, and in one direction that means
+// re-execution. Reading a table this build wrote is benign — the community
+// binary sees versions above everything it has and executes nothing — but
+// reading a table the community binary wrote, every converted file matches no
+// row and the whole directory reads as pending. Measured, both halves:
+//
+//	CREATE TABLE t1 (id int)                 -> `table t1 already exists`, exit 1,
+//	                                            and a dirty revision left behind
+//	CREATE TABLE IF NOT EXISTS + INSERT       -> exit 0, and the seed row inserted twice
+//
+// The second is why this is a refusal and not a diagnostic. Nothing in the exit
+// code, stdout or `migrate status` said the data had been duplicated.
+//
+// WHY REFUSE RATHER THAN MATCH. The community binary exits 0 on that database
+// having executed nothing, so refusing exits 1 where it exits 0. That is the
+// deliberate half (b) choice: the alternatives are to reproduce its version
+// space (which changes migration identity in the migrator, far past this issue)
+// or to record the token alongside the version and match on it (which is the
+// same change seen from the other end, since the revision table would still be
+// keyed on the int64). Refusing removes the data-safety edge without touching
+// the projection, and the parity rule allows being stricter, never looser.
+//
+// THE RULE, and what separates it from the plausible looser one. "A recorded
+// version no converted file produces" would refuse an ordinary baseline squash,
+// which retires applied migrations by design. The rule is narrower: a file's
+// SOURCE TOKEN is recorded AND the version that file converts to is not. A
+// database this build wrote fails the first clause; a database an operator has
+// already migrated forward fails the second.
+//
+// The token is compared as an int64 read by strconv.ParseInt, because that is
+// exactly what the revision reader does with the recorded string
+// (parseAtlasRevisionVersion), so "01" and "1" meet in the same place they meet
+// there. A token the reader cannot parse — "1.5", or a repeatable's empty
+// string — is skipped, and needs no branch here: such a row makes the read
+// itself fail first, with `Atlas revision version "1.5" is not a numeric Ptah
+// migration version`, which is already a refusal before anything executes.
+//
+// ONLY FLYWAY, measured rather than assumed. Of the five converted layouts,
+// only Flyway re-executes. Applying the same two-migration directory with each
+// implementation into its own database and reading each with the other: goose
+// records `00001`/`00002` against this build's `1`/`2`, which meet under
+// ParseInt, so nothing is pending; golang-migrate, dbmate and liquibase record
+// identical versions on both. Those three still refuse cross-tool, with
+// `checksum mismatch: stored <base64>, current <hex>` — the two implementations
+// also disagree about the checksum ENCODING — but that is a refusal, not a
+// re-execution, and it is not this issue. Extending this guard to them would be
+// unreachable code, since their tokens ARE the versions they convert to.
+//
+// There is exactly ONE layout gate, and it is inside
+// [atlasmigrateimport.FlywayCoveredSourceVersions], which reports nothing for
+// any other layout. Repeating it here would make each copy individually
+// unkillable — either alone still suppresses the check — and a rule two
+// redundant guards hold is a rule no test can hold.
+func checkForeignFlywayRevisions(
+	captured fs.FS,
+	format atlasmigrateimport.Format,
+	plan atlasmigrate.ApplyPlan,
+) error {
+	if plan.Status == nil {
+		return nil
+	}
+	applied := plan.Status.AppliedMigrations
+	if len(applied) == 0 {
+		return nil
+	}
+	covered, err := atlasmigrateimport.FlywayCoveredSourceVersions(captured, format)
+	if err != nil {
+		// The conversion error this would report is the one the caller already
+		// surfaces on its own path; nothing to add here.
+		return nil //nolint:nilerr // conversion failures are reported by the apply path itself.
+	}
+	stale := foreignFlywayRevisions(covered, applied)
+	if len(stale) == 0 {
+		return nil
+	}
+	return errors.New(foreignFlywayRefusal(stale, applied))
+}
+
+// foreignFlywayRevisions selects the migrations this database recorded under
+// the source token.
+//
+// The third clause is the one that keeps the detector from inventing a
+// refusal. A converted version and a source token live in the same int64 space,
+// and a baseline converts into the LOW band — B10__base.sql becomes 448844 —
+// so a directory that also holds V448844__x.sql has a token equal to a version
+// this very directory produces. Recording it says nothing about which
+// implementation wrote the row, so that pairing is dropped rather than read as
+// foreign history. Measured without it, `migrate apply 1` on exactly that
+// directory (which runs the baseline alone, since the baseline band sorts
+// first) makes the next apply refuse a database this build had just written
+// itself, blaming another implementation for a row it did not write.
+//
+// The cost is stated rather than hidden: on that same pathological directory
+// the clause also declines to flag the genuine cross-tool case, because the
+// evidence really is ambiguous. That input then behaves exactly as it did
+// before this check existed, so it is a detection this does not add rather than
+// one it takes away.
+func foreignFlywayRevisions(
+	covered []atlasmigrateimport.FlywayCoveredSourceVersion,
+	applied []int64,
+) []foreignFlywayRevision {
+	ours := make([]int64, 0, len(covered))
+	for _, migration := range covered {
+		ours = append(ours, migration.Version)
+	}
+
+	var stale []foreignFlywayRevision
+	for _, migration := range covered {
+		recorded, err := strconv.ParseInt(strings.TrimSpace(migration.Token), 10, 64)
+		if err != nil {
+			continue
+		}
+		if !slices.Contains(applied, recorded) || slices.Contains(applied, migration.Version) {
+			continue
+		}
+		if slices.Contains(ours, recorded) {
+			continue
+		}
+		stale = append(stale, foreignFlywayRevision{
+			source:   migration.Source,
+			recorded: recorded,
+			current:  migration.Version,
+		})
+	}
+	return stale
+}
+
+// foreignFlywayRefusal renders the refusal and the two routes measured to work.
+//
+// The hand-written UPDATE the pre-#982 refusal prints is deliberately NOT
+// offered here, and the last paragraph says why. That repair moves a row
+// between two encodings of ONE implementation, so the recorded hash still
+// covers the same converted body. Across implementations it does not: measured,
+// rewriting `1` to `4611686018427469511` by hand on a community-binary-written
+// table gets `migration 4611686018427469511 checksum mismatch: stored
+// 9mRxDpig5tZbhHslnEoWdaeiT7fFPnkxapig7dheO1w=, current
+// ca5446be32fd97a6194819f463b3321c799f9acd0fb138f62dca0cc48164dfa8` on the next
+// apply — the two implementations record the checksum in different encodings.
+// `migrate set` writes whole revision rows, so it settles both at once.
+//
+// Both routes are measured on the fixture in the reproduction, not inferred:
+// `migrate set <head>` reports `(2 set)` and removes nothing, the next
+// `migrate apply` prints `No migration files to execute.` with the seed row
+// still inserted once, and the community binary afterwards prints
+// `No migration files to execute` at exit 0 — so adopting this build's versions
+// does not take the database away from the implementation that wrote it.
+func foreignFlywayRefusal(stale []foreignFlywayRevision, applied []int64) string {
+	var b strings.Builder
+	fmt.Fprintf(&b, "this database records converted Flyway migrations under their SOURCE version token, "+
+		"which is how another Atlas implementation identifies them; Ptah's migrator identifies a migration "+
+		"by the int64 that token projects to, so %d already-applied migration(s) read as pending here and "+
+		"would run a second time. Nothing has been applied", len(stale))
+
+	b.WriteString("\n\nrecorded version -> version this build uses:\n")
+	head := int64(0)
+	for _, revision := range stale {
+		fmt.Fprintf(&b, "  %-20d -> %-20d %s\n", revision.recorded, revision.current, revision.source)
+		head = max(head, revision.current)
+	}
+
+	b.WriteString("\nways forward:\n")
+	b.WriteString("  - keep applying this database with the implementation that wrote the table: it reads " +
+		"its own versions and is unaffected\n")
+	b.WriteString(foreignFlywaySetRoute(head, applied))
+
+	b.WriteString("\nrewriting the version column by hand is not enough: the two implementations also record " +
+		"the migration checksum differently, so an apply after a bare version rewrite refuses with a " +
+		"checksum mismatch. `migrate set` writes the whole revision row.\n")
+	return b.String()
+}
+
+// foreignFlywaySetRoute renders the second way forward, or says there is none.
+//
+// `migrate set V` moves the database to EXACTLY V: it records everything up to
+// V as applied and REMOVES every revision above it. Offering it unconditionally
+// would print a command that deletes real history on one measurable shape.
+// Converted baselines sit in the low band — B2__base.sql becomes 122412 — and a
+// baseline squashes files whose token sorts at or below its own AS A STRING, so
+// `V1000000__z.sql` (token "1000000" < "2", value 1000000) is squashed out of
+// the covered set while its revision row sits above the head. Recommending
+// `migrate set 122412` there would retire the migration the database actually
+// ran.
+//
+// So the route is offered only when it removes nothing, and named as unsafe
+// with the rows it would delete otherwise — the same choice
+// [flywayBaselineRefusal] makes for the same verb, and for the same reason: a
+// refusal that prints a command which quietly makes it worse is not a way
+// forward.
+func foreignFlywaySetRoute(head int64, applied []int64) string {
+	var removed []int64
+	for _, version := range applied {
+		if version > head {
+			removed = append(removed, version)
+		}
+	}
+	if len(removed) == 0 {
+		return fmt.Sprintf("  - adopt the versions this build uses: `migrate set %d`, with the same --dir "+
+			"and --url, records every migration up to and including that version as applied under those "+
+			"versions. The other implementation still reads the result as up to date.\n", head)
+	}
+
+	labels := make([]string, 0, len(removed))
+	for _, version := range removed {
+		labels = append(labels, strconv.FormatInt(version, 10))
+	}
+	return fmt.Sprintf("\nadopting the versions this build uses has no safe route here: `migrate set %d` "+
+		"would move the database to exactly that version and remove the revision(s) above it — %s — which "+
+		"is history this database really ran.\n", head, strings.Join(labels, ", "))
+}
--- a/cmd/atlas/migrate_apply_foreign_flyway.go
+++ b/cmd/atlas/migrate_apply_foreign_flyway.go
@@ -115,7 +115,7 @@ func checkForeignFlywayRevisions(
 	if len(stale) == 0 {
 		return nil
 	}
-	return errors.New(foreignFlywayRefusal(stale, applied))
+	return errors.New(foreignFlywayRefusal(covered, stale, applied))
 }
 
 // foreignFlywayRevisions selects the migrations this database recorded under
@@ -167,7 +167,7 @@ func foreignFlywayRevisions(
 	return stale
 }
 
-// foreignFlywayRefusal renders the refusal and the two routes measured to work.
+// foreignFlywayRefusal renders the refusal and the ways forward.
 //
 // The hand-written UPDATE the pre-#982 refusal prints is deliberately NOT
 // offered here, and the last paragraph says why. That repair moves a row
@@ -179,14 +179,11 @@ func foreignFlywayRevisions(
 // ca5446be32fd97a6194819f463b3321c799f9acd0fb138f62dca0cc48164dfa8` on the next
 // apply — the two implementations record the checksum in different encodings.
 // `migrate set` writes whole revision rows, so it settles both at once.
-//
-// Both routes are measured on the fixture in the reproduction, not inferred:
-// `migrate set <head>` reports `(2 set)` and removes nothing, the next
-// `migrate apply` prints `No migration files to execute.` with the seed row
-// still inserted once, and the community binary afterwards prints
-// `No migration files to execute` at exit 0 — so adopting this build's versions
-// does not take the database away from the implementation that wrote it.
-func foreignFlywayRefusal(stale []foreignFlywayRevision, applied []int64) string {
+func foreignFlywayRefusal(
+	covered []atlasmigrateimport.FlywayCoveredSourceVersion,
+	stale []foreignFlywayRevision,
+	applied []int64,
+) string {
 	var b strings.Builder
 	fmt.Fprintf(&b, "this database records converted Flyway migrations under their SOURCE version token, "+
 		"which is how another Atlas implementation identifies them; Ptah's migrator identifies a migration "+
@@ -203,7 +200,7 @@ func foreignFlywayRefusal(stale []foreignFlywayRevision, applied []int64) string
 	b.WriteString("\nways forward:\n")
 	b.WriteString("  - keep applying this database with the implementation that wrote the table: it reads " +
 		"its own versions and is unaffected\n")
-	b.WriteString(foreignFlywaySetRoute(head, applied))
+	b.WriteString(foreignFlywaySetRoute(head, covered, stale, applied))
 
 	b.WriteString("\nrewriting the version column by hand is not enough: the two implementations also record " +
 		"the migration checksum differently, so an apply after a bare version rewrite refuses with a " +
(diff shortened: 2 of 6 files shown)


denisvmedia:
Marked draft. The core fix is right and measured — master re-executes and this branch refuses — but the adversarial review found two blocking findings, and both are in the **recovery advice the refusal prints**, not in the refusal itself.

**Finding 1 — following the tool's own advice silently drops a migration.** After taking the printed route, the next apply prints `No migration files to execute.` at exit 0 with table `g2` absent and unreachable forever. On that database the pinned binary still exits 1 where ptah-compat exits 0 — parity (a) violated, reached by doing exactly what the message says. So the change converts a loud failure into a documented procedure that loses a migration: the same class of defect the PR exists to remove.

**Finding 2, same root — after the route, the other implementation stops working.** With both spellings recorded (`1,2,4611686018427469511,4611686018427510315`), adding an ordinary `V3__more.sql` and re-hashing gives ptah-compat exit 0 `Migration complete.` and the pinned binary exit 1 `migration file V3__more.sql was added out of order`. Controls are clean: fresh database with V1,V2,V3 → binary exit 0; a CE-only database `{1,2}` plus V3 → binary exit 0.

That falsifies two sentences that ship in this branch: the refusal's "The other implementation still reads the result as up to date", and the body's "a migration added later still applies". Both were measured with ptah-compat only. `TestCompatMigrateApply_ForeignFlywayRefusalPrintsWorkingRecovery` asserts precisely the state the binary rejects, so the test pins the wrong claim rather than catching it.

The remedy the reviewer names, and it is already the shape this file uses for the other direction: **withdraw the route** when any covered migration below the head is unapplied — or name the rows it would record without running them, so the operator can see what is being asserted rather than being told it is safe.

The author's statement "no (a) violation survives from this change, and none was introduced" is true of the code path and false of the workflow the change publishes. Worth keeping that distinction in the rewritten body: a refusal is only as good as the recovery it hands you.

Everything else re-measured clean and is worth keeping: the CE-vs-Ptah version spellings, the ` 1`/`01` trimming agreement with `parseAtlasRevisionVersion`, repeatables recording `''` on both sides, and goose/golang-migrate/dbmate/liquibase agreeing on versions and refusing cross-tool on checksum.

denisvmedia:
Both blocking findings were reproduced here and both are fixed. Everything below is measured on this branch; the pinned community binary v1.3.0 is invoked by absolute path, sqlite, each cell in its own database.

## Finding 1 — the printed route dropped a migration

`migrate set V` does not only remove revisions above `V`. It writes a revision row for **every covered migration up to `V`**, run or not. The head is the largest version among the migrations the other implementation *ran*, and the covered migrations below it need not all be among them.

`V1__g1.sql` and `V3__g3.sql` recorded as `1`,`3`, with `V2__g2.sql` added afterwards; the route followed verbatim, before and after:

```
before   revisions: [1 3]                     tables: [g1 g3]

$ ptah-compat migrate set 4611686018427551119 --dir 'file://d?format=flyway' --url sqlite://gap.db
Current version is 4611686018427551119 (3 set):
  + 4611686018427469511 (g1)
  + 4611686018427510315 (g2)
  + 4611686018427551119 (g3)

$ ptah-compat migrate apply --dir 'file://d?format=flyway' --url sqlite://gap.db
No migration files to execute.
exit=0

after    revisions: [1 3 4611686018427469511 4611686018427510315 4611686018427551119]
         tables: [g1 g3]
```

`g2` never ran, its version is recorded, and it can never run again. The pinned binary refuses that database too:

```
$ /…/ptah-atlas-conformance/bin/atlas migrate apply --dir 'file://d?format=flyway' --url sqlite://after-route.db
Error: migration file V2__g2.sql was added out of order. See: https://atlasgo.io/versioned/apply#non-linear-error
exit=1
```

**Fix — both halves of the named remedy.** The route is offered only when it removes nothing **and** records nothing that did not execute; otherwise it is withdrawn, naming the rows:

```
adopting the versions this build uses has no safe route here: `migrate set 4611686018427551119`
would move the database to exactly that version, and on this database that is not a no-op:
  - it would record 1 migration(s) as applied WITHOUT running them — V2__g2.sql
    (4611686018427510315) — so that SQL would never execute here and nothing afterwards would
    report it missing
```

Nothing runs and nothing is recorded: `revisions [1 3]`, `tables [g1 g3]` — exactly where the other implementation left it. That is `TestCompatMigrateApply_ForeignFlywaySetRouteWithdrawnWhenItWouldRecordUnrunMigrations`.

Two inputs keep the new clause from over-refusing, each with its own test and its own mutant: a migration **above** the head stays pending and `migrate set` never touches it, and a migration this build recorded **itself** has run even though the foreign half of the table does not say so.

## Finding 2 — "The other implementation still reads the result as up to date" was false

Measured on `V1__a.sql`,`V2__b.sql` recorded as `1`,`2`, after the route (`revisions [1 2 4611686018427469511 4611686018427510315]`):

```
$ atlas migrate apply --dir 'file://d?format=flyway'  --url sqlite://after-route.db
No migration files to execute
exit=0

$ atlas migrate apply --dir 'file://d3?format=flyway' --url sqlite://after-route.db   # + V3__c.sql
Error: migration file V3__c.sql was added out of order. See: …
exit=1

$ ptah-compat migrate apply --dir 'file://d3?format=flyway' --url sqlite://after-route.db
Migrating to version 4611686018427551119 from 1 pending migrations.
Migration complete. Current version: 4611686018427551119
exit=0
```

Controls, both clean, separating that from the added file itself:

```
$ atlas migrate apply --dir 'file://d3?format=flyway' --url sqlite://fresh.db
Migrating to version 3 (3 migrations in total): … exit=0

$ atlas migrate apply --dir 'file://d3?format=flyway' --url sqlite://foreign-only-v3.db   # revisions [1 2]
Migrating to version 3 from 2 (1 migrations in total): … exit=0
```

The sentence is replaced with the measured one: the offer now says it is a **one-way switch**, and that the implementation that wrote the table refuses a migration added afterwards as out of order. The two ways forward are alternatives, not steps. `TestCompatMigrateApply_ForeignFlywayRefusalPrintsWorkingRecovery` no longer ends by adding a migration and asserting it applies — that assertion pinned `ptah-compat` exiting 0 where the pinned binary exits 1 as the desired result.

**That state is not created by this change.** With the refusal call removed from `migrate_apply.go`, a plain apply over the same foreign table lands on the identical four-row set:

```
BASE before: [1 2]
BASE apply : Migrating to version 4611686018427510315 from 2 pending migrations. / Migration complete.
BASE after : [1 2 4611686018427469511 4611686018427510315]
```

Base reaches it by accident on every such apply. This branch removes the accidental path and leaves one deliberate, disclosed path. The residual half-(a) gap on a both-spellings table is now named under "Deliberately left open" instead of being asserted as working; closing it means closing the version projection, which is shapes 2 and 3 from the issue.

## One more sentence corrected on the way

The first version of this fix had the offer say "that set is exactly the N migration(s) listed above". False on a database that recorded part of this directory under this build's own versions and the rest under the source tool's: the refusal lists only the foreign half while `migrate set` writes a row for both, and both have run. The count is gone; what ships is the claim that survives every shape — every migration it would record has already run here.

denisvmedia:
Independently verified the blocking finding on `8b4bf671`, with fixtures I built myself and the pinned community binary by absolute path.

**The safe case still offers the route, and now says why it is safe:**

```
V1__g1.sql, V2__g2.sql — both applied by the community binary, revisions 1,2

  - adopt the versions this build uses: `migrate set 4611686018427510315` …
    On this database every migration it would record has already run here, so
    nothing is recorded that did not execute.
```

**The gap case withdraws it, and names the row that would be lost:**

```
V1__g1.sql and V3__g3.sql applied (revisions 1,3), V2__g2.sql added afterwards

  adopting the versions this build uses has no safe route here:
  `migrate set 4611686018427551119` would move the database to exactly that
  version, and on this database that is not a no-op:
    - it would record 1 migration(s) as applied WITHOUT running them —
      V2__g2.sql (4611686018427510315) — so that SQL would never execute here
      and nothing afterwards would report it missing
```

That is the discriminator the finding asked for: the two databases produce **different** advice, so the check is real rather than a sentence appended to a fixed message. The finding's requirement was "withdraw the route or name exactly which rows are dropped"; this does both, choosing per database, and names the file and the version it projects to.

Both cells exit 1 and apply nothing.

denisvmedia:
## Adversarial re-review of the blocker fix — head `8b4bf671`

I tried to refute "the blocker is gone" and could not. Everything below is re-measured here, sqlite, each cell in its own database, the pinned community binary invoked by absolute path.

### The blocker is fixed — measured, not read

Rebuilt the gap fixture from scratch: `V1__g1.sql` + `V3__g3.sql` applied by the pinned community binary v1.3.0, then `V2__g2.sql` added and re-hashed.

```
BEFORE     revisions: 1 3      tables: g1 g3

ptah-compat @ 8b4bf671:
Error: ... so 2 already-applied migration(s) read as pending here and would run a second time. Nothing has been applied
...
adopting the versions this build uses has no safe route here: `migrate set 4611686018427551119` would
move the database to exactly that version, and on this database that is not a no-op:
  - it would record 1 migration(s) as applied WITHOUT running them — V2__g2.sql (4611686018427510315) — ...
exit=1
AFTER      revisions: 1 3      tables: g1 g3     (untouched)
```

The route is withdrawn and the row it would have asserted is named. Applying the `the lossy migrate set is printed anyway` mutant (`if len(removed) == 0` in place of `if len(removed) == 0 && len(unrun) == 0`) and following the printed route verbatim reproduces the loss the blocker described:

```
$ ptah-compat migrate set 4611686018427551119 --dir '...?format=flyway' --url sqlite://gap-mut.db
Current version is 4611686018427551119 (3 set):
  + 4611686018427469511 (g1)
  + 4611686018427510315 (g2)
  + 4611686018427551119 (g3)
$ ptah-compat migrate apply ...
No migration files to execute.        exit=0
AFTER  revisions: 1 3 4611686018427469511 4611686018427510315 4611686018427551119
       tables:    g1 g3          <- g2 recorded, never created, unreachable
```

That mutant reddens the new test, watched failing, not assumed:

```
--- FAIL: TestCompatMigrateApply_ForeignFlywaySetRouteWithdrawnWhenItWouldRecordUnrunMigrations
    migrate_apply_foreign_flyway_test.go:286: ... want: "has no safe route here"
```

Full sweep re-run independently: **17/17 killed, tree clean after restore.** The three new route guards are each killed by exactly the test the body names, and the `INVERSE` rows are present for both directions.

Gates re-run from the worktree root: `go build` 0, `go vet` 0, `go test ./... -count=1` 0 (176 packages ok), `qtlint` 0, `golangci-lint` 0 issues (private `GOLANGCI_LINT_CACHE`), `check-test-style` OK, `check-public-api` + snapshot 0, all 13 docs `check:*`/selftest scripts 0, `build-feature-matrix.mjs --check: OK (163 rows)`. `qtlint` proved live: rewriting one assertion as `c.Assert(len(userTables(…)), qt.Equals, 2)` prints `qtlint: use qt.HasLen instead of len(x), qt.Equals` at exit 3. `check:responsive` could not be verified here (no `node_modules`); its no-`dist` failure mode is exit 1, as the body says.

The one-way-switch claim also reproduces exactly, controls included: after the route the pinned binary prints `No migration files to execute` at exit 0; add `V3__c.sql` and it prints `migration file V3__c.sql was added out of order` at exit 1 while `ptah-compat` applies at exit 0; the same three-file directory on a fresh database and on a database carrying only `1`,`2` both apply at exit 0 on that binary. And the "pre-existing" argument holds: base `ptah-compat` lands on the identical four-row set `1 2 4611686018427469511 4611686018427510315` on a plain apply (with the seed row duplicated to 2, which is the original defect).

So the blocker is closed. Four things are still wrong, and one of them is a repeat.

---

### 1. Over-refusal on a database `ptah-compat` wrote itself, and the body describes it as the opposite

`B10__base.sql` beside `V448844__x.sql`; `ptah-compat` applies the baseline alone, recording `448844` under **its own** encoding. Then a second baseline `B20__base2.sql` is added — which drops `B10` out of the covered set, so `448844` is no longer in `ours` and the collision clause stops firing:

```
ptah-compat writes it:  migrate apply 1  ->  revisions: 448844   tables: cbase

ptah-compat @ 028ac05d (base):  Migrating to version 4611686036742059283 from 2 pending migrations.   exit=0
ptah-compat @ 8b4bf671 (head):  Error: this database records converted Flyway migrations under their
                                SOURCE version token, which is how another Atlas implementation
                                identifies them ...
                                  448844  ->  4611686036742059283   V448844__x.sql
                                exit=1
pinned community binary v1.3.0: No migration files to execute                                          exit=0
```

Head exits 1 where both base and the pinned binary exit 0, on a revision row `ptah-compat` wrote itself, with a message that blames another implementation and mis-attributes `448844` to `V448844__x.sql`. That is a drop-in regression under rule 7(b), and rule 7(b) says it has to be argued explicitly.

Instead the body's "Deliberately left open" bullet says of this exact directory:

> That input behaves exactly as it does on base — a detection not added, not one taken away. The clause exists because without it `migrate apply 1` on that same directory makes the next apply blame another implementation for a row `ptah-compat` had just written.

Measured, it does not behave as it does on base, and the branch does precisely the thing the bullet says the clause prevents. The clause protects only while the collision partner stays covered; one more baseline retires it. Either fix the clause (compare against the tokens/versions the directory produced *at record time*, not the currently-covered set) or replace that bullet with the measured behavior and argue the 7(b) direction.

### 2. A fabricated FAIL output — same finding as the previous round, unfixed

Body table row, and the comment at `cmd/atlas/migrate_apply_foreign_flyway_test.go:626-627`:

> `…_DetectorDoesNotOverRefuse/a database migrated only by this build` — with the second clause dropped: `1 already-applied migration(s) read as pending`

Applying that exact mutant (`slices.Contains(applied, migration.Version)` removed) and running only that test:

```
--- PASS: TestCompatMigrateApply_ForeignFlywayDetectorDoesNotOverRefuse (0.09s)
    a_database_migrated_only_by_this_build   PASS
```

The whole test passes — the row's tokens are never recorded, so clause 1 short-circuits and the row never reaches clause 2. This PR's own harness agrees: the `reddened` column for `a database already migrated forward is flagged again` lists `BaselineOverForeignHistory, RefusalPrintsWorkingRecovery, RepeatableCarriesNoToken` and `(0 subtests)`. The clause is held — but not by the fixture documented as holding it, and that fixture is a coincidence for this clause. Under rule 8 a documented FAIL output that does not occur is worse than none. Fix the comment and the table row to name the tests that actually go red, or give that row an input that reaches clause 2.

### 3. On the shape where the route is withdrawn, the remaining "way forward" is also false

The gap database gets exactly one way forward printed:

> `- keep applying this database with the implementation that wrote the table: it reads its own versions and is unaffected`

Measured on that same database (`revisions 1 3`, `V2__g2.sql` added):

```
$ atlas migrate apply --dir '...?format=flyway' --url sqlite://gap.db
Error: migration file V2__g2.sql was added out of order.    exit=1
```

"unaffected" is true of the version-encoding divergence and false of the database. On the one shape where the second route is withdrawn, the operator is sent to a tool that also refuses, and is told it will not. The withdrawal branch should either qualify that sentence or say plainly that this database needs the gap resolved first, whichever tool runs it.

### 4. The offer's "nothing is recorded that did not execute" is literally false on the baseline shape

`V1__p.sql`, `V2__q.sql` recorded as `1`,`2`; `B2__base.sql` added. The route **is** offered, and taking it:

```
$ ptah-compat migrate set 122412 --dir '...?format=flyway' --url sqlite://b.db
Current version is 122412 (1 set):
  + 122412 (base)
$ ptah-compat migrate apply ...            No migration files to execute.   exit=0
AFTER  revisions: 1 2 122412     tables: p q      <- table `base` never created
```

`B2__base.sql` is recorded and its SQL never ran, while the message says every migration it would record has already run here. `TestCompatMigrateApply_ForeignFlywayBaselineOverForeignHistory` asserts that same absence, so this is intended — and the outcome matches the pinned binary, which also never creates `base`. Nothing is lost; the sentence is just untrue on this shape, and it is the sentence the blocker was about. Worth narrowing to what `foreignFlywayUnrunBelowHead` actually checks.

---

### Verdict

The blocker is gone: the lossy route is withdrawn, the withdrawal names the file and version it would have asserted, the fixture is discriminating (mutant watched failing), no new parity violation is introduced by the two fix commits, and every gate reproduces green. Items 1 and 2 are pre-existing to the blocker fix but still open on a PR marked ready, and item 1 is an unargued rule-7(b) regression whose body text asserts the opposite of what it measures.

No repo-rule violations found in the diff, the commit messages or the body: no banned tokens, no AI attribution, American English throughout, `quicktest` only, no `if`/`switch` in any test body.

PR #1189, opened 2026-08-06
feat(protobuf-export): add --proto-comments to omit source comments

### LINKED ISSUE #1145: Protobuf export copies source comments into the published contract with no way to suppress them
Split out of #923, which measured the eleven documented protobuf-export limitations and found this one resolvable and untracked.

## What reproduces today

Measured on master `feb6af19` with a `ptah` built from that commit, plus `protoc` 35.1 and `protoc-gen-go` v1.36.11.

Models carrying an internal table comment and a sensitive column comment:

```go
//ptah:schema:table name="users" comment="Internal: sharded by tenant_id; see runbook RB-42"
type User struct {
	//ptah:schema:field name="id" type="SERIAL" primary="true"
	ID int64

	//ptah:schema:field name="email" type="VARCHAR(255)" not_null="true"
	Email string

	//ptah:schema:field name="password_hash" type="VARCHAR(255)" not_null="true" comment="bcrypt hash, never expose"
	PasswordHash string
}
```

`ptah schema export --to protobuf` exits 0 and writes both comments verbatim:

```proto
// Internal: sharded by tenant_id; see runbook RB-42
message User {
  int32 id = 1;
  string email = 2;
  // bcrypt hash, never expose
  string password_hash = 3;
}
```

They travel one hop further than the documentation says. `protoc --go_out` over that file produces a `.pb.go` in which `runbook RB-42` and `bcrypt hash, never expose` each occur once, so publishing generated Go republishes them too.

There is no way to turn this off: `--no-comments`, `--strip-comments`, and `--omit-comments` each exit 2 with `unknown flag` (control: `--include-tables users` exits 0 on the same fixture; `--definitely-not-a-flag` exits 2 with `unknown flag`).

## Acceptance

On the fixture above, with the new flag:

1. The export writes a `.proto` containing **zero** `//` lines other than the three generated header lines.
2. A second run with the flag is a byte-identical no-op.
3. A run **without** the flag restores both comments and **renumbers no field** — `id` stays 1, `email` stays 2, `password_hash` stays 3. Comment emission must not be part of the compatibility state.

Today step 1 exits 2 with `unknown flag`.

The policy decision this issue also has to settle: whether suppression is all-or-nothing or per-object, and whether the generated header lines are in scope (they are covered separately by the header-position issue split from #923).

Documentation to update once this lands: the Limitations bullet about copied comments on `docs/site/src/content/docs/schema/protobuf.md`, which links here.

denisvmedia:
## What reproduced

Measured on `25cf00ff` with a `ptah` built from that commit, plus `protoc` 35.1 and `protoc-gen-go` v1.36.11, on the fixture from the issue.

`ptah schema export --to protobuf` exited 0 and wrote both comments verbatim:

```proto
// Code generated by ptah schema export --to protobuf; DO NOT EDIT.
// ptah:protobuf-export-version=1
// ptah:content-sha256=45d34ff34a925392045e86ab1b2ec9e55fac977725e5d80936dc72b84eb47b5b
edition = "2023";

package acme.inventory.v1;

// Internal: sharded by tenant_id; see runbook RB-42
message User {
  int32 id = 1;
  string email = 2;
  // bcrypt hash, never expose
  string password_hash = 3;
}
```

`protoc --go_out` over that file produced a `schema.pb.go` in which `grep -c "runbook RB-42"` reported `1` and `grep -c "bcrypt hash, never expose"` reported `1`, so publishing generated Go republished them too.

There was no way to turn it off. `--no-comments`, `--strip-comments`, `--omit-comments` and `--proto-comments` each exited 2 with `error: unknown flag`; the control `--include-tables users` exited 0 on the same fixture.

## What this changes

`--proto-comments` takes `all` (default) or `none`. On the same fixture, `--proto-comments=none` exits 0 and writes:

```proto
// Code generated by ptah schema export --to protobuf; DO NOT EDIT.
// ptah:protobuf-export-version=1
// ptah:content-sha256=410a512e42ff37d9e9e53485b91e2f7c96a17dac2aeb2e8113e497611da0d476
edition = "2023";

package acme.inventory.v1;

message User {
  int32 id = 1;
  string email = 2;
  string password_hash = 3;
}
```

`protoc --go_out` over that file reports 0 occurrences of either string (`grep` exit 1).

Acceptance, measured end to end with the built binary:

1. The file carries zero `//` lines other than the three generated header lines.
2. A second run with the flag is a byte-identical no-op — `cmp` exit 0.
3. A run without the flag restores both comments and renumbers nothing: `id` 1, `email` 2, `password_hash` 3, and the digest returns to `45d34ff34a92…`, byte-identical to the pre-flag export.

## The policy decisions the issue asked to settle

- **All-or-nothing, not per-object.** A table comment can carry exactly the internal detail a column comment can, so a per-table or per-column switch would report a boundary the published contract does not have. The value is enumerated rather than a boolean so a further policy can be added later without renaming the flag.
- **Ptah's own generated prose stays.** Suppression is applied where source text enters the model, not in the writer. What a published contract owes its readers is the wire shape plus Ptah's own account of it; the database's internal documentation is neither. The tombstone rationale therefore survives `none` — without it a reader meets a bare `reserved` block with nothing to explain it. The three header lines are out of scope here per the issue.
- **Comment emission is not compatibility state.** `previous.go` reads numbers and reservations only, so toggling the flag can never move a field number.

## Mutants

Reverting `sourceComment` to return its argument unchanged:

```
--- FAIL: TestCommentsNoneOmitsEverySourceComment (0.00s)
    comments_test.go:95:
        error:
          unexpected length
        len(got):
          int(5)
        got:
          []string{"// Code generated by ptah schema export --to protobuf; DO NOT EDIT.", "// ptah:protobuf-export-version=1", "// ptah:content-sha256=45d34ff3...", "// Internal: sharded by tenant_id; see runbook RB-42", "// bcrypt hash, never expose"}
        want length:
          int(3)

--- FAIL: TestSchemaExportProtobufCommentsNoneOmitsSourceComments (0.03s)
    export_protobuf_test.go:140:
        error:
          unexpected length
        len(got):
          int(5)
        want length:
          int(3)
```

Stripping at output time instead, which takes Ptah's own prose with it:

```
--- FAIL: TestCommentsNoneKeepsPtahsOwnAccountOfTheContract (0.01s)
    comments_test.go:118:
        error:
          no substring match found
        container:
          "...\nmessage User {\n  reserved 1 to 3;\n  reserved email, id, password_hash;\n}\n"
        want:
          "// Removed from the source schema; retained for wire compatibility.\nmessage User {"
```

Stripping after the content digest is stamped:

```
--- FAIL: TestCommentPolicyIsNotPartOfTheCompatibilityState (0.00s)
    comments_test.go:141:
        error:
          got non-nil error
        got:
          e"output file was modified since it was generated: recorded 45d34ff34a92..., computed 410a512e42ff...; restore it from version control, or delete it to start a new compatibility history"
```

The first mutant leaves `TestCommentsNoneKeepsPtahsOwnAccountOfTheContract`'s tombstone assertion and the round-trip test green, which is why the second and third exist: each of the three assertions is separated from its plausible alternative implementation.

## Gates

Run from the repository root of the worktree, all exit 0:

`go build ./...`, `go vet ./...`, `go test ./... -count=1`, `go tool qtlint ./...`, `golangci-lint run ./...` (0 issues), `./scripts/check-test-style.sh` (175 conditional groups, 45 white-box files), `./scripts/check-public-api.sh`, `./scripts/check-public-api-snapshot.sh`.

Docs were touched, so also every `check:*` script in `docs/site/package.json` including the four selftests — `check-exit-codes` (60 reference rows), `check-core-doc-links` (7 protected core references), `check-page-health` (59 sidebar entries), `check-links` (60 pages, 575 heading anchors), `check-redirects` (30 redirects), `check-style` (100 documentation files), `check-limitations` (1 governed page), `check-responsive` (60 routes x 2 viewports, after `npm run build`) — and `node docs/site/scripts/build-feature-matrix.mjs --check` (164 rows).

## Documentation

`docs/site/src/content/docs/schema/protobuf.md` gains a "Control what the contract says" section with the before/after output above, a `--proto-comments` row in the flag table, and a rewritten Limitations bullet: the copied-comments gap is closed, and the residual — that suppression has no per-object form — closes with the reason it is permanent.

Closes #1145


denisvmedia:
## Adversarial review

Reviewed `b60d4110` against base `25cf00ff`, in a clean worktree, with a `ptah` built from the branch plus `protoc` 35.1 and `protoc-gen-go` v1.36.11.

I could not refute the central claim. The flag does what the body says, on the issue's own fixture, byte for byte, and all three mutants redden when I run them myself. Three findings below; one is a real discriminator gap, the other two are accuracy.

---

## 1. The fourth production hunk has no discriminating test

Three of the four production hunks are pinned by the new tests. The fourth is not: the `parseCommentPolicy` call added to `validateProtobufExportOptions` (`cmd/schema/export_protobuf.go:64-66`).

Deleting exactly that hunk and nothing else:

```
$ go test ./cmd/schema/... -count=1
ok  	go.5x5.cz/ptah/cmd/schema	3.590s
```

Green. The whole package, not a filtered run. But the hunk is not dead code — it changes observable behavior, which is the property its own doc comment claims ("checks every protobuf-only flag before any file system work happens", line 35). Two binaries, same arguments:

```
$ ptah schema export --to protobuf --root-dir /nonexistent-dir-xyz \
    --out ./out/x.proto --proto-package acme.inventory.v1 --proto-comments some

# branch
error: invalid --proto-comments "some": expected all or none

# hunk deleted
error: migrations directory /nonexistent-dir-xyz: stat /nonexistent-dir-xyz: no such file or directory
```

With the hunk gone the value still gets rejected — `runProtobufExport` parses it again at line 168 and returns the identical message — so every existing assertion in `TestSchemaExportProtobufRejectsInvalidConfiguration` passes. What is lost is the ordering: the flag error stops being the one the user sees.

Honestly: this hole is **not new**. The same deletion applied to the sibling flag also leaves the package green:

```
# parseNameReusePolicy removed from validateProtobufExportOptions
$ go test ./cmd/schema/... -count=1
ok  	go.5x5.cz/ptah/cmd/schema	4.098s
```

So the PR extends an existing untested pattern rather than creating one. It is still a behavior change that shipped without a mutant. A row in `TestSchemaExportProtobufRejectsInvalidConfiguration` that points `--root-dir` at a path that does not exist and asserts the flag error would pin the ordering for `--proto-comments` and close it; doing the same for the three siblings would close the pre-existing half.

## 2. The mutant-1 quote drops a failing test from the same run

Reverting `sourceComment` to return its argument unchanged reddens **three** tests, not the two the body and the commit message quote. The third:

```
--- FAIL: TestCommentsNoneKeepsPtahsOwnAccountOfTheContract (0.01s)
    comments_test.go:121:
        error:
          unexpected length
        len(got):
          int(5)
        got:
          []string{"// Code generated by ptah schema export --to protobuf; DO NOT EDIT.", "// ptah:protobuf-export-version=1", "// ptah:content-sha256=a4c9100c66c966fb35cab55f1abb8e967ec6ff3dfd6023cc42c2ef0272812531", "// Internal: retention is 30 days, enforced by cron", "// Removed from the source schema; retained for wire compatibility."}
        want length:
          int(4)
```

The prose claim is true as written — the *tombstone assertion* at `comments_test.go:118` does stay green, and the round-trip test does stay green. But a reader of the quoted output block concludes the test passed, and it did not. The quote should either carry all three FAIL blocks or say it is filtered.

## 3. Two shared pages still carry the unqualified claim

The intro sentence on `docs/site/src/content/docs/schema/protobuf.md` was correctly qualified with "unless you pass `--proto-comments none`". The same sentence lives on two pages that were not touched, and both now overstate what publishing reveals:

- `docs/site/src/content/docs/schema/export.md:205` — "Publishing any generated schema reveals the selected entity and field names, translated types, enum values, relations, and exported source comments." That page's per-target table names Protobuf explicitly two lines above, so the claim is scoped to include it.
- `docs/site/src/content/docs/schema/export.md:229` — review step 2, "Review exported table and field comments too." For the Protobuf target there is now a mechanical alternative to the manual review this step prescribes.
- `docs/api_schema_export.md:122` — "Generated descriptions also expose table and field comments." and `:141` — "Inspect every generated field, relation, and exported source comment."

`docs/api_schema_export.md:26` already delegates the `--proto-*` flags to the protobuf page, so the flag table needs nothing; it is the disclosure prose that is stale. No gate covers this — `check-style` and `check-limitations` both pass on the branch — so it will not be caught later.

---

## What I verified and could not break

**The claim itself, on the issue fixture.** Both digests reproduce exactly as claimed:

```
// ptah:content-sha256=45d34ff34a925392045e86ab1b2ec9e55fac977725e5d80936dc72b84eb47b5b   (default)
// ptah:content-sha256=410a512e42ff37d9e9e53485b91e2f7c96a17dac2aeb2e8113e497611da0d476   (--proto-comments none)
```

Second `none` run: `cmp` exit 0. Dropping the flag: `cmp` against the pre-flag file exit 0, `password_hash` still 3, no `reserved` anywhere.

**The one hop further.** `protoc --go_out --go_opt=module=…` over the default export:

```
grep -c 'runbook RB-42'            -> 1
grep -c 'bcrypt hash, never expose' -> 1
```

Over the `--proto-comments none` export: `0` and `0`, `grep` exit 1 both times.

**Suppression is complete for the surface that exists.** `goschema.Enum` (`core/goschema/types.go:516`) has no comment field and `registerEnum` never sets `enum.Comment` or `enumValue.Comment`, so `table.Comment` and `field.Comment` are the entire source-comment surface today. All three sites that carry them into the model — `buildDesired` and both returns of `buildField`, including the enum branch — go through `sourceComment`. There is no fourth constructor of `desiredMessage` or `desiredField`.

**Nothing else moves.** On a fixture with two tables, an array column, a multi-line comment and a comment on every object, `diff` between the `all` and `none` outputs (digest line excluded) is six deletions, every one of them a `//` line, and all four field numbers unchanged.

**All three mutants.** Ran each myself, each reddens with the quoted assertion:

- `sourceComment` reverted to identity: `TestCommentsNoneOmitsEverySourceComment` `len(got): int(5)` / `want length: int(3)`, plus the CLI test, plus the one in finding 2.
- stripping in the writer instead: `TestCommentsNoneKeepsPtahsOwnAccountOfTheContract` `no substring match found`, container ends `message User {\n  reserved 1 to 3;\n  reserved email, id, password_hash;\n}\n` — exactly the bare block the design note describes.
- stripping after `stampDigest`: `TestCommentPolicyIsNotPartOfTheCompatibilityState` and the CLI test both fail with `output file was modified since it was generated: recorded 45d34ff34a92..., computed 410a512e42ff...`.

**Gates, exit codes confirmed individually.** `go build ./...` 0, `go vet ./...` 0, `go tool qtlint ./...` 0 (empty output), `golangci-lint run ./...` 0 (`0 issues.`), `./scripts/check-test-style.sh` 0 (`175 conditional groups, 45 white-box files`), `./scripts/check-public-api.sh` 0, `./scripts/check-public-api-snapshot.sh` 0, every `check:*` in `docs/site/package.json` including the four selftests 0, and `node docs/site/scripts/build-feature-matrix.mjs --check` 0 (`164 rows`). `check:responsive` after `npm run build`: `60 routes x 2 viewports`.

**Two of those gates were confirmed live rather than trusted.** Adding an `if` to a test body: `check-test-style.sh` exit 1, `new test conditional violations: internal/protobufrender/comments_test.go TestCommentsNoneOmitsEverySourceComment if 1` — so the gate does reach the new file. Changing "is permanent because" to "stays that way because" in the rewritten Limitations bullet: `check-limitations` exit 1 at `protobuf.md:562` — so the gate does govern the new bullet.

**Merge probe.** The branch is 6 commits behind `origin/master`, which includes #1191 (`Keep protobuf reservation state as ranges`) touching four files in `internal/protobufrender` and the same docs page. `git merge origin/master` exits 0 with no conflict; the merged tree builds, `internal/protobufrender` and `cmd/schema` both `ok`, and every docs gate passes on it (`check-links: 60 pages, 579 heading anchors`, `build-feature-matrix.mjs --check: OK (164 rows)`). The range-form reservations do not disturb the tombstone test.

**Rules.** No banned tokens in the diff, commit message or body; no attribution; American English throughout.

---

## Deliberately not findings

- `go test ./... -count=1` exits 1 on my machine with three failures — `TestPtahLSArgumentHandling` and `TestPtahLSVersionSpellingsPrintIdenticalBytes` in `cmd/ptah-ls`, `TestSVGReportsGraphvizStderrOnFailure` in `cmd/viz`, all `context deadline exceeded` / `signal: killed`. Both packages are untouched by this PR and both are `ok` when run in isolation. Load-induced, not a regression. Flagging only because the body says every gate exits 0 and mine did not on the first attempt.
- Acceptance criterion 1 says "zero `//` lines other than the three generated header lines". With a tombstone present that is four, not three. The PR settles this deliberately, argues it, and documents it, and the issue's own fixture has no tombstone, so the criterion is literally met on the fixture it was written against. Not a defect, but the reviewer of #1145 should confirm they accept the reading.
- `error: migrations directory /nonexistent-dir-xyz: ...` for `schema export --root-dir` is wrong wording from a shared `StatDir` helper. It reproduces on the branch with a perfectly valid `--proto-comments`, so it predates this PR. Out of scope, noted because I had the output in hand.

denisvmedia:
I like this addition.
One thing I'd consider before GA is the default policy.
Source comments frequently contain internal implementation notes, operational details, or security-related context that users may not intend to publish through generated protobuf definitions and downstream generated code.
Have you considered making none the default and requiring --proto-comments=all as an explicit opt-in?
It feels like a safer default for an exported interface.

denisvmedia:
Good catch, and done — pushed as a follow-up branch rather than folded silently into this one, so the flip is reviewable on its own: `issue-1145-proto-comments-default`. Say the word and I fold it in here.

You are right about the direction. A schema comment is written for whoever reads the database and routinely carries operational notes, sharding hints or security context; a `.proto` is published to consumers and copied onward into *their* generated code. Publishing that prose should be a decision rather than something that happens because nobody passed a flag. Pre-GA is exactly when this costs nothing.

## What changed

- `--proto-comments` defaults to `none`; `all` is the opt-in.
- **Flipped in the library too**, not only the CLI: `Options.Comments`'s zero value is now `CommentsNone`. A default that differs between `protobufrender.Render` and `ptah schema export` would be a trap for precisely the caller this protects — an embedder who never set the field.
- Docs updated, including the security-boundary paragraph, which said the file reveals comments "unless you pass `--proto-comments none`" and now says it does not carry them unless you ask.

`--proto-comments all` restores the previous bytes exactly. Numbers never depended on the policy, so an export written before the flip loses its prose on the next run and nothing else — a diff, not a compatibility break.

## Two tests inverted rather than deleted

`TestCommentsAllIsTheDefaultAndCopiesSourceProse` and `TestSchemaExportProtobufCommentsNoneOmitsSourceComments` both asserted the old default by name. They are renamed and inverted so the moved default shows up in the diff instead of vanishing from it, and each now asserts **both** halves — the zero value omits, and `all` brings the prose back. Either half alone is satisfiable by an implementation that ignores the policy entirely: one that never emits comments passes the first, one that always emits them passes the second.

One fixture needed the same reasoning applied to it. `TestSchemaExportProtobufOnlyFlagsRejectedOnOtherTargets` proved that `--proto-comments` is refused on non-protobuf targets by passing `none`; that is now the default, and a flag left at its default is indistinguishable from one never passed, so the guard stopped firing and the test went green for the wrong reason. It passes `all` now.

## Gates

`go build`, `go vet`, `go test ./... -count=1`, `qtlint`, `golangci-lint` (0 issues), `check-test-style.sh`, `check-public-api.sh`, `check-public-api-snapshot.sh` — all pass. Every `check:*` in `docs/site/package.json` plus `build-feature-matrix.mjs --check` and `npm run build`.

denisvmedia:
Closing: every commit on this branch is already on `master`.

The default-flip follow-up (#1203) was branched from this one, so its head has `bd3a97c5` and `c4c6e414` as ancestors:

```
08ffa0e1 Make omitting source comments the default
c4c6e414 test(protobuf): follow the header to the foot of the file
bd3a97c5 feat(protobuf-export): add --proto-comments to omit source comments
```

Squash-merging #1203 landed all three as `272691c6`. Verified on `master` rather than inferred from the ancestry: `--proto-comments` and `parseCommentPolicy` are in `cmd/schema/export_protobuf.go`, both `CommentsAll` and `CommentsNone` are in `internal/protobufrender/protobufrender.go`, and `internal/protobufrender/comments_test.go` carries its four tests.

That leaves nothing here to merge — the branch diff against `master` is now only the *older* shape of those same lines, which is why it reads as conflicting. #1145 is closed.

PR #1191, opened 2026-08-06
Keep protobuf reservation state as ranges

### LINKED ISSUE #1147: Keep protobuf reservation state as ranges instead of expanding it to individual numbers
Split out of #923, which measured the eleven documented protobuf-export limitations and found this one resolvable and untracked. It is the smallest of the five and is self-contained inside `internal/protobufrender`.

## What reproduces today

Measured on master `feb6af19` with a `ptah` built from that commit, using two hand-built baselines that are identical except for the width of one `reserved` range, each re-stamped with a valid `ptah:content-sha256`.

| Previous baseline | Individual numbers | Result |
| --- | --- | --- |
| `reserved 20000 to 20005;` | 6 | exit **0**, range preserved (positive control) |
| `reserved 20000 to 1030000;` | 1,010,001 | exit **0**, rewritten with `reserved 20000 to 1030000;` intact |
| `reserved 20000 to 2100000;` | 2,080,001 | exit **2**, `output file is not valid protobuf: message "Account": reserved ranges expand to more than 1048576 numbers; refusing to truncate compatibility state` |

Negative controls on the same fixture: a file Ptah did not write is refused with `output file was not generated by ptah`, and the 6-number baseline widened by one number without re-stamping its digest is refused with `output file was modified since it was generated: recorded …, computed …`. So the pair above measures the range width and nothing else.

## Cause

`maxReservedNumbers = 1 << 20` in `internal/protobufrender/model.go`, and the loader in `internal/protobufrender/previous.go` expands every range into a `[]int32` of individual numbers, checking the remaining budget at `previous.go:206`. The bound exists because the representation is a list of numbers, not because protobuf limits reservations — the legal field-number space runs to 536,870,911.

Failing closed is the right default for the current representation: truncating the list would let a later run reallocate a number the published wire contract still reserves. Representing reservations as ranges removes the cap without weakening that.

## Acceptance

1. The `reserved 20000 to 2100000;` baseline exits **0**, and the rewritten file still contains `reserved 20000 to 2100000;` — a range, not an expansion.
2. `reserved 20000 to 536870911;` (the whole legal space above the implementation range) also exits 0 and survives the round trip.
3. A live field number falling inside a reserved range is still refused, with the diagnostic it produces today.
4. Loading is linear in the number of ranges, not in the count of numbers they cover.

Today step 1 exits 2 with `reserved ranges expand to more than 1048576 numbers`.

Documentation to update once this lands: the Limitations bullet about the 1,048,576 cap on `docs/site/src/content/docs/schema/protobuf.md`, which links here.

denisvmedia:
Closes #1147.

`internal/protobufrender` read a previous export's `reserved` statements by expanding every range into individual `int32` entries. That representation forced a cap — `maxReservedNumbers = 1 << 20` — and a wider range was refused. The legal field-number space runs to 536,870,911, so a published contract may reserve more than the loader accepted.

`reservations` now holds `[]numberRange` instead of `[]int32`. `addRange` stores the range as the file spells it, `reservations.sort` normalizes by sorting and merging overlapping or adjacent runs, and `previousType.used` (every number ever held) becomes `highestUsed`, with `nextNumber` taking that high-water mark instead of the list.

## Measured

CLI, one `ptah` built from this branch, one models directory (`accounts(id BIGINT, email TEXT)`), three hand-built baselines identical except for the width of one `reserved` range, each re-stamped with a valid `ptah:content-sha256`:

| Previous baseline | Individual numbers | Result |
| --- | --- | --- |
| `reserved 20000 to 20005;` | 6 | exit **0** |
| `reserved 20000 to 2100000;` | 2,080,001 | exit **0**, file byte-identical, digest `50ab9778…` unchanged |
| `reserved 20000 to 536870911;` | 536,850,912 | exit **0**, file byte-identical, digest `6ea7e3d9…` unchanged |

Acceptance 3 is unchanged behavior and is now pinned by a test: a baseline whose live field sits inside a reserved range is still refused with

```
output file is not valid protobuf: acme/inventory/v1/schema.proto:9:14: message acme.inventory.v1.Thing: field id is using tag 5 which is in reserved range 1 to 2100000
```

Acceptance 4 is measured, not asserted from the shape of the code: loading one 40,000,000-number range allocates 119,344 bytes for the whole export, against a 64 MiB budget in the test.

## Mutants

**M1 — restore the retired cap in `addRange`** (revert of the behavior change). Five tests redden; the 6-number and 1,010,001-number rows and the digest control stay green, so the pair measures the range width and nothing else:

```
--- FAIL: TestReservedRangesSurviveRegardlessOfWidth (0.00s)
    --- FAIL: TestReservedRangesSurviveRegardlessOfWidth/2080001_numbers,_past_the_retired_cap (0.00s)
            error:
              got non-nil error
            got:
              e`output file is not valid protobuf: message "Thing": reserved ranges expand to more than 1048576 numbers; refusing to truncate compatibility state`
    --- FAIL: TestReservedRangesSurviveRegardlessOfWidth/536850912_numbers,_the_whole_legal_space_above_the_implementation_range (0.00s)
--- FAIL: TestReservedEnumRangesKeepTheirInclusiveEnd (0.00s)
--- FAIL: TestReservedRangeAllocationClearsTheWholeRange (0.00s)
--- FAIL: TestReservedRangeLoadingDoesNotExpandTheRange (0.00s)
FAIL	go.5x5.cz/ptah/internal/protobufrender	0.380s
```

The same mutant built as a binary reproduces the issue at the CLI, and the narrow baseline separates it from a binary that is simply broken:

```
$ ptah-m1 schema export --to protobuf … --out …/wide/schema.proto
error: output file is not valid protobuf: message "Account": reserved ranges expand to more than 1048576 numbers; refusing to truncate compatibility state
exit=2

$ ptah-m1 schema export --to protobuf … --out …/narrow/schema.proto   # reserved 20000 to 20005
Exported 1 message(s), 2 field(s), 0 enum(s)
exit=0
```

**M2 — drop the cap but keep the expansion** (`for n := start; n <= end; n++ { res.addRange(int32(n), int32(n)) }`). This is the mutant that matters, because `TestReservedRangesSurviveRegardlessOfWidth/2080001_numbers…` **passes** under it — the emitted text is identical, so width alone cannot tell a range from an expansion. Only the allocation assertion separates them:

```
--- FAIL: TestReservedRangeLoadingDoesNotExpandTheRange (1.15s)
    reservedrange_test.go:176:
        error:
          value is not true
        comment:
          export allocated 2519306952 bytes loading one 40000000-number reserved range
        got:
          bool(false)
FAIL	go.5x5.cz/ptah/internal/protobufrender	1.596s
```

The fixture width is 40,000,000 rather than the whole space so an expanding implementation fails the assertion instead of exhausting the machine before it can be measured.

## Other changes

- The rejection case `oversized reserved state` is deleted from `previous_test.go`: it pinned the refusal this PR removes. The `unsupported negative reserved state` case stays, and `reserved -1;` is still refused with `reserved range -1 to -1 is outside Ptah's supported numbering space`.
- `normalizeRanges` merges adjacent runs, so a number retired by this run that sits directly above a range the previous file carried extends that range (`reserved 2 to 7;` + retired 8 → `reserved 2 to 8;`) rather than being written beside it. Pinned by `TestReservedRangesMergeWithNewlyRetiredNeighbors`.
- Docs: the Limitations bullet about the 1,048,576 cap is removed, and the editions section states that reservations round-trip as ranges at any legal width.

## Gates

From the repo root of the worktree, exit code confirmed for each:

- `go build ./...` 0
- `go vet ./...` 0
- `go test ./... -count=1` 0 (176 packages ok)
- `go tool qtlint ./...` 0
- `golangci-lint run ./...` 0 issues
- `./scripts/check-test-style.sh` OK (175 conditional groups, 45 white-box files)
- `./scripts/check-public-api.sh` 0
- `./scripts/check-public-api-snapshot.sh` 0

Docs gates, run because `docs/site` is touched: `check:exit-codes` (+selftest), `check:core-doc-links`, `check:page-health`, `check:links` (+selftest), `check:redirects` (+selftest), `check:style` (+selftest), `check:limitations` OK (1 governed page) and its selftest OK (7 assertions), `versions:selftest`, and `node docs/site/scripts/build-feature-matrix.mjs --check` OK (164 rows).

`check:responsive` did **not** run: it reports `SKIPPED (playwright is not installed…)` and exits 0 without measuring anything. Installing playwright here would write into a `node_modules` outside this worktree. That check reddens on table row count; this diff adds no table and removes no table row.

### DIFF
--- a/docs/site/src/content/docs/schema/protobuf.md
+++ b/docs/site/src/content/docs/schema/protobuf.md
@@ -202,6 +202,12 @@ wire-relevant feature default differs between the two editions, and Edition
 Under editions, `reserved` names are bare identifiers (`reserved nickname;`),
 the inverse of the quoted form `proto2` and `proto3` use.
 
+Reservations are read back and written out as ranges, never as the numbers a
+range covers. A range is accepted at any width the field-number space allows,
+so a previous file carrying `reserved 20000 to 536870911;` round-trips
+unchanged and contiguous retired numbers are written as `reserved 2 to 8;`
+rather than one entry each.
+
 ## Type mapping
 
 The lookup is dialect-agnostic, so the Postgres and MySQL spellings Ptah emits
@@ -511,10 +517,6 @@ permanent, so nothing on this list is an unowned gap.
   next run read it as a type that left the schema: a previous file carrying
   `message AuditMarker {}` for a table still present in the source is refused
   with `types removed from the source schema: AuditMarker`.
-- A previous type whose `reserved` ranges expand past 1,048,576 individual
-  numbers is refused instead of being loaded partially. Partial loading could
-  reallocate a number that remains reserved in the published wire contract.
-  Tracked by [#1147](https://github.com/stokaro/ptah/issues/1147).
 - The three header comment lines are copied into whatever `protoc-gen-go`
   generates, so the content digest appears at the top of the `.pb.go` files
   too. It changes only when the schema changes. Tracked by
--- a/internal/protobufrender/model.go
+++ b/internal/protobufrender/model.go
@@ -1,6 +1,7 @@
 package protobufrender
 
 import (
+	"cmp"
 	"slices"
 	"sort"
 )
@@ -11,14 +12,6 @@ const (
 	reservedRangeStart = 19000
 	reservedRangeEnd   = 19999
 	maxFieldNumber     = 536870911
-
-	// maxReservedNumbers bounds how many individual numbers one type's
-	// reservations may expand to when a previous export is read back. The
-	// exporter only ever writes ranges it built from individual numbers, so
-	// reaching this cap means the input is not something this exporter produced;
-	// the bound keeps a "reserved 1 to max" style range from allocating two
-	// billion entries before the digest gate can reject the file.
-	maxReservedNumbers = 1 << 20
 )
 
 // field is one message field in the generated file.
@@ -40,16 +33,39 @@ type enumValue struct {
 // reservations are the numbers and names a type may never reuse. They are what
 // buys the WIRE_JSON compatibility guarantee: buf reports a removed field as
 // breaking unless both its number and its name are reserved.
+//
+// Numbers are held as ranges, the same shape a .proto spells them in. A single
+// legal range covers the whole 536,870,911-number field space, so expanding one
+// into individual numbers would make both loading and storage proportional to
+// the width of the reservation rather than to the number of ranges.
 type reservations struct {
-	Numbers []int32
-	Names   []string
+	Ranges []numberRange
+	Names  []string
+}
+
+// hasNumbers reports whether the type reserves any number at all.
+func (r *reservations) hasNumbers() bool {
+	return len(r.Ranges) > 0
+}
+
+// highest returns the largest reserved number, or 0 when nothing is reserved.
+func (r *reservations) highest() int32 {
+	var highest int32
+	for _, rng := range r.Ranges {
+		highest = max(highest, rng.End)
+	}
+	return highest
 }
 
 func (r *reservations) addNumber(n int32) {
-	if slices.Contains(r.Numbers, n) {
-		return
-	}
-	r.Numbers = append(r.Numbers, n)
+	r.addRange(n, n)
+}
+
+// addRange records [start, end] inclusive. Overlaps and duplicates are folded
+// away by normalizeRanges before the model is written, so callers may add the
+// same number twice.
+func (r *reservations) addRange(start, end int32) {
+	r.Ranges = append(r.Ranges, numberRange{Start: start, End: end})
 }
 
 func (r *reservations) addName(name string) {
@@ -63,16 +79,12 @@ func (r *reservations) hasName(name string) bool {
 	return slices.Contains(r.Names, name)
 }
 
-func (r *reservations) count() int {
-	return len(r.Numbers)
-}
-
 func (r *reservations) dropName(name string) {
 	r.Names = slices.DeleteFunc(r.Names, func(existing string) bool { return existing == name })
 }
 
 func (r *reservations) sort() {
-	slices.Sort(r.Numbers)
+	r.Ranges = normalizeRanges(r.Ranges)
 	slices.Sort(r.Names)
 }
 
@@ -133,40 +145,42 @@ type numberRange struct {
 	End   int32
 }
 
-// collapseRanges folds a sorted number list into contiguous runs so the writer
-// can emit "reserved 1 to 7;" instead of seven separate entries.
-func collapseRanges(numbers []int32) []numberRange {
-	if len(numbers) == 0 {
+// normalizeRanges sorts ranges ascending and folds overlapping or adjacent runs
+// into one, so two separately retired neighbors are written as "reserved 2 to
+// 3;" rather than "reserved 2, 3;". Merging is a correctness requirement and not
+// only formatting: protoc refuses a file whose reserved ranges overlap, and the
+// same number can reach the model from both a removed field and a range the
+// previous file already carried.
+func normalizeRanges(in []numberRange) []numberRange {
+	if len(in) == 0 {
 		return nil
 	}
-	sorted := slices.Clone(numbers)
-	slices.Sort(sorted)
-
-	ranges := []numberRange{{Start: sorted[0], End: sorted[0]}}
-	for _, n := range sorted[1:] {
-		last := &ranges[len(ranges)-1]
-		switch n {
-		case last.End:
-			// duplicate, already covered
-		case last.End + 1:
-			last.End = n
-		default:
-			ranges = append(ranges, numberRange{Start: n, End: n})
+	sorted := slices.Clone(in)
+	slices.SortFunc(sorted, func(a, b numberRange) int {
+		if a.Start != b.Start {
+			return cmp.Compare(a.Start, b.Start)
 		}
+		return cmp.Compare(a.End, b.End)
+	})
+
+	out := []numberRange{sorted[0]}
+	for _, rng := range sorted[1:] {
+		last := &out[len(out)-1]
+		// Compared in int64 so an end of the maximum field number cannot
+		// overflow the adjacency test.
+		if int64(rng.Start) <= int64(last.End)+1 {
+			last.End = max(last.End, rng.End)
+			continue
+		}
+		out = append(out, rng)
 	}
-	return ranges
+	return out
 }
 
-// nextNumber returns the next free field number for a type, given every number
-// it has ever used. Gaps are never filled and the protobuf implementation range
-// is skipped, so a number retired by a removed column can never come back.
-func nextNumber(used []int32) (int32, bool) {
-	var highest int32
-	for _, n := range used {
-		if n > highest {
-			highest = n
-		}
-	}
+// nextNumber returns the next free field number for a type, given the highest
+// number it has ever used. Gaps are never filled and the protobuf implementation
+// range is skipped, so a number retired by a removed column can never come back.
+func nextNumber(highest int32) (int32, bool) {
 	next := highest + 1
 	if next >= reservedRangeStart && next <= reservedRangeEnd {
 		next = reservedRangeEnd + 1
--- a/internal/protobufrender/previous.go
+++ b/internal/protobufrender/previous.go
@@ -55,15 +55,16 @@ type previousType struct {
 	Reserved reservations
 }
 
-// used returns every number the type has ever held, live or reserved, which is
-// what new allocations must exceed.
-func (p previousType) used() []int32 {
-	out := make([]int32, 0, len(p.Numbers)+len(p.Reserved.Numbers))
+// highestUsed returns the largest number the type has ever held, live or
+// reserved, which is what new allocations must exceed. It is a single number
+// rather than the list of every number ever used because a reserved range may
+// legally cover the whole field-number space.
+func (p previousType) highestUsed() int32 {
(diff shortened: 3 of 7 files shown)


denisvmedia:
## Adversarial review

Re-measured from a clean worktree at `e01e56b`, plus a second worktree at the branch parent `25cf00f` for a before/after operand. The behavior change and both mutants reproduce. Four things do not hold as written; one of them is a real gap against the issue's own framing.

---

### What I reproduced

**M1 (restore the retired cap).** Reintroduced the `1<<20` budget check inside `addRange` on top of the new representation. Five failures, the same set the body lists, and the 6-number and 1,010,001-number rows plus `TestReservedRangesStillGuardTheirBaseline` all stay green:

```
--- FAIL: TestReservedRangesSurviveRegardlessOfWidth/2080001_numbers,_past_the_retired_cap
    got: e`output file is not valid protobuf: message "Thing": reserved ranges expand to more
         than 1048576 numbers; refusing to truncate compatibility state`
--- FAIL: TestReservedRangesSurviveRegardlessOfWidth/536850912_numbers,_the_whole_legal_space...
--- FAIL: TestReservedEnumRangesKeepTheirInclusiveEnd
--- FAIL: TestReservedRangeAllocationClearsTheWholeRange
--- FAIL: TestReservedRangeLoadingDoesNotExpandTheRange
FAIL	go.5x5.cz/ptah/internal/protobufrender	5.054s
```

**M2 (no cap, keep the expansion).** `TestReservedRangesSurviveRegardlessOfWidth/2080001…` passes under it, as claimed; only the allocation assertion separates them:

```
--- FAIL: TestReservedRangeLoadingDoesNotExpandTheRange (2.54s)
    comment: export allocated 2519246448 bytes loading one 40000000-number reserved range
```

**The CLI table.** I rebuilt the three baselines from scratch with an independent digest stamper and got the same content hashes the body quotes — `50ab9778…` for `20000 to 2100000`, `6ea7e3d9…` for `20000 to 536870911`. Ran the same models directory through a binary built from the parent commit and one built from this branch:

```
ptah-base  narrow  exit=0  byte-identical
ptah-base  wide    exit=2  error: ... reserved ranges expand to more than 1048576 numbers ...
ptah-base  whole   exit=2  error: ... reserved ranges expand to more than 1048576 numbers ...
ptah-head  narrow  exit=0  byte-identical
ptah-head  wide    exit=0  byte-identical
ptah-head  whole   exit=0  byte-identical
```

Acceptance 1 and 2 hold. Allocation on this machine is 119,024 bytes against the 64 MiB budget (body says 119,344). `go test ./... -count=1` exits 0, 176 packages. `go build`, `go vet`, `go tool qtlint`, `golangci-lint` (0 issues), `check-test-style.sh` (175 conditional groups, 45 white-box files), `check-public-api.sh`, `check-public-api-snapshot.sh` all exit 0.

---

### 1. The new docs sentence is false for enums, and it replaced the only place the bound was written down

The paragraph added to the editions section says:

> A range is accepted at any width the field-number space allows

An enum's number space is the full `int32` range, not 536,870,911. `reserved 20000 to max;` — the spelling the protobuf language guide uses for an enum reservation — is still refused. Measured with a `ptah` built from this branch against an enum-bearing models directory and three hand-stamped baselines that differ only in the width of one enum reservation:

```
ptah-head  enum-in-space    exit=0    reserved 20000 to 536870911;
ptah-head  enum-past-space  exit=2    error: output file is not valid protobuf: enum "AccountStatus":
                                      reserved range 20000 to 600000000 is outside Ptah's supported numbering space
ptah-head  enum-to-max      exit=2    error: output file is not valid protobuf: enum "AccountStatus":
                                      reserved range 20000 to 2147483647 is outside Ptah's supported numbering space
```

Those two files compile: the diagnostic is Ptah's own, raised in `enumState` after `compileProto` already accepted the input, so protobuf considers the reservation ordinary and Ptah refuses it. `ptah-base` produces the same two refusals, so this is not a regression — but it is the same shape as the gap #1147 was filed to close ("a published contract may legitimately reserve more than the loader accepted"), and this PR deletes the one Limitations bullet that told a reader a bound existed. The replacement sentence lives in the editions section, so `check:limitations` cannot govern it: the remaining refusal is now documented nowhere and actively contradicted.

Either narrow the sentence to the message field-number space and keep a Limitations bullet for the enum half, or carry the enum case.

Related: the same sentence says a file carrying `reserved 20000 to 536870911;` "round-trips unchanged". A message baseline spelled `reserved 20000 to max;` comes back rewritten as `reserved 20000 to 536870911;` (exit 0, re-stamped). Fine behavior, but not "unchanged".

### 2. `check:responsive` does run, and passes — the gate note is wrong on both claims

> `check:responsive` did **not** run: it reports `SKIPPED (playwright is not installed…)` and exits 0 without measuring anything. Installing playwright here would write into a `node_modules` outside this worktree.

Measured in a fresh worktree at `e01e56b`:

```
### check:responsive -> exit 1
check-responsive.mjs: dist not found; run "npm run build" first.

### check:responsive:selftest -> exit 0
check-responsive.mjs: SKIPPED (playwright is not installed; run "npm i -D playwright && npx playwright install chromium")
```

The `SKIPPED`/exit-0 line quoted in the body is the **selftest**, a different script. `check:responsive` itself exits **1** on a clean tree, because the `dist` check runs before the browser check (`scripts/check-responsive.mjs:403` vs `:410`).

`playwright` is a declared devDependency (`"playwright": "^1.62.1"`) in `docs/site/package.json`, and `docs/site/.gitignore` lists `node_modules/` — so it installs to `docs/site/node_modules/playwright`, inside this worktree and untracked, not "outside this worktree". After running the two commands the script asks for:

```
### npm ci -> exit 0
### npm run build -> exit 0        [build] 61 page(s) built in 28.68s
### check:responsive (after build) -> exit 0
check-responsive.mjs: OK (60 routes x 2 viewports, cells within 8 lines)
```

The outcome is benign — it passes — but the body reports an unmeasured skip as a measured fact for a gate this diff is required to run.

### 3. The half of the `addRange` guard that survives is pinned by nothing

The PR rewrote `addRange` and deleted one of its two error-path cases. What is left is `start < 0 || end > maxFieldNumber`, and only the `start < 0` half is asserted anywhere in the tree:

```
internal/protobufrender/previous.go:211:      return fmt.Errorf("reserved range %d to %d is outside Ptah's supported numbering space", start, end)
internal/protobufrender/previous_test.go:79:  "reserved range -1 to -1 is outside Ptah's supported numbering space",
```

Mutant: delete `|| end > maxFieldNumber`.

```
### M3 (upper-bound guard deleted) -> exit 0
ok  	go.5x5.cz/ptah/internal/protobufrender	0.370s
```

The whole package stays green. Built as a binary, the enum baseline head refuses is silently accepted:

```
ptah-m3 enum-to-max exit=0
    enum AccountStatus {
      ...
      reserved 20000 to 2147483647;
    }
```

`wideRangeCases` walks the width axis right up to `20000 to 536870911` and stops there, so the operand `maxFieldNumber` never gets separated from "no bound at all" or from an off-by-one. One row — an enum reserving `536870912` — closes it. Worth doing here specifically because the `//nolint:gosec` directive on the next line rests on that branch ("Both conversions are guarded by the supported-numbering-space check").

### 4. "Other changes": the adjacency merge is preserved behavior, not a change

> `normalizeRanges` merges adjacent runs, so a number retired by this run that sits directly above a range the previous file carried extends that range (`reserved 2 to 7;` + retired 8 → `reserved 2 to 8;`) rather than being written beside it.

Same baseline (`reserved 2 to 7;` with `string email = 8;`), same models directory, both binaries:

```
== ptah-base exit=0
     message Account {
       int64 id = 1;

       reserved 2 to 8;
       reserved email, sku;
     }
== ptah-head exit=0
     message Account {
       int64 id = 1;

       reserved 2 to 8;
       reserved email, sku;
     }
```

`collapseRanges` already folded `last.End + 1`, so the emitted text does not move for anyone. The test is a good regression pin for the new representation — dropping adjacency from the merge reddens it along with three pre-existing tests, and dropping the sort reddens `TestNumberingCollapsesContiguousReservedRuns` with exactly the silent drop you would want caught (`reserved 3 to 4` where `2 to 4` is correct) — but listing it under "Other changes" tells a reviewer the output moved. It did not.

---

### Things I checked and found clean

- `sortForOutput` is the single call site for `reservations.sort`, and it runs immediately before `render`, so the writer emitting `res.Ranges` verbatim has no unsorted path.
- Merging is exact union in every case (overlap and adjacency), so no number is invented into a reservation; `normalizeRanges` compares in `int64`, so an end at `maxFieldNumber` cannot overflow the adjacency test.
- `highestUsed` is equivalent to `max(used())` on both branches, including the dropped `used = append(used, 0)` for the synthesized enum zero value.
- A wide range plus a new column allocates above the range end (`reserved 20000 to 2100000` + new column → `string nickname = 2100001;`), and the saturated case fails closed with `message "Account" has exhausted the protobuf field number space` rather than filling a gap.
- No remaining reference to `maxReservedNumbers` or the cap message anywhere outside one test comment.
- All other docs gates exit 0: `check:exit-codes` (+selftest), `check:core-doc-links`, `check:page-health`, `check:links` (+selftest), `check:redirects` (+selftest), `check:style` (+selftest), `check:limitations` (1 governed page) and its selftest (7 assertions), `versions:selftest`, and `build-feature-matrix.mjs --check` (164 rows).

### Verdict

The representation change is correct and the mutant evidence for it is sound. Blocking on (1) — the docs assert a bound that measurement contradicts, and the deleted bullet was the only record of it. (2) should be corrected before merge because the gate note is not what the scripts print. (3) is a one-row fix I would take with it.

Issue #1196, opened 2026-08-06
Recorded revision error text and partial_hashes diverge from the pinned binary on a failed migration


denisvmedia:
Found while running the conformance `migrate-runtime` tier. **Pre-existing on `ptah-atlas-conformance` `main`**, and currently the only non-OK observation in that tier — measured on a clean detached `origin/main` worktree at `3148945` with nothing else applied, so it is not attributable to any open conformance PR.

## What diverges

Scenario `sqlite / per-file-txmode / revision-bookkeeping`, comparing the revision row written after a migration whose second statement fails (`INSERT INTO txmode_missing` against a table that does not exist).

**1. `error` carries a driver prefix and repeats the statement.**

```
Ptah   failed to execute migration SQL: SQL logic error: no such table: txmode_missing (1)
       SQL: INSERT INTO txmode_missing (id) VALUES (1)
       SQL: INSERT INTO txmode_missing (id) VALUES (1)

Atlas  no such table: txmode_missing
```

Two things are wrong independently of the wording choice: the wrapping prefix, and the `SQL:` line appearing **twice** for one statement.

**2. `error_stmt` loses the terminating semicolon.**

```
Ptah   INSERT INTO txmode_missing (id) VALUES (1)
Atlas  INSERT INTO txmode_missing (id) VALUES (1);
```

**3. `partial_hashes` is `null` where the binary records a hash array.**

```
Ptah   null
Atlas  ["h1:YG6I9i3XJ94/l+rSfna7yLoUh3Z54G71zVeJKNp/IVI="]
```

This is the one with consequences beyond cosmetics: `partial_hashes` is how a partially applied migration is resumed safely. A `null` there means a resume cannot verify which statements already ran.

**4. Some directive combinations write an extra revision row.** Under `global-file-directive-absent`, `global-file-directive-file`, `global-all-directive-absent` and `global-none-directive-file`, Ptah writes one row where the binary writes none (`count=1, Atlas=0`).

Seven of seven body-execution cells differ.

## Direction

Divergence 3 and 4 are the substantive ones; 1 and 2 are text shape. Note that 4 is Ptah recording *more* than the binary, so it does not breach "never exit 0 where the binary exits 1" — but a revision row the binary does not write is a drop-in difference a resuming tool will act on.

## Definition of done

[...]

PR #1197, opened 2026-08-06
Scope migrate lint to the schema the dev URL covers (#1074 S1)

denisvmedia:
Closes item **S1** of #1074 — the one currently-failing DoD line ("the two differential fixtures match without waivers"). Fixture 1 (rename) already matched via #1120; fixture 2, run exactly as the issue body writes it, did not.

## The diagnosis, confirmed

The dev database a lint run replays into is what the run compares against, so it is also what decides which objects are under review. With `?search_path=public`, the pinned community binary's snapshot covers only `public`, so a table the migration creates and drops in schema `app` was never in the before-state and dropping it is not a covered change. Ptah analyzed every object the SQL named, regardless of schema.

The axis is isolated by the single-target control: `DROP TABLE app."Users";` alone splits the same way, so this is not the multi-target defect closed by #1112. The `DROP COLUMN` form splits the same way too, so it is not the `DROP TABLE` grammar either.

## Before / after

All measured on PostgreSQL 16 against the pinned community binary v1.3.0 at `/Users/buster/Work/denis/ptah-atlas-conformance/bin/atlas`, `migrate lint --dir file://… --dev-url … --latest 1`. "identical" = byte-for-byte with only the two elapsed-duration lines differing. Exit codes captured separately from the report.

| # | fixture (earlier file creates, next file changes) | dev URL scope | pinned binary | ptah before | ptah after |
|---|---|---|---|---|---|
| 1 | `DROP TABLE app."Users", app.audit_log;` — the issue's §2 fixture | `search_path=public` | rc=0, `-- no diagnostics found` / `-- 1 version ok` | rc=1, 2×DS102, `2 schema changes`, `2 diagnostics` | **rc=0, identical** |
| 2 | `DROP TABLE app."Users";` — single-target form of the same defect | `search_path=public` | rc=0, silent | rc=1, 1×DS102, `1 schema change` | **rc=0, identical** |
| 3 | `ALTER TABLE app.users DROP COLUMN nick;` | `search_path=public` | rc=0, silent | rc=1, 1×DS103, `1 schema change` | **rc=0, identical** |
| 4 | `CREATE SCHEMA app2; CREATE TABLE app2.t (…); CREATE TABLE keep (…);` | `search_path=public` | rc=0, `-- 1 schema change` | rc=0, `-- 3 schema changes` | **rc=0, identical** |
| 5 | `DROP TABLE users, app.audit_log;` — one statement, both schemas | `search_path=public` | rc=1, 1×DS102 (`users`), `1 schema change` | rc=1, 2×DS102, `2 schema changes` | **rc=1, identical** |
| 6 | **control** `DROP TABLE users, audit_log;` (public) | `search_path=public` | rc=1, 2×DS102, `2 diagnostics` | identical | **identical, unchanged** |
| 7 | **control** `DROP TABLE users;` (public) | `search_path=public` | rc=1, 1×DS102 | identical | **identical, unchanged** |
| 8 | **control** mid/zeta/alpha created, `DROP TABLE zeta, alpha, mid;` | `search_path=public` | rc=1, 3×DS102 (`alpha`,`mid`,`zeta`) | identical | **identical, unchanged** |
| 9 | **control** `DROP TABLE public.users, public.audit_log;` fully qualified | `search_path=public` | rc=1, 2×DS102 | identical | **identical, unchanged** |
| 10 | **control** the row-1 fixture through a dev URL with **no** `search_path` | whole database | rc=1, 2×DS102 | rc=1, identical | **identical, unchanged** |
| 11 | `INSERT INTO t (id) VALUES (1);` only | `search_path=public` | rc=0, **no** schema-change line | rc=0, `-- 0 schema changes` | **rc=0, identical** |

Rows 6–8 are the three controls the issue names; row 9 and row 10 are added because they are the only rows that redden under the inverse mutants below. Row 11 is a pre-existing divergence the scope change made load-bearing: with row 1 filtered to zero changes, ptah would otherwise have printed `-- 0 schema changes` where the pinned binary prints nothing.

Two more measurements pin the boundary's edges rather than its interior:

- `search_path=public,app` is not a search-path list to the pinned binary but a single schema NAME — it refuses the run with `taking database snapshot: postgres: schema "public,app" was not found`. There is no scoping behavior to match, so a comma-carrying value scopes nothing here and every object stays under review.
- A dev URL with no `search_path` reviews the whole database (row 10). This is why the filter is conditional on the URL, not on a default of `public`.

## What changed

- `migration/lint` gains `Options.SchemaScope`. It filters findings by their subjects (a column is measured by its table, never by its own name) and semantic schema changes by the object's qualifier. `CREATE SCHEMA` is measured against the reviewed schema by its own name, and `ALTER TABLE` by its table.
- `internal/atlasurl.SchemaScope` reads the boundary off the dev URL. Only the PostgreSQL family, only a single-schema `search_path`.
- `internal/migrationlintreport` wires it for **both** surfaces from the same URL, so `ptah-compat migrate lint` and native `ptah migrations lint` cannot disagree about what is under review — which is the complaint the issue's last comment ends on.
- `internal/atlasreport` suppresses the schema-change line at zero.
- `docs/site/.../atlas/migrate-commands.md` gains a "Which objects are under review" section recording the boundary.

The filter fails toward reporting at every point where the boundary is unknown: an unparsable reference, an unqualified reference, an object the node grammar exposes no name for, a dialect other than PostgreSQL, and — most importantly — any finding that carries no subjects at all. Statement rules report the statement rather than the objects in it, so `TRUNCATE TABLE app.users` and `DROP FUNCTION app.recalc()` stay reported under `search_path=public`; there is nothing to measure a scope against and a hazard must not be silenced on an unestablished boundary. Unit tests pin both.

## Mutants

**Mutant 1 — revert the scope filter** (`newSchemaScope` always returns the unrestricted scope). Exactly the five scope-dependent cells go red; the four controls stay green, which is the whole reason the controls need inverse mutants.

```
--- FAIL: TestAtlasMigrateLintSchemaScopeE2E/two_targets_outside_the_reviewed_schema
    error: got non-nil error
    got: e"exit status 1"
--- FAIL: TestAtlasMigrateLintSchemaScopeE2E/one_target_outside_the_reviewed_schema
--- FAIL: TestAtlasMigrateLintSchemaScopeE2E/a_column_dropped_outside_the_reviewed_schema
--- FAIL: TestAtlasMigrateLintSchemaScopeE2E/objects_created_outside_the_reviewed_schema_count_no_change
    line diff (-got +want):
      - "  -- 3 schema changes\n",
      + "  -- 1 schema change\n",
--- FAIL: TestAtlasMigrateLintSchemaScopeE2E/one_statement_across_both_schemas_reports_only_the_in-scope_table
```

and at unit level:

```
--- FAIL: TestAnalyzeFS_SchemaScopeFiltersFindings/multi-target_drop_outside_the_scope_is_not_reported
    diff (-got +want):
        []string{
      - 	"DS101",
      - 	"DS101",
        }
```

**Inverse mutant A — an absent `search_path` scopes to `public` anyway.** Reddens exactly the control that proves the filter depends on the dev URL:

```
--- FAIL: TestAtlasMigrateLintSchemaScopeE2E/control:_a_dev_URL_naming_no_schema_reviews_the_whole_database
    error: got nil error but want non-nil
    regexp: "exit status 1"
```

**Inverse mutant B — negate the schema comparison** (`!strings.EqualFold`). Reddens the fully-qualified `public` control. The two unqualified `public` controls stay green here, because an unqualified reference never reaches the comparison — which is exactly why row 9 had to exist:

```
--- FAIL: TestAtlasMigrateLintSchemaScopeE2E/control:_targets_qualified_with_the_reviewed_schema_itself
    error: got nil error but want non-nil
    regexp: "exit status 1"
```

**Inverse mutant C — treat an unqualified reference as out of scope.** Reddens the unqualified `public` controls that B could not reach:

```
--- FAIL: TestAtlasMigrateLintSchemaScopeE2E/control:_two_targets_inside_the_reviewed_schema
    error: got nil error but want non-nil
    regexp: "exit status 1"
--- FAIL: TestAtlasMigrateLintSchemaScopeE2E/control:_one_target_inside_the_reviewed_schema
--- FAIL: TestAtlasMigrateLintSchemaScopeE2E/objects_created_outside_the_reviewed_schema_count_no_change
--- FAIL: TestAtlasMigrateLintSchemaScopeE2E/one_statement_across_both_schemas_reports_only_the_in-scope_table
```

**Mutant 2 — restore the unconditional schema-change line** (`model.changes >= 0`):

```
--- FAIL: TestWriteMigrateLintText_RendersAtlasDiagnostics/a_version_expressing_no_schema_change_prints_no_change_line
    line diff (-got +want):
      - "  -- 0 schema changes\n",
--- FAIL: TestAtlasMigrateLintSchemaScopeE2E/two_targets_outside_the_reviewed_schema
--- FAIL: TestAtlasMigrateLintSchemaScopeE2E/one_target_outside_the_reviewed_schema
--- FAIL: TestAtlasMigrateLintSchemaScopeE2E/a_column_dropped_outside_the_reviewed_schema
```

## Parity rule

This moves ptah from exit 1 to exit 0 on rows 1–3, which is the direction that needs justification. It is justified per cell by the pinned binary's own measured exit 0, and every other cell is verified unchanged — including the two in-scope controls the issue names, the ordering fixture, and the no-`search_path` control that keeps the strict behavior reachable. No cell moved from exit 1 to exit 0 anywhere else, and no diagnostic ptah reports today was silenced except the ones the boundary explicitly excludes.

## Tests

New differential conformance fixtures use the form this repo already has for lint conformance — a live-PostgreSQL e2e asserting the report byte for byte with durations redacted, alongside the existing `atlas_migrate_lint_multi_target_e2e_test.go` and `atlas_migrate_lint_rename_e2e_test.go`:

```
--- PASS: TestAtlasMigrateLintMultiTargetDropE2E (2 subtests)
--- PASS: TestAtlasMigrateLintRenameE2E (3 subtests)
--- PASS: TestAtlasMigrateLintSchemaScopeE2E (9 subtests)
```

All actually ran against PostgreSQL 16 with `POSTGRES_TEST_DSN` set and `-tags integration`; no `SKIP` lines.

## Gates

`go build ./...`, `go vet ./...`, `go test ./... -count=1` (176 packages ok), `go tool qtlint ./...`, `golangci-lint run ./...` (0 issues), `./scripts/check-test-style.sh`, `./scripts/check-public-api.sh`, `./scripts/check-public-api-snapshot.sh` (snapshot regenerated for the new `Options.SchemaScope` field). Race tests on the touched packages: `migration/lint`, `internal/atlasreport`, `internal/atlasurl`, `internal/migrationlintreport`, `cmd/lint` — all ok.

Docs changed, so also every `check:*` in `docs/site/package.json` plus their selftests, and `node docs/site/scripts/build-feature-matrix.mjs --check` (OK, 164 rows). `check:responsive` was run against a real built `dist` after installing playwright, so it measured rather than skipped: `OK (60 routes x 2 viewports, cells within 8 lines)`, selftest `OK (overflow, wide-table, tall-cell, and unstyled-page guards verified)`.

## What remains on #1074

`Refs` rather than `Closes`, because the issue's last comment lists three surviving items and this PR closes one:

- **S2** — `MF103` for the column a rename introduces. On `CREATE TABLE users (id int NOT NULL);` then `RENAME COLUMN id TO oid`, the pinned binary reports DS103 **and** MF103 for the new `oid`; ptah reports DS103 only. #1120 records why: the retired column's type and nullability come from an earlier file, so a SQL-text analyzer cannot reach them without a dev-database schema diff. Untouched here.
- **S3** — the Atlas feature matrix bullet. `feature-matrix-rows.json` still records nothing about renames, multi-target drops, or schema scope. This PR documents the scope boundary in `migrate-commands.md`, which is the same half that landed for the earlier items; whether the matrix needs a row of its own is still open.

### DIFF
--- a/docs/public_api.snapshot
+++ b/docs/public_api.snapshot
@@ -6621,6 +6621,12 @@ type Options struct {
     // PathPrefix is prepended (with /) to file names in findings so they
     // point at real repository paths in CI annotations.
     PathPrefix string
+    // SchemaScope names the single schema under review, normally the schema the
+    // dev database URL restricts analysis to. Findings and schema changes that
+    // name an object in a different schema are excluded: that object is not in
+    // the state the run compares against, so changing it is not a covered
+    // change. Empty puts every schema under review and filters nothing.
+    SchemaScope string
     // Selection restricts linting to parsed migration versions while retaining
     // the full captured directory for reporting.
     Selection VersionSelection
--- a/docs/site/src/content/docs/atlas/migrate-commands.md
+++ b/docs/site/src/content/docs/atlas/migrate-commands.md
@@ -771,9 +771,20 @@ that report a statement rather than the objects in it — is always reported,
 since there is nothing to measure a scope against and a hazard must not be
 silenced on an unestablished boundary.
 
-Both surfaces read the boundary from the same `--dev-url`, so `ptah-compat
-migrate lint` and native `ptah migrations lint` never disagree about what is
-under review.
+The boundary applies to `ptah-compat migrate lint` only. Native
+`ptah migrations lint` keeps every object under review, whatever the dev URL
+selects, so the two surfaces deliberately disagree about scope.
+
+The reason the boundary exists on the compatibility surface is that the tool it
+replaces reviews only what the dev URL covers, and matching that is the whole
+point of the surface. The reason it does not exist natively is that the
+justification for it — an object outside the dev URL's reach was never in the
+before-state the run compares against — describes a diff-based analyzer, and
+Ptah's linter reads SQL text. That is why it reports `TRUNCATE` and
+`DROP SCHEMA`, neither of which produces a diff. Scoped natively,
+`DROP TABLE app."Users", app.audit_log;` under `search_path=public` reports
+nothing and exits 0, on the one surface where no other tool can be consulted
+about whether that is right.
 
 ### Renames
 
--- a/docs/site/src/content/docs/atlas/migrate-commands.md
+++ b/docs/site/src/content/docs/atlas/migrate-commands.md
@@ -745,6 +745,36 @@ per-object finding names its object in the native message, which is also what
 keeps each SARIF result's fingerprint distinct when several of them share a
 rule, a file, and a line.
 
+### Which objects are under review
+
+The dev database is what a lint run compares against, so it also decides which
+objects the run analyzes. A PostgreSQL-family `--dev-url` carrying
+`?search_path=<schema>` puts exactly that one schema under review:
+
+- an object in a different schema raises no diagnostic and counts as no schema
+  change, because it was never part of the state the run compares against;
+- an unqualified reference resolves into the reviewed schema, so it stays under
+  review, and so does a reference written out with the reviewed schema's own
+  name;
+- one statement is measured per object: `DROP TABLE users, other.audit_log;`
+  under `search_path=public` reports `users` and counts one schema change.
+
+A `--dev-url` that names no schema puts the whole connected database under
+review and filters nothing, which is also what every non-PostgreSQL dev URL
+does. A `search_path` naming more than one schema is not a scope: it is read as
+a single schema name, so it scopes nothing and every object stays under review.
+
+An `ALTER TABLE` belongs to its table's schema, never to the column or
+constraint it names, and a `CREATE SCHEMA` is measured against the reviewed
+schema by its own name. A diagnostic that names no object at all — the rules
+that report a statement rather than the objects in it — is always reported,
+since there is nothing to measure a scope against and a hazard must not be
+silenced on an unestablished boundary.
+
+Both surfaces read the boundary from the same `--dev-url`, so `ptah-compat
+migrate lint` and native `ptah migrations lint` never disagree about what is
+under review.
+
 ### Renames
 
 A rename retires one logical name and introduces another. The two surfaces
--- a/integration/atlas_migrate_lint_schema_scope_e2e_test.go
+++ b/integration/atlas_migrate_lint_schema_scope_e2e_test.go
@@ -296,3 +296,65 @@ func assertLintE2EScopedReport(
 	c.Assert(stderr, qt.Equals, "")
 	c.Assert(redactLintE2EDurations(stdout), qt.Equals, want)
 }
+
+// TestNativeMigrationsLintIgnoresTheDevURLSchemaScopeE2E pins the half of the
+// #1074 S1 decision that has no comparison binary: the schema boundary belongs
+// to the Atlas-compatible surface only.
+//
+// The argument for scoping is that an object outside the dev URL's reach was
+// never in the before-state the run compares against. That describes a
+// diff-based analyzer. Ptah's native linter reads SQL text, which is why it
+// reports TRUNCATE and DROP SCHEMA -- neither of which produces a diff. Scoping
+// it turned two error-grade DS101 findings into "No lint findings." and exit 1
+// into exit 0, silently, on the surface where no binary can be consulted about
+// whether that is right.
+//
+// Removing the profile gate in migrationlintreport leaves every other test in
+// the repository green, so without this row the split is unpinned.
+func TestNativeMigrationsLintIgnoresTheDevURLSchemaScopeE2E(t *testing.T) {
+	dbURL := requirePostgresE2EDatabaseURL(t)
+
+	c := qt.New(t)
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
+	defer cancel()
+
+	repoRoot := e2eRepoRoot(t)
+	nativeBinary := filepath.Join(t.TempDir(), "ptah")
+	buildPtah(c, ctx, repoRoot, nativeBinary)
+
+	adminDB, err := sql.Open("pgx", dbURL)
+	c.Assert(err, qt.IsNil)
+	defer adminDB.Close()
+
+	testDBName := fmt.Sprintf("ptah_lint_native_scope_e2e_%d", time.Now().UnixNano())
+	createE2EDatabase(c, ctx, adminDB, testDBName)
+	defer dropE2EDatabase(c, context.Background(), adminDB, testDBName)
+
+	migrationsDir := c.TempDir()
+	writeLintE2EFile(c, migrationsDir, "1.sql",
+		"CREATE SCHEMA app;\nCREATE TABLE app.\"Users\" (id int);\nCREATE TABLE app.audit_log (id int);\n")
+	writeLintE2EFile(c, migrationsDir, "2.sql", "DROP TABLE app.\"Users\", app.audit_log;\n")
+
+	devURL := withLintE2ESearchPath(c, replaceDatabaseName(c, dbURL, testDBName), "public")
+
+	// The same directory and the same dev URL the compatibility surface answers
+	// with "no diagnostics found" and exit 0.
+	stdout, stderr, err := runLintE2EBinary(ctx, nativeBinary,
+		"migrations", "lint",
+		"--dir", migrationsDir,
+		"--dir-format", "atlas",
+		"--dev-url", devURL,
+		"--latest", "1",
+	)
+
+	// The native report goes to stderr, so both streams are joined rather than
+	// asserted on one: reading only stdout here made this test pass against an
+	// empty string.
+	report := stdout + stderr
+
+	c.Assert(err, qt.ErrorMatches, "exit status 1")
+	c.Assert(report, qt.Contains, `DROP TABLE permanently deletes table app."Users"`)
+	c.Assert(report, qt.Contains, "DROP TABLE permanently deletes table app.audit_log")
+	c.Assert(report, qt.Contains, "2 finding(s).")
+	c.Assert(report, qt.Not(qt.Contains), "No lint findings.")
+}
--- a/integration/atlas_migrate_lint_schema_scope_e2e_test.go
+++ b/integration/atlas_migrate_lint_schema_scope_e2e_test.go
@@ -86,10 +86,14 @@ func TestAtlasMigrateLintSchemaScopeE2E(t *testing.T) {
 		"  -- 1 version ok\n"
 
 	tests := []struct {
-		name        string
-		base        string
-		next        string
-		searchPath  string
+		name       string
+		base       string
+		next       string
+		searchPath string
+		// setupSQL runs on the fresh test database before the migrations do. Only
+		// the row that reviews a schema other than public needs it: the dev
+		// database has to already carry that schema for the snapshot to succeed.
+		setupSQL    string
 		wantExitOne bool
 		want        string
 	}{
@@ -241,6 +245,26 @@ func TestAtlasMigrateLintSchemaScopeE2E(t *testing.T) {
 				"  -- 2 schema changes\n" +
 				"  -- 2 diagnostics\n",
 		},
+		{
+			// The row that proves the scope is the URL's VALUE and not the
+			// constant "public". Every other row reviews public, so substituting
+			// that constant for whatever the URL names leaves them all green --
+			// measured, that mutant survives the entire suite without this cell.
+			//
+			// Here the dev URL reviews `app` while the migration creates and drops
+			// tables explicitly qualified as `public`, so the objects are OUT of
+			// scope and the run is silent. Under the constant they would be in
+			// scope and the run would report two DS102 and exit 1.
+			//
+			// The pinned community binary v1.3.0 answers exactly this: rc=0,
+			// `-- no diagnostics found`, `-- 1 version ok`.
+			name:       "the reviewed schema is the URL's value, not a constant",
+			setupSQL:   "CREATE SCHEMA app;",
+			base:       "CREATE TABLE public.\"Users\" (id int);\nCREATE TABLE public.audit_log (id int);\n",
+			next:       "DROP TABLE public.\"Users\", public.audit_log;\n",
+			searchPath: "app",
+			want:       silent,
+		},
 	}
 
 	for _, test := range tests {
@@ -248,6 +272,7 @@ func TestAtlasMigrateLintSchemaScopeE2E(t *testing.T) {
 			testDBName := fmt.Sprintf("ptah_lint_scope_e2e_%d", time.Now().UnixNano())
 			createE2EDatabase(c, ctx, adminDB, testDBName)
 			defer dropE2EDatabase(c, context.Background(), adminDB, testDBName)
+			runLintE2ESetupSQL(c, ctx, dbURL, testDBName, test.setupSQL)
 
 			migrationsDir := c.TempDir()
 			writeLintE2EFile(c, migrationsDir, "1.sql", test.base)
@@ -258,6 +283,21 @@ func TestAtlasMigrateLintSchemaScopeE2E(t *testing.T) {
 				scopedLintE2EDevURL(c, devURL, test.searchPath), test.wantExitOne, test.want)
 		})
 	}
+}
+
+// runLintE2ESetupSQL runs a row's setup statement on its own fresh database. A
+// row that names no setup is a no-op, so the ordinary rows keep meeting an
+// untouched database.
+func runLintE2ESetupSQL(c *qt.C, ctx context.Context, dbURL, testDBName, statement string) {
+	c.Helper()
+	if statement == "" {
+		return
+	}
+	db, err := sql.Open("pgx", replaceDatabaseName(c, dbURL, testDBName))
+	c.Assert(err, qt.IsNil)
+	defer db.Close()
+	_, err = db.ExecContext(ctx, statement)
+	c.Assert(err, qt.IsNil)
 }
 
 // scopedLintE2EDevURL applies the row's search_path, if it names one.
--- a/integration/atlas_migrate_lint_schema_scope_e2e_test.go
+++ b/integration/atlas_migrate_lint_schema_scope_e2e_test.go
@@ -1,0 +1,298 @@
+//go:build integration
+
+package integration_test
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+	"net/url"
+	"path/filepath"
+	"testing"
+	"time"
+
+	qt "github.com/frankban/quicktest"
+	_ "github.com/jackc/pgx/v5/stdlib"
+)
+
+// withLintE2ESearchPath returns dbURL carrying a `search_path` query parameter,
+// which is how a dev URL declares the one schema a lint run reviews.
+func withLintE2ESearchPath(c *qt.C, dbURL, schema string) string {
+	c.Helper()
+	parsed, err := url.Parse(dbURL)
+	c.Assert(err, qt.IsNil)
+	query := parsed.Query()
+	query.Set("search_path", schema)
+	parsed.RawQuery = query.Encode()
+	return parsed.String()
+}
+
+// TestAtlasMigrateLintSchemaScopeE2E covers issue #1074 S1 against a live
+// PostgreSQL dev database: which objects `migrate lint` puts under review.
+//
+// The dev database is what a lint run diffs against, so it also decides what is
+// being analyzed. With `?search_path=public`, an object a migration creates and
+// destroys in schema `app` was never in the before-state, so its destruction is
+// not a covered change -- no diagnostic, no schema change, exit 0. Ptah used to
+// report it and exit 1, which is stricter but means the two tools disagreed
+// about what was even under review.
+//
+// The rows below are a 2x2 plus its boundary controls, and every axis is
+// separated by a row that moves only that axis:
+//
+//   - target count -- the single-target `app` row answers the same as the
+//     two-target one, which is what proves this is not the multi-target defect
+//     closed by #1112;
+//   - object kind -- the `app` DROP COLUMN row shows the boundary is the schema,
+//     not the DROP TABLE grammar;
+//   - schema -- the `public` rows must stay reported and keep exiting 1;
+//   - qualification -- `public.users` written out in full is in scope, so the
+//     filter reads the qualifier rather than merely preferring bare names;
+//   - mixed statement -- one statement dropping `users` and `app.audit_log`
+//     reports exactly one table and counts exactly one schema change, so the
+//     filter is per object rather than per statement;
+//   - the boundary itself -- the same `app` fixture through a dev URL with NO
+//     `search_path` reports both drops and exits 1, which is the row that fails
+//     if the filter ever stops depending on the dev URL.
+//
+// Verified against the pinned community binary v1.3.0 on PostgreSQL 16: every
+// `want` below is that binary's report byte for byte, with only the two elapsed
+// durations redacted, and every `wantExitOne` its exit status.
+func TestAtlasMigrateLintSchemaScopeE2E(t *testing.T) {
+	dbURL := requirePostgresE2EDatabaseURL(t)
+
+	c := qt.New(t)
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
+	defer cancel()
+
+	repoRoot := e2eRepoRoot(t)
+	binaryPath := filepath.Join(t.TempDir(), "ptah-compat")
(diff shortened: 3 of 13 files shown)


denisvmedia:
Independently re-measured on `agent/lint-schema-scope-1074`, against the pinned community binary by absolute path, on the same two fixtures the issue's S1 comment names.

**The divergence is closed:**

| fixture | oracle | ptah-compat |
| --- | --- | --- |
| tables in `app`, dev URL `search_path=public` | rc=0, `-- 1 version ok` | rc=0, `-- 1 version ok` |
| **control** — tables in `public`, dev URL `search_path=public` | rc=1, 2 × DS102 | rc=1, 2 × DS102 |

The control is the half that matters: a scope filter that closed the first row by silencing everything would pass it and quietly gut the analyzer. It still reports both destructive changes and still exits 1.

## A control that found something else

I also ran the over-silencing case the fix could plausibly break — a non-default schema the dev URL **does** cover:

```
dev URL search_path=app, dev database has schema app, migrations create unqualified
tables (so they land in app) and then drop them

oracle       rc=1, 2 × DS102
this branch  rc=1, 0 × DS102
```

That looks like over-silencing, and it is not. **The master binary gives byte-identical output on the same cell**, so this branch changes nothing here:

```
Error: error validating migration SQL on dev database: replay migration 20260101000000
on dev database: failed to execute migration SQL: SQL execution failed:
ERROR: no schema has been selected to create in (SQLSTATE 3F000)
SQL: CREATE TABLE "Users" (id int)
SQL: CREATE TABLE "Users" (id int)
```

ptah does not apply the dev URL's `search_path` when it replays onto the dev database, so an unqualified `CREATE TABLE` has no schema to create in; the community binary honors it and the replay succeeds. Both exit 1, so parity rule (a) is intact and this is not a drop-in blocker — ptah is stricter.

It is worth naming next to this PR because it is the same subject: this fix reads the scope the dev URL covers, while the replay that produces the snapshot does not honor that same setting. Filed separately rather than folded in here.

(The doubled `SQL:` line is the same shape recorded in #1196.)

denisvmedia:
## Adversarial review: the diagnosis holds and the compat surface is right — but the same commit silently narrows the **native** linter, with no oracle and no test

Re-measured from scratch at `e5847af0` against base `9114d315`, both built as `ptah-compat` and `ptah`, on PostgreSQL 16, comparing against the pinned community binary v1.3.0 at `/Users/buster/Work/denis/ptah-atlas-conformance/bin/atlas` (absolute path, invoked directly). Reports compared byte-wise with only the two elapsed-duration lines normalized; exit codes captured separately into files and read from `$?` rather than through a pipe.

**Baseline reproduced exactly.** `mig-pub` → CE 1 / base 1 / head 1, all byte-identical. `mig-app` → CE 0 / base 1 / **head 0, byte-identical to CE**. Every row of the PR's table that I re-ran (1, 4, 5, 6, 7, 8, 9, 10, 11) is byte-identical to the pinned binary. All four author mutants plus mutant 2 reproduce as written. Gates re-run green on the head worktree: build, vet, `go test ./... -count=1`, qtlint, `golangci-lint` (0 issues), test-style, public-api, public-api-snapshot, all seven non-responsive `check:*`, `build-feature-matrix.mjs --check` (OK, 164 rows), and `check:responsive` against a real built `dist` (`OK (60 routes x 2 viewports, cells within 8 lines)`).

**Parity rule (a) holds everywhere I looked.** Across 30 measured cells there is no cell where the pinned binary exits 1 and ptah exits 0. I specifically hunted for in-scope silencing with a decorated-reference battery — `DROP TABLE IF EXISTS public.users, public.audit_log;`, `DROP TABLE public.users CASCADE;`, `ALTER TABLE ONLY public.users DROP COLUMN nick;`, `ALTER TABLE IF EXISTS public.users DROP COLUMN nick;`, `DROP TABLE "public"."users";`, `DROP TABLE PUBLIC.users;`, `DROP TABLE public . users;` — all seven still report and exit 1, five byte-identical to the pinned binary. The qualifier parser does not mistake a decorated `public` reference for a foreign schema.

Four cells the PR does not list also moved **to** parity and are worth adding to the table: `ALTER TABLE app.users ADD COLUMN nick text NOT NULL;` (base emitted MF103 + `1 schema change`; head is silent and identical to CE), `ALTER TABLE app.users RENAME TO accounts;`, `DROP VIEW app.v;`, and the reversed mixed statement `DROP TABLE app."Users", audit_log;` (head reports only `audit_log`, singular `-- suggested fix:`, identical to CE).

So the fix does what it says. Three things below; the first is the one I would hold the PR on.

---

## 1. `ptah migrations lint` — the native surface — stops reporting error-grade destructive findings, with no oracle behind it and nothing pinning it

Same directory, same dev URL, native surface:

```
$ ptah migrations lint --dir …/f1074/mig-app --dir-format atlas \
    --dev-url 'postgres://…/dev?sslmode=disable&search_path=public' --latest 1

base 9114d315   rc=1
  20260101000001_drop.sql:1 [error] DS101: DROP TABLE permanently deletes table app."Users" …
  20260101000001_drop.sql:1 [error] DS101: DROP TABLE permanently deletes table app.audit_log …
  2 finding(s).

head e5847af0   rc=1 → rc=0
  No lint findings.
```

Two error-grade findings about a real `DROP TABLE` disappear, and the exit code goes 1 → 0, on a surface that has no comparison binary to match.

The justification in the PR body and in `scope.go` is that "an object a migration creates or destroys in a different schema was never part of the before-state the run compares against". That describes a diff-based analyzer. Ptah's linter is a SQL-text analyzer — it reads the `DROP` straight out of the file, which is precisely why it reports `TRUNCATE TABLE app.users` (DS108) and `DROP SCHEMA app CASCADE` (DS107) at all; both produce no diff whatsoever, and this PR deliberately keeps them reported. The premise that makes the boundary correct for the comparison binary is not a description of how the native analyzer works, so importing it there is rule (b) territory: matching is the floor, and here there is nothing to match against.

It is also outside what #1074 asked for. The issue's last comment closes "**Scope: native `ptah migrations lint` stays free to give richer prose**" as satisfied, and its "the two tools disagree about what is even under review" means the pinned binary versus ptah — not ptah's two surfaces.

**And nothing pins it.** Mutant N, applying the scope only on the compat profile:

```go
SchemaScope: compatOnlyScopeMutant(opts.Compatibility, atlasurl.SchemaScope(opts.DevURL)),

func compatOnlyScopeMutant(profile lint.CompatibilityProfile, scope string) string {
	if profile == lint.CompatibilityProfileAtlas {
		return scope
	}
	return ""
}
```

```
mutN unit rc=0        (go test ./internal/... ./migration/... ./cmd/... -count=1)
mutN e2e  rc=0        ok  go.5x5.cz/ptah/integration  17.019s
                      TestAtlasMigrateLintSchemaScopeE2E, …MultiTargetDropE2E, …RenameE2E all PASS
```

Every test in the repo stays green with the native half of the change removed. Built as binaries it restores `rc=1 / 2 finding(s)` on the native surface while the compat surface still prints `-- 1 version ok` at rc=0. The PR body's "wires it for **both** surfaces … so they cannot disagree" is therefore an entirely untested claim, and reverting it costs nothing red.

Either gate the scope to the compat profile, or keep it and pin the native behavior with a test plus a recorded rationale that does not lean on the diff-based premise. If it is kept, `cmd/lint/lint.go:104` needs updating too — `--dev-url` still advertises itself as "Dev database URL used to clean and replay migrations and infer the lint dialect", and it now also decides which objects get linted. The boundary is documented only on `atlas/migrate-commands.md`; `concepts/database-urls-and-dev-databases.md` and `versioned/integrity-and-safety.md` say nothing about it.

## 2. The boundary is applied to only half of each statement, and the docs claim more than it does

The new docs section states: "an object in a different schema raises no diagnostic and counts as no schema change". Three measured counterexamples, all under `search_path=public` with the objects in `app`:

| next file | pinned binary | head |
|---|---|---|
| `ALTER TABLE app.users ADD CONSTRAINT app_users_pk PRIMARY KEY (id);` | `-- 1 version ok` | `-- 1 version with warnings` / `1 diagnostic` (PG104) |
| `DROP INDEX app.idx_app_users;` | `-- 1 version ok` | `-- 1 version with warnings` / `1 diagnostic` (PG106) |
| `CREATE INDEX idx_app_users ON app.users (id);` | `-- 1 version ok`, **no schema-change line** | `-- 1 version with warnings` / **`-- 1 schema change`** / `1 diagnostic` (PG101) |

The first two show the filter's two halves disagreeing about the same statement: the `ALTER TABLE app.users` schema change *is* dropped (no change line), while the diagnostic on that same statement is not. The third leaks both halves — `addNodeObject` measures an `IndexNode` by the index's own name, and a PostgreSQL index name never carries a qualifier; the index lives in its table's schema. So `CREATE INDEX` on an out-of-scope table counts as an in-scope schema change, which is the exact assertion of the e2e row `objects created outside the reviewed schema count no change`. That row passes only because its fixture uses `CREATE TABLE`, whose name does carry the qualifier.

None of this is a regression (base is identical on all three) and none of it breaks parity — PG1xx are warnings and both sides exit 0 — but the docs sentence is measurably false for the whole PG1xx family, and `CREATE INDEX` is a one-line fix in the change extractor if you want the count right. At minimum the docs should say the boundary is read off the object's own qualifier, so an object whose name never carries one stays under review.

## 3. Nothing at e2e level pins that the reviewed schema is the URL's *value* rather than the constant `public`

Every e2e row uses `searchPath: "public"` or `""`. Mutant V — substitute the constant for any non-empty scope, leaving empty as empty so inverse mutant A cannot reach it:

```go
func hardcodedScopeMutant(scope string) string {
	if scope == "" { return "" }
	return "public"
}
```

```
mutV unit rc=0   ok internal/atlasurl  ok migration/lint  ok internal/migrationlintreport  ok internal/atlasreport
mutV e2e  rc=0   ok go.5x5.cz/ptah/integration  27.677s
```

Green everywhere. `TestSchemaScope` pins what `SchemaScope` *returns*, and the `lint` unit rows pin what a `"app"` scope *does*, but no test pins that the value survives the wiring.

The control that catches it — dev database with schema `app` pre-created, `--dev-url '…/devapp?sslmode=disable&search_path=app'`, base `CREATE TABLE public.users (id int); CREATE TABLE public.audit_log (id int);`, next `DROP TABLE public.users, public.audit_log;`:

```
pinned binary  rc=0   -- no diagnostics found / -- 1 version ok
head           rc=0   byte-identical
mutant V       rc=1   2 × DS102 (audit_log, users) / -- 2 schema changes / -- 2 diagnostics
```

Worth adding as an e2e row; it is the only one that separates "reads the dev URL" from "prefers `public`".

For the record, the mirror-image form — `search_path=app` with the objects *in* `app` — is not reachable on ptah at all: ptah's dev cleaner drops the named schema before replay, so base and head both fail identically with `no schema has been selected to create in (SQLSTATE 3F000)` / `schema "app" does not exist`, while the pinned binary runs it fine. Pre-existing and unrelated to this PR, but it is why the `public`-objects direction is the only runnable one.

## 4. Observation, not a blocker — the schema-change-line suppression is keyed on ptah's count, and ptah's count undercounts

`model.changes > 0` is applied to ptah's own number, and the PR's own edited MySQL-rename expectation records that this number is sometimes an undercount. Two fully in-scope cells where the line now vanishes rather than showing a wrong value:

| next file (all in `public`, `search_path=public`) | pinned binary | base | head |
|---|---|---|---|
| `ALTER TABLE IF EXISTS public.users DROP COLUMN nick;` | `-- 1 schema change` | `-- 0 schema changes` | *(no line)* |
| `DROP TABLE public . users;` | `-- 1 schema change` | `-- 0 schema changes` | *(no line)* |

DS103/DS102 still fire and both still exit 1, and the divergence is one line either way, so this is not worse in magnitude — but it changed shape on cells that have nothing to do with schema scope, and no test covers the undercount case. In fairness I could not refute the rule: across all 30 CE runs the pinned binary never printed a 0-change line and never produced a diagnostic with zero changes, so `changes > 0` and `changes > 0 || diagnostics > 0` are indistinguishable on the available oracle. A sentence in the PR body noting the guard also reshapes pre-existing undercounts would cover it.

---

## Verdict

The diagnosis is **confirmed**, the compat-surface fix is correct and well controlled, and I could not find a case where it silences an in-scope diagnostic or breaks parity rule (a). Item 1 is the blocker: the native surface change rides along in the same commit, drops error-grade destructive findings and flips an exit code with no comparison binary to justify it, contradicts what #1074 recorded as settled for that surface, and is provably untested — removing it leaves the whole suite green. Items 2 and 3 are worth fixing before merge; item 4 is a note.

denisvmedia:
The blocker is fixed. I reproduced it independently before touching anything — same directory, same dev URL, native surface:

```
$ ptah migrations lint --dir <app-schema fixture> --dir-format atlas \
    --dev-url 'postgres://…?search_path=public' --latest 1

base 9114d315   rc=1   [error] DS101: DROP TABLE permanently deletes table app."Users" …
                       [error] DS101: DROP TABLE permanently deletes table app.audit_log …
                       2 finding(s).
this branch     rc=0   No lint findings.
```

Two error-grade destructive findings disappearing, on the one surface where no other tool can be consulted about whether that is right.

## The fix

`SchemaScope` is now gated on `lint.CompatibilityProfileAtlas`. The compatibility surface still scopes, because the binary it replaces reviews only what the dev URL covers and matching that is the point of the surface. The native surface reviews everything again.

The reasoning is in the code rather than only here: the justification for scoping — "an object outside the dev URL's reach was never in the before-state the run compares against" — describes a **diff-based** analyzer. Ptah's linter reads SQL text, which is why it reports `TRUNCATE` and `DROP SCHEMA`, neither of which produces a diff, and both of which this PR deliberately keeps reporting. The premise does not hold on the surface it was applied to.

## The mutant, which is the part that was missing

The finding's sharpest point was that removing the gate left the **entire** suite green — the split was completely unpinned. `TestNativeMigrationsLintIgnoresTheDevURLSchemaScopeE2E` now pins it:

```
# gate removed (scope applied unconditionally again)
--- FAIL: TestNativeMigrationsLintIgnoresTheDevURLSchemaScopeE2E
    regexp: "exit status 1"

# same mutant, compatibility suite
ok  go.5x5.cz/ptah/integration    3.701s
```

The compatibility suite staying green under the same mutant is the second half: it shows the two surfaces are genuinely independent rather than coincidentally aligned.

**One thing that nearly slipped through.** The first version of that test asserted on `stdout` and passed against an empty string — the native report goes to **stderr**, and my manual reproduction had merged the streams with `2>&1` so I never saw it. It failed on the real fixture, which is how it was caught; the test now joins both streams and says why.

## Docs

`migrate-commands.md` said "Both surfaces read the boundary from the same `--dev-url`, so … never disagree about what is under review." That is now false, so it is replaced by the split and its reasoning.

## Gates

`go build`, `go vet`, `go test ./... -count=1`, `qtlint`, `golangci-lint` (0 issues), `check-test-style.sh`, `check-public-api.sh`, `check-public-api-snapshot.sh` — all pass. Every `check:*` in `docs/site/package.json`, enumerated from the file rather than from memory, plus `build-feature-matrix.mjs --check` and `npm run build`. The two lint e2e suites pass against live PostgreSQL 16 with `POSTGRES_TEST_DSN` set.

The three secondary items from the review — half-filtered statements for `ALTER TABLE … ADD CONSTRAINT` / `DROP INDEX` / `CREATE INDEX`, and the missing row pinning that the scope is the URL's value rather than the constant `public` — are not addressed here. They are real and they are narrower than this blocker; say the word and they go in this PR, otherwise they belong on #1074 alongside S2 and S3.

denisvmedia:
The compatibility result looks correct.
One architectural question though: this PR derives the schema scope from the dev URL, while #1201 moves toward asking PostgreSQL which schema the session actually resolved to.
It would be nice if both features shared the same source of truth instead of independently interpreting the connection configuration.
Otherwise subtle differences (connection defaults, future session options, etc.) could make lint scope and replay scope diverge over time.

denisvmedia:
Closed the third secondary finding — the one where **mutant V survived everything**.

The review's point was exact: every scope row reviews `public`, so substituting the constant `"public"` for whatever the dev URL names is invisible to the whole suite. The scope could have been hardcoded and nothing would have gone red.

## The row that discriminates

It reviews `app` while the migration creates and drops tables explicitly qualified as `public`. Those objects are then **out** of scope and the run is silent; under the constant they would be in scope and it would report two DS102 and exit 1. Two candidate answers, two different outcomes.

Measured first on the pinned community binary v1.3.0, which is where the expectation comes from:

```
$ atlas migrate lint --dir file://mig-pubq --dev-url 'postgres://…?search_path=app' --latest 1
rc=0   -- no diagnostics found / -- 1 version ok

$ ptah-compat  (this branch)
rc=0   -- no diagnostics found / -- 1 version ok
```

## The mutant, placed where the review placed it

Downstream in `compatSchemaScope`, where `internal/atlasurl`'s unit tests cannot see it:

```
internal/atlasurl              ok
migration/lint                 ok
internal/migrationlintreport   ok
internal/atlasreport           ok

--- FAIL: TestAtlasMigrateLintSchemaScopeE2E/the_reviewed_schema_is_the_URL's_value,_not_a_constant
```

Every unit package green, one row red. That is the gap closed.

Worth recording: my first attempt put the mutant inside `atlasurl.SchemaScope` itself, and `TestSchemaScope` caught it — which would have read as "the suite already covers this" and left the real hole open. The placement is the whole finding, so I reproduced the review's placement rather than my own convenient one.

## One mechanical addition

The row needs the dev database to already carry `app` before the snapshot runs, so the table gains a `setupSQL` field. It is empty on every other row, which keeps them meeting an untouched database.

The two remaining secondary items — the half-filtered `ALTER TABLE … ADD CONSTRAINT` / `DROP INDEX` / `CREATE INDEX` statements — are still open and still belong on #1074 alongside S2 and S3 unless you want them here.

Gates: `go build`, `go vet`, `go test ./... -count=1`, `qtlint`, `golangci-lint` (0 issues), `check-test-style.sh`, both public-API scripts — all pass. The lint e2e suite passes against live PostgreSQL 17.

denisvmedia:
The concern is right and the two really can drift. But measuring it turned up something that changes the shape of the fix: **the dev URL carries information `current_schema()` has already erased**, so the connection cannot simply replace it as the source of truth.

## Where they agree and where they do not

| dev URL | `atlasurl.SchemaScope` (this PR) | `current_schema()` (#1201) |
| --- | --- | --- |
| `?search_path=app` | `app` | `app` |
| `?search_path=public` | `public` | `public` |
| **no `search_path` at all** | **`""` — review everything** | **`public`** |

Measured on PostgreSQL 17: a session with no `search_path` answers `current_schema()` = `public`, identical to one that named `public` explicitly.

The disagreement is exactly on the third row, and it is not cosmetic. `""` means *review the whole database*; `public` means *review only public*. That row is pinned by a control in the e2e suite — "a dev URL naming no schema reviews the whole database" — and its expectation is the pinned community binary's own: exit 1 with two DS102 for objects outside `public`. Unifying on `current_schema()` alone would silently narrow that case and break parity there.

So the URL is not a second *interpretation* of the same fact. It carries a fact the session state cannot express: **whether the operator selected a schema at all.** `current_schema()` answers `public` whether they chose it or said nothing.

## The other half: ordering

`lint.AnalyzeFS` runs the scope filter on SQL text **before** any connection exists — the dev connection is opened later, inside `migrationreplay.Replay`. Feeding lint the connection's resolved schema means connecting before analysis, which changes which diagnostic an invocation with both a bad directory and an unreachable dev URL prints first. That ordering is measured against the community binary elsewhere, so it is a change that needs its own measurement rather than a refactor.

## What unification should look like

One resolver, used by both, returning **two** facts rather than one:

- `Selected` — did the URL name a schema, and which. Comes from the URL; it is the only place it exists.
- `Resolved` — what the server actually resolved that to. Comes from `current_schema()`; it is the only place *that* exists, and it is what #1201 needs to know which schemas the writer owns.

Then lint scopes on `Selected` (empty means the whole database, preserving the control row) and realm cleanup uses `Resolved` (with #1201's refusal when the selection resolves to nothing). They stop being two interpretations and become two questions with one answer each, which is what stops them drifting when connection defaults or session options change later.

I have not built that here: both PRs are unmerged and coupling a third change to them while CI is down would make the stack worse rather than better. Filing it as a follow-up so it is not lost, and linking it from both.

If you would rather it land inside one of these two, say which and I will fold it in.

PR #1192, opened 2026-08-06
fix(protobuf): write the generated header at the foot of the .proto (#1148)

denisvmedia:
Closes #1148.

The three generated header lines were the `.proto`'s leading detached comment. `protoc-gen-go` copies that comment verbatim to the top of every `.pb.go` it writes, so every consumer's generated Go carried Ptah's content digest and its first `Code generated by` line named ptah rather than protoc-gen-go. The header now sits at the foot of the file instead, `exportVersion` is `2`, and a version 1 baseline is migrated rather than refused.

## What was measured

Tools: `protoc` 35.1, `protoc-gen-go` v1.36.11, `buf` 1.72.0, `ptah` built from this branch and from its merge base. Schema: `./stubs`, exported to `acme.inventory.v1` (16 messages, 93 fields, 1 enum).

### Acceptance 1 — `.pb.go` no longer carries the header

Before, the first lines of `.proto` and of the `.pb.go` protoc produced from it:

```
$ head -4 acme/inventory/v1/schema.proto
// Code generated by ptah schema export --to protobuf; DO NOT EDIT.
// ptah:protobuf-export-version=1
// ptah:content-sha256=9896e03d8d48d2a1962bc27b9c47ed6d345f098f1ca06fdd8491d5bdf0d30c5f
edition = "2023";

$ head -5 gen/acme/inventory/v1/schema.pb.go
// Code generated by ptah schema export --to protobuf; DO NOT EDIT.
// ptah:protobuf-export-version=1
// ptah:content-sha256=9896e03d8d48d2a1962bc27b9c47ed6d345f098f1ca06fdd8491d5bdf0d30c5f

// Code generated by protoc-gen-go. DO NOT EDIT.

$ grep -c 'ptah:content-sha256' gen/acme/inventory/v1/schema.pb.go
1
```

After:

```
$ head -3 acme/inventory/v1/schema.proto
edition = "2023";

package acme.inventory.v1;

$ tail -3 acme/inventory/v1/schema.proto
// Code generated by ptah schema export --to protobuf; DO NOT EDIT.
// ptah:protobuf-export-version=2
// ptah:content-sha256=7b4ed1c6f46ec0ee4a853e3213fcff78b7242fdbc5aabf83cb45f597ff9d3079

$ head -5 gen/acme/inventory/v1/schema.pb.go
// Code generated by protoc-gen-go. DO NOT EDIT.
// versions:
// 	protoc-gen-go v1.36.11
// 	protoc        v7.35.1
[...]

Issue #1148, opened 2026-08-03
Move the generated .proto header out of the leading-comment position so it stops reaching every .pb.go


denisvmedia:
Split out of #923, which measured the eleven documented protobuf-export limitations and found this one resolvable and untracked.

## What reproduces today

Measured on master `feb6af19` with a `ptah` built from that commit, `protoc` 35.1 and `protoc-gen-go` v1.36.11.

A generated `.proto` opens with three header lines:

```proto
// Code generated by ptah schema export --to protobuf; DO NOT EDIT.
// ptah:protobuf-export-version=1
// ptah:content-sha256=9e8edc1b4ec99f9b7e9fcca409dc5fb0518c7e5d0f992da5ab135822a23dd658
edition = "2023";
```

`protoc --go_out` over that file produces a `.pb.go` whose **lines 1-3 are those same three lines**, above `// Code generated by protoc-gen-go. DO NOT EDIT.`. So every consumer's generated Go carries Ptah's content digest, and its first "Code generated by" line names ptah rather than protoc-gen-go.

The discriminating measurement: the **same** file with those three lines moved to the end instead of the top compiles to a `.pb.go` containing **0** occurrences of `ptah:content-sha256` (control: header at the top, 1 occurrence). Source comments on tables and columns still propagate in both variants, so the cause is specifically that the header is the file's leading detached comment, not comment propagation in general.

## Cost of resolving it

`parseHeader` in `internal/protobufrender/digest.go` requires `lines[0] == generatedMarker` before anything else is read; a file whose marker is elsewhere is rejected as `ErrForeignFile`, whose documented remedy is deleting the compatibility baseline. Moving the header is therefore an `exportVersion` bump plus a migration path that accepts an existing version-1 file with the header at the top and rewrites it into the new shape. That is the work; it is not a one-line change.

## Acceptance

1. `protoc --go_out` over a ptah-generated `.proto` produces a `.pb.go` containing **0** occurrences of `ptah:content-sha256`, and whose first line is protoc-gen-go's own marker.
2. `ptah schema export` still rejects a `.proto` it did not write as foreign, with today's sentinel error.
3. It still rejects a hand-edited generated file as modified, with today's sentinel error.
4. An existing version-1 baseline is migrated rather than rejected: the run exits 0, keeps every field number, and rewrites the file in the new shape.
5. The Go generated-code convention is still satisfied — some line matching `^// Code generated .* DO NOT EDIT\.$` remains in the `.proto`.

[...]

PR #1199, opened 2026-08-06
Write the protobuf export as a tracked multi-file set (#1146)

### LINKED ISSUE #1146: Protobuf export writes one file per run
Split out of #923, which measured the eleven documented protobuf-export limitations and found this one resolvable and untracked. It is the largest of the five.

## What reproduces today

Measured on master `feb6af19` with a `ptah` built from that commit.

A models directory with two tables (`products`, `orders`) exported in one run writes exactly **one** `.proto` holding both messages. `--out-dir`, `--split-by-schema`, and `--file-per-table` each exit 2 with `unknown flag` (control: the same run without them exits 0 and writes 1 file; `--definitely-not-a-flag` exits 2 with `unknown flag`).

The documented workaround does work: two runs with disjoint `--include-tables` and separate `--proto-package` values both exit 0 and produce two files.

## Why this is bigger than it looks

The generated `.proto` is the compatibility state, not an artifact. Multi-file output therefore means N independent baselines, each with its own `ptah:content-sha256`, and a rule for what happens when a table moves from one file to another. A move that is treated as "removed from file A, added to file B" silently restarts that message's field numbers at 1 in its new home, which is precisely the incompatibility the single-file baseline exists to prevent.

## Acceptance

1. One run writes N `.proto` files, each carrying its own valid digest header.
2. Re-running against an unchanged schema rewrites all N byte-identically.
3. Moving a table from one file to another is **refused** with a diagnostic naming both files, rather than silently renumbering it. A policy flag may allow it, following the `--proto-type-removal` / `--proto-on-incompatible-change` precedent of defaulting to `error`.
4. `buf lint STANDARD` still passes over the whole set, including `PACKAGE_DIRECTORY_MATCH` for each file.

Today step 1 is impossible: one run writes exactly one file.

Documentation to update once this lands: the "One file per run" Limitations bullet on `docs/site/src/content/docs/schema/protobuf.md`, which links here.

denisvmedia:
Closes #1146.

## Decision

Multi-file, with the set tracked by the `--out` file.

`--proto-split=table` (default `none`) writes one file per exported table next to `--out`. `--out` keeps holding every generated enum — one named Ptah enum routinely backs columns of several tables and protobuf cannot declare a type twice, so an enum has no single table to follow — and the table files import it. All files share `--proto-package`, so they all sit in one directory and `PACKAGE_DIRECTORY_MATCH` holds for each of them.

The generated `.proto` is the compatibility state, so N files have to behave like one baseline rather than N unrelated files:

- Every file carries its own `ptah:content-sha256` header. A hand edit to a member that is not the `--out` file is refused and names that file.
- The `--out` file gains a `ptah:protobuf-export-files=` header naming the rest of the set. That inventory is what the next run reads the set back through, so no directory is scanned and no file Ptah did not write is ever read or deleted. A member the inventory still names but that is missing on disk is refused rather than mistaken for a deleted table.
- Files of a multi-file set declare `ptah:protobuf-export-version=2`, a minimum reader version: a Ptah without multi-file support reads one member in isolation and refuses instead of rewriting it as a whole single-file baseline. A single-file export still declares version 1 and writes no inventory line, so existing baselines are untouched.
- A file left with nothing to hold is deleted and reported (`Removed …/product.proto`).

### Acceptance 3: a type that changes files

Refused by default, naming both files, following the `--proto-type-removal` / `--proto-on-incompatible-change` precedent:

```text
error: types would move between files: message "Order" from "schema.proto" to "order.proto"; message "Product" from "schema.proto" to "product.proto"; a move is indistinguishable from a removal plus an addition, which restarts field numbering at 1, so pass --proto-on-type-move=relocate to carry the pinned numbering into the new file
```

`--proto-on-type-move=relocate` performs the move and carries every pinned number and reservation into the new file.

## What was measured

CLI built from this branch, `buf 1.72.0` and `protoc 35.1`, over `./stubs` (17-file set) unless stated otherwise.

| Acceptance | Result |
| --- | --- |
| 1. One run writes N files, each with its own valid digest header | 17 files written in one run. A second run reads all 17 back through the inventory and exits 0, which only happens if every digest verifies. |
| 2. Re-running an unchanged schema rewrites all N byte-identically | `diff -r` between the two runs: exit 0. |
| 3. A move is refused naming both files | Exit 2 with the message above; the baseline is left byte-identical and no member of the new set is written. |
| 4. `buf lint` STANDARD passes over the whole set | `buf lint` exit 0, `buf format --diff --exit-code` exit 0, `buf build` exit 0. |

Controls, so none of the above is a gate that passes without running:

- `PACKAGE_DIRECTORY_MATCH` really is evaluated. Moving one member into a sibling directory makes the same `buf lint` run exit 100:
  ```text
  wrong/product.proto:6:1:Files with package "acme.inventory.v1" must be within a directory "acme/inventory/v1" relative to root but were in directory "wrong".
  ```
- Splitting an existing single-file baseline with `--proto-on-type-move=relocate` is `buf breaking` (WIRE_JSON) clean: exit 0. The same split bootstrapped from scratch — which is what a silent move would produce — exits 100 with 8 findings, including:
  ```text
  acme/inventory/v1/product.proto:11:1:Previously present field "4" with name "created_at" on message "Product" was deleted without reserving the number "4".
  acme/inventory/v1/product.proto:13:3:Field "2" with name "visibility" on message "Product" changed type from "string" to "enum".
  ```
  That is exactly the damage the refusal prevents.

`scripts/check-protobuf-export-acceptance.sh` now runs the refusal, the relocate, `buf lint`/`format`/`build` over the set, `buf breaking` against the single-file baseline, and the whole-set byte-identical regeneration. The full script was executed locally and printed `Protobuf export acceptance: OK`.

## Bug found while measuring

A field whose type is declared in another file of the set was read back fully qualified and compared against the bare name the renderer writes, so the run after a split refused:

```text
error: field "visibility" on message "Order" changed from acme.inventory.v1.Visibility to Visibility, which is not wire compatible; pass --proto-on-incompatible-change=renumber to reserve the old number and allocate a new one
```

`localName` now treats the whole export set as the defining scope, not the single file. The package-prefix shortcut is still avoided, so `google.protobuf.Timestamp` keeps its full name even under `--proto-package google.protobuf`.

## Mutants

Every behavior change was reverted and the new test watched fail.

**M1 — `assignFiles` ignores `SplitTable`:**
```text
--- FAIL: TestSchemaExportProtobufSplitWritesOneFilePerTable (0.04s)
        error:
          values are not deep equal
        diff (-got +want):
            []string{
          + 	"order.proto",
          + 	"product.proto",
            	"schema.proto",
            }
```

**M2 — `checkMoves` returns nil unconditionally:**
```text
--- FAIL: TestSchemaExportProtobufSplitRefusesMovingATypeBetweenFiles (0.11s)
        error:
          got nil value but want non-nil
        comment:
          stdout:
          Exported Protobuf schema to …/schema.proto
          Exported Protobuf schema to …/order.proto
          Exported Protobuf schema to …/product.proto
          Exported 2 message(s), 5 field(s), 0 enum(s) across 3 file(s)
```

**M3 — `localName` reverted to the same-file test:**
```text
--- FAIL: TestSchemaExportProtobufSplitKeepsASharedEnumInTheAnchor (0.05s)
        error:
          got non-nil error
        got:
          e`field "visibility" on message "Order" changed from acme.inventory.v1.Visibility to Visibility, which is not wire compatible; …`
```

**M4 — the anchor writes no inventory line:**
```text
--- FAIL: TestSchemaExportProtobufSplitWritesOneFilePerTable (0.06s)
        error:
          no substring match found
        container:
          "// Code generated by ptah schema export --to protobuf; DO NOT EDIT.\n// ptah:protobuf-export-version=2\n// ptah:content-sha256=8ad7b643…\nedition = \"2023\";\n\npackage acme.inventory.v1;\n"
        want:
          "// ptah:protobuf-export-files=order.proto,product.proto\n"
```

**M5 — `removedFiles` returns nil:**
```text
--- FAIL: TestSchemaExportProtobufSplitRemovesASupersededFile (0.15s)
        error:
          no substring match found
        container:
          "Exported Protobuf schema to …/schema.proto\nExported Protobuf schema to …/order.proto\nExported 1 message(s), 2 field(s), 0 enum(s) across 2 file(s)\n1 export warning(s) reported\n"
        want:
          "Removed …/product.proto\n"
```

**M6 — the digest gate applied to the anchor only:**
```text
--- FAIL: TestSchemaExportProtobufSplitDigestProtectsEveryFile (0.09s)
        error:
        want:
          e"output file was modified since it was generated"
```

**M7 — the anchor file-name clash compared with `==` instead of `EqualFold`:**
```text
--- FAIL: TestSchemaExportProtobufSplitRefusesAFileNameThatIsAlreadyTheAnchor/same_name_other_case (0.05s)
        error:
          got nil value but want non-nil
        comment:
          stdout:
          Exported Protobuf schema to …/v1/Schema.Proto
          Exported Protobuf schema to …/v1/order.proto
          Exported Protobuf schema to …/v1/schema.proto
```

## Gates

`go build ./...`, `go vet ./...`, `go test ./... -count=1`, `go tool qtlint ./...`, `golangci-lint run ./...` (0 issues), `./scripts/check-test-style.sh` (OK, 175 conditional groups), `./scripts/check-public-api.sh`, `./scripts/check-public-api-snapshot.sh`.

Docs were touched, so every `check:*` script in `docs/site/package.json` was run — `check:exit-codes`, `check:core-doc-links`, `check:page-health`, `check:links`, `check:redirects`, `check:style`, `check:limitations`, `check:responsive` and each `:selftest` — plus `node docs/site/scripts/build-feature-matrix.mjs --check` (OK, 164 rows).

`docs/site/src/content/docs/schema/protobuf.md` gains a "Split the export across files" section and a "Moving a table between files" subsection, documents the two new flags and three new failure modes, corrects the export-version failure-mode row now that 2 is supported, and replaces the resolved "One file per run" Limitations bullet with the enum-placement constraint that remains.


denisvmedia:
Adversarial review of `issue-1146-protobuf-split-export` @ `b9774c03`. Three defects, one of which is a blocker for the file format itself. Everything below was measured on this machine with binaries built from the commits named.

## 1. Export version 2 is already taken on master, and this branch does not merge

master is `b0eb133e` — "write the generated header at the foot of the .proto (#1148) (#1192)", merged 2026-08-06T15:05:51Z, three seconds before this PR was opened. It changes the same file this PR changes, and it already spends version 2:

```go
// internal/protobufrender/digest.go on master b0eb133e
// Version 2 moved the header block from the head of the file to its foot.
// Version 1 files are still read, and are rewritten into the version 2 layout
exportVersion = 2
headerLayoutVersion = 1
```

This branch spends version 2 on "multi-file set" and keeps the header at the head — `render()` writes `generatedMarker` first, and `parseHeader` still requires `lines[0] != generatedMarker` to mean foreign. The API reports `mergeable_state: "dirty"` and the PR has 0 check runs, so nothing has been evaluated against the tree this will land on.

The collision is not textual. Two binaries, one from `b0eb133e` and one from `b9774c03`, same models directory, same `--proto-package`:

**master's baseline handed to this branch:**

```text
error: output file was not generated by ptah; delete it to start a new compatibility history, or point --out somewhere else
```

Every `.proto` written since #1192 reads as `ErrForeignFile` here, and the remedy printed to the operator is to delete the compatibility baseline.

**this branch's multi-file anchor handed to master's Ptah** — which is precisely "a Ptah without multi-file support":

```text
Exported Protobuf schema to …/v1/schema.proto
Exported 2 message(s), 5 field(s), 0 enum(s)
exit 0
```

`schema.proto` came back as a whole single-file baseline. `// ptah:protobuf-export-files=order.proto,product.proto` is gone, `Order` and `Product` are now declared in `schema.proto` *and* in the untouched `order.proto` and `product.proto`, and the directory no longer compiles:

```text
error: output file is not valid protobuf: acme/inventory/v1/order.proto:13:9: symbol "acme.inventory.v1.Product" already defined at acme/inventory/v1/product.proto:8:9
```

That refutes the PR body and the `:::caution` block added to `docs/site/src/content/docs/schema/protobuf.md`:

> Files of a multi-file set declare `ptah:protobuf-export-version=2`, a minimum reader version: a Ptah without multi-file support reads one member in isolation and refuses instead of rewriting it as a whole single-file baseline.

The gate is `version > exportVersion`. It fires only against a reader whose `exportVersion` is 1. Master's is 2, so `2 > 2` is false and nothing fires. A rebase alone does not fix this: the multi-file format needs a version master has not claimed, `parseHeader`/`render` have to adopt the foot layout, and the manifest line has to move with it.

## 2. `--proto-split=table` writes over files it did not write, with no gate at all

In single-file mode `--out` is the only path written, and if something is there it goes through `parseHeader` and the digest before anything is replaced. Under `--proto-split=table`, `publishProtobufSet` writes N−1 sibling paths that were never read, never stat'ed, and never validated.

A hand-written `order.proto` sitting in the package directory:

```text
$ cat proto/acme/inventory/v1/order.proto
// my hand written file, not generated by ptah
edition = "2023";

package acme.inventory.v1;

message HandWritten {
  string keep_me = 1;
}

$ ptah schema export --to protobuf --root-dir ./models \
    --out ./proto/acme/inventory/v1/schema.proto \
    --proto-package acme.inventory.v1 --proto-split table
warning: …/v1/schema.proto: no previous export found at …
Exported Protobuf schema to …/v1/schema.proto
Exported Protobuf schema to …/v1/order.proto
Exported Protobuf schema to …/v1/product.proto
Exported 2 message(s), 5 field(s), 0 enum(s) across 3 file(s)
bootstrapped new compatibility history
exit 0
```

`message HandWritten` is gone. Nothing in the output names `order.proto` as a file that already existed.

The compatibility-damaging variant is renaming `--out` inside an existing set. Given a set whose `Product.email` is pinned at 3 with 2 retired:

```text
=== product.proto BEFORE ===
message Product {
  int32 id = 1;
  string email = 3;

  reserved 2;
  reserved nickname;
}

$ ptah schema export … --out ./proto/acme/inventory/v1/api.proto --proto-split table
RENAMED_ANCHOR_EXIT=0

=== product.proto AFTER ===
message Product {
  int32 id = 1;
  string email = 2;
}
```

`email` moved from 3 back to 2 and both reservations were dropped, at exit 0, with the only warning naming `api.proto` — not `product.proto`, whose baseline was destroyed. Running through the original anchor again afterwards exits 0 and accepts the corrupted state permanently: `Product` now pins `email = 2` and there is nothing left to notice.

The PR body says "no file Ptah did not write is ever read or deleted". That is literally true and materially misleading — the file is *written over*, which is strictly worse than deletion, because a deletion is at least reported (`Removed …/product.proto`).

## 3. Pointing `--out` at a member of a set converts it into an anchor and bricks the set

```text
$ ptah schema export … --out ./proto/acme/inventory/v1/order.proto --proto-package acme.inventory.v1
Exported Protobuf schema to …/v1/order.proto
Exported 2 message(s), 4 field(s), 0 enum(s)
exit 0

$ head -2 proto/acme/inventory/v1/order.proto
// Code generated by ptah schema export --to protobuf; DO NOT EDIT.
// ptah:protobuf-export-version=1
```

`order.proto` dropped from version 2 to version 1 and now declares `Order` *and* `Product`, while `product.proto` still declares `Product` and `schema.proto` still lists both in its manifest. The next run through the anchor fails with `symbol "acme.inventory.v1.Product" already defined`, and the only recovery is deleting the whole compatibility history. A member carries no back-reference to its anchor, so nothing can detect this. It happens to be caught when the member imports the anchor — the dangling import fails to link — but only then; the import-free case above is silent.

## 4. Branches with no discriminator

`manifestNames` (`internal/protobufrender/digest.go:128-158`) is new and carries four distinct refusals — more than one manifest line, empty value, an entry that is not a plain `.proto` name, a duplicate entry. No test in the diff names any of them, and it runs on unauthenticated bytes (`readPreviousSiblings` calls it before `Render` reaches the digest gate), so it is the boundary deciding which files get opened. Same for the `messages map to the same protobuf file name` collision in `assignFiles` and the non-`ErrNotExist` branch of `readPreviousSiblings`. Nothing anywhere pins the behavior in (2) and (3): no fixture states whether Ptah may write a path it has not validated.

## What does hold

- M1 (`assignFiles` ignores `SplitTable`), M2 (`checkMoves` returns nil), M7 (`EqualFold` → `==`) reproduced; each reddens exactly the named test with the output quoted in the PR body. M7 reddens only the `same_name_other_case` subtest, so the case axis is genuinely separated.
- `go build`, `go vet`, `go test ./... -count=1`, `go tool qtlint ./...`, `golangci-lint run ./...` (0 issues), `check-test-style.sh` (OK, 175 conditional groups, 45 white-box files), `check-public-api.sh`, `check-public-api-snapshot.sh` — all exit 0 on this commit in isolation.
- Acceptance 1 and 2 over `./stubs`: one run wrote 17 files, `Exported 16 message(s), 93 field(s), 1 enum(s) across 17 file(s)`; the second run reproduced all 17 byte for byte (`diff -r` exit 0); `protoc 35.1` compiles the whole set together, exit 0.
- `buf` is not installed on this machine, so I did not re-run the `buf lint` / `buf format` / `buf breaking` half of acceptance 4 or `scripts/check-protobuf-export-acceptance.sh`, and I am not claiming anything about them either way.

Blocking on (1). (2) and (3) are new ways for the exporter to destroy pinned wire numbers at exit 0, which is the one outcome this target exists to prevent.

denisvmedia:
Rebasing this branch onto current master is not a merge conflict to resolve — it is a design collision, and resolving it hunk by hunk would silently undo #1148. Recommending re-derivation from master instead. Six conflict hunks across four files; here is what each one actually is.

## 1. The version field means two different things

```
master (#1192)                       this branch
  version 1 = header at the head       exportVersionSingle = 1  (one file)
  version 2 = header at the foot       exportVersionMulti  = 2  (a set)
                                       exportVersion = exportVersionMulti
```

Same field, two independent axes. A file declaring `2` would have to mean both "header at the foot" and "part of a set", and there is no spelling in this scheme for "single file, header at the foot" — which is what every export written today is.

This is sharper now than when the branch was cut: #1200 retires version 1 entirely, per Denis on #1148 — pre-general-availability, no legacy readers. With version 1 gone, `exportVersionSingle = 1` has nothing left to name, and the multi-file marker needs its own encoding rather than a version number. The manifest line this branch already adds is the obvious candidate: its presence is the fact, and no version number has to carry it.

## 2. The branch writes the header at the top of the file

`render.go:16-23` on this branch:

```go
sb.WriteString(generatedMarker + "\n")
sb.WriteString(versionPrefix + strconv.Itoa(version) + "\n")
sb.WriteString(digestPrefix + "\n")
if len(f.Siblings) > 0 {
    sb.WriteString(manifestPrefix + strings.Join(f.Siblings, ",") + "\n")
}
sb.WriteString("edition = \"2023\";\n\n")
```

That is the pre-#1148 layout, and taking this side of the conflict re-introduces the defect #1148 closed: the block becomes the file's leading detached comment, so `protoc-gen-go` copies it verbatim into the top of every `.pb.go`. The new **manifest** line rides along with it, putting the export set's file list into every consumer's generated Go.

Neither side of that hunk is right on its own. The manifest has to move to the foot with the rest of the header, and it has to stay covered by the digest — which the current `hasHeaderMarker` walks backwards from the end, so a fourth line in that block is fine, but the walk has to be told about it.

## 3. `previousFile` versus `previousSet`

Master's `previousFile` carries a `Version` field, kept so the caller can report which layout it read. This branch replaces the type with `previousSet` and drops that field. Reconciling those means deciding what a set of files declares as a whole — one version for the set, or per file — which is the same question as item 1 and cannot be answered inside a conflict marker.

## Recommendation

Re-derive from current master, after #1200 lands, with these decisions made explicitly:

- [ ] The multi-file marker is the manifest line's presence, not a version number, unless there is a reason a reader needs to refuse before parsing the manifest.
- [ ] The manifest is written in the foot block with the other generated lines, and the digest still covers it. Verify with `protoc-gen-go` that it does not reach the `.pb.go` — `grep -c 'ptah:' *.pb.go` must print 0, the assertion `scripts/check-protobuf-export-acceptance.sh` already makes.
- [ ] Whether one file of a set read in isolation is refused, and by which sentinel. The branch's stated reason for `exportVersionMulti` is real — a reader without set support would rewrite one member as a whole single-file baseline — so whatever replaces the version number must still produce that refusal, with a test that reverts it and watches the refusal disappear.

The measured work in this branch — the 307-line `split.go`, the acceptance additions — should carry over largely intact; it is the header/version plumbing that has moved underneath it.

I am not closing this PR: unlike #1170 the body of the work is still applicable, so it is a rebase-with-decisions rather than a rewrite. Say the word and I will do the re-derivation.

denisvmedia:
I like the overall direction. Tracking the export as an explicit file set instead of scanning a directory is a good design choice.
One thing I'd like to understand better is the publication model.
Atomic rename solves a single file, but here the compatibility state is an entire set of files. What guarantees do we have if the process is interrupted halfway through publishing?
For example:
anchor written, members not yet written;
some members updated, others still old;
obsolete members deleted before the new set is fully published.
It may already be handled internally, but I think the PR would benefit from documenting the atomicity guarantees of the multi-file publish step.

denisvmedia:
Good question, and the answer is worse than the code comment implies. I read the publish path on `b9774c03` and measured each of your three cases. Short version: **one of the three is genuinely guaranteed, the other two are not, and one of them silently reallocates field numbers** — the exact damage the whole target exists to prevent.

## What is actually guaranteed

Per *file*, the guarantee is real. `writeProtobufAtomically` does `CreateTemp` in the destination directory → `Chmod` → `Write` → `Sync` → `Close` → `fsdurable.ReplaceFile` → `SyncDir`. No reader ever sees a half-written `.proto`, and a failed render never truncates a baseline.

Across the *set*, there is no atomicity at all. `publishProtobufSet` is three sequential phases, each a loop of independent renames:

1. every non-anchor member, in sorted order;
2. the anchor;
3. `os.Remove` for each superseded member.

Measured mtimes from one 4-file export (`stat -f "%Fm %N"`):

```
1786050912.267466369  invoice.proto
1786050912.286519790  order.proto
1786050912.295873231  product.proto
1786050912.308858426  schema.proto   <- anchor, last
```

So the anchor lands ~41 ms after the first member here, and that window grows with the table count. Every state I describe below is a state reachable inside that window. I reproduced each one by constructing it on disk rather than by signalling the process mid-run, but the ordering above is what makes each one reachable.

## Case 1 — anchor written, members not yet: guaranteed

This one is handled, and deliberately. The anchor is written *last* precisely so its manifest never names a file that is not on disk. Your first case therefore cannot arise from an interrupted publish.

And if it arises another way (a partial checkout, someone deleting a member), the next run refuses loudly rather than renumbering:

```
error: .../schema.proto lists product.proto as part of its export set, but that file is missing;
restore it from version control, or delete the whole set to start a new compatibility history
EXIT=2
```

There is a test for it (`TestSchemaExportProtobufSplitRefusesAMissingMemberOfTheSet`). No complaint here.

## Case 2 — some members updated, others still old: not handled

The digest is **per file**. `computeDigest` hashes one file's own bytes; the anchor's manifest records sibling *names* only, no digests. So every file in a torn set validates individually and the set-level inconsistency is invisible by construction. Two shapes, with very different outcomes.

**2a. The tear leaves a dangling cross-file reference.** Generated enums live only in the anchor, and the anchor is written last — so an interrupted run that adds an enum column *always* has this shape. Member on disk referencing an enum the old anchor does not declare yet:

```
$ protoc --proto_path=. ... acme/inventory/v1/product.proto
acme/inventory/v1/product.proto:13:3: "ProductStatus" is not defined.
EXIT=1
```

The next Ptah run does refuse — but incidentally, and with the wrong diagnosis:

```
error: output file is not valid protobuf: acme/inventory/v1/product.proto:13:3:
field acme.inventory.v1.Product.status: unknown type ProductStatus
EXIT=2
```

That is the `protocompile` step tripping over a dangling reference, not an atomicity check. The file is not malformed; the *set* is torn. The message carries no remedy, and the fix that actually works (re-run, or restore the whole set) is not stated.

**2b. The tear adds a member the anchor has not learned about yet. This is silent and permanent.**

A new member file is renamed into place before the anchor's manifest names it. Crash in that window and the member is an orphan: `readPreviousSiblings` only reads names from the manifest, so the orphan's pinned numbering is never read back, and the message is treated as brand new.

Measured. `Invoice` pinned `alpha = 1; beta = 2`. Crash before the anchor write. Then the schema drops `alpha` and adds `gamma`, and the same command is run against the intact set and the torn set:

```diff
--- proto-control/acme/inventory/v1/invoice.proto
+++ proto-torn/acme/inventory/v1/invoice.proto
 message Invoice {
-  string beta = 2;
-  string gamma = 3;
-
-  reserved 1;
-  reserved alpha;
+  string beta = 1;
+  string gamma = 2;
 }
```

Both runs exit 0. The torn run emits no diagnostic at all. `beta` silently moved from field 2 to field 1 and took the number `alpha` had retired — and because both are `string`, a consumer holding the pre-crash numbering does not get a decode error, it reads the `beta` value as `alpha`. Silent misfiling, not a visible failure.

Nothing recovers from this. Once the second run writes an anchor whose manifest names `invoice.proto`, the wrong numbering is the pinned baseline.

## Case 3 — obsolete members deleted before the set is published: the stated order is safe, the inverse is not

Deletions run last, after the anchor, so your literal case cannot happen. But the deletion loop is itself not atomic and runs *after* the anchor has stopped naming those files. Interrupt mid-loop and the survivors are orphans Ptah can never see again.

Collapsed a 4-file set to single-file with `--proto-on-type-move=relocate`, then restored two members (a crash after the first `os.Remove`):

```
$ protoc --proto_path=. acme/inventory/v1/{schema,order,product}.proto
acme/inventory/v1/order.proto:8:9: "acme.inventory.v1.Order" is already defined in file "acme/inventory/v1/schema.proto".
EXIT=1
```

Re-running the export does not repair or even mention it:

```
Exported Protobuf schema to .../schema.proto
Exported 3 message(s), 6 field(s), 0 enum(s)
EXIT=0
```

Orphans still present, directory still does not compile. This one is permanent by construction: the new anchor is `version=1` with no manifest line, so the record of the old set is gone. There is no state left from which Ptah could know those files were ever its own.

## The gap in the reasoning, not just the code

The doc comment on `publishProtobufSet` is accurate about the invariant it holds:

> the old inventory never names a file that is not on disk

That is true, and it is what makes case 1 safe. What it does not say is that the invariant is one-directional. It prevents the manifest from *over*-naming; it says nothing about *under*-naming, and under-naming is the direction that renumbers silently (2b) and orphans permanently (3). The PR would benefit less from documenting the current guarantees than from noting that two of the three cases you asked about have no guarantee to document.

Minor, while it is in view: on a hard kill between `CreateTemp` and the rename, `.<name>.proto.NNNN.tmp` stays in the output directory forever. It is dot-prefixed and `.tmp`-suffixed so `*.proto` globs miss it, but nothing ever cleans it up.

Also worth flagging for the re-cut: `writeProtobufAtomically` calls `fsdurable.ReplaceFile`, the unconditional variant, even though `internal/fsdurable` already exposes a conditional `PublishFileAt` that binds the expected destination state and returns `ErrDestinationChanged` instead of clobbering. Multi-file publish is exactly where that precondition earns its keep.

## Coverage

`publishProtobufSet` has no direct test. Of the ten tests in `export_protobuf_split_test.go`, one covers a missing member (case 1); none constructs a torn publish or an orphaned member.

## What the re-derivation has to guarantee

This branch cannot be rebased as it stands — it spends the export version field on single-vs-multi while master already spends it on header layout (details in https://github.com/stokaro/ptah/pull/1199#issuecomment-5207444545). Since it is being re-cut anyway, the atomicity work belongs in the re-derivation rather than as a patch here. I have written this list onto #1146 so it is found when the work is picked up; restating it here:

1. **No interrupted publish may ever reallocate a field number.** This is the one invariant the target exists to hold. Every torn state must end in either identical numbering or a non-zero exit. Exit 0 with a silent renumber (2b) has to be unreachable.
2. **The anchor must pin member *content*, not just member names** — a digest per member, or one set-level digest. That is what makes a stale or newly-torn member detectable instead of individually-valid.
3. **A member Ptah wrote must be recognizable even when the manifest does not name it**, or orphans stay invisible. Matching Ptah's own generated marker plus the declared package is much narrower than the directory scanning this PR rightly rejected, and it is the direction that closes both 2b and 3.
4. **Deletion needs an intent record.** Write the new anchor still naming the files pending removal, delete, then rewrite the anchor without them — so an interrupted delete is a state the next run can finish rather than an orphan set it cannot see.
5. **Say which contract is on offer.** POSIX gives no multi-file atomic rename, so "the next run deterministically repairs any torn state" and "a torn state is refused with a restore-from-version-control remedy" are both defensible. Silence is not. Whichever is chosen should be stated in the docs page, because callers commit this output.
6. **Test it by injecting a failure between each pair of publish steps**, asserting the next run either reproduces byte-identical output or exits non-zero. That is the shape that would have caught 2b.

## What I did not run

I constructed the reachable interruption states on disk rather than killing the process mid-publish with a real signal; the mtime measurement above is what establishes those states are reachable. `protoc` is 35.1. `buf` is not installed on this machine, so `buf lint` / `buf breaking` over the torn sets was not run — only `protoc` compilation.

denisvmedia:
Adversarial re-check of the atomicity answer above (https://github.com/stokaro/ptah/pull/1199#issuecomment-5209192818). I rebuilt `b9774c03`, wrote my own fixtures, and re-derived all three cases rather than re-reading the earlier ones. **The answer holds. Every load-bearing claim reproduces.** One real rule violation and one overstatement, both listed at the end.

## Verified against the source

`origin/issue-1146-protobuf-split-export` is `b9774c03`, worktree clean, 0 commits ahead — no code change was made, which is correct given the version collision.

- `publishProtobufSet` (`cmd/schema/export_protobuf.go:355`) is three phases exactly as described: non-anchor members, then the anchor, then `os.Remove` of superseded members.
- Per-file chain is as quoted: `CreateTemp` → `Chmod` → `Write` → `Sync` → `Close` → `fsdurable.ReplaceFile` → `SyncDir`.
- **The central claim — `computeDigest` hashes one file's own bytes, and the manifest carries names only.** Confirmed in `internal/protobufrender/digest.go` and in the generated bytes:

```
// ptah:content-sha256=6a9aacb54fed8e4a03a28a66ab5bf48abaf0814ce4817041da1080fb5e1f9d86
// ptah:protobuf-export-files=order.proto,product.proto
```

No digest anywhere in the manifest line. Every member carries its own `content-sha256` over its own bytes. A torn set is individually valid by construction, as stated.

- Ordering re-measured independently on a 4-file export (`stat -f "%Fm %N"`):

```
1786051665.978519488  invoice.proto
1786051665.991327892  order.proto
1786051666.001271248  product.proto
1786051666.010309227  schema.proto   <- anchor, last
```

Members alphabetical, anchor last, window ~32 ms here against the ~41 ms quoted. Same ordering, same magnitude.

- `publishProtobufSet` appears only at `export_protobuf.go:226/343/355` — no test file references it. `export_protobuf_split_test.go` has exactly 10 `func Test`. Both coverage claims are accurate.
- `internal/fsdurable` does expose the conditional `PublishFileAt` (`doc.go:4`) while the branch uses unconditional `ReplaceFile`. The flag for the re-cut is correct.

## Re-measured, all exit codes captured unpiped

I redirected to files and read `$?` on the next line in every case, so none of these are a pipeline's status.

**Case 1 — anchor names a missing member.** `EXIT=2`, verbatim:

```
error: .../schema.proto lists product.proto as part of its export set, but that file is missing;
restore it from version control, or delete the whole set to start a new compatibility history
```

**Case 2a — dangling cross-file reference.** `EXIT=2`, diagnosed only by `protocompile`:

```
error: output file is not valid protobuf: acme/inventory/v1/order.proto:12:3: field acme.inventory.v1.Order.visibility: unknown type Visibility; acme/inventory/v1/product.proto:12:3: field acme.inventory.v1.Product.visibility: unknown type Visibility
```

No remedy clause, unlike case 1. Confirmed. I also checked the supporting claim that enums always land in the anchor: an enum referenced by **one** table still goes to `schema.proto` with the member importing it, so the member→anchor edge exists regardless of sharing.

**Case 2b — the silent renumber.** This is the finding that matters, so I built the discriminator explicitly rather than taking it on trust. Identical schema change (`nickname` dropped), identical command, **only the anchor's manifest differs**:

| | manifest | result | exit | stderr |
|---|---|---|---|---|
| control (intact) | `order.proto,product.proto` | `email = 3` + `reserved 2; reserved nickname;` | 0 | empty |
| torn (under-named) | `order.proto` | `email = 2`, no reservations | 0 | empty |

The torn state was produced faithfully: run the real export, then restore the anchor to its pre-run bytes — which is exactly "members renamed, anchor not yet renamed". `email` moved from 3 to 2 and both reservations vanished, at exit 0 with nothing on stderr and no `bootstrapped` line. Permanent once the next anchor names the file. **Confirmed as described.**

**Case 3 — interrupted deletion.** Confirmed on both variants. Dropping a table and interrupting before `os.Remove`: re-run is `EXIT=0`, orphan never mentioned, still on disk. Re-adding the table later over the orphan renumbers `email` 3 → 2 at `EXIT=0`, empty stderr — the same silent renumber compounding. On the split→none collapse the anchor does come back as `ptah:protobuf-export-version=1` with no manifest line, and:

```
$ protoc --proto_path=. acme/inventory/v1/schema.proto acme/inventory/v1/product.proto
acme/inventory/v1/product.proto:8:9: "acme.inventory.v1.Product" is already defined in file "acme/inventory/v1/schema.proto".
PROTOC_EXIT=1
```

Re-running the export: `EXIT=0`, silent, orphan still present. "Permanent by construction" is accurate. `protoc` here is `libprotoc 35.1`, matching what the answer stated.

## Behavior vs intent

I checked specifically for intent standing in for behavior. The answer does not do this. Case 1's "the anchor is written last precisely so..." is a design statement, but it is backed by a measured exit 2 and a named test, and the reasoning section correctly separates the invariant the doc comment *holds* (no over-naming) from the one it does not (under-naming). The "unhandled" cases are called unhandled plainly, not described as intended-to-be-handled.

## Two things to fix

1. **American English.** "rather than by **signalling** the process mid-run" — should be `signaling`. Probed the full prose for 16 British/American divergences with the probe validated against a positive control first; this is the only hit. Repo-banned tokens and AI attribution: all zero.

2. **"always" is slightly too strong** in 2a: "an interrupted run that adds an enum column *always* has this shape". An interrupt landing before the first referencing member is renamed leaves a clean, untorn state. It should read "any tear after the first referencing member is written". The mechanism and the conclusion are unaffected.

Neither changes the verdict. One of three guaranteed, 2b silent and permanent, 3 permanent and invisible — all reproduced independently. Acceptance 5-10 on #1146 is the right place for the requirement given the re-cut.

denisvmedia:
Rebuilt on top of `master` (`272691c6`). Three changes had landed underneath this branch and two of them collided with it head-on, so this is a re-derivation rather than a merge.

**#1148 moved the header block to the foot of the file.** This branch wrote the manifest as a fourth line of a header block at the *top*. The manifest now goes where the rest of the header went, via `renderHeader(siblings)` — still inside the digest-covered block, which is the property it needs.

**#1200 retired export version 1.** This branch used the version number as a set marker: 1 for a single file, 2 for every file of a multi-file export, so a Ptah without multi-file support would refuse a member read in isolation. That refusal was aimed at an older Ptah, and compatibility with an older Ptah is not owed before v1 (AGENTS.md, "Compatibility with older Ptah is a different axis"). It is also the exact trade #1200 already made when it deleted the v1 reader instead of carrying it.

So `exportVersionSingle`/`exportVersionMulti` are gone and `exportVersion = 2` is the single constant `master` has. The manifest line — not the version number — is what marks a set. `parseHeader` keeps `master`'s strict `!= exportVersion` refusal rather than this branch's looser `> exportVersion`, which would have accepted the retired v1 again.

**#1203 added `--proto-comments`** to the same three files. Both sides are additive there and both are kept.

One claim in the docs needed a test rather than an assertion, so `TestSchemaExportProtobufSplitDigestCoversTheManifest` is new: it drops `product.proto` from the anchor's manifest line and measures that the next run refuses with `ErrModified` and leaves every file byte-identical, instead of reading the survivors as a smaller set and restarting the dropped file at field number 1. Mutating `computeDigest` to strip manifest lines the way it strips the digest line reddens it.

Gates run on the merged tree: `go build`, `go vet`, `go test ./... -count=1`, `qtlint`, `check-test-style.sh`, `check-public-api.sh`, `check-public-api-snapshot.sh`, `golangci-lint`, all seven `check:*` docs scripts and `build-feature-matrix.mjs --check`. `check:responsive` needs a built `dist` and the protobuf export acceptance script needs `buf`, which is not installed here — both run in CI.

PR #1200, opened 2026-08-06
Retire the version 1 protobuf export layout instead of reading it

denisvmedia:
Per Denis on #1148: Ptah is pre-general-availability, so we do not carry legacy. #1148 introduced format version 2 while keeping a reader for version 1; this removes the version 1 reader entirely.

## What changes

- `exportVersion` stays `2` and is now the **only** version read or written. `parseHeader` refuses anything else, where it previously refused only *higher* versions.
- The migration path and its warning are gone.
- `headerLayoutVersion` is gone.

**Renumbering to 1 was considered and rejected.** A real version 1 file declares `version=1`; if the current format also claimed 1, that file would pass the version gate and then be misread. Keeping the number at 2 means the version line keeps meaning exactly one thing.

## The refusal is by version, not as a foreign file

This is the substance of the change rather than a detail. Version 1 put the generated header at the **head** of the file, so the obvious way to retire it — stop recognizing the marker in that position — makes a version 1 file read as *foreign*:

> `output file was not generated by ptah; delete it to start a new compatibility history, or point --out somewhere else`

That is a true sentence about the wrong problem, and it sends the reader hunting for whoever else wrote their `.proto`. So `hasHeaderMarker` still recognizes the retired position, and the diagnosis lands on the version instead:

```
$ ptah schema export --to protobuf --root-dir ./stubs --out <a version 1 file> …
rc=2
error: unsupported ptah protobuf export version: file declares export version 1,
this Ptah reads and writes only 2; delete the file and export again to start a
new compatibility history

$ cmp before.proto after.proto && echo untouched
untouched
```

Measured with a binary built from this branch, over `./stubs`, on a baseline produced by `scripts/downgrade-protobuf-export-to-v1.py` from the current exporter's own output.

## Mutants — both directions

The refusal test needs both, because reverting a guard cannot make its non-interference control fail.

**Accept version 1 again** (`version != exportVersion` → `version > exportVersion`):

```
--- FAIL: TestPreviousRefusesTheRetiredVersionOneLayout
    want: e"unsupported ptah protobuf export version"
```

**Stop recognizing the marker in the retired position** — the inverse mutant, which proves the control is not decorative:

```
--- FAIL: TestPreviousRefusesTheRetiredVersionOneLayout
    wanted error is not found in error chain
    got:  e"output file was not generated by ptah; delete it to start a new compatibility history, …"
    want: e"unsupported ptah protobuf export version"
```

Exactly the wrong-remedy failure the test exists to prevent. Both restore green.

## A fixture the #1192 review flagged is retired with it

That review found the acceptance script's version 1 migration case could not discriminate: its baseline was a fresh bootstrap of the same schema, so "every number survived the rewrite" and "numbers reallocated from 1" produce identical bytes, and `cmp` plus `buf breaking` passed either way. The case is now a refusal case, which has no such coincidence — the export either runs or it does not, and the script additionally asserts the file was left byte-identical and that the message is *not* the foreign-file one.

The `downgrade-protobuf-export-to-v1.py` helper stays: building the retired shape is the only way to test the refusal.

## Gates

```
```

All 14 `check:*` scripts in `docs/site/package.json` pass, enumerated from `package.json` rather than from memory — that is how `check:limitations` got run — plus `node docs/site/scripts/build-feature-matrix.mjs --check` and `npm run build`.

`scripts/check-protobuf-export-acceptance.sh` could not run locally: `buf` is not installed on this machine, and the script exits 1 naming it. The block this PR changes was therefore executed by hand with the real binary, which is the transcript quoted above; CI's `Export Acceptance` job runs the whole script.


denisvmedia:
I like this direction.
For a pre-GA project, removing legacy readers instead of accumulating compatibility code keeps the implementation much simpler, and I agree that diagnosing the retired format as "unsupported version" is significantly better than treating it as a foreign file.

PR #1201, opened 2026-08-06
Resolve the PostgreSQL connection schema from search_path

### LINKED ISSUE #1198: Dev-database preparation drops the schema the dev URL selects, then the replay fails with no schema to create in
Found while building an over-silencing control for #1197. **Pre-existing** — the master binary and that branch produce byte-identical output, so nothing in flight caused it.

## Reproduction

A PostgreSQL dev database that has schema `app`, a dev URL that selects it, and migrations whose DDL is unqualified so it should land there:

```sql
-- 20260101000000_base.sql
CREATE TABLE "Users" (id int);
CREATE TABLE audit_log (id int);
-- 20260101000001_drop.sql
DROP TABLE "Users", audit_log;
```

```
$ atlas       migrate lint --dir file://mig --dev-url 'postgres://…/dev?search_path=app' --latest 1
rc=1   -- 1 version with errors / 2 schema changes / 2 diagnostics   (2 × DS102)

$ ptah-compat migrate lint --dir file://mig --dev-url 'postgres://…/dev?search_path=app' --latest 1
rc=1   0 diagnostics
Error: error validating migration SQL on dev database: replay migration 20260101000000
on dev database: failed to execute migration SQL: SQL execution failed:
ERROR: no schema has been selected to create in (SQLSTATE 3F000)
SQL: CREATE TABLE "Users" (id int)
SQL: CREATE TABLE "Users" (id int)
```

The community binary honors `search_path=app` on the replay connection, so the unqualified `CREATE TABLE` lands in `app` and the lint proceeds. Ptah does not, so the very first statement fails and no analysis happens at all.

## Why it matters, and why it is not a drop-in blocker

Both exit 1, so ptah never exits 0 where the binary exits 1 — this is the stricter direction. But the failure is total rather than partial: a project whose dev URL selects a non-default schema gets **no lint at all**, with an error that names a SQL problem rather than the configuration difference that caused it. A user reading that message will look at their migration, which is fine.

It also sits directly under #1074's S1 fix (#1197): that change reads the schema scope the dev URL covers, while the replay that builds the snapshot it analyzes ignores the same setting. The two should agree about what the dev URL selects.

## Definition of done

- The replay connection applies the dev URL's `search_path` (and the equivalent selector on other engines), so the fixture above lints instead of erroring.
- A fixture pinning it, with the **negative control** kept: a dev URL selecting a schema that does not exist must still fail, and fail naming the schema rather than the SQL.
- The `public`-schema cells stay byte-identical and keep exiting 1 — a change to connection setup is exactly the kind of edit that silently moves them.

## Also visible in this output

The `SQL:` line is printed twice for one statement, and the driver prefix wraps the server message. Same shape as #1196; fixing either should check the other.

denisvmedia:
Closes #1198.

`info.Schema` for PostgreSQL was the constant `"public"` — the code assigned `"public"`, then branched on the URL path to assign `"public"` again, under a comment saying the schema "is usually specified via search_path or defaults to 'public'". The search_path was never read.

## Why that is worse than ignoring a flag

A writer decides from `info.Schema` which schemas belong to the caller and which are strangers it may drop. So a dev URL carrying `?search_path=app` did not merely get ignored — `DropDatabaseRealm` treated `app` as a stranger and **dropped a schema Ptah did not create**, then the replay ran under a search_path resolving to nothing and failed on the first unqualified `CREATE TABLE`:

```
ERROR: no schema has been selected to create in (SQLSTATE 3F000)
SQL: CREATE TABLE "Users" (id int)
```

A message that describes neither step, so the operator goes and looks at their migration.

## Measured

Same database, same fixture, schema counted before and after each run, pinned community binary v1.3.0 by absolute path:

| | `app` before → after | result |
| --- | --- | --- |
| `ptah-compat`, master | **1 → 0** | exit 1, 0 diagnostics, the error above |
| pinned binary | 1 → 1 | exit 1, 2 × DS102, 2 schema changes |
| `ptah-compat`, this branch | **1 → 1** | exit 1, 2 × DS102, 2 schema changes |

## The fix

`SELECT current_schema()` — what the server resolves `search_path` to, which is what the URL selects.

`current_schema()` is NULL when `search_path` names only schemas that do not exist. `"public"` is kept for that case: it is what every caller assumed before, and inventing a refusal there would change far more than the drop. The third row of the new resolution test pins that fallback.

## Mutants

**Direct** — hardcode `"public"` again:

```
--- FAIL: TestPostgresConnectionResolvesTheSearchPathSchemaE2E/a_search_path_naming_another_schema_resolves_to_it
--- FAIL: TestPostgresRealmCleanupKeepsTheSelectedSchemaE2E
```

**Inverse, and the reason the second test's comment says what it says.** The realm test also asserts a `bystander` schema is still cleaned. That is a non-interference claim, so I ran the inverse mutant to see whether it discriminates — spare every schema but the root — and it does **not** belong to the assertion I wrote it for. It is caught one line earlier, by the production code's own verification:

```
e`PostgreSQL database realm cleanup left residual user schema "bystander"`
```

returned from `DropDatabaseRealm`, which the test already asserts is nil. So the bystander is there to give the cleanup something to do and make that verification reachable; the count restates the outcome for a reader rather than adding coverage the error check does not already have. The comment in the file says exactly that instead of claiming a control it cannot prove.

(The first inverse attempt — drop nothing at all — reddens the test too, but through `failed to recreate root schema "app": schema "app" already exists`, a third path again. Recorded because it is the kind of thing that reads as confirmation if you stop at "the test went red".)

## Controls

With no `search_path`, and with `search_path=public`, `migrate lint` output is byte-identical to master modulo elapsed durations, same exit code, and the realm is still cleaned to `public` alone.

## Gates

```
integration, -tags integration, live PostgreSQL 17: the four new cells pass
```

One thing worth recording about that unit run: with `POSTGRES_TEST_DSN` exported for the integration tests, `TestMigrateGenerateShadowVerificationWithRealDB` fails with `shadow database must be distinct from target database`. That is my shell pointing target and shadow at one database, not this change — unset it and the suite is clean. Checked rather than assumed, because "a test failed after my change" is the easiest thing to misattribute in either direction.

**CI note:** GitHub Actions has been failing to resolve action download info repo-wide for several hours (`Service Unavailable` / `Bad Gateway` at `Set up job`, before any code runs), including on master with no code change. This PR should not be merged until CI can produce a green that means something; the evidence above is local.

### DIFF
--- a/dbschema/connection.go
+++ b/dbschema/connection.go
@@ -749,17 +749,27 @@ func getDatabaseInfo(ctx context.Context, db *sql.DB, dialect string, parsedURL 
 		// resolving to nothing (stokaro/ptah#1198).
 		//
 		// current_schema() is NULL when search_path names only schemas that do
-		// not exist. "public" is kept for that case: it is what every caller
-		// assumed before, and inventing a refusal here would change far more
-		// than the drop.
+		// not exist. That is refused rather than folded back to "public": a
+		// caller who named a schema and silently got a different one is the
+		// failure this whole change is about, and answering "public" would
+		// resume dropping the schemas that one does not cover. Ptah is
+		// pre-general-availability, so the previous fallback is not owed
+		// compatibility.
+		//
+		// The message names the schema, because the operator's mistake is in the
+		// URL and nothing downstream can say so: without this, the run reaches
+		// the replay and fails on a CREATE TABLE with "no schema has been
+		// selected to create in", which sends them to their migration.
 		var currentSchema sql.NullString
 		if err := db.QueryRowContext(ctx, "SELECT current_schema()").Scan(&currentSchema); err != nil {
 			return info, fmt.Errorf("failed to resolve PostgreSQL current schema: %w", err)
 		}
-		info.Schema = "public"
-		if currentSchema.Valid && currentSchema.String != "" {
-			info.Schema = currentSchema.String
+		if !currentSchema.Valid || currentSchema.String == "" {
+			return info, fmt.Errorf(
+				"database URL selects schema %q, which does not exist in this database",
+				postgresSearchPathSelection(parsedURL))
 		}
+		info.Schema = currentSchema.String
 
 	case platform.MySQL, platform.MariaDB:
 		// Get MySQL/MariaDB version
@@ -1029,4 +1039,19 @@ func removePostgresPoolParams(dbURL string) string {
 	q.Del("pool_min_conns")
 	parsedURL.RawQuery = q.Encode()
 	return parsedURL.String()
+}
+
+// postgresSearchPathSelection returns the schema the URL's search_path names, so
+// a refusal can quote the operator's own value back at them. It is only ever
+// called on the refusal path, where current_schema() resolved to nothing.
+//
+// An absent search_path cannot reach that path: PostgreSQL's own default
+// resolves to "public", which exists in every database Ptah connects to, so the
+// empty string here means the URL carried no selection and the server still
+// answered nothing -- a shape worth reporting verbatim rather than guessing at.
+func postgresSearchPathSelection(parsedURL *url.URL) string {
+	if parsedURL == nil {
+		return ""
+	}
+	return parsedURL.Query().Get("search_path")
 }
--- a/dbschema/connection.go
+++ b/dbschema/connection.go
@@ -740,14 +740,26 @@ func getDatabaseInfo(ctx context.Context, db *sql.DB, dialect string, parsedURL 
 		info.Dialect = detectPostgresWireDialect(dialect, version)
 		info.IdentifierSemantics = identifier.ForDialect(info.Dialect)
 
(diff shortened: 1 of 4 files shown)


denisvmedia:
We should be strict about what issue requests. The or is not following it fully in case of non-existing search path. No backwards compatibility is needed.

denisvmedia:
Nice fix. I think one product-level decision is still worth reconsidering.
current_schema() returning NULL means the session does not currently resolve to an existing schema. Falling back to "public" silently changes the user's explicit configuration (search_path=...) into a different execution context.
Since Ptah uses info.Schema to decide ownership during realm cleanup and replay, I think failing fast would be a safer contract than silently switching to public.
In a pre-GA product I'd rather see:
current_schema() resolves → continue;
current_schema() is NULL → return a clear configuration error.
That avoids potentially operating on the wrong schema while also making configuration mistakes immediately visible.

PR #1204, opened 2026-08-06
probe: does a fresh PR get Actions check-suites

denisvmedia:
Throwaway probe to establish whether the push-to-Actions path is working for new pull requests. Will be closed and the branch deleted as soon as it answers.


denisvmedia:
Probe answered: a brand-new PR gets only a `renovate` check-suite and no `github-actions` one, so the push-to-Actions path is broken repo-wide right now rather than specific to any branch. Re-creating stalled PRs on fresh branches would not have helped. Closing; branch deleted.

Issue #1042, opened 2026-08-02
atlas.hcl file() sandbox is stricter than CE: CE reads absolute and parent-traversal paths, Ptah refuses


denisvmedia:
Surfaced during the second review pass on #1040. **Pre-existing** — not introduced there — but it needs a decision, because it is a genuine tension between the governing CE rule and a security control.

## Measured

Pinned CE v1.3.0 binary vs `ptah-compat`, `file()` inside `atlas.hcl`:

| Expression | CE | ptah-compat |
| --- | --- | --- |
| `file("local.txt")` | 0 | 0 |
| `file("/etc/passwd")` | **0** | **1** — `absolute paths are not supported` |
| `file("../../../../etc/passwd")` | **0** | **1** — `path escapes atlas.hcl directory` |
| `getenv("HOME")` | 0 | 0 |

CE reads absolute paths and parent-traversal paths out of an `atlas.hcl` without complaint. Ptah sandboxes `file()` to the config file's directory and refuses both.

## Why it was invisible until now

In a *known* position (`env { url = file("/etc/passwd") }`) master already refused, so the divergence existed. In an *unknown* position it was masked: master rejected the whole construct as an unsupported name before ever evaluating the body. #1040 makes unknown top-level names tolerated with their bodies still evaluated — matching CE — so the sandbox refusal is now what surfaces there.

**The exit code is unchanged either way (1 on both master and #1040), so this is not a regression.** What changed is that the reason is now accurate and visible.

## The tension

The standing rule is *"Atlas CE — full drop-in, no exceptions. A measured CE behavior is the contract, even when it is a footgun."* Applied literally, compat should read `/etc/passwd` when an `atlas.hcl` asks it to.

Against that: an `atlas.hcl` is frequently repository-controlled, and `file()` evaluation happens before anything is applied. Removing the sandbox means a config file in a pull request can read any file the process can, and place its contents somewhere observable — a database URL, an error message. That is a different class of cost from the footguns the rule was written about (a typo'd block doing nothing, a baseline silently skipped).

## Options

1. **Keep the sandbox on both surfaces**, and record it as a deliberate CE divergence with its reason. Compat stays stricter in exactly one respect, loudly.
2. **Surface split**, per the existing principle: `ptah-compat` matches CE (sandbox off) so drop-in fidelity is exact, native `ptah` keeps the sandbox as a documented advantage. This is what was done for failed `migrate down` bookkeeping — but there the worse behavior was *bookkeeping*, not file access.
[...]

Issue #1186, opened 2026-08-06
A --dir without a scheme is accepted everywhere the pinned binary refuses it, and writing verbs create the directory


denisvmedia:
Found while reviewing #1172, and deliberately filed separately rather than fixed inside it — the class is pre-existing and uniform across verbs, while that PR only newly *reaches* it.

## The rule

`internal/atlasargs/args.go`, `ParseLocalDir`: *"Plain filesystem paths are preserved verbatim."* The pinned community binary v1.3.0 requires the scheme and says so:

```
$ atlas migrate hash --dir mig --dir-format goose
exit 1   missing scheme for dir url. Did you mean "file://mig"?
```

`ptah-compat` accepts it. Measured on master, all exit 0 where the binary exits 1: `migrate hash`, `migrate validate`, `migrate status`, `migrate lint`, and `migrate new --dir mig --dir-format atlas`. Positive control: `--dir file://mig` exits 0 on both.

That is parity rule (a) — exit 0 where the binary exits 1 — on five verbs.

## Why the writing verbs are the sharp end

Reading a directory under a lenient spelling is a divergence. **Creating** one is a side effect:

```
$ ptah-compat migrate new addcol --dir /tmp/nosuchxyz --dir-format goose
exit 0, and it wrote to /private/tmp/nosuchxyz
```

The binary exits 1 and writes nothing. So a typo'd or scheme-less `--dir` on `migrate new` or `migrate diff` silently materialises a migration directory somewhere the user did not point at — outside the working directory in the case above, because the bare path resolved through the system temp symlink.

## What the fix has to decide

Whether `ParseLocalDir` should require the scheme at all, or only on the verbs that write. Requiring it everywhere matches the binary exactly and is the smaller rule, but it is a breaking change for anyone whose scripts pass a bare path today — and this repository's own tests and probe scripts do so in places, so the blast radius is measurable before deciding rather than guessable.

Requiring it only where a directory can be created is narrower and closes the destructive half immediately. It leaves the read verbs diverging, which is honest as long as it is written down rather than left as an accident of one helper.

Completion criterion either way: `migrate new --dir mig --dir-format goose` and `migrate diff --dir mig …` exit non-zero with the binary's message and create nothing, and `--dir file://mig` keeps working on every verb.

## Not in scope here

[...]

Issue #1143, opened 2026-08-03
Epic: complete the OCI artifact lifecycle and promotion path


denisvmedia:
#928 fixes unevenness in the OCI surface that exists. This epic covers what does not exist: the lifecycle between publishing an artifact and consuming it.

## The shape of the gap

OCI is implemented as **transport for individual commands**, not as an artifact lifecycle. Measured on a binary built from master:

```
$ ptah oci --help
Available Commands:
  referrers   List metadata attached to an OCI artifact
```

One verb. `push` and `pull` live in `ptah migrations` and `ptah schema` instead, so there is no namespace that owns an artifact between the moment it is published and the moment something consumes it.

The consequence is concrete rather than aesthetic: **Ptah cannot read the reports Ptah writes.** It publishes `lint.json`, `plan.json` and `deployment.json` as referrers, and `ptah oci referrers` returns descriptor metadata only. Reading back a report Ptah itself produced requires `oras` or a raw registry call.

### 1. Fetch referrer payloads

A verb that returns the attached JSON, not just its descriptor. Selection must be **deterministic**: with several `deployment` or `lint` referrers on one subject, silently taking "the latest" is the failure mode to avoid — the caller names the digest, or the command refuses and says why.

### 2. `resolve` and `inspect`

`resolve` turns a mutable tag into a digest. `inspect` reports the descriptor, artifact type, subject, annotations, layer names/media types/sizes/digests, and **which referrer discovery mechanism was used**.

Without these the central CI operation has no spelling: *resolve mutable tag → record immutable digest → plan and apply that digest*. A future Kubernetes reconciler cannot be built without it either.

### 3. Promotion without repack or re-push

`build → staging → production` currently has no digest-preserving path. Needed: a `tag` that moves an alias onto the **same manifest digest** without rebuilding the artifact, and a recursive `copy` that carries lint, plan, deployment and signature referrers with it.

Ptah-specific wrinkle: a generic recursive copy does not know about Ptah's content-derived fallback tags, so either promotion is implemented inside Ptah or compatible behavior is guaranteed explicitly.

### 4. An explicit Referrers API policy

Today Ptah writes the standard tag-schema index **and** a content-derived tag, and merges both when listing. That makes Ptah-to-Ptah discovery robust and leaves cross-client discovery unguaranteed: another OCI client can miss referrers after a concurrent publish.

[...]

Issue #1118, opened 2026-08-03
Keep migration writer publication bound to the opened project root


denisvmedia:
## Problem

Atlas project configuration preserves migration-directory provenance through an opened project root while resolving and capturing the source, but writer paths eventually discard that rooted handle and continue with a string path.

The affected flows include at least:

- `ptah-compat migrate diff`: `atlasProject.captureLocal` validates through the opened project root, then `atlasmigrate.GenerateDiff` receives only `Dir string` and reopens it for locking, replay checks, staging, publication, recovery, and `atlas.sum` updates.
- forwarded `ptah-compat migrate new`: project resolution constrains the source, but the native generator ultimately writes through a resolved string path.

If the selected directory or one of its ancestors is replaced after validation, a writer can verify one filesystem object and publish to another. The cross-process lock coordinates Ptah writers, but it does not bind later pathname lookups to the directory object that was validated.

This is pre-existing writer architecture debt surfaced while reviewing #1086. It should not be solved by weakening the checksum gate or by adding another preflight path check.

## Required design

- Carry an opened/rooted migration-directory capability through the complete writer transaction: capture, checksum verification, lock scope, staging, optional editor preparation, migration publication, checksum commit, rollback, and recovery.
- Once a rooted handle exists, do not fall back to an unconstrained string-path reopen for any write or verification step.
- Support a missing migration directory without escaping the selected project root.
- Use conditional publication against the expected destination state. A concurrent or hostile replacement must fail without overwriting the replacement.
- Publish `atlas.sum` through the same rooted boundary and bind it to the exact migration snapshot being committed.
- Preserve current durability and commit-uncertain semantics.
- Keep direct CLI absolute-path behavior explicit; project-config relative paths must remain confined to the project root.

## Acceptance criteria

- Add Unix regression tests that replace the migration directory symlink or an ancestor after capture and before publication for both `migrate new` and `migrate diff`.
- Prove that no migration file, checksum, journal, or recovery artifact appears outside the originally opened root.
- Add destination-replacement tests at the final migration filename and `atlas.sum` commit points.
- Cover successful create/diff, editor preparation, rollback, interrupted-publication recovery, and initially missing directories.
- Keep the tests declarative and black-box by default; justify any white-box fault-injection tests per `AGENTS.md`.
- Update the migration generator/public API and safety documentation if the rooted writer contract becomes externally visible.

denisvmedia:
[...]

PR #1205, opened 2026-08-06
docs: state that compatibility with older Ptah is not owed before v1

denisvmedia:
Per Denis: mark explicitly that backward compatibility with older Ptah is not needed, and write the rule into AGENTS.md. The rule holds until v1.

## What was already there

A pre-GA rule existed at `AGENTS.md`, under **Code Style And Linting**:

> Ptah is pre-GA. Do not preserve old command aliases, compatibility wrappers, fallback APIs, or backward-compatibility behavior only to keep an older internal shape.

Read literally that is a rule about *code shape*. It says nothing about the cases where the question actually comes up: a default that would change existing output, a retired file format, an error message, a refusal that would break something an earlier Ptah accepted. Three separate decisions today ran into exactly those and had to reason it out from scratch.

## What this adds

A first-class section in **Compatibility Policy**, which is where the distinction it needs is available. Everything else in that section is about the pinned community binary — which **is** a contract with users of that CLI. Compatibility with Ptah's own history is the other axis and is not a contract. Conflating the two is how a change gets designed around a migration path nobody is owed.

The scope is stated in both directions, because a rule with only the permissive half gets over-applied:

- **It licenses** changing Ptah's defaults, spellings, internal formats and error text without a migration path.
- **It does not license** breaking parity with the community binary, nor silently discarding user data.
- **It expires at v1.**

The existing code-shape paragraph stays where it is and now points at the rule, since it is that rule applied to one narrow case.

## Note on how this was nearly shipped wrong

The first attempt wrote the section in a checkout sitting on a different branch and copied the file across. That silently **deleted** the `### A second worked example` section — the `file()` confinement write-up for #1042 — which exists on master but not on that branch. Caught by reading the diff rather than the stat line: an additive change showing `30 insertions, 27 deletions` is not additive. Redone in place; the diff is now 29 insertions and the single replaced paragraph, and `A second worked example` is still there.

`node docs/site/scripts/check-style.mjs` — OK (100 documentation files).

A sweep over the open issues, adding the note where a proposal genuinely changes existing behavior, is running separately.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -24,6 +24,31 @@ When the two halves pull apart, say so in the commit and in the issue rather
 than picking silently. "We are stricter here, deliberately, and here is the
 measurement" is a complete answer; quietly matching is not.
 
+### Compatibility with older Ptah is a different axis, and it is not owed
+
+Everything above is about the community binary. Compatibility with **Ptah's own
+previous behavior is a separate question, and until Ptah ships v1 the answer is
+no.** There is no supported upgrade path to preserve, so:
+
+- Do not keep a fallback, an alias, a tolerated old spelling, or a second reader
+  for a retired format only because an earlier Ptah produced it.
+- Do not soften a refusal because it would break something an earlier Ptah
+  accepted.
+- Do not carry a default only because changing it would alter existing output.
+  Pick the default that is right for a reader meeting it for the first time.
+
+Changing behavior is the normal, cheap thing to do right now, and the cost of
+not changing it compounds. When a change alters behavior, say so plainly in the
+issue and the commit -- "this changes behavior; pre-v1, so no compatibility is
+owed" -- rather than quietly designing around it.
+
+This does **not** license breaking parity with the community binary, which is a
+contract with users of that CLI rather than with Ptah's own history, nor
+silently discarding user data. It licenses changing Ptah's defaults, spellings,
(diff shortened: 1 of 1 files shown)


denisvmedia:
**The heading in this PR is now load-bearing — please do not reword it during review without saying so.**

The sweep over the open issues has landed, and **fifteen** of them now cite this section by anchor:

`#905 #906 #916 #929 #931 #938 #995 #996 #997 #1028 #1035 #1106 #1118 #1143 #1186`

Every one links `AGENTS.md#compatibility-with-older-ptah-is-a-different-axis-and-it-is-not-owed`, which GitHub derives from the heading text. Two independent verification passes flagged the same thing: those anchors are **dead until this merges**, and they break permanently if the heading is reworded here.

The failure is soft rather than a 404 — the link lands on `AGENTS.md` and simply does not jump — and each comment quotes the rule's substance inline, so a reader today is not left guessing. But it is worth knowing before anyone edits the heading.

If the wording should change, say so and I will update the fifteen comments in the same pass rather than leaving them pointing at nothing.

For the record, the sweep was selective on purpose: 15 issues noted, 36 skipped. The skips are the additive work, the parity-with-the-community-binary work, the epics that propose no specific change, and the docs-only issues — where a compatibility note is noise that trains the next reader to skim past them.

PR #1212, opened 2026-08-06
Make omitting source comments the default (#1145 follow-up)

denisvmedia:
Same commit as #1203, re-opened on a fresh branch because that PR is stuck: GitHub stopped creating `github-actions` check-suites for its head during today's Actions outage and has not resumed for it. A `--force-with-lease` to a new sha and a close/reopen — both after Actions recovered and was demonstrably creating suites for other PRs — each produced zero suites, while #1205's head has nine. #1203 is closed in favor of this one.

---

Follows #1189, on Denis's review comment there:

> Source comments frequently contain internal implementation notes, operational details, or security-related context that users may not intend to publish through generated protobuf definitions and downstream generated code. Have you considered making `none` the default and requiring `--proto-comments=all` as an explicit opt-in?

Yes — and it is the right call.

## What changed

| | before | after |
| --- | --- | --- |
| `--proto-comments` default | `all` | **`none`** |
| `protobufrender.Options.Comments` zero value | `CommentsAll` | **`CommentsNone`** |

**Flipped in the library too**, deliberately. A default that differs between `protobufrender.Render` and `ptah schema export` is a trap for exactly the caller this protects: an embedder who never sets the field would keep publishing prose while the CLI stopped.

`--proto-comments all` restores the previous bytes exactly. Field and enum numbers never depended on the policy, so an export written before this flip loses its source prose on the next run and nothing else — a diff, not a compatibility break. **This changes behavior and no backward compatibility is owed: pre-v1**, per the compatibility-with-older-Ptah rule added in #1205.

## Tests inverted rather than deleted

`TestCommentsAllIsTheDefaultAndCopiesSourceProse` (library) and `TestSchemaExportProtobufCommentsNoneOmitsSourceComments` (CLI) both named the old default. They are renamed and inverted, so the moved default appears in the diff instead of disappearing from it.

Each now asserts **both** halves — the zero value omits, and `all` brings the prose back — because either half alone is satisfiable by an implementation that ignores the policy entirely: one that never emits comments passes the first, one that always emits them passes the second.

## A fixture that would have gone green for the wrong reason

`TestSchemaExportProtobufOnlyFlagsRejectedOnOtherTargets` proves `--proto-comments` is refused on non-protobuf targets. It passed `none` — now the default — and a flag left at its default is indistinguishable from one never passed, so the guard stopped firing and the test passed while measuring nothing. It passes `all` now.

## Gates

```
```

Every `check:*` in `docs/site/package.json`, enumerated from the file, plus `node docs/site/scripts/build-feature-matrix.mjs --check` and `npm run build`.

This commit was also merged with the other six open PRs into one tree and the full gate set re-run there, so the protobuf cluster is known to hold together — three of those PRs touch the same package and this one flips a default another asserts.


denisvmedia:
Closing: this PR was created on a wrong conclusion of mine.

I read #1205 having check-suites as "Actions recovered" and inferred that #1201, #1202 and #1203 were specially stuck — plausibly by the close/reopen I did during the outage — so I re-cut #1203's commit onto a fresh branch to route around it.

That inference was wrong, and the measurement that settles it is one query: the newest workflow runs in the whole repository are from **21:45:43**, all on `docs-pre-v1-compat-rule`. Nothing has been created since — not the three refreshed branches, not this brand-new PR. #1205 caught a brief window; the service did not come back.

So there is nothing special about those three PRs, this branch adds nothing #1203 does not already have, and the duplicate is worse than the problem. #1203 stays.

Issue #942, opened 2026-07-31
Spanner: no live coverage


denisvmedia:
Spanner status rests entirely on offline paths, with no live container or live test.

Found by a deep audit of every capability the feature matrix marked as partial.
Each item below was reproduced against a binary built from this repository.

Severity assessed as **incomplete-capability**.

### 1. Add live Spanner PostgreSQL-interface coverage

Spanner has no live container, no dedicated introspection package, and no live test. Its entire status rests on offline capability, planning, rendering and URL-detection tests, so no executed evidence shows that generated SQL is accepted by a real Spanner PostgreSQL endpoint. CockroachDB and SQL Server both have compose services; Spanner does not.

Reproduction:

```text
$ grep -ci spanner docker-compose.yaml
0
$ ls internal/dbschema/
clickhouse  dbtest  mssql  mysql  postgres  sqlite
(no spanner package; dbschema/connection.go:170 routes spanner through the PostgreSQL pgx driver)
$ ptah schema render --root-dir ./engmodels --dialect spanner
error: error rendering spanner schema: error rendering column id: unsupported feature: spanner does not support sequence-backed type SERIAL; use a platform-specific type override
```

Evidence: docker-compose.yaml declares cockroachdb (line 71) and sqlserver (line 101) but no spanner service; docs/site/src/content/docs/databases/support-matrix.md:116-118 states Spanner coverage is offline and advises reviewing generated SQL

## Notes

The matrix row for each of these now states the specific limitation rather than
the word "partial". Closing an item here should update the corresponding row on
`docs/site/src/content/docs/atlas/feature-matrix.md`.

denisvmedia:
## Frame: split-required — two workstreams, and the blocker is upstream of the container

Measured on master `d29a44f4` (`Decide Flyway linearity on the version token, not the ordering key (#1098) (#1132)`), against a binary built from that tree in a clean worktree. **Every claim in the body still reproduces.** One sentence in the *matrix row this issue points at* is now false.

But the interesting result is not the re-measurement. **I stood up the live Spanner stack this issue says does not exist, ran Ptah against it, and it does not get past `db read`.** So "add a live container and a live test" is not one piece of work — the container is the easy half, and the test it would run cannot be written today.

---

### The stack, since its feasibility is the load-bearing unknown

Two Apache-2.0 images, both pulled and run locally under Docker 29.4.0:

[...]

PR #1216, opened 2026-08-07
Derive the dev URL selection and the resolved schema in one place (#1207)

### LINKED ISSUE #1207: Lint scope and realm ownership read the dev URL twice, and the two readings disagree when no schema is selected
Raised by Denis on #1197: that PR derives the lint schema scope by parsing `search_path` out of the dev URL, while #1201 makes the connection ask PostgreSQL what the session actually resolved to. Two independent interpretations of one connection configuration, which can drift as connection defaults or session options change.

Measuring it turned up the part that decides the shape of the fix.

## The dev URL carries a fact `current_schema()` has erased

| dev URL | `atlasurl.SchemaScope` | `current_schema()` |
| --- | --- | --- |
| `?search_path=app` | `app` | `app` |
| `?search_path=public` | `public` | `public` |
| **no `search_path`** | **`""` — review everything** | **`public`** |

Measured on PostgreSQL 17: a session with no `search_path` answers `current_schema()` = `public`, the same as one that named `public` explicitly.

The two agree wherever a schema is named and disagree exactly where none is — and that row is load-bearing. `""` means *review the whole database*, `public` means *review only public*. It is pinned by an e2e control whose expectation is the pinned community binary's own: with a dev URL naming no schema, exit 1 with two DS102 for objects outside `public`.

So this is not one fact read twice. **`current_schema()` cannot express whether the operator selected a schema at all** — it answers `public` either way. Replacing the URL reading with it would silently narrow that case and break parity there.

## Ordering, the second constraint

`lint.AnalyzeFS` runs the scope filter over SQL text **before** any connection exists; the dev connection is opened later inside `migrationreplay.Replay`. Giving lint the connection's resolved schema means connecting before analysis, which changes which diagnostic an invocation carrying both a bad directory and an unreachable dev URL prints first. That ordering is measured against the community binary elsewhere.

## Proposed shape

One resolver used by both, returning two facts rather than one:

- **`Selected`** — did the URL name a schema, and which. Only the URL knows this.
- **`Resolved`** — what the server resolved that to. Only `current_schema()` knows this, and it is what realm cleanup needs to decide which schemas the writer owns.

Lint scopes on `Selected`, where empty keeps meaning the whole database. Realm cleanup uses `Resolved`, with #1201's refusal when a selection resolves to nothing. Two questions, one answer each, instead of two interpretations of the same string.

## Definition of done

- One exported resolver is the only place either fact is derived; `atlasurl.SchemaScope` and the `current_schema()` query are both reached through it.
- The e2e control "a dev URL naming no schema reviews the whole database" stays byte-identical and keeps exiting 1. **This is the row a naive unification breaks**, so it is the discriminator, not a formality.
- The `?search_path=app` scope row and #1201's refusal row both still pass.
- If the connection moves earlier than the analysis, the diagnostic ordering for a bad directory plus an unreachable dev URL is re-measured against the pinned community binary and recorded, rather than assumed unchanged.
- **This changes behavior and no backward compatibility is owed:** pre-v1, per the compatibility-with-older-Ptah rule in `AGENTS.md`. What must not change is parity with the community binary, which the control rows above pin.

Blocked on #1197 and #1201 landing first; coupling a third change to them while they are open makes the stack worse.

denisvmedia:
Closes #1207.

`migrate lint` asked the dev URL which schema it names (`atlasurl.SchemaScope`); the PostgreSQL realm cleanup asked the server what the session resolved to (`current_schema()`, added by #1201). Two independent readings of one connection configuration, which is what Denis raised on #1197.

## They are not one fact read twice

Measured on PostgreSQL 17:

| dev URL | what the URL names | `current_schema()` |
| --- | --- | --- |
| `?search_path=app` | `app` | `app` |
| `?search_path=public` | `public` | `public` |
| **no `search_path`** | **nothing — review everything** | **`public`** |

They agree wherever a schema is named and disagree exactly where none is. `current_schema()` cannot express whether the operator selected a schema at all, so unifying on it silently narrows the last row — and that row is pinned by an e2e control whose expectation is the pinned community binary's own: one DS102 per dropped table, exit 1.

## Shape

One type, `internal/schemaselection.Selection`, carrying both:

- `Scope` — the one schema the URL restricts analysis to, empty when it restricts nothing. `migrate lint` filters on this, so the "reviews the whole database" case keeps its meaning.
- `Resolve(ctx, db)` — asks the server which schema the session landed in, and carries #1201's refusal when a selection resolves to nothing. The realm cleanup uses this.
- `Raw` — the `search_path` value as written. The refusal quotes it, because `Scope` is deliberately empty for `search_path=a,b` and a refusal reading `selects schema ""` would name nothing back to an operator who plainly named something.

`atlasurl.SchemaScope` is retired into it rather than left as a second door onto the same question, and its table moves across unchanged.

## The discriminator the issue asked for

The DoD names the row a naive unification breaks. Making an empty selection scope to `public` — exactly what reading `current_schema()` for both would do — reddens that row and only that row:

```
--- FAIL: TestAtlasMigrateLintSchemaScopeE2E
    --- FAIL: .../control:_a_dev_URL_naming_no_schema_reviews_the_whole_database
```

Two more mutants, on the unit tests: folding a NULL `current_schema()` back to `"public"` reddens both refusal rows, and making `Raw` follow `Scope` reddens the comma row.

## Ordering did not move

The DoD asks for a re-measurement *if* the connection moves earlier than the analysis. It does not: lint still scopes from the URL, before any connection exists, so the diagnostic order for a bad directory plus an unreachable dev URL is untouched.

## Gates

`go build`, `go vet`, `go test ./... -count=1`, `qtlint`, `check-test-style.sh`, `check-public-api.sh`, `check-public-api-snapshot.sh`, `golangci-lint` — all clean. Integration run against a live PostgreSQL 17 with `POSTGRES_TEST_DSN`: the ten `TestAtlasMigrateLintSchemaScopeE2E` rows, `TestNativeMigrationsLintIgnoresTheDevURLSchemaScopeE2E`, the three `TestPostgresConnectionResolvesTheSearchPathSchemaE2E` rows and `TestPostgresRealmCleanupKeepsTheSelectedSchemaE2E` all pass.

Pre-v1, no backward compatibility with older Ptah is owed; parity with the pinned community binary is what the control rows above pin.


denisvmedia:
One property this does not change, recorded rather than left implicit: the `case` that runs `current_schema()` covers `platform.Spanner` alongside PostgreSQL, CockroachDB and YugabyteDB. That grouping predates this PR — Spanner was already in that branch when #1201 replaced the hardcoded `"public"` — and this change only moves the call, so behavior there is identical either way. Spanner has no live coverage at all (#942), so whether its PostgreSQL interface answers `current_schema()` is untested rather than known. Narrowing the branch here would be an unmeasured behavior change riding along on a refactor.

PR #1217, opened 2026-08-07
Record the migrate lint scope and analyzer-shape boundary (#1074 S3)

denisvmedia:
Closes the S3 half of #1074: the feature matrix and lint documentation now carry the support boundary the rest of that issue established.

Two rows in `feature-matrix-rows.json`, both measured against the pinned community binary v1.3.0 rather than read off the code.

**Dev-URL schema scope on `migrate lint`** — ✅/✅/✅. `ptah-compat migrate lint` reviews only the schema the dev URL's `search_path` names; native `ptah migrations lint` reads SQL text and deliberately does not scope. That split is what the #1197 review caught before it shipped, and a reader otherwise finds it by running both.

**Analyzers that need a dev-database schema diff** — 🟡/✅/✅, the open half. Re-measured today on PostgreSQL 17:

```
CREATE TABLE users (id int NOT NULL);
ALTER TABLE users RENAME COLUMN id TO oid;

pinned binary v1.3.0   DS103 + MF103 "Adding a non-nullable \"integer\" column \"oid\""   rc=1
ptah-compat            DS103                                                            rc=1
```

The renamed column's type and nullability come from an earlier file or the base schema rather than from the statement, and the binary spells the type as the database canonicalizes it (`integer`, not `int`), so reaching that needs the replayed dev schema rather than the SQL text. `migration/lint/rules.go` `notNullWithoutDefaultFindings` scans `ADD COLUMN` clauses only.

Worth stating plainly because it changes what the gap *is*: Ptah is one diagnostic short here, not differently classified. The rename classification itself now matches exactly — same rule, same subject, same suggested fix, same exit code — which I re-measured on `master` rather than assuming from #1074's 2026-08-03 table:

```
CREATE TABLE users (id int); ALTER TABLE users RENAME COLUMN id TO oid;

both binaries: DS103, "Dropping non-virtual column \"id\"", the NULL-precheck fix, rc=1
```

The docs half is a paragraph in the Renames section of `atlas/migrate-commands.md` saying the same thing in prose.

#1074 stays open on S2 — closing it needs the dev-database schema diff, which is a real engine change rather than a wording one.

Gates: `node scripts/build-feature-matrix.mjs --check` plus all seven `check:*` docs scripts. `check:responsive` needs a built `dist` and runs in CI. No Go code changed.

### DIFF
--- a/docs/site/scripts/data/feature-matrix-rows.json
+++ b/docs/site/scripts/data/feature-matrix-rows.json
@@ -864,6 +864,24 @@
     "evidence": "migration/lint/rules.go:20 Register, migration/lint/lint.go:107 Options.ExtraRules; migration/lint/config.go Config has only Dialect/DisabledRules/Rules{severity,exclude} — no rule definition. config/projectconfig/atlas.go parseLintPolicyBlocks accepts only concurrent_index, condrop, data_depend, destructive, git, nestedtx, incompatible; parseLintAnalyzer tolerates `force` without acting on it. Atlas features page classifies \"Custom linting rules\" as Pro. `lint { rule \"hcl\" \"custom\" { src = [...] } }` accepted and the rules execute during migrate lint; rule severity is only `error=true/false` (measured; observation study, scenario, lint-test-pro)"
   },
   {
+    "area": "Lint analyzers",
+    "feature": "Dev-URL schema scope on `migrate lint`",
+    "ptah": "yes",
+    "atlas_oss": "yes",
+    "atlas_pro": "yes",
+    "note": "`ptah-compat migrate lint` reviews only the schema the dev URL's search_path names, matching the pinned CE binary. Native `ptah migrations lint` reads SQL text and deliberately does not scope.",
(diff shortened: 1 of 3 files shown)


denisvmedia:
Conformance check for this change, and the honest answer is that nothing there moved.

I looked for the place a "CE emits MF103 on a rename, Ptah does not" row would be pinned automatically, and there is not one:

- `ptah-atlas-conformance` has no fixture mentioning `search_path` or `RENAME COLUMN` outside vendored build trees. Its `lint-analyzer-catalog` probe (`internal/probe/atlas_lint_fidelity.go`) calls `lint.LintFS` in process, so it measures Ptah's analyzer behavior against itself, not against the binary. Its `ce-gating` lint case runs the CE binary alone, to decide whether CE gates the command.
- In this repo, the `differential and corpus` job (`.github/workflows/atlas-oracle.yml`) is scoped to `TestSumFileNamesDifferentialFuzz*` plus the schema-HCL unknown-name policy. There is no lint differential harness for it to join.

So the divergence is recorded in prose — the matrix row and the Renames paragraph — and tracked by #1074 S2, rather than pinned by a fixture. #1074's Scope does ask for differential fixtures for both inputs; the multi-target one landed as e2e rows in #1197, and the rename one has no harness to land in yet. Building that harness is S2-sized work, not a docs change, so I have left it there rather than widening this PR.

One thing worth flagging for a reader of the matrix: the neighbouring "Default-firing Atlas analyzer concern mapping" row stays ✅ and does not contradict the new 🟡. MF103 *is* mapped — `internal/atlaslint/rules.go` maps Ptah's `DD101` onto it. What is missing is not the mapping but the input: `notNullWithoutDefaultFindings` reads `ADD COLUMN` clauses, and a `RENAME COLUMN` statement has none.

Issue #1214, opened 2026-08-06
Post-parity audit: unify native and compatibility capabilities over one core


denisvmedia:
## Timing

**Do not start this issue until the current Atlas compatibility campaign has stabilized.**

This issue is intentionally a post-parity consolidation pass.

Start it only after:

1. the targeted Atlas CE compatibility level has been reached and the relevant conformance gates are green or have explicit intentional-divergence records;
2. the desired Atlas Pro best-effort scope has either been implemented or explicitly declined/deferred;
3. major compatibility work is no longer rapidly changing the capability surface.

The purpose is to avoid repeatedly restructuring the architecture while agent-driven compatibility work is still discovering required semantics.

Until then, the prospective architecture contract established by the native/compat capability-ownership issue should prevent new divergence.

## Problem

Ptah has evolved through two user-facing surfaces:

```text
ptah
ptah-compat
```

Compatibility work has been intentionally aggressive: Atlas CE behavior is reproduced where appropriate, selected Pro capabilities are implemented openly, and some compatibility work has discovered or introduced useful functionality that did not previously exist in native Ptah.

This creates two possible forms of architectural debt:

### Capability asymmetry

A semantic capability may exist through one surface but not the other.

Examples of shapes to look for:

```text
shared implementation exists
        |
        +--> ptah-compat exposed
        |
        +--> native Ptah missing
```

or the reverse:

```text
shared/native capability exists
        |
        +--> ptah exposed
        |
        +--> ptah-compat missing despite an applicable Atlas surface
```

Not every asymmetry is a bug: Atlas-specific adapters and Ptah-native-only product concepts can legitimately exist.

The audit must distinguish deliberate interface-specific behavior from missing product functionality.

### Duplicated semantics

Because many features were implemented concurrently and agentically, equivalent behavior may have grown independently under native and compatibility paths.

For example:

```text
cmd/ptah -> implementation A
cmd/atlas -> implementation B
```

when the intended architecture is:

```text
                 shared implementation
                    /          \
                   /            \
              cmd/ptah        cmd/atlas
```

The second shape is required wherever both commands perform the same semantic operation.

## Goal

Perform a repository-wide post-parity audit and establish this architecture:

```text
                    Ptah semantic core
                  /                    \
                 /                      \
       native interface             compat interface
          `ptah`                    `ptah-compat`
```

The command/client layers should primarily be responsible for:

- parsing their own interface;
- resolving configuration;
- translating external representations;
- selecting compatibility policy;
[...]

Issue #1213, opened 2026-08-06
Establish native/compat capability parity as a project architecture contract


denisvmedia:
## Problem

Ptah intentionally exposes two different user-facing command surfaces:

* `ptah` — the native Ptah CLI;
* `ptah-compat` — the Atlas-compatible CLI used for drop-in migration of existing Atlas workflows.

The two surfaces should remain different interfaces. Atlas command names, flags, URI spellings, configuration quirks, output compatibility, and other Atlas-specific contracts do not belong in the native `ptah` CLI.

However, the project currently lacks the corresponding architectural rule for **semantic capabilities**.

As Atlas CE parity and selected Atlas Pro compatibility have expanded, useful functionality has sometimes been implemented while working primarily in the compatibility tree. Without an explicit project rule, it is possible for a real product capability to become reachable only through `ptah-compat`, while other capabilities remain reachable only through native `ptah`.

That would gradually turn the repository into two partially independent products.

The intended architecture is instead:

```text
                         shared Ptah capabilities
                        /                        \
                       /                          \
             native Ptah surface          compatibility surface
                  `ptah`                    `ptah-compat`
```

The compatibility layer may translate Atlas interfaces into Ptah concepts, but it should not become the owner of general product behavior.

## Goal

Establish a repository-wide contract that:

> Every semantic capability exposed through `ptah-compat` must also be available through the native Ptah product when that capability has meaningful use outside Atlas compatibility.

The native surface does **not** need to use the same command names, flags, configuration syntax, file formats, or other interface conventions.

This is **functional parity, not interface parity**.

Atlas-specific compatibility machinery is exempt when it exists only to interpret, reproduce, or bridge an Atlas interface or persisted representation.

## Required `AGENTS.md` rule

Add a dedicated section to `AGENTS.md` covering native/compat capability ownership.

The exact editorial wording may be adjusted to fit the document, but the rule must communicate all of the following:

> `ptah-compat` is an adapter over Ptah capabilities, not an independent product implementation.
>
> When implementing behavior for `ptah-compat`, distinguish between:
>
> 1. a general semantic capability; and
> 2. Atlas-specific interface or compatibility machinery.
>
> A general semantic capability must live in a reusable Ptah package below the CLI layer and, where meaningful to a native Ptah user, must be reachable through the native Ptah surface as well.
>
> Do not implement general product behavior exclusively inside `cmd/atlas` or another compatibility-only package.
>
[...]

PR #1247, opened 2026-08-07
Report the add side of a rename in migrate lint

denisvmedia:
Closes the last open scenario of #1074 — S2, the `MF103` a rename introduces.

A rename retires one logical name and introduces another. The retirement has been reported since #1120. The introduction was unreachable, because the retired column's type, nullability and default live in an earlier migration file or in the base schema, never in the rename statement. Reaching it needs a dev-database schema read, which is why this scenario waited.

## What was measured, and how

Pinned comparison binary: `atlas community version v1.3.0` at `/Users/buster/Work/denis/ptah-atlas-conformance/bin/atlas`, invoked by absolute path. Dev database: PostgreSQL 16 in a container, one freshly created database per fixture, `atlas migrate hash` then `migrate lint --dir file://… --dev-url … --latest 1` on both binaries. Reports compared byte-wise with only the two elapsed-duration lines normalized; exit codes captured from a redirected run, never from a pipe.

`ptah-compat` was built twice — once from the commit before this change, once from this branch — so every row below has a before and an after rather than only an after.

### The scenario

`CREATE TABLE users (id int NOT NULL);` then `ALTER TABLE users RENAME COLUMN id TO oid;`

| | report | exit |
|---|---|---|
| community binary | `DS103` "id" + NULL precheck, **and** `MF103` `Adding a non-nullable "integer" column "oid" …`, `2 diagnostics` | 1 |
| ptah-compat, before | `DS103` alone, `1 diagnostic` | 1 |
| ptah-compat, after | **byte-identical to the community binary** | 1 |

The completion criterion this issue's S2 comment set — that line present under a `-- data dependent changes detected:` heading with the trailer reading `-- 2 diagnostics`, and the nullable fixture unchanged at `-- 1 diagnostic` — is met.

### Full fixture table, before and after

`yes` means byte-identical to the community binary modulo the two duration lines.

| fixture | community | before | after |
|---|---|---|---|
| `id int NOT NULL` rename | rc=1 | no | **yes** |
| `id varchar(20) NOT NULL` rename | rc=1 | no | **yes** |
| two renames in one file | rc=1 | no | **yes** |
| `ALTER TABLE public.users` rename | rc=1 | no | **yes** |
| rename across two analyzed versions (`--latest 2`) | rc=1 | no | **yes** |
| quoted `"Users"."Id"` rename | rc=1 | no | **yes** |
| `-- atlas:nolint destructive` above the rename | rc=0 | no | **yes** |
| add on L1, drop on L2 (analyzer group order) | rc=1 | no, groups reversed | groups now match; still differs only by `"int"` vs `"integer"`, see below |
| `id int` nullable rename (control) | rc=1 | yes | yes |
| `id int NOT NULL DEFAULT 7` rename (control) | rc=1 | yes | yes |
| same-file `CREATE` then rename (control) | rc=0 | yes | yes |
| `ALTER TABLE users RENAME TO accounts` (control) | rc=1 | yes | yes |
| `ALTER TABLE users DROP COLUMN nick` (control) | rc=1 | yes | yes |
| `DROP TABLE users` (control) | rc=1 | yes | yes |
| `DROP TABLE users, audit_log` (control) | rc=1 | yes | yes |
| `mid`/`zeta`/`alpha` ordering fixture (control) | rc=1 | yes | yes |
| rename in a schema the dev URL excludes (control) | rc=0 | rc=0 | rc=0 |

Seven fixtures moved to byte-identical; nine controls that already matched still match, and the group-order fixture moved from wrong-order to right-order. No fixture moved the other way, and no fixture changed exit code.

### Ordering was measured, not chosen

`ALTER TABLE users RENAME COLUMN zeta TO zz;` then `… alpha TO aa;` in one file gives four diagnostics on the pinned binary: two `DS103` in statement order, then two `MF103` in statement order, with `-- suggested fixes:` pluralized. A rule emitting one finding carrying both introduced columns, or sorting them by name the way a multi-target `DROP TABLE` is sorted, produces a different report on that fixture; a single-rename fixture cannot tell those apart.

### Type spellings were measured one by one

Renaming a NOT NULL column of each type and reading the diagnostic:

`int` → `integer`, `bigint`, `smallint`, `text`, `boolean`, `date`, `json`, `jsonb`, `uuid`, `real`, `bytea`, `double precision`, `varchar(20)` → `character varying(20)`, `char(5)` → `character(5)`, `numeric(10,2)`, `numeric(10)` — and `numeric(10,0)` also prints `numeric(10)`, so a zero scale is not printed.

Three spellings are **not** the catalog's and are deliberately absent from the allow-list: `timestamptz` (catalog says `timestamp with time zone`), `timestamp` (catalog says `timestamp without time zone`), and `integer[]` (catalog says the bare category `ARRAY`). Those keep the diagnostic — same code, same line, same severity, same count, same exit status — under Ptah's own visibly labeled wording rather than a guess at the other tool's. That is the renderer's existing answer for a type it cannot spell, not a new behavior.

## What changed

The dev database the run **already replays on** now also answers what the versions start from. The analysis reports which versions it could use a starting state for (`Analysis.BaselineVersions`), the replay reads those states while standing on them — before the version runs, while the retired column still exists — and the analysis runs once more with them in hand.

- A directory with no renames asks for nothing, so it pays no introspection and no second analysis.
- A run without `--dev-url` reports exactly what it did before.
- The analysis still runs before the replay, so error precedence does not move: a directory that fails to analyze reports that, not a replay error it would also have hit.

The add side is decided by one function, `renameAddSideCandidates`, which both the request for the state and the finding built from it go through. That started as two copies of the same condition, and the first mutation run showed why that was wrong — see the mutation section.

**Native `ptah migrations lint` deliberately does not move.** It models a rename as a rename, and a rename does not fail on a populated table, so claiming it might would be false of the statement its analyzer describes. `BC101` stays its single finding, and an e2e row pins that: removing the surface gate leaves it red.

Two neighboring fixes the scenario needs:

- **Analyzer groups now print in analyzer order**, not in the order their first diagnostic appears. Measured: `ALTER TABLE users ADD COLUMN x text NOT NULL;` on line 1 and `ALTER TABLE users DROP COLUMN nick;` on line 2 print `destructive changes detected` above `data dependent changes detected` on the pinned binary; ordering by first appearance printed them the other way round. Only the two analyzers whose relative order was measured are ranked; Ptah-owned groups keep first-appearance order behind them, because there is no binary to measure them against.
- **The renderer's Atlas type-name allow-list gained the measured catalog spellings.** It stays fail-closed for everything else.

## Deliberately not matched, with the measurement

Three rows still differ. Each was reproduced against a build of the commit before this change, so each is pre-existing rather than introduced here.

1. **`"int"` vs `"integer"` on the plain `ADD COLUMN` path.** `ALTER TABLE users ADD COLUMN nick int NOT NULL;` prints `"int"` on ptah and `"integer"` on the pinned binary. That path reads the type from the statement text, where `int` is what is written. Fixing it properly means resolving the added column's type from the state **after** the version, which is a second read and a separate decision; this change reads only the before-state. Already recorded on #1074 as a pre-existing divergence belonging to its own issue.

2. **An `MF103` message whose rendered content lands on exactly 89 columns wraps on ptah and not on the pinned binary.** Reproduced without any rename: `ALTER TABLE t ADD COLUMN cccccc text NOT NULL;` diverges identically on the pre-change binary. The pinned binary's wrapping is not a single greedy width — solving the constraint system over fourteen measured diagnostics gives `smallint` requiring width ≥ 89 and `character varying(20)` requiring ≤ 88, which no one width satisfies — so matching it needs its own investigation rather than a constant change. `lintWrapWidth` is left at its measured 88.

3. **`ALTER TABLE users ADD COLUMN x int NOT NULL; ALTER TABLE users RENAME COLUMN x TO y;` in one file.** The pinned binary exits 0 and reports two `MF103`; ptah exits 1 and reports `MF103` for `x` plus `DS103` for `x`. The pinned binary exempts a rename of a column **added** in the same file, where Ptah's exemption covers only a column whose **table** the file created. This change does not touch it: ptah is stricter, which is the permitted direction, and relaxing a destructive check is not something to do as a side effect of a rename change. Same exit codes and same diagnostic count as before this branch.

None of these is in the forbidden direction. There is no fixture in this measurement where ptah-compat exits 0 and the community binary exits 1.

## Tests

- `migration/lint/baseline_test.go` — the analyzer. Seven silent rows, each one axis away from the reporting row: nullable retired column, retired column with a default, table created in the same file, table rename, baseline for another table, baseline for another version, and a bare table name carried by two schemas. Plus the surface split, the qualified-reference resolution, and what `BaselineVersions` asks for.
- `internal/atlasreport/migrate_lint_text_internal_test.go` — three rendering rows, verbatim from the pinned binary: the rename's two halves, the nullable control, and the analyzer group order.
- `integration/atlas_migrate_lint_rename_add_side_e2e_test.go` — ten rows against a live PostgreSQL 16, every `want` verbatim from the pinned binary, including both suppression selectors: `-- atlas:nolint destructive` silences the retirement and leaves the addition, `-- atlas:nolint data_depend` does the reverse. An add side carrying a DS code would pass one and fail the other. Plus the native-surface control.

### The tests were broken on purpose and watched go red

Eight mutants, each applied alone and reverted afterwards:

| mutant | red |
|---|---|
| analyzer stops emitting the add side | `TestAnalyzeFS_RenameAddSideNeedsTheBaseline`, the renderer row, four e2e rows |
| surface gate removed | `TestAnalyzeFS_RenameAddSideIsCompatibilitySurfaceOnly`, `TestAnalyzeFS_BaselineVersionsIsNativeSurfaceEmpty`, `TestNativeMigrationsLintRenameHasNoAddSideE2E` |
| analyzer group ordering reverted | the group-order renderer row, the rename renderer row, four e2e rows |
| catalog type spellings dropped from the renderer | the `character varying` e2e rows |
| qualified-reference fallback removed | `TestAnalyzeFS_RenameAddSideResolvesQualifiedReference` |
| nullability and default conditions dropped | both silent-case unit rows, both e2e control rows |
| the dev-database read never happens | four e2e rows |

The surface-gate mutant is the reason `renameAddSideCandidates` exists. On the first run it left the native control **green**: the gate was written twice, and with the version request still gated, removing the rule's own gate changed nothing observable. One gate now serves both, and the same mutant is red.

## Gates

All green on this branch:

`go build ./...` · `go vet ./...` · `go test ./... -count=1` · `go tool qtlint ./...` · `golangci-lint run ./...` · `scripts/check-test-style.sh` · `scripts/check-public-api.sh` · `scripts/check-public-api-snapshot.sh`

Documentation was touched, so every `check:*` script in `docs/site/package.json` was run as well, plus `node docs/site/scripts/build-feature-matrix.mjs --check`. `check:responsive` needs a built site; `npm ci && npm run build` were run so it could execute rather than be skipped.

`docs/public_api.snapshot` is regenerated for the two additions to `migration/lint`: `BaselineColumn` and `Analysis.BaselineVersions`.

## What this does not cover

- The three divergences listed above, each reproduced on the pre-change binary.
- PostgreSQL only. The type spellings and the whole scenario were measured on PostgreSQL 16; MySQL and MariaDB were not measured, and an unmeasured catalog spelling falls through to Ptah's prose rather than to a wrong Atlas sentence.
- Identity and `serial` columns diverge in the reporting-less direction: the pinned binary reports `MF103` for a renamed `serial` column, and Ptah does not, because the catalog gives it a `nextval(…)` default. A renamed `GENERATED ALWAYS AS IDENTITY` column has no catalog default and does report.
- S3 of #1074, the feature-matrix row, is untouched.

### DIFF
--- a/docs/public_api.snapshot
+++ b/docs/public_api.snapshot
@@ -6441,6 +6441,7 @@ func Describe(f Finding) string
 func Register(rule Rule) error
 type Analysis struct{ ... }
     func AnalyzeFS(fsys fs.FS, opts Options) (Analysis, error)
+type BaselineColumn struct{ ... }
 type CompatibilityProfile string
     const CompatibilityProfileNative CompatibilityProfile = "" ...
 type Config struct{ ... }
@@ -6478,10 +6479,50 @@ type Analysis struct {
     snapshot.
 
 func AnalyzeFS(fsys fs.FS, opts Options) (Analysis, error)
+func (a Analysis) BaselineVersions() []int64
 func (a Analysis) Files() []File
 func (a Analysis) Findings() []Finding
 func (a Analysis) SelectedFiles() []File
 func (a Analysis) SnapshotFS() fs.FS
+
+### go.5x5.cz/ptah/migration/lint.BaselineColumn
+
+package lint // import "go.5x5.cz/ptah/migration/lint"
+
+type BaselineColumn struct {
+    // Version is the migration version whose starting state this column belongs
+    // to: the state read BEFORE that version is applied, not after.
+    Version int64
+    // Schema is the schema of the owning table, as the server spells it.
+    Schema string
+    // Table is the owning table, as the server spells it.
+    Table string
+    // Name is the column name, as the server spells it.
+    Name string
+    // DataType is the column type spelled the way the compatibility surface
+    // prints it. Empty when the caller could not establish that spelling, which
+    // leaves the diagnostic reported under Ptah's own wording rather than
+    // inventing an Atlas sentence for a type nobody measured.
+    DataType string
+    // NotNull reports whether the column rejects NULL.
+    NotNull bool
+    // HasDefault reports whether the column carries a DEFAULT expression.
+    HasDefault bool
+}
+    BaselineColumn is one column of the schema state a migration version starts
+    from, as read from the dev database the run validates against.
+
+    Reading SQL text is enough to see that `ALTER TABLE users RENAME COLUMN id
+    TO oid` retires the name `id`. It is not enough to see what the new name
+    introduces: the retired column's type, nullability and default live in an
+    earlier migration file or in the base schema, never in the rename statement.
+    A caller that replays the directory on a dev database can read them there
+    and hand them over here, and only then can the add side of a rename be
+    reported.
+
+    Callers ask Analysis.BaselineVersions which versions are worth reading,
+    so a directory with no renames costs no introspection at all.
+
 
 ### go.5x5.cz/ptah/migration/lint.CompatibilityProfile
 
@@ -6665,6 +6706,11 @@ type Options struct {
     // RuleConfigs carries per-rule severity and path-scoping overrides,
     // normally loaded from .ptah-lint.yaml.
     RuleConfigs map[string]RuleConfig
+    // Baseline carries the schema state each analyzed version starts from,
+    // normally read from the dev database after replaying the migrations that
+    // precede that version. Empty analyzes SQL text alone, which is what every
+    // run without a dev database does. See [BaselineColumn].
+    Baseline []BaselineColumn
 }
     Options configures a lint run.
 
--- a/docs/site/src/content/docs/atlas/migrate-commands.md
+++ b/docs/site/src/content/docs/atlas/migrate-commands.md
@@ -853,18 +853,34 @@ incompatible` does not.
 Index, key and constraint renames are not reported at all: deployed application
 code does not name them.
 
-What a rename does **not** report on either surface is the column it
-introduces. Renaming a `NOT NULL` column draws the destructive diagnostic for
-the retired name from Ptah and from the CE binary alike, but the binary adds a
-second one — `MF103`, "adding a non-nullable column will fail in case the table
-is not empty" — for the new name, and Ptah does not. Ptah's analyzers read the
-migration's SQL text, and a `RENAME COLUMN` statement carries neither the
-column's type nor its nullability: both come from an earlier file or from the
-base schema, and the binary's message spells the type as the database canonicalizes
-it. Reaching that needs the replayed dev schema rather than the statement, which
-is [#1074](https://github.com/stokaro/ptah/issues/1074). Until then a rename is
-reported as destructive on both surfaces and as a data-dependent addition on
-neither, so it is one diagnostic short rather than differently classified.
+On the compatibility surface a rename also reports the column it **introduces**.
+Renaming a `NOT NULL` column with no `DEFAULT` produces a second diagnostic —
+`MF103`, "adding a non-nullable column will fail in case the table is not
+empty" — naming the new column and the retired column's type. A `RENAME COLUMN`
+statement carries neither that type nor its nullability, so this one is not read
+from the migration text: `ptah-compat migrate lint` reads the schema state the
+version starts from off the `--dev-url` database during the replay it already
+performs, before the version runs and while the retired column still exists. A
+run without `--dev-url` reports the retirement alone.
+
+Three consequences follow from where the facts come from:
+
+- The type is spelled the way the database canonicalizes it, not the way the
+  migration writes it: `int` reports as `integer` and `varchar(20)` as
+  `character varying(20)`.
+- A retired column that is nullable, or that carries a `DEFAULT`, has no such
+  diagnostic — the introduced column cannot fail on existing rows.
+- A type whose diagnostic spelling Ptah has not measured keeps the diagnostic but
+  prints Ptah's own labeled wording for it rather than a guess at the other
+  tool's.
+
+The two halves belong to different analyzers, so `-- atlas:nolint destructive`
+silences the retirement and leaves the addition, and `-- atlas:nolint
+data_depend` does the reverse.
+
+Native `ptah migrations lint` reports no addition. It models a rename as a
+rename, and a rename does not fail on a populated table, so `BC101` stays its
+single finding.
 
 The report is written to stdout even when findings fail, and error-severity
 findings still exit with code 1. The native `ptah migrations lint` output is
--- a/integration/atlas_migrate_lint_rename_add_side_e2e_test.go
+++ b/integration/atlas_migrate_lint_rename_add_side_e2e_test.go
@@ -1,0 +1,370 @@
+//go:build integration
+
+package integration_test
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+	"path/filepath"
+	"testing"
+	"time"
+
+	qt "github.com/frankban/quicktest"
+	_ "github.com/jackc/pgx/v5/stdlib"
+)
+
+// TestAtlasMigrateLintRenameAddSideE2E covers the last open scenario of
+// stokaro/ptah#1074 against a live PostgreSQL dev database.
+//
+// A rename retires one name and introduces another. The retirement is reachable
+// from the statement text and has been reported since #1120; the introduction is
+// not, because the retired column's type, nullability and default live in an
+// earlier migration file. Reaching it needs the dev database: the linter reads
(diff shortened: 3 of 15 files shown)


denisvmedia:
## Capability classification under #1213

Reviewed after merge, against the ownership rule in #1213 rather than against the Atlas parity behavior itself.

```text
Classification:
GENERAL CAPABILITY:
  baseline/before-state reconstruction for migration analysis

COMPATIBILITY-SPECIFIC:
  Atlas rename add-side interpretation and MF103 rendering
```

**The current implementation satisfies that boundary.** No adjustment is warranted, and the general half was not embedded in the compatibility layer. Evidence, since #1213 asks the PR to answer three questions rather than assert:

### 1. Where does the semantic implementation live?

- The model is `lint.BaselineColumn` in `migration/lint/baseline.go`, exported and recorded in `docs/public_api.snapshot`, alongside `Analysis.BaselineVersions()`.
- Production is `internal/migrationlintreport.lintDirectory` plus `baselineCollector` (`report.go`), reading through `migrationreplay.Options.ObserveVersion` — a general replay hook with no notion of a profile.

None of this is in `cmd/atlas`.

### 2. How can native Ptah consume it?

Through the same public contract the compatibility surface uses, and through the same code path:

- `Analysis.BaselineVersions()` declares which versions need starting state; the caller feeds them back via `Options.Baseline`.
- `cmd/lint/lint.go` (native `ptah migrations lint`) and `cmd/atlas/migrate_lint.go` both go through `internal/migrationlintreport`. `lintDirectory` is one function serving both — there is no second, compat-only baseline path.

Dependency direction verified rather than assumed:

```
go list -deps ./migration/lint        | grep -E 'cmd/atlas|atlasreport|atlasmigrate'  -> nothing
go list -deps ./internal/migrationreplay | grep -E 'cmd/atlas|atlasreport|atlasmigrate' -> nothing
```

The collection mechanism is **demand-driven, not profile-driven**: `lintDirectory` collects exactly the versions the analysis asked for, hands the replay a `nil` observer when it asked for none, and skips the second analysis entirely. A native rule that wanted before-state would get it without touching the compatibility layer.

### 3. If no native surface is added here, why is that reasonable, and what tracks the gap?

Reasonable, and deliberately so. Native `ptah migrations lint` on this PR's own fixture reports:

```
20240102000000_b.sql:1 [warning] BC101: renames are not backwards compatible: ...
1 finding(s).      rc=0
```

That is a rename modeled as a rename. Copying `MF103` there would be a claim about the statement that is not true of it, and #1213 asks for shared capability, not mirrored diagnostics.

The compatibility-specific half is gated in three separate places, all correctly:

- the interpretation — `renameAddSideCandidates` returns nil unless `file.compatibility == CompatibilityProfileAtlas` (`migration/lint/rename.go:217`)
- the code mapping — `DD101 -> MF103` in `internal/atlaslint/rules.go`
- the wording, wrapping and grouping — `internal/atlasreport/migrate_lint_atlas_text.go`

Note the shared layer emits Ptah's own `DD101`, not `MF103`. The Atlas code never exists below the adapter.

## The one seam worth recording

`baselineVersions()` (`migration/lint/baseline.go:144`) is the demand function, and today it enumerates demand through exactly one source — `renameAddSideCandidates`, which is Atlas-gated. So:

- native **access** to the capability exists and is public;
- native **demand** does not exist yet, because no native rule asks.

That is deferred exposure, not compat-only architecture: adding a native clause to `baselineVersions()` is the whole of what a native consumer would need. It is worth recording explicitly so it is a decision rather than a thing nobody noticed, and I have added it to #1214 rather than opening a duplicate issue — #1213 is the prospective rule and explicitly defers the repository-wide audit.

## Not introduced here

`internal/migrationlintreport` pulls `internal/atlaslint` transitively. Checked directly: no file in that package imports it, and this PR added zero such lines. Pre-existing, and out of scope per #1213's own statement that existing debt may remain.

Issue #1245, opened 2026-08-07
A MySQL unique index is added as an index and dropped as a constraint in the same plan


denisvmedia:
Measured on a live MySQL 9.7.1 (`mysql:9.7`, host port 33110) against the pinned Atlas community binary v1.3.0.

In MySQL a `UNIQUE` constraint and its backing unique index are one object. Introspection reports it twice — once as a `DBIndex`, once as a `DBConstraint` of type `UNIQUE`. The two representations are compared in two separate pools, and an unchanged unique index falls between them: it is reported **added** by the index pool and **removed** by the constraint pool, in the same plan.

## Measurement

Target `w3a98_tgt`, desired state is that database's own unedited `atlas schema inspect` output:

```sql
CREATE TABLE users (
  id BIGINT NOT NULL AUTO_INCREMENT,
  email VARCHAR(255) NOT NULL,
  PRIMARY KEY (id),
  UNIQUE KEY uq_users_email (email)
);
```

| binary | exit | output |
| --- | --- | --- |
| pinned Atlas community v1.3.0 | 0 | `Schema is synced, no changes to be made` |
| `ptah-compat` | 0 (`--dry-run`) | see below |

```sql
CREATE UNIQUE INDEX `uq_users_email` ON `w3a98_tgt`.`users` (`email`);
-- ALTER statements: --
ALTER TABLE `users` DROP INDEX `uq_users_email`;
```

The real apply exits 1 — the dev-database simulation replays the baseline and then hits its own plan:

```
Error: dev database simulation failed during plan: failed to execute SQL statement:
SQL execution failed: Error 1061 (42000): Duplicate key name 'uq_users_email'
SQL: CREATE UNIQUE INDEX `uq_users_email` ON `w3a98_tgt`.`users` (`email`);
the plan was not applied to the target database
```

## Mechanism

Instrumented dump of the `SchemaDiff` in `internal/atlasschema/apply.go`, immediately after `applyDiffPolicy`:

```
IndexesAdded=[{Name:uq_users_email TableName:w3a98_tgt.users}]
IndexesRemoved=[]
ConstraintsAdded=[]
ConstraintsRemovedWithTables=[{Name:uq_users_email TableName:users Type:UNIQUE}]
TablesAdded=[] TablesRemoved=[]
```

Three steps, all in `migration/schemadiff/internal/compare`:

[...]

Issue #1240, opened 2026-08-07
The dev-database simulation writes into the target on MySQL, and never cleans up on any dialect


denisvmedia:
Two defects in the same subsystem, both measured against a live MySQL 9.7 and a live PostgreSQL 17, both distinct from #1232 (they survive its fix).

## Defect 1 — on MySQL the simulation executes against the target database

`schema apply` renders the plan with the **target's** schema name baked into every statement, then runs that plan against the dev connection to simulate it. In MySQL a schema *is* a database, so a statement reading `` CREATE TABLE `cr1232`.`users` `` lands in `cr1232` no matter which connection issues it. The simulation therefore mutates the target.

Clean room — two freshly created databases, one apply:

```
ptah-compat schema apply \
  -u      'mysql://.../cr1232' \
  --to    'file://cr.hcl' \
  --dev-url 'mysql://.../cr1232dev' --auto-approve
rc=1
Error: apply schema changes: failed to execute SQL statement:
       SQL execution failed: Error 1050 (42S01): Table 'users' already exists
```

Afterwards:

| database | role | contains |
| --- | --- | --- |
| `cr1232` | **target** | `users` |
| `cr1232dev` | dev | *(nothing)* |

The table exists in the target, created by the step that was supposed to be a simulation, and the command reported that the apply **failed**. The operator is told nothing was applied while the target has been changed.

It then compounds. On the next run the target already has the table, so the plan becomes a `CREATE` plus

```
-- WARNING: This will delete all data!
DROP TABLE IF EXISTS `users`;
```

and the run dies with

```
Error: dev database simulation failed during plan: ... Table 'users' already exists
       ...; the plan was not applied to the target database
```

The pinned Atlas community binary v1.3.0, identical inputs, its own fresh pair:

```
apply #1  rc=0   CREATE TABLE `users` (...)
apply #2  rc=0   Schema is synced, no changes to be made
```

and its dev database is left **empty**.

## Defect 2 — the dev database is never cleaned, on any dialect

The pinned binary leaves its dev database with no objects in it after a successful apply. Ptah leaves its own work behind.

PostgreSQL 17, after three successful applies (`rc=0/0/0`):

| database | role | contains |
| --- | --- | --- |
| `fix1138` | target | `users` |
| `fixdev1138` | **dev** | `users` |

PostgreSQL tolerates the leftover — later applies still report "Schema is synced". MySQL does not: the leftover is part of why every subsequent command against that dev-url fails.

[...]

Issue #1232, opened 2026-08-07
Applying an unchanged schema twice removes constraints: silently on PostgreSQL, with an error on SQLite


denisvmedia:
Found by the same SQLite compatibility sweep as #1231 (947 invocations, both binaries on identical fixtures). This is the sharpest of the 19 cases where **ptah exits 1 and the pinned community binary v1.3.0 exits 0**, and it is filed separately because it breaks the declarative workflow's central promise: converge, then stay converged.

## Reproduction

```hcl
schema "main" {
}
table "t" {
  schema = schema.main
  column "id" {
    null = false
    type = integer
  }
  primary_key {
    columns = [column.id]
  }
}
```

```
<BIN> schema apply --url sqlite://t.db --to file://t.hcl --dev-url "sqlite://dev?mode=memory" --auto-approve   # twice

pinned binary   first 0   second 0   "Schema is synced, no changes to be made"
ptah-compat     first 0   second 1   "Error: generate schema apply SQL:
                                      sqlite: rebuilding table t requires the retained table definition"
```

The first apply succeeds and writes `CREATE TABLE IF NOT EXISTS "t" ("id" integer PRIMARY KEY)`. Every apply after that fails against the database Ptah itself just produced.

## Bisected

I bisected this by hand rather than taking the sweep's word for it.

| desired schema | second apply |
| --- | --- |
| one `text` column, no primary key | **exit 0**, "Schema is synced" |
| one `integer` column + `primary_key` | exit 1 |
| one `text` column + `primary_key` | exit 1 |

So it is not the column type, and not the rowid-alias special case — it is **any inline PRIMARY KEY**. A table without one is idempotent.

## Where it comes from

`internal/planner/dialects/sqlite/sqlite.go` schedules a rebuild when a table diff carries modified columns, removed columns, or added/removed constraints, and then `rebuildTable` refuses when the desired schema does not hold the table:

```go
table := findTable(generated.Tables, target.tableName)
if table == nil {
    return nil, unsupportedFeaturef("rebuilding table %s requires the retained table definition", target.tableName)
}
```

[...]

PR #1254, opened 2026-08-07
Let a MySQL connection name the schema that owns its tables (#1244)

### LINKED ISSUE #1244: MySQL schema apply plans CREATE and DROP for every table: nothing ever names the MySQL-family default schema
Measured on a live MySQL 9.7.1 (`mysql:9.7`, container `mylive-mysql`, host port 33110) against the pinned Atlas community binary v1.3.0.

Fed a canonical Atlas-HCL description of the schema that is **already in the database**, `ptah-compat schema apply` plans `CREATE TABLE` *and* `DROP TABLE` for every table, plus add/drop of every index and foreign key. The pinned binary answers "Schema is synced, no changes to be made" and exits 0.

This is **not** #1232. It was reproduced with `ptah-compat` built from `issue-1232-implicit-schema-constraints` (PR #1238) and the plan is byte-identical to the one from `master`.

## Measurement

Target `w3a98_tgt`:

```sql
CREATE TABLE users (
  id BIGINT NOT NULL AUTO_INCREMENT,
  email VARCHAR(255) NOT NULL,
  name VARCHAR(120) NULL,
  PRIMARY KEY (id),
  UNIQUE KEY uq_users_email (email)
);
CREATE TABLE posts (
  id BIGINT NOT NULL AUTO_INCREMENT,
  user_id BIGINT NOT NULL,
  title VARCHAR(200) NOT NULL,
  PRIMARY KEY (id),
  KEY idx_posts_user (user_id),
  CONSTRAINT fk_posts_user FOREIGN KEY (user_id) REFERENCES users (id)
);
```

Desired state is that database's own `atlas schema inspect` output, unedited.

| binary | exit | plan |
| --- | --- | --- |
| pinned Atlas community v1.3.0 | 0 | `Schema is synced, no changes to be made` |
| `ptah-compat` @ `master` | 0 (`--dry-run`) | CREATE + DROP of every table |
| `ptah-compat` @ `issue-1232-implicit-schema-constraints` | 0 (`--dry-run`) | byte-identical to `master` |

The real apply exits 1:

```
Error: dev database simulation failed during plan: failed to execute SQL statement:
SQL execution failed: Error 1050 (42S01): Table 'users' already exists
SQL: CREATE TABLE `w3a98_tgt`.`users` (...); the plan was not applied to the target database
```

The plan itself:

```sql
CREATE TABLE `w3a98_tgt`.`users` (...);
CREATE TABLE `w3a98_tgt`.`posts` (...);
CREATE INDEX `idx_posts_user` ON `w3a98_tgt`.`posts` (`user_id`);
CREATE UNIQUE INDEX `uq_users_email` ON `w3a98_tgt`.`users` (`email`);
ALTER TABLE `w3a98_tgt`.`posts` ADD CONSTRAINT `fk_posts_user` FOREIGN KEY ...;
ALTER TABLE `posts` DROP FOREIGN KEY `fk_posts_user`;
DROP INDEX `idx_posts_user` ON `posts`;
-- WARNING: This will delete all data!
DROP TABLE IF EXISTS `posts`;
-- WARNING: This will delete all data!
DROP TABLE IF EXISTS `users`;
```

The qualification tells the story: everything created is `w3a98_tgt`-qualified, everything dropped is bare.

## Root cause

Instrumented dump of the two comparison sides at `internal/atlasschema/apply.go`, immediately before `schemadiff.CompareWithDatabase`:

```
dialect="mysql" info.Schema="w3a98_tgt" sem.DefaultSchema=""
desired table: Schema="w3a98_tgt" Name="users"
desired table: Schema="w3a98_tgt" Name="posts"
current table: Schema=""           Name="posts"
current table: Schema=""           Name="users"
```

MySQL introspection reports every table with an **empty** schema; a desired state written as Atlas HCL carries `schema = schema.<database>`. Both sides are keyed through `newTableIdentity` (`migration/schemadiff/internal/compare/column_uniqueness.go:175`), which resolves an absent schema to `semantics.DefaultSchema` — but `identifier.ForDialect` returns `DefaultSchema: ""` for MySQL and MariaDB (`core/platform/identifier/identifier.go:80-86`). The normalization is a no-op, the keys never meet, and every table reads as both added and removed.

The empty value is not itself wrong: a MySQL-family schema *is* a database, so no static dialect rule can name it. The defect is that nothing then supplies it from the place that does know — `DBInfo.Schema`, which `getDatabaseInfo` already sets to the connected database name (`dbschema/connection.go:761-783`). SQL Server has exactly this shape and already pins the field from the same place (`dbschema/connection.go:844-845`):

```go
info.IdentifierSemantics = identifier.ForSQLServerCatalog(catalogCollation)
info.IdentifierSemantics.DefaultSchema = info.Schema
```

`internal/atlasschema/scope.go:79-92` already documents the rule this violates: *"MySQL-family schemas are databases, so only a database-backed side can name the default."* The filter path honors it; the comparison path does not.

### A second cause behind it

Setting `DefaultSchema` on the connection is necessary but not sufficient, because `compare.Constraints` **discards the resolved semantics and rebuilds static ones from the dialect string**. On `issue-1232-implicit-schema-constraints`, `migration/schemadiff/internal/compare/constraints.go:68`:

```go
semantics := identifier.ForDialect(dialect)
```

`CompareWithOptions` computes and validates an `identifierSemantics` value and threads it into `TablesAndColumnsWithSemantics` and `IndexesWithSemantics`, but passes `compare.Constraints` only `opts` (`migration/schemadiff/schemadiff.go:181`). So the constraint half re-derives `DefaultSchema: ""` no matter what the connection resolved, and #1232's normalization cannot bite on MySQL.

With the connection fix alone, the table churn disappears but this remains:

```sql
ALTER TABLE `posts` DROP PRIMARY KEY;
ALTER TABLE `posts` DROP FOREIGN KEY `fk_posts_user`;
ALTER TABLE `users` DROP PRIMARY KEY;
```

which is the silent-corruption shape #1232 describes on PostgreSQL. Both causes have to be fixed together; shipping only the first trades a loud failure for a quiet one.

## Verification

With both fixes, on a control schema of PK + FK + plain index, `ptah-compat` and the pinned binary agree:

| binary | exit | output |
| --- | --- | --- |
| pinned Atlas community v1.3.0 | 0 | `Schema is synced, no changes to be made` |
| `ptah-compat` (both fixes) | 0 | `Schema is synced, no changes to be made.` |

A residue remains when the schema carries a **unique index**; that is a different mechanism and is filed separately.

## Definition of done

- A MySQL or MariaDB connection resolves the default schema that owns unqualified objects from the connected database, the way SQL Server already does.
- Constraint comparison receives the same validated identifier semantics that table and index comparison receive, rather than re-deriving static ones from the dialect name.
- `schema apply` fed a database's own inspected description reports synced and exits 0, matching the pinned binary.

## Notes

- Depends on #1232 / PR #1238: without that branch's normalized `tableMemberKey`, constraint comparison consults no semantics at all and the connection-side fix cannot reach it.
- The dev-database halves of this failure — the simulation executing target-qualified DDL against the dev connection, and the dev database never being cleaned — are #1240, independently reproduced here.

denisvmedia:
Fixes #1244. Supersedes #1250, which was branched from `issue-1232-implicit-schema-constraints` before #1238 merged; this is the same commit rebased onto master, where that work now lives.

## What was wrong

Fed a canonical Atlas-HCL description of the schema **already in the database**, `ptah-compat schema apply` on MySQL planned `CREATE TABLE` *and* `DROP TABLE` for every table, plus add and drop of every index and foreign key, then exited 1 and applied nothing. Measured on a live MySQL 9.7.1 against the pinned Atlas community binary v1.3.0, which answers `Schema is synced, no changes to be made` and exits 0.

Reproduced with `ptah-compat` built from `issue-1232-implicit-schema-constraints`, byte-identical to the plan from `master` — so it is not the constraint-keying defect #1232 fixed.

## Cause

Everything created was database-qualified, everything dropped was bare. Instrumented dump of the two sides immediately before `schemadiff.CompareWithDatabase`:

```
dialect="mysql" info.Schema="w3a98_tgt" sem.DefaultSchema=""
desired table: Schema="w3a98_tgt" Name="users"
current table: Schema=""           Name="users"
```

A MySQL reader reports every table with an empty schema; a desired state written as Atlas HCL carries `schema = schema.<database>`. Both sides key through `newTableIdentity`, which resolves an absent schema to `semantics.DefaultSchema` — but a MySQL-family schema *is* a database, so `identifier.ForDialect` has no static default to give and returns an empty one. The normalization was a no-op and the keys never met.

The empty value is correct. Nothing was supplying the real one from the place that has it.

## What changed

Three fixes, one family, all needed together.

1. **`dbschema/connection.go`** — the MySQL/MariaDB branch pins `IdentifierSemantics.DefaultSchema` from `info.Schema`, exactly as `getSQLServerDatabaseInfo` already does for the dialect with the same shape. `internal/atlasschema/scope.go` already stated the rule this violated: *"MySQL-family schemas are databases, so only a database-backed side can name the default."* The filter path honored it; the comparison path did not.

2. **`ConstraintsWithSemantics`** — constraint comparison discarded the resolved semantics and rebuilt static ones from the dialect string (`semantics := identifier.ForDialect(dialect)`), so it saw the empty default no matter what the connection had resolved. Table and index comparison already take the validated value; constraints now do too.

3. **`tablePrimaryKeyConstraintName`** — the same raw-string table comparison, in a spot #1232 did not reach. It adopts the name the database gave the primary key; comparing the two spellings directly meant the lookup never matched and the name always came from the fallback. Invisible wherever the fallback agrees with the engine (MySQL always says `PRIMARY`, PostgreSQL usually says `<table>_pkey`), and wrong the moment a schema names its own primary key — surfacing as a drop of the real constraint plus an add of a differently named one. Found by the test written for (1) and (2).

**Half of this is worse than none.** With only (1), the table churn disappears and `ALTER TABLE users DROP PRIMARY KEY` stays behind — the silent-corruption shape #1232 describes on PostgreSQL, reached from another direction. Both halves land together.

## Verification

Live MySQL 9.7.1, target of primary key + foreign key + plain index, desired state its own unedited `atlas schema inspect` output:

| binary | exit | output |
| --- | --- | --- |
| pinned Atlas community v1.3.0 | 0 | `Schema is synced, no changes to be made` |
| `ptah-compat` before | 1 | `CREATE TABLE` + `DROP TABLE` for every table |
| `ptah-compat` after | 0 | `Schema is synced, no changes to be made.` |

The target database was unchanged after the run.

Each fix was checked against its own **inverse mutant** — reverting any one of the three reddens the test that covers it, and only that test. Notably, reverting (3) leaves the `postgres conventional name` row green and reddens only the schema-chosen-name rows, which is exactly why nobody caught it.

The two controls (a constraint on a table the desired schema does not have; a table in another database) fail any change that buys idempotency by making removal impossible.

`dbschema/mysql_identifier_live_test.go` is env-gated on `MYSQL_TEST_URL` / `MARIADB_TEST_URL`; it was run against the live MySQL 9.7.1 and passes, and fails when (1) is reverted. The MariaDB row was skipped — not measured.

## What is not fixed here

A residue remains when the schema carries a **unique index**. MySQL reports it both as an index and as a `UNIQUE` constraint; the object is added by the index pool and removed by the constraint pool. Different mechanism, filed as #1245 with the instrumented diff.

The dev-database halves of the original failure — the simulation executing target-qualified DDL against the dev connection, and the dev database never being cleaned — are #1240, independently reproduced during this work. Neither is touched here.

## Gates

Run on the rebased tree, each exit status read on its own line.

| gate | exit |
| --- | --- |
| `go build ./...` | 0 |
| `go vet ./...` | 0 |
| `go test ./... -count=1 -p 2` | 0 |
| `golangci-lint run ./...` | 0 |

`go test ./...` at default parallelism is load-flaky on the machine this ran on — `cmd/viz` and `cmd/ptah-ls` time out, and both reproduce on an unmodified base checkout. `-p 2` is clean.

No documentation change is owed: no flag, output format, config key or documented behavior changes, and nothing in the docs claimed the broken shape.

### DIFF
--- a/dbschema/connection.go
+++ b/dbschema/connection.go
@@ -781,6 +781,18 @@ func getDatabaseInfo(ctx context.Context, db *sql.DB, dialect string, parsedURL 
 			}
 			info.Schema = dbName
 		}
+		// A MySQL-family schema is a database, so no static dialect rule can
+		// name the one that owns an unqualified table the way "public" and
+		// "main" do; only the connection knows it. Leaving the field empty is
+		// what made the two comparison sides key differently: the catalog
+		// reports every table with no schema, while a desired state written as
+		// Atlas HCL carries `schema = schema.<database>`. Measured on live
+		// MySQL 9.7.1, `schema apply` fed the database's own inspected HCL
+		// planned CREATE TABLE and DROP TABLE for every table where the pinned
+		// Atlas community v1.3.0 binary answered "Schema is synced, no changes
+		// to be made" (stokaro/ptah#1244). SQL Server pins the same field from
+		// the same place, in getSQLServerDatabaseInfo.
+		info.IdentifierSemantics.DefaultSchema = info.Schema
 	case platform.ClickHouse:
 		var version string
 		if err := db.QueryRow("SELECT version()").Scan(&version); err != nil {
--- a/dbschema/mysql_identifier_live_test.go
+++ b/dbschema/mysql_identifier_live_test.go
@@ -1,0 +1,84 @@
+package dbschema_test
+
+import (
+	"net/url"
+	"os"
+	"strings"
+	"testing"
+
+	qt "github.com/frankban/quicktest"
+
+	"go.5x5.cz/ptah/dbschema"
+)
+
+// TestMySQLLiveConnection_DefaultSchemaIsTheConnectedDatabase pins that a
+// MySQL-family connection reports the connected database as the schema that
+// owns unqualified objects (stokaro/ptah#1244).
+//
+// A schema on these engines is a database, so the offline dialect rules have no
+// static default to offer and deliberately carry an empty one -- unlike
+// PostgreSQL's "public" or SQLite's "main". The connection is the only place
+// that knows the answer, and until it filled the field in, schema comparison
+// keyed a table the catalog reports with no schema differently from the same
+// table a desired state names explicitly, and planned to create and drop every
+// one of them. SQL Server resolves the same field from the same place.
+//
+// Live-only: the value comes from the URL path or from SELECT DATABASE(), and
+// asserting it against anything but a real connection would only restate the
+// test's own fixture.
+func TestMySQLLiveConnection_DefaultSchemaIsTheConnectedDatabase(t *testing.T) {
+	tests := []struct {
+		name           string
+		environmentKey string
+	}{
+		{name: "mysql", environmentKey: "MYSQL_TEST_URL"},
+		{name: "mariadb", environmentKey: "MARIADB_TEST_URL"},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			c := qt.New(t)
+			databaseURL := requireLiveMySQLURL(c, test.environmentKey)
+			wantSchema := databaseNameFromURL(c, databaseURL)
+
+			conn, err := dbschema.ConnectToDatabase(c.Context(), databaseURL)
+			c.Assert(err, qt.IsNil)
+			c.Cleanup(func() {
+				dbschema.CloseAndWarn(conn)
+			})
+
+			info := conn.Info()
+
+			c.Assert(info.Schema, qt.Equals, wantSchema)
+			c.Assert(info.IdentifierSemantics.DefaultSchema, qt.Equals, wantSchema)
+			// The pairing is the invariant, not either value alone: comparison
+			// resolves an absent schema through DefaultSchema, and a connection
+			// whose two fields disagree would key its own tables against a
+			// database it is not connected to.
+			c.Assert(info.IdentifierSemantics.DefaultSchema, qt.Equals, info.Schema)
+		})
+	}
+}
+
+func requireLiveMySQLURL(c *qt.C, environmentKey string) string {
+	c.Helper()
+	databaseURL := os.Getenv(environmentKey)
+	if databaseURL == "" {
+		c.Skip(environmentKey + " is not set")
+	}
+	return databaseURL
+}
+
+// databaseNameFromURL reads the database the URL selects. A MySQL URL without a
+// path leaves the connection to resolve it with SELECT DATABASE(), which this
+// test cannot predict, so it is skipped rather than guessed at.
+func databaseNameFromURL(c *qt.C, rawURL string) string {
+	c.Helper()
+	parsed, err := url.Parse(rawURL)
+	c.Assert(err, qt.IsNil)
+	name := strings.TrimPrefix(parsed.Path, "/")
+	if name == "" {
+		c.Skip("the configured URL names no database")
+	}
+	return name
+}
--- a/migration/schemadiff/internal/compare/constraint_synthesis.go
+++ b/migration/schemadiff/internal/compare/constraint_synthesis.go
@@ -104,7 +104,7 @@ func synthesizeTablePrimaryKeyConstraints(
 			continue
 		}
 
-		name := tablePrimaryKeyConstraintName(table, database.Constraints, dialect)
+		name := tablePrimaryKeyConstraintName(table, database.Constraints, dialect, semantics)
 		synthesized = append(synthesized, goschema.Constraint{
 			StructName: table.StructName,
 			Name:       name,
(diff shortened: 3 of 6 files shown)


denisvmedia:
Duplicate of #1250 — the two branches have byte-identical trees (`git diff origin/issue-1244-mysql-default-schema origin/issue-1244-mysql-connection-default-schema` is empty).

#1250 is the one being carried forward: it was rebased onto master after #1232 merged, so it no longer duplicates that commit, and its CI has run on the rebased tree. Closing this one to avoid two reviews of the same change.

Issue #1251, opened 2026-08-07
compat schema inspect: suppress the Ptah superset, and fix the bare identifiers that make our own blocks unreadable


denisvmedia:
Product decision taken: `ptah-compat schema inspect` should not be called compatible while its output is unreadable to the tool it stands in for. Suppression belongs on the **compatibility surface only** — native `ptah` inspect must keep everything it models.

Measured against the pinned Atlas community binary v1.3.0 on a live PostgreSQL 17 database carrying an extension, an enum, a sequence, a table, a view, a materialized view, a function, a trigger, RLS with a policy, a role and two grants.

## The measurement changes the plan

Suppression alone will **not** make the output readable. Most of the unreadability is Ptah's own rendering, not the superset.

Method: build a base file the binary accepts, then add one block type back at a time. My first attempt had an invalid base — the table has an enum-typed column and the base did not declare the type — so every row inherited that error. Reported here only after the base was made valid; the rows below are one changed operand each.

### Category A — Ptah rendering defects, unreadable because of us

Same class as the `to = PUBLIC` / `privileges = [USAGE]` defect fixed in #1248: a bare word in an attribute position is an HCL variable reference with nothing behind it, and the binary refuses the whole file.

| block | what Ptah emits | the binary's answer |
| --- | --- | --- |
| `sequence` | `type = bigint` | `There is no variable named "bigint"` |
| `function` | `lang = PLpgSQL`, `return = trigger`, `security = INVOKER`, `volatility = VOLATILE` | `There is no variable named "trigger"` |
| `trigger` | `for = ROW` | `There is no variable named "ROW"` |
| `policy` | `for = ALL` | `There is no variable named "ALL"` |

Two more that are not about bare words:

| defect | detail |
| --- | --- |
| `enum` block has no `schema` | the binary answers `extract schema name from enum reference` |
| an enum-typed column renders as `type = sql("status")` | the type does not exist in the dev database, so it fails with `pq: type "status" does not exist`. The binary writes `type = enum.status` — a reference to the declared block |

**These should be fixed, not suppressed.** Suppressing a block that is only unreadable because of our own rendering would hide a Ptah defect behind a compatibility argument.

### Category B — the genuine superset

Three block types answer with an explicit feature message rather than a parse error:

```
extension   postgres: extensions are not supported by this version.
sequence    postgres: sequences are not supported by this version.
policy      postgres: policies are not supported by this version.
```

[...]

PR #1256, opened 2026-08-07
compat inspect: omit the block types the pinned binary refuses (#1251, Category B)

denisvmedia:
Closes the Category B half of stokaro/ptah#1251. Category A (the bare identifiers in `sequence`, `function`, `trigger` and `policy`, the `enum` block's missing schema, and the enum-typed column) is a separate front and is **not** touched here.

## What this is: a COMPATIBILITY ADAPTER

Under stokaro/ptah#1213 this is a compatibility adapter, not a general capability. Nothing here improves Ptah's model of a database — it strictly *reduces* what one binary describes, and the reduction's only justification is a property of the tool `ptah-compat` stands in for. A general capability would apply on both binaries and would survive that tool changing; this must not, and is measured so it does not. The whole point of the oracle run below is that the day the pinned binary starts modeling one of these blocks, the adapter stops filtering it.

## The measurement

PostgreSQL 17, live database carrying an extension, an enum, a sequence, a table, a view, a materialized view, a function, a trigger, RLS with a policy, a role and two grants. Base = the pinned binary's **own** inspect output for that database, asserted at exit 0 first; then one bare block added at a time.

| block added | exit | answer |
| --- | --- | --- |
| *(base alone)* | **0** | — |
| `extension "pgcrypto" {}` | 1 | `postgres: extensions are not supported by this version` |
| `sequence "order_seq" {}` | 1 | `postgres: sequences are not supported by this version` |
| `policy "accounts_all" {}` | 1 | `postgres: policies are not supported by this version` |
| `role`, `function`, `view`, `materialized`, `trigger`, `permission`, `range` | 0 | dropped, file read |
| `wibble "x" {}` | 0 | dropped, file read |

Two things make these rows mean what they say.

**The blocks are stripped to their label.** Carrying the attributes Ptah writes today, `sequence` comes back with `There is no variable named "bigint"` — Ptah's own rendering, same class as the `to = PUBLIC` defect fixed in #1248, and not this front's to suppress. Bare, what is left is a refusal of the block *type*, which no spelling can lift.

**`wibble` is why exit 0 on the others is not evidence of support.** That binary drops a top-level block whose name it does not model and carries on. Exit 0 says only "harmless to the file" — which is the entire test a compatibility surface needs. Omitting `role` or `view` would lose description and buy nothing.

Also recorded, not acted on:

- `domain`/`composite` carrying `type = int` are refused with `schemahcl: failed reading spec as *postgres.doc: set field "type"`. Bare, both are accepted. That is an attribute spelling inside an ignored block — Category A's class, not a feature gap.
- Ptah's unlabeled `data {` collides with a real Atlas `data` block (`must have exactly 2 labels`). Out of scope: an inspected database read never produces `ManagedData`, so `data` cannot appear in inspect output.
- **The binary's own inspect** of that same rich database emits exactly three block types: `enum`, `schema`, `table`. It silently omits the extension, the sequence, the view, the materialized view, the function, the trigger, the policy, the role and both grants. We do not copy that; we omit only what it cannot *read*.

## Scope, as narrow as the measurement

- **Compatibility surface only.** `cmd/atlas` sets `OmitAtlasRefusedBlocks: true`. Native `ptah schema inspect` sets nothing. Verified against the live database: native output is byte-identical to before this change.
- **HCL only.** `--format sql` keeps all three. SQL output is read by a database, not by that binary; filtering it would lose data for no reader.
- **PostgreSQL only.** On SQLite the same three blocks are accepted and dropped exactly like `wibble`, all at exit 0 — the refusal is the PostgreSQL driver's, so the list is keyed by dialect and SQLite has no entries. MySQL/MariaDB were **not** measured and suppress nothing.

## Suppression is visible

Every omitted object is reported through the renderer's existing loss-diagnostic channel — the one already used for constructs Ptah cannot represent — which `schema inspect` writes to stderr:

```
warning: extensions.pgcrypto: omitted from Atlas-compatible schema inspect output: the Atlas community CLI refuses a postgres schema file that declares any extension block, and one of them makes the whole document unreadable to it
warning: sequences.order_seq: ...
warning: rls_policies.accounts_all: ...
```

## The list is re-measured, not frozen

Two oracle runs in `internal/atlashclrender`, following `TestOracleModeledColumnTypesMatchTheBinary`, both enumerated by name in `.github/workflows/atlas-oracle.yml` (count raised 2 → 4, plus per-subtest `grep -Fq` lines so a selection that matched nothing cannot pass):

- **`TestOracleAtlasRefusedBlockTypesMatchTheBinary`** — each entry must still be refused **and** still refused with `are not supported by this version`. A build that starts modeling one turns the job red rather than withholding it forever; a row that starts failing over one of *our* spellings stops counting as a feature gap. A `base` subtest asserts the base exits 0 before any row is believed — the earlier attempt at this measurement built an invalid base and every row inherited its error.
- **`TestOracleToleratesTheBlockTypesPtahStillRenders`** — the opposite direction. `role`, `function`, `view`, `materialized`, `trigger`, `permission` and `wibble` on PostgreSQL, and `extension`/`sequence`/`policy` on SQLite, must all stay readable; each also asserts it is *absent* from the suppression list, so the list keeps a boundary.

Probes run serially: each dialect's probes share one dev database and the binary materializes the file there.

Mutation-checked, all four directions:

| mutant | result |
| --- | --- |
| suppression reverted (`omitsBlock` → false) | 5 red, incl. `...OmitsTheBlocksTheBinaryRefuses`, `...ReportsEveryOmission` |
| guard dropped (suppression leaks to native) | red: `...KeepsEveryBlockTypeOnTheNativeSurface`, `...NativeReportsNoOmission` |
| `role` added to the refused list | red: oracle `refused/role` **and** control `tolerated/role` |
| `sequence` moved out of the refused list | red: oracle `tolerated/sequence` |

## What this does NOT do

**It does not make compat inspect readable on its own, and it is not claimed to.** With the three blocks gone, the pinned binary refuses the real `ptah-compat` output at the next bare identifier:

```
Error: compat.hcl:43,10-17: Unknown variable; There is no variable named "PLpgSQL".
```

The first refusal moved from `bigint` (the sequence) to `PLpgSQL` (the function), which is itself the evidence the Category B blocks are gone. Hand-fixing the remaining Category A spellings on that same file — the four bare words, the `enum` block's `schema`, and `type = sql("status")` → `type = enum.status` — brings it to **exit 0**. So the two fronts are jointly sufficient and neither is sufficient alone. The issue's definition of done ("no hand editing") needs Category A to land too.

## Not verified

- MySQL, MariaDB, ClickHouse: not measured. They suppress nothing, which is the status quo.
- PostgreSQL-family dialects other than `postgres` (`cockroachdb`, `yugabytedb`, `spanner`): not measured, not suppressed.
- Whether that binary can use an `enum`-declared type for a column was not separately established here either; it writes `type = enum.status` itself and read the hand-fixed file at exit 0, which is consistent but is not an isolated fixture.

### DIFF
--- a/.github/workflows/atlas-oracle.yml
+++ b/.github/workflows/atlas-oracle.yml
@@ -133,18 +133,24 @@ jobs:
       # The set of column types the HCL schema models is the third contract in
       # this repository whose authority is the community binary rather than any
       # document (stokaro/ptah#1138). A wrong row emits HCL that binary refuses
-      # to read, so the set is re-measured here instead of being trusted.
-      # Listed before it is run, for the same reason as above.
-      - name: Verify column-type conformance selection
+      # to read, so the set is re-measured here instead of being trusted. The
+      # set of top-level BLOCK types it refuses is the same kind of contract
+      # (stokaro/ptah#1251): ptah-compat omits those blocks, and a list nothing
+      # re-measures would go on withholding a construct that binary started
+      # modeling. Both live in internal/atlashclrender and are listed before
+      # they are run, for the same reason as above.
+      - name: Verify HCL render conformance selection
         run: |
           GOWORK=off go test -list '^TestOracle' \
             ./internal/atlashclrender > "$RUNNER_TEMP/atlas-column-type-tests"
           cat "$RUNNER_TEMP/atlas-column-type-tests"
           grep -Fx 'TestOracleModeledColumnTypesMatchTheBinary' "$RUNNER_TEMP/atlas-column-type-tests"
           grep -Fx 'TestOracleAcceptsTheWrapItFallsBackTo' "$RUNNER_TEMP/atlas-column-type-tests"
-          test "$(grep -c '^TestOracle' "$RUNNER_TEMP/atlas-column-type-tests")" -eq 2
+          grep -Fx 'TestOracleAtlasRefusedBlockTypesMatchTheBinary' "$RUNNER_TEMP/atlas-column-type-tests"
+          grep -Fx 'TestOracleToleratesTheBlockTypesPtahStillRenders' "$RUNNER_TEMP/atlas-column-type-tests"
+          test "$(grep -c '^TestOracle' "$RUNNER_TEMP/atlas-column-type-tests")" -eq 4
 
-      - name: Run column-type conformance tests
+      - name: Run HCL render conformance tests
         run: |
           GOWORK=off go test -count=1 -v -run '^TestOracle' ./internal/atlashclrender \
             > "$RUNNER_TEMP/atlas-column-type-run" 2>&1 || {
@@ -156,6 +162,15 @@ jobs:
           # database URL -- which would leave every PostgreSQL row unmeasured
           # while the job stayed green. Both are fatal here.
           ! grep -q 'SKIPPED' "$RUNNER_TEMP/atlas-column-type-run"
+          # A selection that matched no subtest is the other way to stay green
+          # while measuring nothing: the refused list is the whole subject here,
+          # so its rows are named rather than counted in bulk.
+          grep -Fq 'TestOracleAtlasRefusedBlockTypesMatchTheBinary/postgres/base' "$RUNNER_TEMP/atlas-column-type-run"
+          grep -Fq 'TestOracleAtlasRefusedBlockTypesMatchTheBinary/postgres/refused/extension' "$RUNNER_TEMP/atlas-column-type-run"
+          grep -Fq 'TestOracleAtlasRefusedBlockTypesMatchTheBinary/postgres/refused/policy' "$RUNNER_TEMP/atlas-column-type-run"
+          grep -Fq 'TestOracleAtlasRefusedBlockTypesMatchTheBinary/postgres/refused/sequence' "$RUNNER_TEMP/atlas-column-type-run"
+          grep -Fq 'TestOracleToleratesTheBlockTypesPtahStillRenders/postgres/tolerated/wibble' "$RUNNER_TEMP/atlas-column-type-run"
+          grep -Fq 'TestOracleToleratesTheBlockTypesPtahStillRenders/sqlite/tolerated/sequence' "$RUNNER_TEMP/atlas-column-type-run"
 
       - name: Regenerate Atlas CE corpus
         run: >-
--- a/cmd/atlas/schema_inspect.go
+++ b/cmd/atlas/schema_inspect.go
@@ -38,6 +38,12 @@ require --dev-url: the dev database is reset, the source is materialized on
 it, and the result is introspected, mirroring Atlas dev-database
 normalization.
 
+On PostgreSQL, HCL output omits ` + "`extension`" + `, ` + "`sequence`" + ` and
+` + "`policy`" + ` blocks: Atlas CE refuses a whole schema file that declares
+any one of them, so emitting one would make the output unreadable to the tool
+this binary stands in for. Each omitted object is reported on standard error.
+SQL output keeps them, and native ` + "`ptah schema inspect`" + ` omits nothing.
+
 The default output is HCL. SQL output is supported with --format sql or
 --format '{{ sql . }}', JSON with --format json, and custom Go templates
 through the same --format flag. Split/write exports support the documented
@@ -156,7 +162,13 @@ func runAtlasSchemaInspect(cmd *cobra.Command, opts atlasSchemaInspectOptions) e
 
 		// Atlas-compatible surface; see cmd/atlas/schema_apply.go.
 		IgnoreUnknownHCLNames: true,
-		Vars:                  schemaVars,
+		// This surface stands in for a binary that refuses a whole schema file
+		// for containing an extension, sequence or policy block, so it renders
+		// what that binary can read and reports what it left out on stderr
+		// (stokaro/ptah#1251). Native `ptah schema inspect` sets nothing here
+		// and keeps describing every construct Ptah models.
+		OmitAtlasRefusedBlocks: true,
+		Vars:                   schemaVars,
 	})
 	if err != nil {
 		return cmdutil.Fail(cmd, err)
--- a/docs/site/src/content/docs/atlas/schema-commands.md
+++ b/docs/site/src/content/docs/atlas/schema-commands.md
@@ -142,6 +142,38 @@ community binary applies on a schema-bound URL. A pattern too deep for any
 scope, such as `*.*.*.*`, is refused before a database is contacted, so its
 message carries no schema prefix.
 
+### Blocks the compatibility surface leaves out
+
+`ptah-compat schema inspect` omits three top-level HCL block types on
+PostgreSQL: `extension`, `sequence`, and `policy`. The pinned Atlas community
+binary refuses an entire schema file that declares any one of them, answering
+`postgres: extensions are not supported by this version` and the equivalent for
+the other two. Output a drop-in replacement writes has to be output the tool it
+replaces can read back, and one such block costs the whole document.
+
+Nothing is dropped quietly. Every omitted object is reported on standard error,
+one line each, alongside the other loss diagnostics inspection already writes:
+
+```console
+$ ptah-compat schema inspect --url "$PG_URL" > schema.hcl
+warning: extensions.pgcrypto: omitted from Atlas-compatible schema inspect output: ...
+warning: sequences.order_seq: omitted from Atlas-compatible schema inspect output: ...
+warning: rls_policies.accounts_all: omitted from Atlas-compatible schema inspect output: ...
+```
+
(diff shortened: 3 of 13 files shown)


denisvmedia:
## Please hold this — it discards a capability rather than gating it

The measurement is excellent and I verified the load-bearing parts myself, including the `wibble` control that stops the list over-reaching. What needs to change is the shape, not the evidence.

As written, `cmd/atlas/schema_inspect.go:170` sets `OmitAtlasRefusedBlocks: true` unconditionally and nothing turns it off. After this, `ptah-compat schema inspect` **cannot express** an extension, a sequence or a row-level security policy at all. Ptah models all three; the compatibility surface would stop being able to say so.

That crosses a standing project rule, which until an hour ago lived only in the #951 umbrella and in project knowledge rather than in the repository. It is now written down in #1259 (`AGENTS.md` plus the user documentation):

> Compatibility never removes a capability. `ptah-compat` is the migration path for Atlas **Pro** users' scripts as much as CE users'. A capability reachable only through native `ptah` verbs is not a migration path — it is a rewrite.

The three block types this omits are exactly the ones the community CLI answers `... are not supported by this version` for. "This version" is the point: they are where Ptah reaches past CE. Deleting them from compat is the one direction we do not go.

## What to change

Keep everything measured. Add the gate.

- **Default stays as you have it** — omit, so the output is readable by the pinned binary. CE drop-in is not optional and unreadable output is a real defect.
- **Add a `PTAH_*` environment variable** that restores the full superset on the same surface. Not a flag: the conformance `cli-surface` tier asserts flag parity with the pinned binary, so a flag Atlas does not have would break the promise the change exists to serve. Follow `PTAH_ALLOW_EXTERNAL_SCHEMA` (`internal/atlassource/atlassource.go:74`) for the spelling and the plumbing.
- **Keep the stderr diagnostics** exactly as they are — they are what makes the default honest, and they should name the environment variable so an operator learns how to get the full output from the message itself.
- **Test both states**, and make the restored path assert the omitted blocks are present rather than only that the flag was read.
- **Document it** on the Atlas overview page next to the new rule, and give it a feature-matrix row, so the capability is constituted rather than implied by a branch.

## Not in dispute

The measurement stands and should survive the reshape unchanged: the three-block list, the `wibble` control proving exit 0 means "harmless to the file" rather than "supported", stripping blocks to their label to separate a block-type refusal from Category A attribute spellings, the dialect keying, HCL-only scope, and the oracle runs enumerated in `atlas-oracle.yml` with per-subtest `grep -Fq`. That last part is the piece I most want kept: the day the pinned binary starts modeling one of these, the run goes red instead of the adapter quietly filtering forever.

Your #1213 classification is right too — this is a compatibility adapter. The environment variable does not change that; it keeps the adapter from becoming a product decision nobody made.

denisvmedia:
## Second, independent reason to hold: the output references an object it deleted

One of the three reviewers found this and I reproduced it on the branch (`a172cc0d`) against a live PostgreSQL 17.

A sequence backing a column default is an ordinary shape:

```sql
CREATE SEQUENCE order_seq;
CREATE TABLE orders (id integer NOT NULL DEFAULT nextval('order_seq'::regclass));
```

`ptah-compat schema inspect` on this branch emits:

```hcl
    default = sql("nextval('order_seq'::regclass)")
```

and **no** `sequence "order_seq"` block, because the block was suppressed and the reference to it was not.

The consequence is not a cosmetic one:

```
ptah-compat schema inspect -u file://<that output>   rc=1
Error: materialize schema on dev database: failed to execute SQL statement ...
```

**Ptah can no longer read its own output.** Before this change that file round-tripped at exit 0. So for any PostgreSQL database with an explicitly declared sequence behind a column default, the change makes the output worse, not more compatible — it is unreadable by the pinned binary *and* by Ptah, where before it was at least readable by Ptah.

This is separate from the capability question in my previous comment. Even with the `PTAH_*` gate added, the default path still has to be self-consistent.

## What the reshape has to satisfy

A suppressed block must leave no dangling reference. Three candidate shapes, and the measurement should pick between them rather than intuition:

1. drop the referencing attribute too — but then the emitted schema silently loses the column's default, which is data loss of a different kind;
2. keep the reference and inline whatever the binary would accept in that position, if anything does;
3. do not suppress a block that something else in the document references, and report *that* on stderr — narrower suppression, self-consistent output.

Option 3 is the one I would measure first: it preserves the property that the emitted document is valid, and it makes the suppression list dynamic rather than a fixed set of three, which is closer to what "omit what the binary refuses" actually means.

## On the vote

The workflow recorded this front as surviving 1-of-3 refutation, i.e. majority. That aggregation is wrong for a finding of this shape: two reviewers confirming the measurement and one finding a regression neither looked for is not two-against-one. The measurement in this PR is genuinely good and I confirmed the load-bearing parts myself — the three-block list, the `wibble` control, stripping blocks to their label, the oracle runs. None of that is in dispute. The regression is, and it only takes one person to find it.

denisvmedia:
## Measured the design question, and the answer changes what this PR can promise

I said the reshape should pick between three options by measurement rather than intuition. Here is the measurement, and it rules out the framing rather than one of the options.

**The pinned community binary v1.3.0 emits the same dangling reference, and cannot read its own output.**

Same database — `CREATE SEQUENCE order_seq; CREATE TABLE orders (id integer NOT NULL DEFAULT nextval('order_seq'::regclass));`

```hcl
# atlas schema inspect -u postgres://.../seqref   rc=0
table "orders" {
  schema = schema.public
  column "id" {
    null    = false
    type    = integer
    default = sql("nextval('order_seq'::regclass)")
  }
}
schema "public" {
  comment = "standard public schema"
}
```

No `sequence` block — it does not model them. Feed that straight back:

```
atlas schema inspect -u file://<its own output> --dev-url ...
rc=1
Error: create "orders" table: pq: relation "order_seq" does not exist
```

## What follows

For a sequence-backed column default there is **no faithful output that the pinned binary can read** — not Ptah's, and not its own. Suppression cannot make one; it can only choose which kind of broken.

That collapses my three options into a real trilemma:

1. **suppress the sequence, keep the default** — what this branch does, and what the binary does. Output readable by neither.
2. **keep the sequence block when something references it** — output readable by Ptah, refused by the binary on a feature it does not have. Loses nothing.
3. **suppress the sequence and the default together** — readable by both, and silently drops the column's default. That is data loss in the description of the user's database.

Option 1 reproduces a defect, which the compatibility policy's second half forbids: *matching is the floor, not the ceiling; we do not copy defects.* Option 3 tells the operator less than the truth, which the same policy's first half is about.

**Option 2 is the one that survives both halves**, and the newly merged capability rule points the same way: default to what the binary accepts, but never at the cost of emitting something nothing can consume.

## So the suppression list has to be dynamic

"Omit these three block types" is the wrong shape. The right one is closer to:

> omit a block the binary refuses **unless** something else in the document references it; report either way.

That keeps the emitted document valid, makes the common case (extensions, standalone sequences, policies) readable by the binary as this PR intends, and stops the referenced case from producing a file with a hole in it.

It also means this PR should say plainly what it cannot deliver: a PostgreSQL database with a sequence-backed default will not produce Atlas-readable inspect output, because no such output exists. Better to state that than to imply the surface is now compatible.

## Still not in dispute

The three-block measurement, the `wibble` control, stripping to labels, the dialect keying, the stderr diagnostics and the oracle runs in `atlas-oracle.yml`. All of that should survive the reshape unchanged.

denisvmedia:
Hold merge on #1256 until the new #1251 classification is resolved. The implementation is useful as a strict CE projection, but do not assume that CE refusal means the block should disappear from the default ptah-compat surface. Also measure round-trip safety of the projection before treating stderr warnings as sufficient.

denisvmedia:
Superseded by #1266, which is merged. Closing.

#1266 keeps every measurement from this branch — the three-block list, the `wibble` control, stripping blocks to their label, dialect keying, HCL-only scope, and the oracle runs in `atlas-oracle.yml` — and changes only the shape, per the two reviews here:

- suppression is now reference-aware: a standalone sequence is omitted, one a column default names is kept, so the emitted document is never left with a hole;
- `PTAH_ATLAS_INSPECT_ALL_BLOCKS=1` restores every block on the same surface, satisfying the capability rule written down in #1259;
- the diagnostics name the variable, so an operator learns from the message how to get the full output.

Verified independently before merge: standalone sequence omitted and the pinned binary reads that document at exit 0; referenced sequence kept with an explanatory diagnostic; the variable restores the block.

**One correction to my own review here.** I wrote that with the dangling reference "Ptah can no longer read its own output ... before this change that file round-tripped at exit 0." The first half was right and the second was not — I inferred it rather than measuring it. Measured now, `master` fails to read that same output too, and identically, for an unrelated reason: Ptah's inspect emits every role in the PostgreSQL **cluster**, and replaying it hits `role "..." already exists`. That is filed separately. The dangling reference was still real and worth fixing; my claim about what preceded it was not.

Issue #1272, opened 2026-08-07
PostgreSQL schema diff: compare the full semantic definition of existing indexes


denisvmedia:
## Summary

Follow up on #1242, #1246, and #1271.

Those changes make PostgreSQL introspection preserve substantially more of an index's semantic definition:

* access method;
* key expressions;
* operator classes;
* sort direction;
* explicit non-default `NULLS FIRST` / `NULLS LAST`;
* `INCLUDE` payload columns;
* domain column types relevant to faithful index/schema reconstruction.

The inspect/replay path is now substantially more faithful.

However, #1271 explicitly leaves one important question unresolved:

> Does Ptah's schema comparator actually use this information when comparing an **existing database index** with a desired index?

Current inspection suggests that it does not fully do so. In particular, `indexDefinitionsChanged` appears to compare only a subset of the modeled definition outside SQL Server.

That can create a dangerous class of false convergence:

```text
database:
    CREATE INDEX i ON t USING gin (...);

desired:
    CREATE INDEX i ON t USING gist (...);

Ptah:
    Schema is synced
```

even though the two indexes are semantically different.

The goal of this issue is to make PostgreSQL index comparison use the complete semantic definition Ptah now preserves.

## Why this is separate from #1271

#1271 fixes the **read/reconstruction path**:

```text
PostgreSQL catalog
        ↓
Ptah DB model
        ↓
schema representation / emitted SQL
```

This issue concerns the **comparison path**:

```text
current database index
        +
desired index
        ↓
semantic comparison
        ↓
no-op or replacement plan
```

A faithful reader is necessary but not sufficient.

The information added by #1271 should not become dead metadata that inspect can reproduce while schema diff ignores it.

## Required investigation

Before changing comparison behavior, reproduce each axis independently against a live PostgreSQL database.

For every fixture:

1. create the current index in PostgreSQL;
2. define a desired schema differing in exactly one semantic attribute;
3. run `ptah schema diff/apply --dry-run` or the appropriate native equivalent;
4. run `ptah-compat schema diff/apply --dry-run`;
5. measure the pinned Atlas CE binary where applicable;
6. record the actual PostgreSQL-valid transition required.

Do not infer comparator behavior from code inspection alone.

## Axis 1: access method

Example:

```sql
CREATE INDEX i ON t USING btree (value);
```

desired:

```sql
CREATE INDEX i ON t USING hash (value);
```

Also test at least one pair involving:

* `gin`;
* `gist`.

[...]

PR #1273, opened 2026-08-07
Describe only the roles the inspected scope uses (#1267)

### LINKED ISSUE #1267: PostgreSQL inspect emits every role in the cluster, not the ones the inspected database uses
Found while controlling a review of #1266. Pre-existing on master (`0b21edc0`), unrelated to that PR — both fail identically, which is how it surfaced.

## What happens

Inspecting a database with **one table** and no roles of its own:

```
ptah-compat schema inspect -u 'postgres://.../seqref?sslmode=disable&search_path=public'
```

emits four `role` blocks — `advsup_reader`, `app_reader`, `appuser`, `w1251c_reader`. None of them belong to `seqref`. They are every role in the PostgreSQL **cluster**, created by unrelated work in other databases on the same server.

The pinned Atlas community binary v1.3.0, same database: **zero** role blocks.

## Why it matters, in three ways

**The description is wrong.** The output claims these objects are part of a database that has nothing to do with them. A schema description that names objects the schema does not contain is not a description of that schema.

**The output is not replayable.** Roles are cluster-scoped and `CREATE ROLE` is not idempotent, so feeding Ptah's own inspect output back fails:

```
       SQL execution failed: ERROR: role "advsup_reader" already exists (SQLSTATE 42710)
```

That happens on a **freshly created** dev database, because creating a new database does not clear cluster-level roles. So a self-round-trip of an inspect is broken on any server that already has roles — which is every real server.

**It leaks.** Inspecting one database exposes every role name on the instance to whoever reads the output. On a shared server that is information about other tenants.

## Reproduction

```sh
createdb seqref
psql -d seqref -c 'CREATE SEQUENCE order_seq; CREATE TABLE orders (id integer DEFAULT nextval(''order_seq''::regclass));'
# any role existing anywhere in the cluster is enough
psql -d postgres -c 'CREATE ROLE someone_elses'

ptah-compat schema inspect -u 'postgres://.../seqref?sslmode=disable&search_path=public' | grep '^role'
atlas       schema inspect -u 'postgres://.../seqref?sslmode=disable&search_path=public' | grep -c '^role'   # 0
```

## Definition of done

- An inspect names only the roles the inspected scope actually uses — those holding privileges on its objects — or none, matching the pinned binary.
- A decision, recorded in code, on what "uses" means: a role granted on an object in scope is the obvious candidate; a role that merely exists is not.
- Ptah's own inspect output replays into a clean dev database at exit 0. That is the property worth pinning, since it is what a round-trip test actually needs.
- Whether the native surface should keep emitting cluster roles is a separate question and should be answered deliberately — see the capability rule in `AGENTS.md`. Native describing more than CE does is allowed; describing objects that are not in the inspected database is a different thing and is wrong on both surfaces.

## Note on how this was found

I first reported the failed self-read as a regression caused by #1256's dangling sequence reference. It was not: controlling against `master` produced the identical error. The dangling reference was real and is fixed by #1266; this is a second, independent cause of the same symptom that predates both.

## Related

- #1264 — the JSON surface drops schemas; same family of scope defects
- #1240 — dev-database isolation

denisvmedia:
Closes #1267.

PostgreSQL roles are cluster-wide. `readRoles` read `pg_roles` with no restriction beyond dropping the reserved `pg_` names and `postgres`, so inspecting a one-table database emitted every role anyone had created anywhere on the server.

## The rule

A description defines a role when some other statement in it names that role, and not otherwise. Concretely a role is described when it holds a privilege on a relation in a schema in scope (`aclexplode` of `pg_class.relacl`) or on one of the schemas (`aclexplode` of `pg_namespace.nspacl`), when it granted one, or when a row-level security policy on a table in scope applies to it (`pg_policy.polroles`, read in the shape `readRLSPolicies` already uses and gated on the same capability).

**Ownership is deliberately not a reason, and that was measured rather than assumed.** Ptah describes no ownership — no `OWNER TO`, no `CREATE SCHEMA ... AUTHORIZATION` — so an owner is a role the description creates and then never refers to. Because the connecting superuser owns every object it creates, an earlier revision of this PR that counted ownership made every inspect describe the connecting role, and CI caught it:

```
TestAtlasMigrateDiffConcurrentIndexAndQualifierE2E
Error: failed to apply migration: ERROR: role "ptah_user" already exists (SQLSTATE 42710)
SQL: CREATE ROLE "ptah_user" WITH LOGIN SUPERUSER CREATEDB CREATEROLE INHERIT REPLICATION
```

An owner a description *does* refer to still appears, through the privilege clauses: granting anything on a relation makes its ACL explicit, and an explicit ACL always carries the owner's own privileges.

## The grant side had to move too

Removing ownership alone left the document inconsistent, and the pinned Atlas community binary v1.3.0 found it — an emitted permission block said `to = role.p1267_owner` with no such role block:

```
Error: Unsupported attribute; This object does not have an attribute named "p1267_owner"
```

Master's document for the same fixture reads at exit 0, so this was a regression rather than a pre-existing hole. The source is `information_schema.role_table_grants`, which reports an owner's built-in privileges as grants even for a table whose `pg_class.relacl` is null. Measured directly — a table nobody has granted anything on still lists seven privileges for its owner:

```
ptah_ref_untouched_137 → p1267_owner: DELETE INSERT REFERENCES SELECT TRIGGER TRUNCATE UPDATE
```

No `GRANT` produced those, and `CREATE TABLE` re-establishes them for the new owner on replay. Table-grant introspection now reads only relations carrying an explicit ACL — the same source the schema and sequence grant queries already read, so all three agree.

## Measured

PostgreSQL 17.10, container `ptah1074-dev`, against master built from `f2e7fa0c` in its own worktree. Cluster holding nineteen non-system roles.

| scope | master | after |
| --- | --- | --- |
| `search_path=public` | 19 roles | **3** — `p1267_policy`, `p1267_select`, `p1267_usage` |
| `search_path=app1267` | 19 roles | **1** — `p1267_appusage` |

Absent after, present in both master columns: `p1267_owner` (owns a table nobody granted anything on), `p1267_stranger` (touches nothing — the control this issue is about), and the sixteen roles belonging to other databases on the same server. The two scopes differing is the point: a filter that merely dropped a fixed set of names would answer both the same way.

Replay, `psql -v ON_ERROR_STOP=1`, exit status read on its own line:

| | master | after |
| --- | --- | --- |
| one table, no roles of its own (this issue's shape) | 19 role blocks, **exit 3** on `role "advsup_reader" already exists` | 0 role blocks, **exit 0** |
| round trip *with* roles — inspect, drop the database and its own roles, replay into a fresh one | **exit 3** | **exit 0** |

The second row is the discriminating one: only an output naming exactly the roles it creates can reach 0, because the leaked roles belong to other databases and cannot be created twice.

## The pinned binary is a readability reference, not an oracle

Atlas community v1.3.0 emits no `role` block for any database, so it cannot say which roles are right. It reads the post-change documents (0 and 3 role blocks) at **exit 0**.

One document it still refuses, **unchanged from master and out of scope here**: a non-default schema scope emits `for = schema.app1267` with no `schema` block, and the binary answers `Unknown variable; There is no variable named "schema"`. Master's output for the same scope fails identically, so it is pre-existing and deserves its own issue.

## An integration test pinned the defect

`TestPostgreSQLDBReadRoleCommentReplayIntegration` created a role owning nothing and holding no privilege, read `--schemas ptah_db_read_empty_schema_137` — a schema it never created, so the read had no schema at all — and asserted the role came back. That is what this issue calls wrong, written down as an expectation. Pre-v1, so the expectation is corrected rather than the fix weakened.

The role now holds a privilege on the schema instead. That needs no `SET ROLE`, so it works whether or not the connection is a superuser, and all four properties the test existed for survive: the role statements, their replay into a cleaned cluster, the fail-closed collision on a second `CREATE ROLE`, and the comment restored via `COMMENT ON ROLE`. The role now appears with the `GRANT` that is its reason, so three statements are replayed rather than two. A second role, `ptah_db_read_stranger_137`, exists in the same cluster, is used by nothing in the schema being read, and must not be named.

## Tests

The white-box guard drives `readRoles` through a fake server that answers a role only when one single `UNION` branch of the query reads the catalog expression that is that role's reason, only when the query binds the schema the reason lives in, and only when the outer statement consumes the set those branches build. Comments are stripped first, so prose naming a catalog column cannot stand in for reading it. `TestReadRolesDoesNotTreatOwnershipAsUse` pins the decision by requiring the query *not* to read `nspowner`, `relowner`, `typowner` or `proowner`.

The new live guard is the one that matters: it asserts the invariant directly — every role a grant or a policy names is a role the description defines — over a fixture holding both a granted role and a role that only owns an ungranted table. Reverting the `relacl` condition turns it red and says so:

```
the description names role "ptah_ref_owner_137" but does not define it
```

Nine query mutants red with a green control in the same sweep: the whole restriction removed; the relation-grantee and relation-grantor branches stopped one at a time; the schema-grantee and schema-grantor branches likewise; the policy branch never added; the scope CTE widened to every schema on the server; the reserved-role exclusions dropped; and the restriction dropped while the schema names stay bound.

Three of the mutants in this PR escaped an earlier version of a guard rather than being predicted:

- the grantee and grantor branches read the same catalog, so a fake matching anywhere in the statement stayed green when either alone was broken — it now matches per `UNION` branch;
- dropping the restriction while leaving the schema names bound passed the fake and was caught only by live PostgreSQL, which answered with every role on the server — binding an argument is not a restriction, so the fake now requires the outer statement to consume the scoped set;
- removing ownership without moving the grant side passed every Go test and was caught only by the pinned Atlas binary refusing the document.

## Not covered

Whether MySQL, MariaDB, or any other dialect has the same server-versus-database leak was **not** measured. #1267 is scoped to PostgreSQL and `readRoles` is PostgreSQL-only, but the shape is worth checking on the other readers separately.

### DIFF
--- a/docs/site/src/content/docs/databases/postgresql.md
+++ b/docs/site/src/content/docs/databases/postgresql.md
@@ -65,6 +65,16 @@ Ordering is dependency-aware: roles are created before the functions and
 policies that reference them, and grants are emitted after the roles and
 target objects exist.
 
+Reading a live database describes only the roles the schemas being read
+actually use, because a PostgreSQL role belongs to the cluster rather than to
+one database. A role counts as used when it owns one of those schemas or an
+object in them, when it holds or granted a privilege on one of them, or when a
+row-level security policy on a table in them names it. A role that merely
+exists elsewhere on the server is not part of the schema being described, so
+it is left out — of `ptah db read` and of `ptah-compat schema inspect` alike.
+Every role the emitted grants and policies refer to is described, so a
+description never names a role it does not also create.
+
 :::caution
 Ptah never drops a role automatically. A role that disappears from the desired
 schema stays in the database, because roles may be shared with DBAs,
--- a/docs/site/src/content/docs/databases/postgresql.md
+++ b/docs/site/src/content/docs/databases/postgresql.md
@@ -67,13 +67,24 @@ target objects exist.
 
 Reading a live database describes only the roles the schemas being read
 actually use, because a PostgreSQL role belongs to the cluster rather than to
-one database. A role counts as used when it owns one of those schemas or an
-object in them, when it holds or granted a privilege on one of them, or when a
-row-level security policy on a table in them names it. A role that merely
-exists elsewhere on the server is not part of the schema being described, so
-it is left out — of `ptah db read` and of `ptah-compat schema inspect` alike.
-Every role the emitted grants and policies refer to is described, so a
-description never names a role it does not also create.
+one database. A role counts as used when it holds a privilege on a relation in
+those schemas or on one of the schemas themselves, when it granted one, or when
+a row-level security policy on a table in them applies to it. A role that
+merely exists elsewhere on the server is not part of the schema being
+described, so it is left out — of `ptah db read` and of `ptah-compat schema
+inspect` alike.
+
+The rule is exact in both directions: a description defines a role when some
+other statement in it names that role, and not otherwise. Ownership alone does
+not qualify, because Ptah describes no ownership — it writes no `OWNER TO` and
+no `CREATE SCHEMA ... AUTHORIZATION` — so an owner would be created and then
+never referred to.
+
+For the same reason, grant introspection reports privileges somebody granted
+rather than the built-in privileges an owner holds by default. A relation whose
+`pg_class.relacl` is null has had no `GRANT` run against it, and its owner's
+implicit privileges are no longer emitted as `GRANT` statements; replaying
+`CREATE TABLE` gives the new owner exactly those privileges again.
 
 :::caution
 Ptah never drops a role automatically. A role that disappears from the desired
--- a/docs/site/src/content/docs/direct/inspect.md
+++ b/docs/site/src/content/docs/direct/inspect.md
@@ -55,12 +55,20 @@ database schemas to read; when empty, Ptah reads the connection's default
 schema.
 
 PostgreSQL roles are cluster-wide rather than per-database, so a read reports
-only the roles the schemas being read actually use: a role that owns one of
-those schemas or an object in them, that holds or granted a privilege on one
-of them, or that is named by a row-level security policy on a table in them. A
+only the roles the schemas being read actually use: a role that holds a
+privilege on a relation in them or on one of the schemas, a role that granted
+one, or a role a row-level security policy on a table in them applies to. A
 role that merely exists elsewhere on the server belongs to no schema being read
-and is not described. Every role the emitted grants and policies name is still
-described, so the output never refers to a role it does not create.
+and is not described.
+
+Equivalently, a read describes a role exactly when some other statement in the
+same output names it, so the output never refers to a role it does not create.
+Ownership alone is not a reason: Ptah writes no `OWNER TO` and no
+`CREATE SCHEMA ... AUTHORIZATION`, so an owner would be a role the output
+creates and never mentions again. For the same reason a read no longer reports
+the built-in privileges an owner holds on a relation nobody has granted
+anything on -- no `GRANT` produced them, and `CREATE TABLE` re-establishes them
+for the new owner when the output is replayed.
 
 Role creation statements in PostgreSQL output are intended for
 a clean target. If a role already exists, applying the SQL
--- a/docs/site/src/content/docs/direct/inspect.md
+++ b/docs/site/src/content/docs/direct/inspect.md
@@ -52,7 +52,17 @@ sqlite3 restored.db <schema.sql
 
 On PostgreSQL-family databases, `--schemas` accepts a comma-separated list of
 database schemas to read; when empty, Ptah reads the connection's default
-schema. Cluster-role creation statements in PostgreSQL output are intended for
+schema.
+
+PostgreSQL roles are cluster-wide rather than per-database, so a read reports
+only the roles the schemas being read actually use: a role that owns one of
+those schemas or an object in them, that holds or granted a privilege on one
+of them, or that is named by a row-level security policy on a table in them. A
+role that merely exists elsewhere on the server belongs to no schema being read
+and is not described. Every role the emitted grants and policies name is still
+described, so the output never refers to a role it does not create.
+
+Role creation statements in PostgreSQL output are intended for
 a clean target. If a role already exists, applying the SQL
 fails before its description or grants are changed; this prevents privileges
 from being attached to a role with unverified security attributes. Role
--- a/integration/gonative/roles_grants_integration_test.go
+++ b/integration/gonative/roles_grants_integration_test.go
@@ -79,9 +79,9 @@ func TestPostgreSQLDBReadRoleCommentReplayIntegration(t *testing.T) {
 	cleanupDBReadRoleIntegration(c, db)
 	c.Cleanup(func() { cleanupDBReadRoleIntegration(c, db) })
 
-	// The described role owns the schema being read, which is why the read
-	// describes it. Ownership adds no statement of its own, so the role is
-	// still named by exactly the two statements this test replays.
+	// The described role holds a privilege on the schema being read, which is
+	// why the read describes it, and the GRANT that says so is emitted beside
+	// the role -- a description names no role it does not also create.
 	//
 	// The stranger role exists in the same cluster and is used by nothing in
 	// the schema being read. PostgreSQL roles are cluster-wide, so before
@@ -91,7 +91,8 @@ func TestPostgreSQLDBReadRoleCommentReplayIntegration(t *testing.T) {
 CREATE ROLE ptah_db_read_role_137;
 COMMENT ON ROLE ptah_db_read_role_137 IS 'Database read replay role';
 CREATE ROLE ptah_db_read_stranger_137;
-CREATE SCHEMA ptah_db_read_empty_schema_137 AUTHORIZATION ptah_db_read_role_137;`)
+CREATE SCHEMA ptah_db_read_empty_schema_137;
+GRANT USAGE ON SCHEMA ptah_db_read_empty_schema_137 TO ptah_db_read_role_137;`)
 	c.Assert(err, qt.IsNil)
 
 	cmd := readdb.NewReadDBCommand()
@@ -118,9 +119,13 @@ CREATE SCHEMA ptah_db_read_empty_schema_137 AUTHORIZATION ptah_db_read_role_137;
 			return !strings.Contains(statement, "ptah_db_read_role_137")
 		},
 	)
-	c.Assert(roleStatements, qt.HasLen, 2)
+	// CREATE ROLE, COMMENT ON ROLE, and the GRANT that is the role's reason
+	// for being described at all.
+	c.Assert(roleStatements, qt.HasLen, 3)
 
-	cleanupDBReadRoleIntegration(c, db)
+	// Drop the roles but keep the schema: the grant among the replayed
+	// statements has to have somewhere to land.
+	cleanupDBReadRolesOnly(c, db)
 	for _, statement := range roleStatements {
 		_, replayErr := db.Exec(statement)
 		c.Assert(replayErr, qt.IsNil, qt.Commentf("role restore failed: %s", statement))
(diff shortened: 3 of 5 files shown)


denisvmedia:
## Please hold — `integration-tests` fails, and it is this change

```
Error: error applying migrations: failed to apply migration 20260807131728:
       ERROR: role "ptah_user" already exists (SQLSTATE 42710)
Error: error applying migrations: failed to apply migration 20260807131738:
       ERROR: role "yugabyte" already exists (SQLSTATE 42710)
FAIL	go.5x5.cz/ptah/integration	44.802s
```

Not a flake and not pre-existing: the same job succeeded on the three most recent master runs, including current master `3acddbdf`.

## The mechanism

The scoping is right for **describing** a schema and wrong for **comparing** one. The comparator reads "not reported" as "does not exist":

- before, the reader returned every role in the cluster, so a role like `ptah_user` was present on the database side and the diff planned nothing;
- after, a role outside the inspected scope is not returned, the diff concludes it is missing, and plans `CREATE ROLE` — which the server refuses because the role is there.

So role management now plans to create roles that already exist. That is worse than the leak it fixes, because the leak was cosmetic and privacy-adjacent while this breaks apply.

## This is the third instance of one pattern today

Worth naming, because the fix should be shaped against the pattern rather than this one symptom:

- #1266 omitted an extension whose name is not a token in the document — and the document still depended on what the extension **provides**. Live regression on master.
- #1272 is the comparator reading only a subset of what the reader preserves, so semantically different indexes compare equal.
- This one: the comparator reading a **scoped-out** object as an **absent** object.

Each time, one side of the read/compare boundary changed and the other side kept its old assumption about what silence means.

## What to change

Keep the scoping — the leak in #1267 is real and the per-clause fixture work is good. Separate the two questions:

- **description**: which roles does this schema use? Scoped, as you built it.
- **comparison**: does this role exist on the server? Not scoped — a role is cluster-wide, and "out of scope" is not "absent".

Concretely, the comparator needs a source of truth for role existence that is independent of the inspect scope, or role additions must be suppressed when the role is out of scope rather than planned. Measure which shape the pinned binary implies, if it implies one — it emits no role blocks at all, so it may have no opinion here, in which case say so and pick on Ptah's own terms.

Add an integration-level guard so this cannot regress silently: a database where a cluster role exists outside the inspected scope, and an apply that must plan nothing for it.

## Not in dispute

The per-clause fixture table, the two-scope discrimination, the round-trip measurement, the twelve-mutant sweep, and the catch that the first guard survived two mutants because it matched tokens anywhere in the query rather than within one `UNION` branch. All of that should survive the reshape unchanged. Adding the grantor side of both ACL clauses so the description never names a role it does not define is a genuinely good refinement.

denisvmedia:
## Third red run, same root — this needs the comparator half, not another reader iteration

```
❌ dynamic_roles_advanced_postgres      role "admin_user" already exists
❌ dynamic_roles_cross_database_postgres role "admin_user" already exists
❌ dynamic_roles_modification_postgres   role "admin_user" already exists
❌ dynamic_roles_rollback_postgres       role "admin_user" already exists
```

Each iteration has fixed a real thing — the ownership clause that created a role nothing referenced, the `role_table_grants` owner built-ins, the fake server that let a mutant pass. The reader is now genuinely better: 19 roles down to 3, no dangling reference, and the guard I ran independently shows this branch reduces a self-apply plan from 20 statements to 6 while introducing none.

But the failure has been the same shape all three times, and it is not in the reader.

**The comparator reads "not described" as "does not exist."** A role the reader no longer reports is a role the diff plans to create, and the server refuses it because it is there. Narrowing the reader can never satisfy that, because the comparator's question — *does this role exist on the server* — is not the reader's question — *does this schema use this role*. Roles are cluster-wide; "out of scope" is not "absent".

That is requirement 2 of #1276, filed while this PR was in flight, and this is now the clearest instance of it:

> Make "not described" and "not present" distinguishable in whatever the comparator consumes. A reader that was not asked about a schema must not be indistinguishable from a reader that looked and found nothing.

## Recommendation

Fold this into #1276 rather than iterating here. The reader change is sound and should survive as-is; what it needs is a comparator that either asks the server about role existence independently of the inspect scope, or suppresses role additions for roles it was not asked about.

I have left `hold-off-merge` on. Nothing here should be thrown away — three separate mutants escaped a guard in this PR and were caught by measurement, and the invariant test that came out of it ("the description names a role it does not define") is worth keeping whatever shape the final fix takes.

## One thing worth recording separately

A non-default schema scope emits `for = schema.app1267` with no matching `schema` block. That is **identical on master**, so it is not this PR's, and it is the same missing-declaration shape as #1248 fixed for tables. It deserves its own issue.

denisvmedia:
This is now DIRTY against master, and the conflict is mine: #1292 rewrote the four `LIKE 'pg_%'` sites in `internal/dbschema/postgres/reader.go`, including the role predicate this branch also edits.

That is worth rebasing onto rather than around, because it removes one of the things the review of this branch flagged. The exclusion both role reads carry was `rolname NOT LIKE 'pg_%'`, and an unescaped `_` is a single-character wildcard, so it was dropping `pgbouncer`, `pgadmin`, `pgpool` and `pguser` — ordinary user roles that landed in neither `Roles` nor `RolesOutOfScope`, and were therefore read as absent and planned as `CREATE ROLE`. Exactly the statement this PR is named after, arriving through a different door.

Master now has `NOT LIKE 'pg\_%' ESCAPE '\'` at all four sites plus the `ptah_trigger_` filter, with a live fixture that reddens per half. So after the rebase, the partition this branch asserts is closer to true than it was — though not yet true, because `queryRoles` still excludes `postgres` from both reads, and the doc comment stating an unqualified partition ships publicly in `docs/public_api.snapshot`. That correction still stands.

denisvmedia (commented):
Independent review confirms that this PR should be closed as superseded by #1305, not rebased.

This branch scopes rendered roles but loses out-of-scope role existence, so replay can emit duplicate `CREATE ROLE` and fail with SQLSTATE 42710. It also predates the corrected `pg\_%` filtering and reintroduces `LIKE 'pg_%'`, where `_` is a wildcard that excludes ordinary names such as `pgadmin` and `pgbouncer`.

#1305 carries the separate description/existence model and the corrected filter. It has its own unresolved replay blocker, but it is still the only branch worth fixing. Close #1273 as stale/superseded.

denisvmedia:
Closing as superseded — the content of this branch is already on \`master\`.

\`f9e21a86 Scope the described roles, and take existence from both reads (#1267) (#1305)\` carries this PR's first commit verbatim and adds the half that was missing here: the comparator side (\`readRolesOutOfScope\` / \`DBSchema.RolesOutOfScope\` / \`internal/rolescope\`). Issue #1267 is closed.

**Why the integration job was red, and why narrowing the reader could not fix it.** Scoping \`readRoles\` is correct for *describing* a schema and wrong for *comparing* one: the comparator read "not reported" as "does not exist", so a cluster role outside the inspected scope was planned as \`CREATE ROLE\` and the server refused it — \`role "ptah_user" already exists (SQLSTATE 42710)\`, across four \`dynamic_roles_*_postgres\` scenarios. Out of scope is not absent.

**Containment verified two independent ways.** Rebasing onto \`origin/master\` conflicts on all three commits; dropping each leaves \`HEAD == origin/master\` with an empty \`git diff\`. Separately, of the 604 lines this branch adds, 588 appear verbatim in \`master\`; all 16 that do not are superseded rather than missing — a doc paragraph \`master\` replaced with a longer one covering \`PTAH_POSTGRES_INSPECT_ALL_ROLES\`, an \`IN (SELECT ...)\` that became \`EXISTS (...)\` plus its \`NOT EXISTS\` complement, and this branch's \`AND r.rolname NOT LIKE 'pg_%'\` — the unescaped-underscore defect that #1292 fixed on \`master\` as \`NOT LIKE 'pg\\_%' ESCAPE '\\'\`.

Merging this would reintroduce that defect and delete the comparator half.

**Confirmed on master against live PostgreSQL 17.10**, executing and asserting the catalog rather than reading rendered output: with \`search_path=public\`, five roles are described (SELECT grantee, USAGE grantee, RLS-policy referent, a \`pgbouncer\`-prefixed grantee, and the relation owner) and eleven are not; with \`search_path=app1273\` the answer changes to two, so no fixed-name filter can produce it. In both scopes every \`role.X\` a \`permission\` block references has a matching \`role "X"\` block. Replaying issue #1267's shape: default read plans 0 \`CREATE ROLE\` and exits 0, while \`PTAH_POSTGRES_INSPECT_ALL_ROLES=1\` reproduces the original failure verbatim — 17 \`CREATE ROLE\`, exit 3, \`ERROR: role "advsup_reader" already exists\`.

All eleven repository gates and the twelve \`check:*\` docs scripts exit 0 on \`master\`.

PR #1274, opened 2026-08-07
compat inspect: ask what an extension provides, not what it is called

denisvmedia:
## The regression

#1266 made `ptah-compat schema inspect` omit a refused block only when nothing else in the document names the object. For **extensions** that asks the wrong question. The scan matched the extension's own label, but a document that depends on an extension almost never spells its name: `isn` supplies the type `isbn`, `pgcrypto` supplies the function `gen_salt`. So the extension block was omitted while the column that needs it stayed behind.

#1266's own report listed this as a documented limitation. It is worse than documented: it breaks a round trip that worked before.

## Reproduced first, both directions

On freshly created PostgreSQL 17.10 databases, objects verified through the catalog rather than trusting the DDL echo:

```sql
CREATE EXTENSION isn;
CREATE TABLE books (id integer PRIMARY KEY, code isbn NOT NULL);
```

| | render | Ptah reads its own output back |
| --- | --- | --- |
| master before #1266 (`0b21edc0`) | rc=0 | **rc=0** |
| master now (`3acddbdf`) | rc=0, `extensions.isn: omitted` | **rc=1** `type "isbn" does not exist (SQLSTATE 42704)` |

And on that same emitted document:

| reader | result |
| --- | --- |
| pinned community binary v1.3.0 | rc=1 `pq: type "isbn" does not exist` |
| pinned binary on **its own** output for the same database | rc=1 `pq: type "isbn" does not exist` |

That last row is decisive. The omission was **not** a compatibility win — neither implementation can read the result. For this shape there is no faithful output the pinned binary can read, *including the one it writes itself*, so suppression can only choose which kind of broken. Copying that hole reproduces a defect (policy half two) and costs a working round trip.

## The fix

The catalog knows what an extension supplies, so ask it. The PostgreSQL reader reads extension membership from `pg_depend` against `pg_extension` — types, functions, relations, operator classes, operator families — and carries it on `DBExtension.Provides` / `goschema.Extension.Provides`. The renderer keeps the block when the surviving document uses the extension's name **or anything it supplies**.

Measured while designing it: no `pg_namespace` row is ever an extension member for an extension installed into an existing schema, so `public` does not enter the set and naming the schema does not pin every extension in the database.

The set is deliberately coarse and errs toward keeping. `pgcrypto` supplies a `gen_random_uuid` that PostgreSQL 13+ also supplies from core, so a document using it keeps `pgcrypto` although it would have resolved without it. Keeping a droppable block costs compatibility; dropping a needed one costs a document nobody can read.

**This is not "never omit an extension."** An extension nothing depends on is still omitted, and the pinned binary still reads that document at exit 0 — pinned by a control test in both the unit and e2e suites.

## After

| shape | extension block | Ptah round trip | pinned binary on the document |
| --- | --- | --- | --- |
| `isn` behind a column type | kept | **rc=0** | rc=1 (feature refusal, as before) |
| `pgcrypto` behind `gen_salt()` | kept | **rc=0** | rc=1 (feature refusal, as before) |
| `isn` nothing depends on | **omitted** | rc=0 | **rc=0** |

Parity rule (a) holds at the invocation level: inspecting the live database directly, both binaries exit 0.

## Does this affect sequences or policies?

No, and the reason is structural rather than lucky:

- **Sequence** — the only thing it supplies *is* its name; `nextval('order_seq')` names the sequence itself. Label-matching already asks the right question.
- **Policy** — nothing in an HCL document can reference a policy by name, so the reference test is vacuous for policies and they are always omitted. That is sound: omitting one leaves no dangling reference. It loses description (which RLS rules exist), which `PTAH_ATLAS_INSPECT_ALL_BLOCKS=1` recovers.

Verified on a live database: the sequence is kept behind its default, the policy is omitted, and the round trip still exits 0.

## Coverage

A live round-trip e2e test renders each shape and feeds Ptah's own output back to Ptah **against a fresh dev database** where the extension is not installed, requiring exit 0. `gen_salt` is used rather than `gen_random_uuid` for the function-only shape because the latter also exists in core and would pass without testing anything.

Three mutants, each observed red:

1. label-only matching → both dependent shapes red (unit + e2e); the `citext` row stays green, confirming it was a false comfort — extension and type are the same word there
2. never omit an extension → the omission controls and the env-var gate red
3. `pg_depend` never consulted → both dependent e2e shapes red

`scripts/check-test-style.sh` was also watched go red (an `if` in a test body) and then green.

## Capability

`PTAH_ATLAS_INSPECT_ALL_BLOCKS=1` still restores every block Ptah models; verified it emits both extension blocks where the default omits them. The kept-block diagnostic now reads "depends on it" rather than "names it", which is what it always meant and is now what it measures.

Behavior change; pre-v1, so no compatibility is owed.

## Gates

| gate | exit |
| --- | --- |
| `go build ./...` | 0 |
| `go vet ./...` | 0 |
| `scripts/check-public-api-snapshot.sh` | 0 (regenerated; +22 lines, 0 deletions — the two new fields only) |
| every `check:*` in `docs/site/package.json` | 0, except `check:responsive` |
| `build-feature-matrix.mjs --check` | 0 |

`check:responsive` could not run: it needs a built site and `docs/site/node_modules` is not installed in this worktree. The docs change is prose inside two existing pages — no new page, component, or CSS.

Not verified: MySQL/MariaDB/SQLite behavior (the refusal list is PostgreSQL-keyed and untouched for other dialects); non-catalog sources of `goschema.Extension` (Go annotations, YAML) leave `Provides` empty and degenerate to the previous label-only rule, which is the most that can be known about such a source.

### DIFF
--- a/cmd/atlas/schema_inspect.go
+++ b/cmd/atlas/schema_inspect.go
@@ -40,12 +40,14 @@ it, and the result is introspected, mirroring Atlas dev-database
 normalization.
 
 On PostgreSQL, HCL output omits an ` + "`extension`" + `, ` + "`sequence`" + ` or
-` + "`policy`" + ` block that nothing else in the document names: Atlas CE
+` + "`policy`" + ` block that nothing else in the document depends on: Atlas CE
 refuses a whole schema file that declares any one of them, so emitting one
 would make the output unreadable to the tool this binary stands in for. A block
-something DOES name — a sequence behind a column default, for instance — is
-kept, because a document that references an object it did not declare is
-readable by nobody. Every decision is reported on standard error. Set
+something still NEEDS — a sequence behind a column default, an extension
+supplying a column's type — is kept, because a document that references an
+object it did not declare is readable by nobody. For an extension the test is
+what the catalog says it supplies, not its name: ` + "`isn`" + ` supplies the
+type ` + "`isbn`" + `. Every decision is reported on standard error. Set
 ` + "`PTAH_ATLAS_INSPECT_ALL_BLOCKS=1`" + ` to emit every block Ptah models on
 this surface. SQL output keeps them all, and native
 ` + "`ptah schema inspect`" + ` omits nothing.
--- a/core/goschema/types.go
+++ b/core/goschema/types.go
@@ -386,6 +386,16 @@ type Extension struct {
 	IfNotExists bool   // Whether to use IF NOT EXISTS clause
 	Version     string // Specific version requirement (optional)
 	Comment     string // Extension comment/description
+
+	// Provides lists the catalog names this extension supplies -- types,
+	// functions, relations, operator classes and operator families. It is filled
+	// in when the schema was read from a live PostgreSQL catalog and answers
+	// "what stops resolving without this extension", which is almost never the
+	// extension's own name: `isn` supplies the type `isbn`.
+	//
+	// Empty means not measured. Annotation and YAML sources have no catalog to
+	// ask and leave it unset.
+	Provides []string
 }
 
 // Table represents a database table configuration parsed from Go struct annotations.
--- a/dbschema/types/types.go
+++ b/dbschema/types/types.go
@@ -310,6 +310,18 @@ type DBExtension struct {
 	Comment          *string `json:"comment"`           // Extension comment/description
 	DefaultVersion   *string `json:"default_version"`   // Default version available
 	InstalledVersion *string `json:"installed_version"` // Currently installed version (may differ from default)
+
+	// Provides lists the catalog names this extension supplies -- its types,
+	// functions, relations, operator classes and operator families -- read from
+	// pg_depend. It answers "what would stop resolving if this extension were
+	// not here", which is a different question from the extension's own name and
+	// usually a disjoint set of words: `isn` supplies the type `isbn`, and
+	// `pgcrypto` supplies the function `gen_salt`.
+	//
+	// Empty means not measured rather than "supplies nothing" -- every extension
+	// supplies something, and readers that do not consult a catalog (Go
+	// annotations, YAML) leave this unset.
+	Provides []string `json:"provides,omitempty"`
 }
 
 // DBSequence represents a standalone PostgreSQL sequence read from the database.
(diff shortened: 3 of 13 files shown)


denisvmedia:
Correcting the record on this one after merge.

The fix here is real and I re-measured it: `isn` keeps its block for a column typed `isbn`, `pgcrypto` for a default calling `gen_salt`, and the round trip that #1266 broke works. But it took the provides set from `pg_depend` unfiltered, and contrib extensions overwhelmingly supply **overloads of core functions**, so those lists are full of ordinary words. Measured on PostgreSQL 17.10, `citext` supplies fifteen names `pg_catalog` also supplies — `max`, `min`, `strpos`, `replace`, `split_part`, `translate` and the `regexp_*` family.

Three shapes that this branch turned from readable into refused, all on fresh databases where nothing depends on the extension, oracle = pinned community binary v1.3.0 reading Ptah's output against a separate clean dev database:

| fixture | ext 3acddbdf → 4c4a8a1d | orc 3acddbdf → 4c4a8a1d |
|---|---|---|
| `citext` + `CHECK (strpos(label,'x') = 0)` | 0 → 1 | 0 → 1 |
| `citext` + `VIEW v AS SELECT max(n) FROM t` | 0 → 1 | 0 → 1 |
| `pgcrypto` + `DEFAULT gen_random_uuid()` | 0 → 1 | 0 → 1 |

The reasoning that let this through is in `documentNamesAny`, and it is the part worth naming: it held that a false positive "costs compatibility" while a false negative "costs correctness", and therefore erred toward keeping. That asymmetry does not exist. The binary refuses **any** file declaring an extension block, so a spurious keep does not shrink the compatibility win — it removes it, landing on the same rc=1 as the hole. The comment named `pgcrypto`/`gen_random_uuid` as its own example of "the harmless direction"; that is the third row above.

The diagnostic was false too — `kept … because another object in this document depends on it`, on databases where nothing did.

Filed as #1280, fixed in #1282 by excluding, per object class, member names `pg_catalog` also supplies. A second mechanism is left open in #1281: `hstore` supplies three functions named `delete`, which `pg_catalog` does not, so a `plpgsql` body doing `DELETE FROM audit` still pins it.

One process note, since it is the reusable part. This PR's report listed `check-public-api.sh` and `check-public-api-snapshot.sh` as green and did not name `check-public-api-released.sh`, which was red here — `Provides []string` made `Extension` and `DBExtension` non-comparable. The approval landed separately. Three API scripts, three separate steps in `go-lint.yml`; the first two do not subsume the third.

Issue #1234, opened 2026-08-07
Ptah's inspect output is unreadable to the pinned binary for two reasons that are not column types


denisvmedia:
Split out of #1233, which fixed the **column type** half of #1138. With that change in place, `ptah-compat schema inspect` still emits HCL the pinned Atlas community binary v1.3.0 refuses, for two remaining reasons. Both are isolated below by peeling one at a time until the file reads back with `rc=0`.

## Reproduction

```
# a live PostgreSQL 17 source database
createdb src1138
psql -d src1138 -c 'CREATE TABLE t (
  id integer NOT NULL, name character varying(100), amount numeric(10,2),
  active boolean, created timestamp with time zone, payload jsonb,
  ratio double precision, tags text[], ident uuid);'

ptah-compat schema inspect -u "postgres://.../src1138?sslmode=disable&search_path=public" > out.hcl
atlas schema inspect -u "file://out.hcl" --dev-url "postgres://.../dev?sslmode=disable&search_path=public"
```

Read the exit status on its own line, not through a pipe.

## Defect 1 — the `permission` block

Ptah renders:

```hcl
permission {
  to = PUBLIC
  for = schema.public
  privileges = [USAGE]
}
```

The binary answers:

```
Error: out.hcl:48,8-14: Unknown variable; There is no variable named "PUBLIC".
```

`PUBLIC` and `USAGE` are written as bare identifiers, which HCL evaluates as variable references. There is nothing to resolve them to.

This block is a Ptah extension — the binary has no `permission` block of its own — so the question is not "match it" but "what should a Ptah-rendered grant look like in a file that stays readable". Worth deciding deliberately, since the answer also governs `grants` on every other dialect.

## Defect 2 — the missing schema declaration

With the `permission` block removed by hand:

```
Error: specutil: failed converting to *schema.Realm: cannot extract schema name for table "t": schemahcl: type "schema" was not found in nil
```

Ptah's inspect emits no `schema "public" { }` block, and no `schema = schema.public` on the table. Adding both by hand takes the same file to `rc=0`.

This is the same defect on SQLite, where the missing pieces are `schema "main" { }` and `schema = schema.main`.

## Measured isolation

| file | rc | first error |
| --- | --- | --- |
| as inspected, before #1233 | 1 | `There is no function named "NUMERIC"` |
| as inspected, with #1233 | 1 | `There is no variable named "PUBLIC"` |
| minus the `permission` block | 1 | `cannot extract schema name for table "t"` |
[...]

Issue #1257, opened 2026-08-07
The not-clean gate misses the binary's realm scope: a plain PostgreSQL URL still applies into a database with other schemas


denisvmedia:
Split out of #1231 case 1. #1252 closed that case for schema-scoped connections only; this is the other half, and it is a live parity-rule (a) violation on master.

## Measured on master (`17b05f52`)

Plain URL, **no `search_path`** — the form the compatibility documentation itself uses:

```
--url postgres://postgres:pw@localhost:5432/<db>?sslmode=disable
```

| database state | pinned community binary v1.3.0 | ptah-compat |
| --- | --- | --- |
| empty | 0 | 0 |
| an **empty** extra schema exists | **1** | **0, applies every migration** |
| a table living only in another schema | **1** | **0, applies every migration** |

```
Error: sql/migrate: connected database is not clean: found schema "extra". baseline version or allow-dirty is required
```

An empty schema is enough. The binary is not looking for tables in that mode — it is looking for schemas.

## Why the existing gate cannot see it

`internal/migrateclean/migrateclean.go` resolves scope through `current_schema()` unconditionally, so it always evaluates one schema and always counts tables. The binary has two modes:

- **schema scope** — a schema is pinned on the connection. Look for tables in it. Refusal: `found table %q in schema %q` / `found multiple tables: %d`. This is what #1252 implemented, correctly and thoroughly.
- **realm scope** — no schema pinned. Look for schemas. Refusal: `found schema %q`.

Both were verified independently by three reviewers on their own containers.

## Definition of done

- A second predicate keyed on the connection having no pinned schema: refuse when any schema other than the connection's default exists, or when the default schema holds a table other than this run's revision table.
- The `found schema %q` refusal reproduced byte-identically.
- MySQL's realm-scope shape measured too — reviewers report `found multiple schemas: %d` as a third message shape, which was **not** verified by me and should be measured rather than assumed.
- Controls in both directions: the schema-scoped cells #1252 already pins must not move, and an empty database must still exit 0 in realm scope.
- A test whose fixture uses the **plain** URL. #1252's entire PostgreSQL table used `?search_path=public`, which is exactly why this was invisible — a suite that only ever pins a schema cannot fail on this.

## How it was missed, worth keeping

[...]

PR #1275, opened 2026-08-07
Describe every schema `schema inspect` can see (#1264)

### LINKED ISSUE #1264: JSON inspect drops the schema on an empty database, and drops other schemas entirely in realm scope
`schema inspect --format '{{ json . }}'` is the scripted surface — it is what a consumer parses. Three divergences from the pinned Atlas community binary v1.3.0, measured on live PostgreSQL 17 and SQLite.

The register in #1235 recorded the first of these from a SQLite sweep. The second is new and is the serious one; it could not be seen from SQLite, which has no second schema.

## 1. An empty database loses its schema entirely

| target | pinned binary | ptah-compat |
| --- | --- | --- |
| empty SQLite | `{"schemas":[{"name":"main"}]}` | `{}` |
| empty PostgreSQL | `{"schemas":[{"name":"public","comment":"standard public schema"}]}` | `{}` |

Anything walking `.schemas` breaks on an empty database rather than seeing an empty schema. The realm's schema list is built purely from tables (`atlasSchemaInspectJSON` in `internal/atlasreport/schema_inspect.go`), so no tables means no schemas, and `json:"schemas,omitempty"` then removes the key.

This is the same defect class as the missing HCL `schema` block fixed in #1248, in the JSON surface.

## 2. Realm scope drops every schema but the connection's default

Database with `public.a` and `extra.b`, **plain URL, no `search_path`**:

```
pinned binary   {"schemas":[{"name":"extra","tables":[{"name":"b",...}]},{"name":"public",...}]}
ptah-compat     {"schemas":[{"name":"public","tables":[{"name":"a",...}]}]}
```

`extra` and its table are simply absent. The tool describes less of the database than it found and says nothing about it.

Control, same database with `?search_path=public`: both emit only `public`, and they agree. So the divergence is precisely the realm-scope mode, not schema filtering in general.

This is the same root distinction as #1257 — the binary has two scopes and Ptah collapses them into one — appearing here in the inspect surface rather than in the adoption gate. Fixing either does not fix the other, but whoever takes one should read the other.

## 3. The schema comment is not carried

`"comment":"standard public schema"` on PostgreSQL's `public`. Ptah omits it. Cosmetic beside items 1 and 2, but it is part of the same object and should come with whichever fix lands.

## Reproduction

```sh
createdb jempty
atlas       schema inspect -u 'postgres://.../jempty?sslmode=disable&search_path=public' --format '{{ json . }}'
ptah-compat schema inspect -u 'postgres://.../jempty?sslmode=disable&search_path=public' --format '{{ json . }}'

createdb jmulti
psql -d jmulti -c 'CREATE SCHEMA extra; CREATE TABLE public.a (id integer); CREATE TABLE extra.b (id integer);'
atlas       schema inspect -u 'postgres://.../jmulti?sslmode=disable' --format '{{ json . }}'
ptah-compat schema inspect -u 'postgres://.../jmulti?sslmode=disable' --format '{{ json . }}'
```

Exit statuses read on their own line, never through a pipe. Both binaries exit 0 on every cell above — these are output-shape divergences, not exit-code ones, which is why the sweep filed them as a register rather than as parity violations.

## Definition of done

- An empty database reports its schema, on every dialect.
- A realm-scoped connection reports every schema it can see, and a schema-scoped one keeps reporting exactly the one it is scoped to — the control that stops item 2's fix from breaking item 2's own control row.
- The schema comment is carried where the catalog has one.
- Tests use both URL forms. A suite that only ever pins `search_path` cannot fail on item 2 — the same blind spot that hid #1257.

## Related

- #1235 — the SQLite output-shape register, where item 1 was first recorded
- #1257 — the same realm-versus-schema scope distinction, in the adoption gate
- #1248 — the same defect class in the HCL surface

denisvmedia:
Closes #1264.

**The semantic hold is lifted.** Both halves of the fix are in this branch, and
the branch now sits on top of the PRESENT / AUTHORITATIVELY ABSENT / NOT
DESCRIBED model that #1298 merged, rather than beside it.

## What the hold was for

Describing more of the database moved one side of the read/compare boundary and
left the other where it was:

1. `schema inspect` began seeing both `public` and `extra`.
2. `schema apply`, when comparing, still read only `public`.
3. So `extra.b` looked non-existent, and Ptah planned to CREATE a schema and a
   table the database already had:

```
CREATE SCHEMA IF NOT EXISTS extra;
CREATE TABLE "extra"."b" ("id" integer);

Error: apply schema changes: failed to execute SQL statement:
       ERROR: relation "b" already exists (SQLSTATE 42P07)
```

It never converged: three consecutive applies planned the same thing.

## The two halves, and the property that shapes them

**The reader reports the schemas it read.** `readSchemas` gated on `r.scoped` and
answered nothing for any read not handed an allow-list, while every other read
below it iterated `schemasToRead()`. One list, two answers. Which schemas an
unscoped read *covers* does not move; what moves is that the read says so.

**`schema apply` reads the database at the scope the DESIRED STATE names** — not
at the URL's realm scope, and that choice is the safety property rather than an
implementation detail. A URL-derived realm scope would widen the read for every
document, including one describing a single schema of a multi-schema database,
and a schema the database has that the document does not describe is planned for
removal. The read is always at least what it was and at most what it was plus
the schemas the document itself names; `--schema` still outranks both.

`cmd/schema/apply.go` and `cmd/atlas/schema_apply.go` share the entry point, so
the native surface gets it too.

**The saved-plan fingerprint follows the read that will verify it.** A plan file
records an exclude list and no schema scope, so `schema apply --plan` re-reads at
the connection's default. Fingerprinting the wider planning read made a freshly
saved plan report itself stale against the database it had just been computed
from.

## How this sits on #1298

#1298 gave the comparator the third state this work needed. The two changes
compose rather than compete, and the boundary guard is where that is visible:

| row | #1234 gave | #1264 gives | together |
| --- | --- | --- | --- |
| `extension_nothing_references` | the document declares `public` | the reader says it read `public` | `wantDocumentOnly` is empty |
| `empty_database` | the document declares `public` | the reader says it read `public` | `wantDocumentOnly` is empty |

Each half alone leaves the pair disagreeing, which is the `[schema:public]`
difference #1276 said must become empty. With both, it is empty. Property 2 now
holds on every fixture and on both surfaces: #1298 turned the extension row's
compatibility plan from `DROP EXTENSION IF EXISTS "pgcrypto"` to nothing, and
this branch keeps it there.

The coverage header rides through unchanged — an inspected multi-schema document
carries `ptah:not-described extension/policy/sequence` **and** declares every
schema the URL reaches.

## What the rebase had to reconcile

Rebased onto `master`. Three conflicts, all resolved as unions of meaning rather
than by choosing a side, plus one defect no conflict reported.

**`internal/atlasschema/apply.go`.** #1329 widened `scopeGeneratedSide` to return
an exclude-match report; this branch moved the desired-state load *above* the
database read, because the read now needs it. Taking either side whole breaks:
master's side duplicates the load and redeclares `desired`, the branch's side
drops `desiredReport` and `UnmatchedAcrossStates` loses the desired half.
Resolved by keeping the hoisted load and master's three-return call.

**`integration/gonative/schema_boundary_guard_postgres_test.go`**, three hunks.
The header comment is the union of #1234's paragraph and this issue's, with one
sentence corrected: it claimed property 2 held "except the extension row", which
#1298 has since closed. The two data hunks keep master's explanation of
`wantDescribedSchemas` — #1234 reached that value first, by collecting what the
document references, so the branch's older prose claiming the row "read nil" is
no longer true of this base — and take this branch's `wantReadSchemas`.

The consequence neither parent states is `wantDocumentOnly` on those two rows.
Git auto-merged it to master's `[schema:public]` because this branch never edited
that line. With both halves present the observed value is `nil`, and putting
master's value back reddens the guard:

```
--- FAIL: .../empty_database/property_1
    diff (-got +want):
      - nil,
      + {"schema:public"},
```

**A clean merge that did not compile.** `NewSchemaInspectReport` took a sixth
parameter here while master grew two more call sites — #1306 in
`schema_inspect_test.go` and #1298 in `schema_inspect_split_coverage_test.go`.
Different files, different lines, no conflict, red tree:

```
vet: not enough arguments in call to atlasreport.NewSchemaInspectReport
     have (*goschema.Database, *types.DBSchema, types.DBInfo, nil, bool)
     want (*goschema.Database, *types.DBSchema, types.DBInfo, io.Writer, bool, bool)
```

Both get `false`, the connected-schema case each fixture represents. Neither
assertion can move on the value: `describeSchemas` is read only by `sqlSource`,
which only `MarshalSQL` calls, and those two cases render `{{ json . }}` and
`{{ hcl . | split }}`.

## Boundary matrix, measured on the rebased tree

Live PostgreSQL 17.10, MySQL 9.7 and SQLite, 2026-08-09. Exit status read on its
own line, never through a pipe. Throwaway databases, one dev database dropped and
recreated between runs.

| | property | result |
| --- | --- | --- |
| A | plain PostgreSQL URL describes ALL schemas | `extra`, `public`, `third`, with `public`'s comment |
| B | `?search_path=public` describes ONLY `public` | one schema |
| C | inspect a multi-schema database, apply back | `Schema is synced`, compat and native |
| D | a document about only `public` does NOT delete `extra` | `Schema is synced`; catalog still holds `extra.b`, `public.a`, `third.c` |
| E | a fresh target receives ALL schemas | catalog holds `extra.b`, `public.a`, `third.c` |
| F | a saved multi-schema plan is not stale at birth | no staleness refusal |
| G | #1298 coverage and #1329 unmatched-exclude survive | below |
| H | SQLite and MySQL empty databases | `{"schemas":[{"name":"main"}]}` · `{"schemas":[{"name":"…","charset":"utf8mb4","collate":"utf8mb4_0900_ai_ci"}]}` |

**D was executed, not rendered** — `--auto-approve` against a live database, then
the catalog asserted. Collapsing `applyReadScope` to the URL's realm scope turns
that exact run into

```
DROP TABLE IF EXISTS "extra"."b" CASCADE;
DROP TABLE IF EXISTS "third"."c" CASCADE;
```

so the synced answer is the guard working, not the guard being vacuous.

**G, #1298 half.** Inspecting a database holding an unreferenced `pgcrypto`, a
standalone `order_seq` and a policy `p`, then applying that document back to it:
`Schema is synced`. Same document with the three header lines deleted:
`DROP POLICY` / `DROP SEQUENCE` / `DROP EXTENSION`. The suppression is not
universal.

**G, #1329 half.** `schema inspect --exclude nosuchobject_zzz` warns and exits 0;
`schema apply` with the same selector exits 1; `PTAH_ATLAS_ALLOW_UNMATCHED_EXCLUDE=1`
restores the permissive behavior at exit 0.

## Corrected: a one-part `--exclude` now drops a schema

An earlier revision of this description recorded that "at realm scope the binary
reads a one-part pattern as a schema name; here it never addresses a schema", and
said it wanted its own issue. **That claim is withdrawn.** #1329 closed it, and on
a plain URL — the form this branch makes describe the whole realm —
`--exclude extra` now removes `schema "extra"` and its table, with no unmatched
warning:

```
$ ptah-compat schema inspect --url "$PLAIN_URL" --exclude extra --format '{{ json . }}'
{"schemas":[{"name":"public",…},{"name":"third",…}]}
```

The two changes meet here: before #1264 a plain URL never described `extra`, so
#1329's schema filter had nothing to reach.

## Gates, literal exit codes read directly

| gate | exit |
| --- | --- |
| `go build ./...` | 0 |
| `go vet ./...` | 0 |
| `go test ./... -count=1` | 0 (186 packages) |
| `golangci-lint run ./...` | 0 (0 issues) |
| docs `check:exit-codes` `check:core-doc-links` `check:page-health` `check:links` `check:redirects` `check:style` `check:limitations` `versions:selftest` | 0 each |
| `build-feature-matrix.mjs --check` | 0 (177 rows) |
| live `go test -tags integration ./integration/gonative/` | 0 |
| live `go test -tags integration ./internal/atlasschema/` | 0 |

`check:responsive` exits 1 with `dist not found` on an unbuilt tree; `npm run
build` exits 0 and it then exits 0. Both results reported. `git status --short`
is empty after the full `go test` sweep.

The four live fixtures this work turns on were confirmed to RUN rather than skip:
`TestInspectLive_ScopeSelectsSchemas`, `TestInspectLive_SQLSchemaStatements`,
`TestInspectLive_HCLRealmScope`, `TestPlanLive_SavedPlanIsNotStaleAgainstItsOwnSource`,
all PASS.

## Still diverging, deliberately, and not touched here

- **`two_schemas_plain_url`'s `wantDocumentOnly` is a DEFAULT-SCOPE difference,
  not a fidelity one.** The document reaches `extra` because inspection resolves
  realm scope from the URL; the reader called with no allow-list still covers the
  connected schema alone. Every object listed is one the database really has.
  Widening the reader's default is what produces the `DROP TABLE` row above, so
  the caller names the scope it wants — which inspection and apply now both do.
- **Column comments are dropped.** `dbschema.DBColumn` has no field for one, so
  this is a reader change rather than a rendering change.
- **`?search_path=public --schema extra`** reports `extra` here. Pre-existing;
  honoring the operator's flag is the direction that describes what was asked for.
- **`--format hcl` / `format { schema { inspect = "json" } }`** are normalized to
  the format here rather than treated as literal template text. Pre-existing.
- **`ptah introspect` on an unscoped connection writes one more annotation**,
  because the reader no longer denies the schema it read:
  `//ptah:schema:schema name="public" comment="standard public schema"`.
  Rendering it emits `CREATE SCHEMA IF NOT EXISTS "public"` and
  `COMMENT ON SCHEMA "public"`; the latter requires schema ownership. Both were
  already reachable through `--schemas public`; what changed is that the default
  reaches them. Whether the connected schema's comment belongs in rendered DDL at
  all deserves its own decision.

## Pre-existing defect found while verifying, NOT introduced here

`schema apply --plan --to` does not converge for a desired document that spans
more schemas than the connection's default. `RehearsePlanStatements` seeds the dev
database with `ReadSchemaWithSchemas(conn, nil)` — deliberately the default scope,
so it matches what `VerifyPlanTarget` will read — and the end-state comparison
reads the dev connection the same way, so objects in the other schemas are absent
from the rehearsal and reported as drift.

Measured on the same fixture with a binary built from `master` at `fc77c58c`:
master fails the identical check with the identical error, listing five objects
of drift where this branch lists two. So the branch narrows the failure and does
not cause it. It wants its own issue; the fix is for the rehearsal seed and the
plan's read scope to be the same decision.

### DIFF
--- a/cmd/atlas/atlas_test.go
+++ b/cmd/atlas/atlas_test.go
@@ -1325,7 +1325,15 @@ func TestCompatCommand_SchemaInspectUsesAtlasProjectFormatAndSchemaMode(t *testi
 	err := cmd.Execute()
 
 	c.Assert(err, qt.IsNil)
-	c.Assert(out.String(), qt.Equals, "{}")
+	// `mode { tables = false }` drops the tables, not the schema they lived in.
+	// The document still describes `main`, which is what the pinned community
+	// binary v1.3.0 does with the same project file: measured on SQLite, it
+	// renders `schema "main" {}` — in fact it renders the tables too, so it
+	// ignores this mode block entirely on inspect. Rendering `{}` for a
+	// database that has a schema is the defect stokaro/ptah#1264 is about, and
+	// this row is the one place it survived a filter rather than an empty
+	// database.
+	c.Assert(out.String(), qt.Equals, `{"schemas":[{"name":"main"}]}`)
 }
 
 // TestCompatCommand_SchemaInspectRejectsProOnlyOutputFlags pins the inspect
--- a/cmd/atlas/schema_inspect_refused_blocks_internal_test.go
+++ b/cmd/atlas/schema_inspect_refused_blocks_internal_test.go
@@ -81,6 +81,7 @@ func TestAtlasInspectRefusedBlockGate(t *testing.T) {
 				types.DBInfo{Dialect: "postgres", Schema: "public"},
 				&diagnostics,
 				atlasInspectOmitsRefusedBlocks(),
+				true,
 			)
 
 			hcl, err := report.MarshalHCL()
@@ -119,6 +120,7 @@ func TestAtlasInspectKeepsAReferencedBlockInEitherState(t *testing.T) {
 				types.DBInfo{Dialect: "postgres", Schema: "public"},
 				&diagnostics,
 				atlasInspectOmitsRefusedBlocks(),
+				true,
 			)
 
 			hcl, err := report.MarshalHCL()
--- a/core/goschema/types.go
+++ b/core/goschema/types.go
@@ -600,8 +600,36 @@ func QualifyTableName(schema, table string) string {
 //	MySQL:
 //	  CREATE TABLE users (status ENUM('active', 'inactive', 'suspended') DEFAULT 'active');
 type Enum struct {
-	Name   string   // The generated enum type name (e.g., "enum_user_status")
+	Name string // The generated enum type name (e.g., "enum_user_status")
+	// Schema owns the enum, empty for the connection's or document's default
+	// schema. It is a field rather than a qualifier folded into Name because
+	// Name is what a column's declared type is matched against, and a domain,
+	// composite and range each keep the two apart for the same reason.
+	//
+	// An enum is a TYPE, so its identity is (schema, name) exactly as a
+	// domain's is: public.mood and extra.mood are two types with different
+	// value sets. Without this field an enum read out of a non-default schema
+	// was described as belonging to the connected one, and applying that
+	// description built the type in the wrong schema and typed the column that
+	// uses it against the wrong type (stokaro/ptah#1276).
+	Schema string
 	Values []string // The allowed enum values (e.g., ["active", "inactive", "suspended"])
+}
+
+// QualifiedName returns schema.name when Schema is set, or Name otherwise.
+//
+// Name is returned VERBATIM when Schema is empty, rather than through
+// QualifyTableName, because Name is not always a bare identifier here. A SQL
+// schema file loaded through internal/convert/toschema parks the qualifier in
+// Name -- `public.e1` -- and an enum may legitimately be named with a literal
+// dot, which QualifyTableName canonicalizes by quoting. Running either through
+// it changes the identifier: `public.e1` becomes the single quoted name
+// "public.e1", and the already-quoted "tenant.data" gains a second layer.
+func (e Enum) QualifiedName() string {
+	if strings.TrimSpace(e.Schema) == "" {
+		return e.Name
+	}
+	return QualifyTableName(e.Schema, e.Name)
 }
 
 // Domain represents a PostgreSQL domain type parsed from Go annotations.
--- a/core/renderer/internal/dialects/sqlite/sqlite.go
+++ b/core/renderer/internal/dialects/sqlite/sqlite.go
@@ -38,7 +38,21 @@ func (r *Renderer) Render(node ast.Node) (string, error) {
 	return r.Output(), nil
 }
 
+// defaultSchema is the namespace every unqualified SQLite object already
+// belongs to. It is not a schema anybody creates.
+const defaultSchema = "main"
+
 func (r *Renderer) VisitCreateSchema(node *ast.CreateSchemaNode) error {
+	if node.Name == defaultSchema {
+		// `main` is where the connection already is, so there is nothing to
+		// create and nothing to refuse. An introspected SQLite database
+		// describes it as a schema — the pinned Atlas community binary v1.3.0
+		// renders `schema "main" {}` in HCL and no statement at all in SQL —
+		// and turning that into a "not supported" comment would put a refusal
+		// in the output for something the author never asked for
+		// (stokaro/ptah#1264).
+		return nil
+	}
 	r.notSupported("schemas", node.Name)
 	return nil
 }
--- a/core/renderer/renderer.go
+++ b/core/renderer/renderer.go
@@ -759,6 +759,13 @@ func GetOrderedCreateStatementsWithCapabilities(
 		if err != nil {
 			return nil, err
 		}
+		// A node a dialect renders as nothing is not a statement. Keeping it
+		// put a bare `;` in front of the script — SQLite renders no statement
+		// for its `main` namespace, which an introspected database now
+		// describes as a schema (stokaro/ptah#1264).
+		if strings.TrimSpace(sql) == "" {
+			continue
+		}
 		statements = append(statements, sql)
 	}
 
--- a/dbschema/types/types.go
+++ b/dbschema/types/types.go
@@ -189,6 +189,9 @@ type DBEnum struct {
 	Values []string `json:"values"`
 }
 
+// QualifiedName returns schema.name when Schema is set, or Name otherwise.
+func (e DBEnum) QualifiedName() string { return QualifyTableName(e.Schema, e.Name) }
+
 // DBDomain represents a PostgreSQL domain type read from the database.
 type DBDomain struct {
 	Name     string `json:"name"`
@@ -594,6 +597,9 @@ type DBFunction struct {
 	Body       string `json:"body"`       // Function body/implementation
 	Comment    string `json:"comment"`    // Function comment/description
 }
+
+// QualifiedName returns schema.name when Schema is set, or Name otherwise.
+func (f DBFunction) QualifiedName() string { return QualifyTableName(f.Schema, f.Name) }
 
 // DBView represents a database view read from the database.
 type DBView struct {
--- a/docs/public_api.snapshot
+++ b/docs/public_api.snapshot
@@ -3930,7 +3930,19 @@ type EmbeddedSources struct {
 package goschema // import "go.5x5.cz/ptah/core/goschema"
 
 type Enum struct {
-    Name   string   // The generated enum type name (e.g., "enum_user_status")
+    Name string // The generated enum type name (e.g., "enum_user_status")
+    // Schema owns the enum, empty for the connection's or document's default
+    // schema. It is a field rather than a qualifier folded into Name because
+    // Name is what a column's declared type is matched against, and a domain,
+    // composite and range each keep the two apart for the same reason.
+    //
+    // An enum is a TYPE, so its identity is (schema, name) exactly as a
+    // domain's is: public.mood and extra.mood are two types with different
+    // value sets. Without this field an enum read out of a non-default schema
+    // was described as belonging to the connected one, and applying that
+    // description built the type in the wrong schema and typed the column that
+    // uses it against the wrong type (stokaro/ptah#1276).
+    Schema string
     Values []string // The allowed enum values (e.g., ["active", "inactive", "suspended"])
 }
     Enum represents a global enumeration type definition that can be shared
@@ -3977,6 +3989,7 @@ type Enum struct {
         MySQL:
           CREATE TABLE users (status ENUM('active', 'inactive', 'suspended') DEFAULT 'active');
 
+func (e Enum) QualifiedName() string
 
 ### go.5x5.cz/ptah/core/goschema.Extension
 
@@ -5656,6 +5669,7 @@ type DBEnum struct {
 }
     DBEnum represents a database enum type (PostgreSQL)
 
+func (e DBEnum) QualifiedName() string
 
 ### go.5x5.cz/ptah/dbschema/types.DBExtension
 
@@ -5724,6 +5738,7 @@ type DBFunction struct {
 }
     DBFunction represents a PostgreSQL custom function read from the database
 
+func (f DBFunction) QualifiedName() string
 
 ### go.5x5.cz/ptah/dbschema/types.DBGrant
 
--- a/docs/site/src/content/docs/atlas/schema-commands.md
+++ b/docs/site/src/content/docs/atlas/schema-commands.md
@@ -86,6 +86,15 @@ ptah-compat schema inspect --url "postgres://…/app?sslmode=disable"
 ptah-compat schema inspect --url "postgres://…/app?sslmode=disable&search_path=public"
 ```
 
+Every object in the document names the schema that owns it, not the one the
+connection happens to be on. That applies to the non-table kinds as much as to
+tables: an `enum` block carries `schema = schema.<name>`, a `function`, `view`,
+`materialized`, `domain`, `composite`, `range` and `sequence` block each carry
+the attribute wherever the object is outside the document's default schema, and
+a column declared against a type in another schema is written against that
+schema's type. Applying such a document therefore rebuilds each object where it
+was, and applying it back to the database it describes plans nothing.
+
 `--schema` / `-s` narrows inspection when the underlying database reader supports
 schema scoping, and outranks the URL's scope: naming a schema that does not
 exist renders an empty document rather than falling back to the connection's
--- a/docs/site/src/content/docs/atlas/schema-commands.md
+++ b/docs/site/src/content/docs/atlas/schema-commands.md
@@ -70,8 +70,26 @@ ptah-compat schema inspect \
   --dev-url "$DEV_DATABASE_URL" > schema.hcl
 ```
 
+### What a URL puts under inspection
+
+A PostgreSQL-family URL that pins no `search_path` describes the whole
+database — every non-system schema, with its tables — and one that pins a
+schema describes exactly that schema. An empty database still describes the
+schema it has rather than rendering an empty document, so a consumer walking
+`.schemas` does not break on a database that is merely empty.
+
+```bash
+# Every schema of the database, `public` and `extra` alike.
+ptah-compat schema inspect --url "postgres://…/app?sslmode=disable"
+
+# Only `public`.
+ptah-compat schema inspect --url "postgres://…/app?sslmode=disable&search_path=public"
+```
+
 `--schema` / `-s` narrows inspection when the underlying database reader supports
-schema scoping. `--format`
+schema scoping, and outranks the URL's scope: naming a schema that does not
+exist renders an empty document rather than falling back to the connection's
+own schema. `--format`
 accepts Atlas-style Go templates with `.MarshalHCL`, `hcl`, `sql`, `json`,
 `base64url`, `mermaid`, `split`, and `write`. Split-write exports are
 supported for HCL and SQL output with the documented Atlas split strategies:
--- a/docs/site/src/content/docs/atlas/schema-commands.md
+++ b/docs/site/src/content/docs/atlas/schema-commands.md
@@ -836,6 +836,16 @@ schema files, prints the planned SQL, and applies it after interactive
 confirmation. Use `--dry-run` to print the plan without applying it, or
 `--auto-approve` to skip the prompt explicitly.
 
+### Which schemas the current side is read at
+
+The database is read at the scope the desired state names, so both sides of the
+diff cover the same schemas. A desired state that names schemas beyond the one
+the connection is on — as an inspected document of a multi-schema database does
+— is compared against those schemas too, which is what makes inspecting a
+database and applying its own output back a no-op. A desired state that names
+only the connected schema reads only that one, so a schema the document never
+mentions is never planned for removal. `--schema` outranks both.
+
 Use `--tx-mode=file` or `--tx-mode=all` to execute the generated plan in one
 transaction, or `--tx-mode=none` to execute statements without transaction
 wrapping. With `--edit`, the planned SQL opens in `$VISUAL` or `$EDITOR`
--- a/docs/site/src/content/docs/reference/hcl-schema.md
+++ b/docs/site/src/content/docs/reference/hcl-schema.md
@@ -67,6 +67,7 @@ table "users" {
 | `unique` | `columns`; PostgreSQL also supports `include` and `nulls_distinct`. |
 | `foreign_key` | One local `columns` entry and one table-qualified `ref_columns` entry. |
 | `check` | `expr`. |
+| `enum` | `values`, plus the `schema` that owns the type. A PostgreSQL enum is created in that schema and a column declared against it is qualified with it. |
 | `extension` | PostgreSQL `if_not_exists`, `version`, and comments. |
 | `role` | PostgreSQL role attributes, including `password`. |
 | `permission` | PostgreSQL table, schema, and sequence permissions. |
--- a/integration/gonative/roles_grants_integration_test.go
+++ b/integration/gonative/roles_grants_integration_test.go
@@ -665,7 +665,7 @@ CREATE ROLE pgbouncer_undescribed_137 LOGIN;`)
 	c.Assert(err, qt.IsNil)
 	c.Cleanup(func() { c.Check(conn.Close(), qt.IsNil) })
 	var inspectDiag bytes.Buffer
-	inspected, err := atlasschema.Inspect(conn, atlasschema.InspectOptions{
+	inspected, err := atlasschema.Inspect(t.Context(), conn, atlasschema.InspectOptions{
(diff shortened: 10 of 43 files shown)


denisvmedia:
## Please hold — the description is right and it breaks the comparison

The three JSON cells are correct and I verified them independently against the pinned community binary v1.3.0:

| cell | pinned binary | master | this branch |
| --- | --- | --- | --- |
| empty database | `{"schemas":[{"name":"public","comment":"standard public schema"}]}` | `{}` | **identical to the binary** |
| plain URL, two schemas | lists `extra` and `public` | lists only `public` | **identical to the binary** |
| `?search_path=public` | only `public` | only `public` | only `public` |

The scoped row is the control and it holds. The schema comment lands too.

## What breaks

Inspecting a two-schema database on a plain URL now describes `extra.b`. Applying that description back to the database it came from:

```
Planned schema changes:
CREATE SCHEMA IF NOT EXISTS extra;
-- POSTGRES TABLE: extra.b --
CREATE TABLE "extra"."b" ("id" integer);

Error: apply schema changes: failed to execute SQL statement:
       ERROR: relation "b" already exists (SQLSTATE 42P07)
```

rc=1, and it never converges — three consecutive applies plan the same thing. On master the same fixture is `Schema is synced, no changes to be made`.

The reader now sees `extra.b`; the comparator's **database side** does not, because it was never taught realm scope. So the diff reads a schema it can now describe as one that does not exist, and plans to create it.

## This is the fourth instance of one pattern today

- #1266 omitted an extension whose name is not a token in the document, while the document still depended on what it **provides**. Live regression, fixed in #1274.
- #1272: the comparator reads only a subset of what the reader preserves, so semantically different indexes compare equal.
- #1273: scoping the role reader made the comparator read **out of scope** as **absent**, so it planned `CREATE ROLE` for roles that exist. Held.
- This one: widening the schema reader makes the comparator read **present** as **absent** on the other side.

Every time, one side of the read/compare boundary moved and the other kept its assumption about what silence means. I have filed that as its own issue rather than leaving it implicit in four PRs.

## What to change

The JSON work should land — it is correct. It needs the comparator taught the same scope in the same change, or the widened description must not reach the comparison path until it is.

Concretely: `schema apply` and `schema diff` must resolve realm-versus-schema scope the way `internal/migrateclean` and `schemaselection` already do for the adoption gate — #1257 solved exactly this distinction for `migrate apply` and the answer should be reused rather than re-derived.

Add a guard at the level where it broke: inspect a multi-schema database on a plain URL, apply the result back, and require a no-op. A unit test on the JSON shape cannot fail on this, which is why it got this far.

## Note on the oracle

The pinned binary is not a reference for this fixture: its own `schema apply` refuses first with `sql/migrate: connected database is not clean: found schema "extra"`. So the convergence question here has to be answered on Ptah's own terms.

denisvmedia:
Pushed a21933b6, which fixes the `{{ sql . }}` regression this was held for.

Describing the connected schema is what #1264 required — a document claiming an empty database had no schema at all broke anything walking `.schemas`. But the row that fixes JSON and HCL also reached the SQL renderer, which then prefixed `CREATE SCHEMA IF NOT EXISTS "public";` and its `COMMENT ON SCHEMA` to output the pinned community binary v1.3.0 emits without either.

That binary does not treat the two formats the same. `CREATE SCHEMA` counts in `--format '{{ sql . }}'`, PostgreSQL 17.10, database holding `public` and `extra`, base = merge-base 3acddbdf:

| cell | oracle | base | df7506d0 | a21933b6 |
|---|---|---|---|---|
| plain URL | 2 | 0 | 2 | **2** |
| `?search_path=public` | 0 | 0 | **1** | **0** |
| `--schema public` | 1 | 1 | 1 | **1** |
| `--schema public --schema extra` | 2 | 2 | 2 | **2** |
| MySQL 9.7, connected database | 0 | 0 | **1** | **0** |

`COMMENT ON SCHEMA` at the scoped PostgreSQL cell moves 0 → 1 → 0 alongside it.

The rule is **not** realm scope alone, which is worth stating because that was my first guess and the `--schema` rows refute it: naming a schema renders it. What that binary leaves out is the schema it was merely *connected* to. That is the branch `inspectSchemaNames` already takes — explicit `--schema` wins, otherwise `schemaselection.Realm` decides — so asking it a second time gives the SQL format its answer with no new plumbing.

Dropping the rows at render time rather than never reading them keeps JSON and HCL seeing everything the reader described. Verified: JSON schema names match the oracle at plain, `search_path` and MySQL scope, and the empty database still renders `{"schemas":[{"name":"public","comment":"standard public schema"}]}` on both.

Four rows added to `TestInspectLive_SQLSchemaStatements`. Forcing the gate open reddens the `search_path` row and only that one. Every row also asserts `CREATE TABLE` is present, so a rendering that lost its whole body cannot pass by having zero schema statements.

Gates on the pushed tree, each read on its own line: `go build` 0 · `go vet` 0 · `go test ./... -count=1` 0 (178 packages) · `go tool qtlint` 0 · `golangci-lint run ./...` 0 ("0 issues") · `check-test-style.sh` 0 · `check-public-api.sh` 0 · `check-public-api-snapshot.sh` 0 · `check-public-api-released.sh` 0. Integration with live PostgreSQL 17.10 and MySQL 9.7: `internal/atlasschema` 0, `internal/atlasreport` 0, `cmd/atlas` 0, `internal/migrateclean` 0.

`NewSchemaInspectReport` took a sixth parameter. It is in `internal/`, so no public API gate moves; four test call sites were updated.

denisvmedia:
The `{{ sql . }}` regression is fixed by a21933b6, but merging master brought in #1279's boundary guard and it goes red. I reproduced it locally and the guard is right. Keeping the hold.

## Reproduction environment, because it mattered

The guard passes on master **only** against a CI-shaped database. Run as the bootstrap `postgres` superuser it fails on master too, on `property 2`, in four fixtures. Faithful local repro:

```
docker run -d --name <c> -e POSTGRES_PASSWORD=pw -p 15997:5432 postgres:17
CREATE ROLE ptah_user LOGIN PASSWORD 'ptah_password' SUPERUSER;
CREATE DATABASE ptah_test OWNER ptah_user;
POSTGRES_TEST_DSN='postgres://ptah_user:ptah_password@localhost:15997/ptah_test?sslmode=disable'
```

Master: `ok`. This branch merged with master: three fixtures red. That the guard's `property 2` is sensitive to the connecting role is worth a separate look — it is a fixture that can be green for the wrong reason.

## What the guard says, per assertion, on `two_schemas_plain_url`

| assertion | guard's pinned master value | its stated target | this branch |
|---|---|---|---|
| `wantDescribedSchemas` | `["public"]` | `["extra","public"]` | **`["extra","public"]`** |
| `wantReadSchemas` | `nil` | `["extra","public"]` | `nil` — unchanged |
| `wantDocumentOnly` | `["schema:public"]` | empty | `["column:extra.b.id","schema:extra","schema:public","table:extra.b"]` |
| `wantNativePlan` / `wantCompatPlan` | owner-grant churn | `nil` | churn **plus** `CREATE SCHEMA IF NOT EXISTS extra`, `CREATE TABLE "extra"."b"`, and five `GRANT ... ON TABLE "extra"."b"` |

So this branch hits the guard's first stated target exactly and moves the other three the wrong way. It is a half-fix, and the guard's comment already predicted which half: *"Must become empty **once both sides read the same list**."*

The document side now reads both schemas. The reader side (`wantReadSchemas`) still reports none, so the two halves are further apart than before, not closer — and `property 2` shows the cost concretely: applying a database's own description back to it now plans to CREATE a schema and a table that already exist, where before it only churned grants.

`extension_nothing_references` and `empty_database` fail the same way: `wantDescribedSchemas` moves from `nil` to `["public"]`, which is the #1264 target met, while the read side stays `nil`.

## What this means for the PR

Nothing here says the change is wrong. `schema inspect` describing every schema it can see is correct, it matches the pinned community binary v1.3.0 at every cell I measured, and the `{{ sql . }}` half is now correct too. What the guard establishes is that the change cannot land alone: it widens the reader/comparator asymmetry that #1276 is about, and it does so in the destructive direction on the apply path.

Two things have to move with it:

1. `observed.live` — the reader's own schema list — has to report what the document reports. Today the renderer synthesizes schema blocks from the tables it was handed while the reader answers no schemas at all, which is the second `WRONG (#1276)` row in that fixture and is the direct cause of `wantDocumentOnly` growing.
2. The apply path has to read the database at the scope the document describes. Otherwise every schema the document gains is a schema apply thinks is missing.

I would rather land those together than update the guard's expectations to match a widened asymmetry — the guard exists precisely so that cannot happen quietly. Folding this into #1276 alongside #1273, which is held for the mirror-image reason on roles.

denisvmedia (commented):
Independent review found two blockers on this stale head.

**Blocker — only the inspected/document side is widened.** `internal/atlasschema/inspect_scope.go:42` reads every selected schema, but apply/current-state comparison remains narrow. Inspecting a two-schema database and applying that output back to the same database plans duplicate `CREATE SCHEMA extra` / `CREATE TABLE extra.b`; the integration guard fails with PostgreSQL SQLSTATE 42P07. Scope/coverage must travel with the observed state so “not described” differs from “absent.” Integrate this with a corrected #1276 coverage model and require native and compat inspect -> apply no-op tests.

**Major — empty SQL output is still `";\n"`.** `internal/atlasreport/schema_inspect.go:293` appends a semicolon to `strings.Join(nil,...)`. The new test creates a table, so it never exercises the claimed empty case. Return `""` for zero statements and add a genuinely empty-database test.

This PR should remain blocked; green focused tests do not satisfy its end-to-end contract.

denisvmedia:
Rebased onto master (9519bc18) and repaired. The hold was right and the guard was right; both things it asked for have moved, so `integration-tests` is green rather than the guard's expectations being relaxed around a widened asymmetry.

## What was actually red

`TestPostgreSQLSchemaBoundaryGuardIntegration`: four assertions on `two_schemas_plain_url`, two each on `extension_nothing_references` and `empty_database`. The destructive one was property 2 — applying a database's own description back to it planned

```
CREATE SCHEMA IF NOT EXISTS extra
CREATE TABLE "extra"."b" ("id" integer)
```

on both surfaces, for a schema and a table the database already had. Fixed in code, not in an expectation.

## The two things that had to move with it

**1. The reader reports the schemas it read.** `readSchemas` gated on `r.scoped` and answered nothing for any read not handed an allow-list, while every other read below it iterated `schemasToRead()`. One list, two answers. Which schemas an unscoped read *covers* does not move — still the connected schema. What moves is that the read says so. That takes `wantReadSchemas` from `nil` to `["public"]` on all six fixtures and `wantDocumentOnly` to `nil` on four.

**2. `schema apply` reads the database at the scope the desired state names.** Not at the URL's realm scope, and the difference is the safety property. Measured on PostgreSQL 17.10, same database, `--to` a document declaring only `public`:

| database side read at | plan |
|---|---|
| the URL's realm scope | `DROP TABLE IF EXISTS "extra"."b" CASCADE` |
| the document's scope | `Schema is synced, no changes to be made` |

The read is always at least what it was and at most what it was plus the schemas the document itself names. Nothing outside the document can be reached, `--schema` still outranks both, and a desired state naming nothing beyond the connected schema reads exactly what it read before. `cmd/schema/apply.go` and `cmd/atlas/schema_apply.go` share the entry point, so the native surface gets it too.

## One more boundary the same widening crossed

`schema plan` fingerprinted its own planning read while `schema apply --plan` re-reads at the connection's default, because a plan file records an exclude list and no schema scope. A plan computed from a two-schema document therefore reported itself stale against the database it had just been computed from:

```
Error: pre-planned migration is stale: the target database schema does not match
the plan's source fingerprint (plan sha256:6bcb8672…, database sha256:3f67ccda…)
```

with nothing changed in between. The from-fingerprint now comes from the read the verifier will make. Pinned by `TestPlanLive_SavedPlanIsNotStaleAgainstItsOwnSource`, with the single-schema document beside it as the control.

## Guard rows

Twelve values move, measured on live PostgreSQL 17.10 and 18.4, which agreed. Ten reach the target the guard's own comments stated; property 2 stays `nil` everywhere except the extension row, which is #1276's own and untouched.

One row records a difference rather than a target, and it is a difference in DEFAULT SCOPE rather than in fidelity:

```
two_schemas_plain_url  wantDocumentOnly  [schema:public]
                                      -> [column:extra.b.id schema:extra table:extra.b]
```

Every object listed is one the database really has. The document reaches `extra` because inspection resolves realm scope from the URL; the reader called with no allow-list does not ask for it. Closing that by widening the reader's DEFAULT is what produces the `DROP TABLE` row above, so the caller has to name the scope it wants — which inspection and apply now both do. The file header says so where the next reader will look.

## Verified by executing, not by rendering

PostgreSQL 18.4, a database holding `public.a`, `extra.b` and `third.c`:

| step | result |
|---|---|
| `schema inspect --url <plain>` | three schemas and three tables, byte-identical to the pinned Atlas community binary v1.3.0 |
| `schema inspect --url <?search_path=public>` | only `public`, byte-identical |
| `schema apply` of that document to a fresh database | catalog holds `extra.b`, `public.a`, `third.c` |
| `schema apply` of that document back to its source | `Schema is synced`, twice in a row |
| `schema apply` of a `public`-only document to it | `Schema is synced` |
| native `ptah schema apply` of the same document | `Schema is synced` |
| MySQL 9.7 empty database | `{"schemas":[{"name":"…","charset":"utf8mb4","collate":"utf8mb4_0900_ai_ci"}]}`, byte-identical |

## Inverse mutants, each killed by a named test

| mutant | killed by |
|---|---|
| `applyReadScope` collapsed to `return nil` | `two_schemas_plain_url` property 2 on both surfaces, and only those two |
| `readSchemas` gated on `r.scoped` again | `TestPostgreSQLReaderReadSchemasReportsWhatItRead`'s two unscoped rows, and twelve assertions across all six guard fixtures |
| inspection's realm decision collapsed to the connected schema | `TestInspectLive_ScopeSelectsSchemas`, `TestInspectLive_SQLSchemaStatements`, `TestInspectLive_HCLRealmScope`, plain-URL row each, plus two guard assertions |
| the plan's from-fingerprint taken from the planning read | `TestPlanLive_SavedPlanIsNotStaleAgainstItsOwnSource`'s two-schema row; the control stayed green |

Restoring all four returns every suite to green.

## Gates

`go build` 0 · `go vet` 0 · `go vet -tags integration ./integration/... ./internal/...` 0 · `go tool qtlint` 0 · `golangci-lint run ./...` 0 ("0 issues") · `check-test-style.sh` 0 · `check-public-api.sh` 0 · `check-public-api-snapshot.sh` 0 · `check-public-api-released.sh` 0 · `migrations lint --format sarif` 0 · `check-hcl-export-acceptance.sh` 0 · `check-api-export-acceptance.sh` 0 · docs site `check:exit-codes`, `check:core-doc-links`, `check:page-health`, `check:links`, `check:redirects`, `check:style`, `check:limitations`, `check:responsive` 0 each.

Live, against a CI-shaped cluster (`ptah_user` LOGIN SUPERUSER owning `ptah_test`): `./integration/gonative -tags=integration` 0 on PostgreSQL 18.4 and MySQL 9.7 — the job that was failing · `./integration -tags=integration` 0 · `internal/atlasschema -tags=integration` 0 · `cmd/atlas -tags=integration` 0 · `internal/migrateclean` 0 · `internal/atlasreport` 0 · the `internal/dblock` and `internal/dbschema/postgres` realm-cleanup steps 0.

## Consequence worth naming

`ptah introspect` on an unscoped connection now writes one more annotation, because the reader no longer denies the schema it read:

```
//ptah:schema:schema name="public" comment="standard public schema"
```

Rendering that emits `CREATE SCHEMA IF NOT EXISTS "public"` and `COMMENT ON SCHEMA "public"`. Both were already reachable through `--schemas public`; what changed is that the default reaches them. The `COMMENT` requires schema ownership — `must be owner of schema public` for a role that is not — so a native introspect-then-apply run as a non-owner meets a statement it did not meet before. That deserves its own decision about whether the connected schema's comment belongs in rendered DDL at all.

Issue #1276, opened 2026-08-07
The reader's scope and the comparator's scope are set independently, and neither knows what the other did


denisvmedia:
Four separate defects today share one root. Filing it once so the next change of this shape is designed against the pattern rather than rediscovering it.

## The pattern

Ptah has a **read** path that describes a database and a **compare** path that decides what to change. Both make scope decisions — which schemas, which objects, which attributes — and they make them independently. When one side moves, the other keeps its old assumption about what silence means, and the result is always the same class of bug: something absent from a description is read as something absent from the database.

| # | what moved | what broke | state |
| --- | --- | --- | --- |
| #1266 | inspect omitted a block nothing **names** | the document still depended on what the extension **provides**; Ptah could not read its own output | fixed in #1274 |
| #1272 | reader preserves access method, opclass, sort order, `INCLUDE` (#1246, #1271) | comparator compares one attribute outside SQL Server, so `USING gin` and `USING gist` are equal | open |
| #1273 | role reader scoped to the inspected schema | comparator read **out of scope** as **absent** and planned `CREATE ROLE` for roles that exist | held |
| #1275 | schema reader widened to realm scope | comparator read a described schema as **absent** and planned `CREATE TABLE` for a table that exists | held |

Two of the four moved the reader narrower and two moved it wider. Both directions break, which is the tell: the defect is not the direction, it is that the two sides are not required to agree.

## Why the tests did not catch any of them

Each was covered by a test at the level that changed — the rendered document, the JSON shape, the role list. None had a test that crossed the boundary. The shapes that would have failed:

- inspect a database, feed the output back to Ptah, require exit 0;
- inspect a database, apply the output to the database it came from, require a no-op;
- change one semantic attribute and require the comparator to notice.

The third and fourth were found by running exactly those by hand after the PRs were green.

## What to build

1. **Name the scope decision once.** `internal/schemaselection` and `internal/migrateclean` already resolve realm-versus-schema scope for the adoption gate (#1257). Both paths should consume that answer rather than each deriving one.
2. **Make "not described" and "not present" distinguishable** in whatever the comparator consumes. A reader that was not asked about a schema must not be indistinguishable from a reader that looked and found nothing — that single conflation is #1273 and #1275.
3. **A boundary-crossing guard in CI.** One test per surface: inspect → parse back → same IR; inspect → apply → no-op. These are cheap and each of the four would have failed one of them.

## Definition of done

- The scope decision has one owner and both paths cite it.
[...]

PR #1283, opened 2026-08-07
Key grants and RLS enablement by table identity, not by spelling (#1276)

denisvmedia:
Refs #1276. Two more instances of its shared root: a comparator that builds its own map key from a raw string, which is exactly what #1232 fixed for `tableMemberKey` and what the doc comment there warns the next caller about.

## The defect

`compare.Grants` keyed on the object name as written. The database side comes from `DBGrant.QualifiedTarget`, which carries the schema the reader found; the declared side carries whatever the author wrote. So `granted` and `"public"."granted"` were two grants — the declaration absent from the database and therefore GRANTed, the database row absent from the declaration and therefore REVOKEd, on every run of an unchanged schema. The grant-option paths key through the same identity, so a qualified target reached neither `GrantOptionsAdded` nor `GrantOptionsRevoked`.

`compare.RLSEnabledTables` had the same shape against `DBTable.QualifiedName`: `secured` and `public.secured` were two tables, so an unchanged schema planned ENABLE on a table that already had it and DISABLE on that same table.

## The fix

Both key through `newQualifiedTableIdentity`, following the package's existing convention — `GrantsWithSemantics` / `RLSEnabledTablesWithSemantics` take the resolved rules, the old names delegate with `identifier.ForDialect("")`, and `schemadiff.go` passes the `identifierSemantics` it already has in scope. Same shape as `ConstraintsWithSemantics` / `Constraints`.

A SCHEMA grant is the one target that is not a table, so it goes in the schema slot with the table slot left empty. That also stops `GRANT ... ON SCHEMA app` from colliding with `GRANT ... ON TABLE app`, which a single normalized string key would have allowed.

Reported names stay the strings each side supplied, because the planner renders them. Only the matching is normalized.

## Mutation, both directions

Twelve rows added across two files.

| mutant | rows that go red |
|---|---|
| grant key reverted to the raw string | bare table, bare sequence, grant option — 3 of 7 |
| RLS key reverted to the raw string | bare declaration — 1 of 5 |
| schema component dropped from both keys | the two cross-schema controls, and nothing else |

The controls that stay green under the first two mutants are the ones that must: a target in a different schema is still a different target, and a SCHEMA grant still does not match a TABLE grant of the same name.

## Reachability — measured, not assumed

On a single-schema PostgreSQL 17.10 database the defect does **not** surface, and it is worth saying why: `Reader.outputSchema` returns an empty schema for the connection's own schema, so both sides spell the table bare and match. Reading two schemas is what separates them. `ptah introspect --schemas public,other` against a database with `other.granted` and `other.secured` writes:

```
//ptah:schema:grant role="rg1276_app" privilege="SELECT" on_table="other.granted"
//ptah:schema:table name="granted" schema="other"
//ptah:schema:rls:enable table="secured"
```

The grant is qualified and matches. The RLS enablement is bare, against a reader that reports `other.secured`.

## What I could not demonstrate, and why it matters

I could not show either defect or its fix through `ptah schema compare`. That command reports `[]` on this fixture — and it still reports `[]` after I disabled RLS in the database with the annotation demanding it enabled, which is the control that says the surface is not merely agreeing with me. The comparator's RLS and grant results do not reach that command's rendering at all.

So this PR's evidence is the comparator's own behavior plus the introspect spellings above. `internal/planner/dialects/postgres`, `migration/generator` and `migration/safety` do read these fields; `cmd`'s compare rendering does not. That gap is its own defect and I have not fixed it here.

## Gates

`go build` 0 · `go vet` 0 · `go test ./... -count=1` 0 · `go tool qtlint` 0 · `golangci-lint run ./...` 0 · `check-test-style.sh` 0 · `check-public-api.sh` 0 · `check-public-api-snapshot.sh` 0 · `check-public-api-released.sh` 0.

`compare` is an internal package, so the two new exported names move no public API gate.

### DIFF
--- a/integration/gonative/schema_boundary_guard_postgres_test.go
+++ b/integration/gonative/schema_boundary_guard_postgres_test.go
@@ -130,10 +130,13 @@ func boundaryCases() []boundaryCase {
 			wantDocumentOnly: []string{"schema:public"},
 			wantDatabaseOnly: nil,
 
-			// WRONG (#1276): applying a database's own description back to it
-			// plans fourteen statements. Must become nil on both surfaces.
-			wantNativePlan: boundaryOwnerGrantChurn("public", "a"),
-			wantCompatPlan: boundaryOwnerGrantChurn("public", "a"),
+			// FIXED (#1283): applying a database's own description back to it
+			// plans nothing. The fourteen statements this row used to expect
+			// were one grant revoked under the unqualified spelling and
+			// re-granted under the qualified one; keying the comparison by
+			// table identity collapsed them.
+			wantNativePlan: boundaryNoPlan,
+			wantCompatPlan: boundaryNoPlan,
 		},
 		{
 			// The control. The same content reached with an explicit
@@ -153,9 +156,10 @@ func boundaryCases() []boundaryCase {
 			wantDocumentOnly: []string{"schema:public"},
 			wantDatabaseOnly: nil,
 
-			// WRONG (#1276): must become nil on both surfaces.
-			wantNativePlan: boundaryOwnerGrantChurn("public", "a"),
-			wantCompatPlan: boundaryOwnerGrantChurn("public", "a"),
+			// FIXED (#1283): the grant churn is gone here too, on the URL form
+			// that pins a single schema.
+			wantNativePlan: boundaryNoPlan,
+			wantCompatPlan: boundaryNoPlan,
 		},
 		{
 			// An extension supplying a type a column uses. #1266: inspect
@@ -175,11 +179,11 @@ func boundaryCases() []boundaryCase {
 			wantDocumentOnly: []string{"schema:public"},
 			wantDatabaseOnly: nil,
 
-			// WRONG (#1276): must become nil on both surfaces. No extension
-			// statement appears here -- the plan is the same fourteen-statement
-			// grant churn every table gets.
-			wantNativePlan: boundaryOwnerGrantChurn("public", "books"),
-			wantCompatPlan: boundaryOwnerGrantChurn("public", "books"),
+			// FIXED (#1283): no extension statement appeared here even when
+			// this row was red -- the plan was the grant churn every table got,
+			// and it is gone. An extension a column uses stays described.
+			wantNativePlan: boundaryNoPlan,
+			wantCompatPlan: boundaryNoPlan,
 		},
 		{
 			// An extension nothing references. This is the fifth defect, and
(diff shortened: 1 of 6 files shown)


denisvmedia:
Pushed 41e7115a. The end-to-end demonstration I said this PR could not give turns out to exist already — #1279's boundary guard had been recording this exact defect since it landed, under a different name.

That guard pinned a fourteen-statement plan for four of its six fixtures and marked it `WRONG (#1276)`: applying a database's own description back to it revoked seven privileges and granted seven back. Here is the plan, and the reason it is one grant and not fourteen:

```
REVOKE DELETE ON TABLE "a"          FROM "ptah_user"
GRANT  DELETE ON TABLE "public"."a" TO   "ptah_user" WITH GRANT OPTION
```

The helper's own doc comment had already diagnosed it:

> the two sides key a grant on how the table is NAMED, and they name it differently. The reader reports the grant against `pk_t`, exactly as the catalog spelled it; the same grant parsed back out of the document resolves through `for = table.pk_t` to the qualified `public.pk_t`. […] **Must become nil.**

With this PR it is nil. Four rows move from `boundaryOwnerGrantChurn` to `boundaryNoPlan` and lose their WRONG marker; the helper has no remaining caller and goes with them.

### Mutation, end to end

Reverting `newGrantIdentity` to the raw string key reddens `property 2: inspect then apply back is a no-op` on both surfaces in exactly those four fixtures — `two_schemas_plain_url`, `two_schemas_search_path_public`, `extension_used_by_a_column`, `table_with_primary_key` — and nothing else in the file. Restored, the whole file is `ok`.

So this supersedes the "What I could not demonstrate" section of the PR body: the defect and the fix are both observable through `ptah schema apply` against a live database. What remains true is that `ptah schema compare` shows neither, which is #1284.

### A note on the measurement environment, because it cost me an hour

This guard is green on master **only** against a CI-shaped database. Connected as the bootstrap `postgres` superuser it fails on master too, in the same four `property 2` rows — the expected plan encodes one particular ownership arrangement. Faithful local repro:

```
docker run -d --name <c> -e POSTGRES_PASSWORD=pw -p <port>:5432 postgres:17
CREATE ROLE ptah_user LOGIN PASSWORD 'ptah_password' SUPERUSER;
CREATE DATABASE ptah_test OWNER ptah_user;
POSTGRES_TEST_DSN='postgres://ptah_user:ptah_password@localhost:<port>/ptah_test?sslmode=disable'
```

A fixture that is green for a reason unrelated to what it asserts is worth fixing on its own; I have not done it here.

### Still red-capable

Every other `WRONG (#1276)` row is untouched: the reader still reports no schemas, the document still declares a schema the reader never reported, and the compatibility surface still plans `DROP EXTENSION`. This PR does not relax a single expectation to make a red row green — it removes expectations that record a defect that is gone.

PR #1288, opened 2026-08-07
Bind the native migration writers to the directory they verified (#1118)

denisvmedia:
## What was already done, and what was not

Two of the three surfaces #1118 names were fixed before this branch. Naming them precisely, because claiming credit for them would be worse than fixing nothing:

| surface | status on master `4fd9f18a` |
| --- | --- |
| `ptah-compat migrate diff` | **already fixed** by 028ac05d, "fix(compat): bind migrate diff publication to the opened directory (refs #1118)". `internal/atlasmigrate/diff_root_unix_test.go` holds the three replacement regressions. |
| `ptah-compat migrate new`, foreign `--dir-format` | **already fixed** by e4b6f4c1, "Write migrate new skeletons through a rooted handle (#1222)". `cmd/atlas/migrate_new.go` calls `atlasmigrate.WriteSkeletonMigration(migrateDiffWriterRoot(...), …)`. The issue's last status comment lists this as open; it is not, as of that commit. |
| `migration/generator` | **still broken.** This PR. |

## The defect

Both of `migration/generator`'s writers resolved the output directory to a string and then reopened it by pathname for every later step.

`ptah migrations create` — and therefore the forwarded `ptah-compat migrate new` for a plain Atlas directory — reopened the pathname for the mkdir, the version scan, each file it created, and the `atlas.sum` commit. `ptah migrations generate` verified the directory's contents by pathname and then let the publication resolve the pathname a second time. `AllowedOutputRoot` was compared against once by `pathguard.ResolveWithinRoot` and then dropped.

Measured on master `4fd9f18a`:

| probe | master `4fd9f18a` | this branch |
| --- | --- | --- |
| `GenerateEmptyMigration --dir-format atlas`, `AllowedOutputRoot` set, racing a symlink swap of its own migration directory to a decoy **outside** that root; 40 runs | **4 of 40 published outside the allowed root.** Escaped artifacts included `20260807155608_added.sql` and `.atlas.sum.3342714190.tmp` | **0 of 40** |
| plan against `root/migrations -> real`, replace the symlink with an empty decoy, then `WriteFilesContext` | **exit nil.** `real=[]`, `decoy=[1700000000_create_users.down.sql 1700000000_create_users.up.sql]` | `migration directory changed after migration planning`; `real=[]`, `decoy=[]` |

The second row needs no fault injection. `PlanMigration` and `WriteFilesContext` are separate public calls, and the pre-publication comparison only asked whether the contents matched — two empty directories match.

## The change

Both writers bind the migration directory and its parent once and run the whole transaction through that binding: materializing a missing directory through the bound parent, scanning versions, creating files exclusively, committing `atlas.sum`, snapshotting before publication, publishing the batch, and rolling back. `AllowedOutputRoot` is *opened*, so the binding itself is confined. The plan additionally records the planned directory's filesystem identity, and publication refuses a directory that is no longer the same object — a substitute holding exactly the files the plan verified is still a different destination.

`atlasmigrate.PublishArtifactsLocked` is **removed** rather than kept beside the new entry point. It took a directory pathname and bound its own handle, which is precisely the unconstrained second reader the issue rules out; `MigrationWriter.PublishArtifacts` is now the only way in, and `migrationWriterDir.create` is the single place a migration directory is materialized through a bound parent.

The public `migration/generator` API is unchanged — the rooted handle is an `internal/atlasmigrate` type reached only on the internal path, which is the "internal seam" option from the issue's own analysis. `docs/public_api.snapshot` does not move and `check-public-api-released` reports no `migration/generator` entry.

**This changes behavior.** A `migrations generate` run that previously published into a recreated directory now fails and writes nothing. Pre-v1, so no compatibility with Ptah's own past is owed; documented in `docs/site/src/content/docs/concepts/migration-directory.md` and `migration/generator/README.md`.

## Regressions

`migration/generator/rootedwriter_unix_internal_test.go` (white-box, `//go:build unix`, justified after the package clause — the moment it measures, after the directory is bound and before anything is read through it, has no exported name because nothing in the product needs one):

- `TestGenerateEmptyMigrationReplacedDirectoryCannotRedirectTheWrite` — 4 rows: the migration directory replaced under an allowed root, with no allowed root, in the paired layout, and an ancestor directory replaced.
- `TestGenerateEmptyMigrationCreatesMissingDirectoryThroughTheBoundParent`
- `TestMigrationPlanWriteFilesReplacedDirectoryCannotRedirectPublication`

Black-box, no seam: `TestMigrationPlanWriteFiles_RefusesDirectoryReplacedBeforePublication` (2 rows, separating the identity guard from the rooted-open guard) and the portable `TestMigrationPlanWriteFiles_RefusesRecreatedDirectory`.

Every row asserts both halves the acceptance criteria ask for: the artifacts land in the bound directory, and the directory the replacement pointed at is left completely untouched.

## Mutation, both directions

| mutant | red | green |
| --- | --- | --- |
| **M1** revert: `WriteNew` writes by pathname | all 4 `ReplacedDirectory…` rows, `CreatesMissingDirectoryThroughTheBoundParent` | every plan test, the other 6 `GenerateEmptyMigration*` tests |
| **I1** over-correct: `WriteNew` revalidates and refuses on change | the same 5 rows, now with `failed to write up migration file: opened directory path changed: …` instead of files in the decoy | same as M1 |
| **M2** revert: `equal` drops `os.SameFile` | `RefusesRecreatedDirectory`, `RefusesDirectoryReplacedBeforePublication/no_allowed_root` | `…/under_an_allowed_root`, every create row, every happy-path publication |
| **I2** over-correct: identity always mismatches | `…ReplacedDirectoryCannotRedirectPublication`, `PublishesMultiplePairsAndReports`, `CollisionLeavesNoPartialArtifacts`, `HappyPath`, `FailurePath` | the two refusal tests, every create row |
| **M3** revert: `openOutputRoot` always nil | `RefusesDirectoryReplacedBeforePublication/under_an_allowed_root` **only** | everything else |
| **I3** over-correct: an empty root confines to the working directory | 13 rows including every direct-CLI create and publication | the rows that already run under a root |

M3 is the honest one: the create-path rows stay green without the confinement root, because the bound handle alone already keeps those writes in the retained directory. The root's unique contribution is refusing, at bind time, a directory that now resolves outside it, and exactly one row pins that.

## Gates

`go build` 0 · `go vet` 0 · `qtlint` 0 · `golangci-lint` 0 · `check-test-style` 0 · `check-public-api` 0 · `check-public-api-snapshot` 0 · `check-public-api-released` 0 · all eight `docs/site` `check:*` scripts 0 (`check-responsive` after `npm ci && npm run build`) · `build-feature-matrix --check` 0.

`go test ./... -count=1` is **exit 1**, from three packages this change does not touch:

- `cmd/ptah-ls` — a 20s process budget missed under load; passes alone (`ok 46.236s`).
- `cmd/viz/TestSVGReportsGraphvizStderrOnFailure` — `signal: killed`. Reproduced identically in a clean worktree at `4fd9f18a`.
- `cmd/atlas` — `signal: terminated` at ~10 minutes with no test output. Reproduced identically at `4fd9f18a` (`FAIL 642.051s`). On this branch, `-run Migrate -parallel 1` — every command that goes through the writer touched here — is `ok 110.521s`.

Every package this change touches passes: `internal/atlasmigrate`, all of `migration/…`, `cmd/migrate`, `cmd/migratecheckpoint`, `cmd/migratedata`.

## Still open on this issue

- `cmd/migrate/new.go --edit` rehashes through `migrateops.Rehash(dir, format)` by pathname after the external editor exits. Separate transaction, outside `migration/generator`; the acceptance criteria's editor bullet is half covered.
- `WriteCheckpointFiles`, `WriteDataMigrationFiles` and `WriteAtlasCheckpointFile` (`ptah migrations checkpoint`, `ptah migrations data`) still write by pathname. Not named by #1118 and carrying no `AllowedOutputRoot`, but the machinery to convert them is now in place.

Refs #1118.
---

**On the gate list**: every gate is 0 except `go test ./... -count=1`, which exited 1 on three subprocess-heavy packages with `signal: killed` / `signal: terminated`. That was measured on a machine carrying roughly twenty concurrent `go test` runs, and an unmodified base checkout reproduced the same failures at the same assertion lines. No package this change touches is among them. CI is the arbiter here rather than my local run.

### DIFF
--- a/docs/site/src/content/docs/concepts/migration-directory.md
+++ b/docs/site/src/content/docs/concepts/migration-directory.md
@@ -97,18 +97,30 @@ bytes.
 
 ## Writing back to the directory
 
-The verbs that write a migration directory — `migrate diff` and native
-`ptah migrate generate` — bind it the same way and keep the binding. They open
-the directory and its parent once, before staging, and every staged file,
-published migration, `atlas.sum`, journal, commit marker, rollback quarantine
-and cleanup entry is named as a direct child of one of those two handles.
-Recovery of an interrupted batch runs through the same handles.
+The verbs that write a migration directory — `migrate diff`, `migrate new`, and
+native `ptah migrations generate` and `ptah migrations create` — bind it the
+same way and keep the binding. They open the directory and its parent once,
+before staging, and every staged file, published migration, `atlas.sum`,
+journal, commit marker, rollback quarantine and cleanup entry is named as a
+direct child of one of those two handles. A migration directory that does not
+exist yet is created through the bound parent, so it is materialized where the
+run looked rather than where the pathname points by then. Recovery of an
+interrupted batch runs through the same handles.
 
 Replacing the directory after the run validated it therefore cannot redirect
 what it writes, and a directory configured through `atlas.hcl` stays inside the
 opened project root. See
 [the publication boundary](../../atlas/migrate-commands/#the-publication-boundary)
 for what remains keyed to the pathname and why.
+
+`ptah migrations generate` plans and publishes in two steps, so the directory
+can be replaced between them by something outside the run's control. The plan
+records the planned directory's filesystem identity as well as its contents, and
+publication refuses a directory that is no longer the same object — including a
+substitute that holds exactly the files the plan verified. The refusal is
+`migration directory changed after migration planning`, and nothing is written
+([#1118](https://github.com/stokaro/ptah/issues/1118)). This changes behavior:
+a run that previously published into a recreated directory now fails instead.
 
 ## Consequences
 
--- a/internal/atlasmigrate/diff_internal_test.go
+++ b/internal/atlasmigrate/diff_internal_test.go
@@ -150,13 +150,15 @@ func TestWritePublicationJournal_FallsBackWithoutHardLinks(t *testing.T) {
 	c.Assert(removePublicationJournal(writer), qt.IsNil)
 }
 
-func TestPublishArtifactsLocked_PublishesCompleteBatch(t *testing.T) {
+func TestMigrationWriterPublishArtifacts_PublishesCompleteBatch(t *testing.T) {
 	c := qt.New(t)
 	dir := c.TempDir()
+	migrationWriter, err := OpenMigrationWriter(nil, dir)
+	c.Assert(err, qt.IsNil)
+	defer func() { _ = migrationWriter.Close() }()
 
-	paths, err := PublishArtifactsLocked(
+	paths, err := migrationWriter.PublishArtifacts(
 		t.Context(),
-		dir,
 		[]PublicationArtifact{
 			{Name: "1_change.up.sql", Contents: []byte("SELECT 1;\n")},
 			{Name: "1_change.down.sql", Contents: []byte("SELECT 2;\n")},
@@ -183,15 +185,17 @@ func TestPublishArtifactsLocked_PublishesCompleteBatch(t *testing.T) {
 	c.Assert(err, qt.ErrorIs, os.ErrNotExist)
 }
 
-func TestPublishArtifactsLocked_CollisionRollsBackWholeBatch(t *testing.T) {
+func TestMigrationWriterPublishArtifacts_CollisionRollsBackWholeBatch(t *testing.T) {
 	c := qt.New(t)
 	dir := c.TempDir()
 	existingPath := filepath.Join(dir, "1_change.down.sql")
 	c.Assert(os.WriteFile(existingPath, []byte("existing\n"), 0o600), qt.IsNil)
+	migrationWriter, err := OpenMigrationWriter(nil, dir)
+	c.Assert(err, qt.IsNil)
+	defer func() { _ = migrationWriter.Close() }()
 
-	paths, err := PublishArtifactsLocked(
+	paths, err := migrationWriter.PublishArtifacts(
 		t.Context(),
-		dir,
 		[]PublicationArtifact{
 			{Name: "1_change.up.sql", Contents: []byte("SELECT 1;\n")},
 			{Name: "1_change.down.sql", Contents: []byte("SELECT 2;\n")},
--- a/internal/atlasmigrate/migrationwriter.go
+++ b/internal/atlasmigrate/migrationwriter.go
@@ -1,0 +1,172 @@
+package atlasmigrate
+
+import (
+	"context"
+	"errors"
+	"io/fs"
+
+	"go.5x5.cz/ptah/internal/migratesum"
+	"go.5x5.cz/ptah/internal/pathguard"
+)
+
+// MigrationWriter is the rooted migration-directory capability a writer
+// transaction outside this package runs through: one binding of the directory
+// and its parent, reused for every read, create, publication and rollback the
+// transaction performs.
+//
+// It exists because the native generator had the shape stokaro/ptah#1118
+// describes. `ptah migrations create` resolved its output directory once, then
+// reopened it by pathname for the mkdir, for the version scan, for each file it
+// created and for the atlas.sum commit; `ptah migrations generate` verified the
+// directory's contents by pathname and then let the publication resolve the
+// pathname a second time. A directory or ancestor replaced between any two of
+// those steps takes the write somewhere the verification never looked.
+//
+// The methods are deliberately thin: this type is the boundary, not a second
+// filesystem abstraction. Everything behind it is the same migrationWriterDir
+// the `migrate diff` publication already runs through, so both writers share one
+// answer to "which filesystem object is this transaction committing to".
+//
+// A nil root keeps direct-CLI behavior, where an explicit absolute directory is
+// the operator's own choice of destination. The binding still happens; it is
+// simply not confined to a project root.
+type MigrationWriter struct {
+	dir *migrationWriterDir
+}
+
+// OpenMigrationWriter binds dir and its parent without creating dir. A missing
+// migration directory is reported through Exists rather than as an error, so a
+// caller can compare the directory's absence against what it planned for before
+// deciding to materialize it.
+//
+// The parent must already exist; call EnsureMigrationParent first when the
+// caller accepts a directory nested below missing levels.
+func OpenMigrationWriter(
+	root *pathguard.OpenedDirectory,
+	dir string,
+) (*MigrationWriter, error) {
+	bound, err := openMigrationWriterDir(root, dir)
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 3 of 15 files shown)


denisvmedia:
All three red jobs — `unit-tests`, `coverage` and `windows-publication` — are the same test, `TestMigrationPlanWriteFiles_RefusesRecreatedDirectory`, and it is not flaky. It fails for a reason that undercuts the guard it asserts.

### Why it passes here and fails in CI

`migrationDirectoryState.sameAs` compares the planned directory to the published one with `os.SameFile`, against an `fs.FileInfo` taken by `os.Stat` at planning time (generator.go:585, "before any writer handle exists"). `os.SameFile` is device + inode on Unix and volume + file index on Windows. It answers "is this the same filesystem object" only for as long as the OS does not hand the identifier out again.

Measured on this machine, 20 remove-and-recreate cycles of the same pathname:

```
first run: before ino=323537633 dev=16777234 | after ino=323537634 dev=16777234
os.SameFile said SAME for 0 of 20 remove+recreate cycles
```

macOS APFS never reused the inode, so the guard fires every time and `go test ./...` is green locally — I ran it and got exit 0 across 178 packages, which is why I pushed. CI's Linux and Windows runners reuse the identifier, `os.SameFile` answers "same", and the guard stays silent.

### What that means for the change

The distinction matters and the two probes in the PR body are not equally affected:

- **Symlink swapped to a different directory** — a genuinely different object, a different inode, so `SameFile` is false and the refusal fires. That probe, and the 4-of-40 escape it fixes, stand.
- **Same pathname removed and recreated** — the object the attacker substitutes may land on the recycled identifier, and then the publication proceeds into it. This is the shape the failing test asserts, and CI is telling us the protection is not there.

So the guard is inode luck on exactly the shape an attacker controls most easily: they do not have to find a different directory to point at, they only have to delete and recreate.

### What it needs

Not a build tag. Gating the test to Unix would have hidden the Linux failure too, and the Linux failure is the important one.

The PR already introduces rooted handles; the planning side is what still resolves by pathname. Holding an open directory handle from planning through publication makes the question unnecessary rather than answering it more cleverly — a recreated directory is simply not the one the process holds, whatever identifier the OS gave it. `migrationDirectoryState` recording `identity fs.FileInfo` is the part to replace.

Until then this should not merge: it would ship a documented refusal that does not happen on either platform CI runs.

I am leaving the test exactly as it is. It is asserting the right property and it is correctly red.

denisvmedia (commented):
Independent review confirms that this head is not mergeable.

**Blocker — detached `FileInfo` cannot prove directory continuity.** `migration/generator/generator.go:643` stores an `os.Stat` result after releasing all handles and later trusts `os.SameFile`. Once the original is removed, inode/file-index reuse can make a replacement appear identical. Current Linux and Windows CI reproduce this exact failure: `TestMigrationPlanWriteFiles_RefusesRecreatedDirectory` gets nil instead of `ErrMigrationDirectoryChanged`.

Keep an opened directory capability from planning through publication; do not bridge the phases with detached identity metadata.

The security documentation also overclaims `migrate new`: `cmd/migrate/new.go:146` still performs pathname-based rehash after the editor exits, so a replacement during editing can redirect the final `atlas.sum` write. Carry the rooted writer through editing/rehash or narrow the claim. PR #1304 is the later retained-handle attempt; this branch should not be merged.

denisvmedia:
Closing as superseded by #1304, which merged to \`master\` on 2026-08-08 as \`49367482\`.

#1304 was opened as a follow-up on top of this branch's head \`d9d2a1de\` and carries the same design: \`PlanMigration\` binds the migration directory through the confinement root and **holds** the binding; \`WriteFilesContext\` reopens nothing, calls \`MigrationWriter.Revalidate()\` under the lock, compares contents, publishes through that handle, then releases it. \`MigrationWriter.Identity\` became \`Revalidate\` and \`migrationDirectoryState\` is gone. \`git diff origin/master d9d2a1de\` is 577 insertions against 3133 deletions — this branch is behind master, not ahead of it.

**The red jobs were one test, and it was correct to be red.** \`windows-publication\`, \`unit-tests\` and \`coverage\` all failed on \`TestMigrationPlanWriteFiles_RefusesRecreatedDirectory\`: \`got nil error but want "migration directory changed after migration planning"\`.

Measured on a loopback **ext4** filesystem in a privileged Linux container — Docker's own \`/tmp\` is overlayfs and reproduces nothing:

| regime | \`os.SameFile\` said SAME |
|---|---|
| ext4, \`os.Stat\` identity, no handle held | **20 of 20** |
| ext4, identity from a held root handle | 0 of 20 |
| APFS (macOS), either regime | 0 of 20 |
| ext4, move-aside then destroy then recreate, no handle | **20 of 20** |
| ext4, same shape, handle held | 0 of 20 |

So the guard was relying on the filesystem not recycling an identifier, and Linux recycles it every time. macOS never does, which is why this was green locally and red in CI.

There is a second mechanism on Windows that the original diagnosis did not name: in Go 1.26.5 \`os/types_windows.go\`, \`sameFile\` calls \`loadFileId()\`, which **reopens the recorded pathname** for a \`fileStat\` produced by \`os.Stat\`. On Windows the old guard therefore held no identifier at all — it re-resolved the pathname at comparison time, and would have failed regardless of file-index reuse.

#1304 also handles the Windows removal semantics better than a rename-aside would: the test's hostile step is \`os.RemoveAll\`, which routes through \`removedirat\` to \`windows.Deleteat\` with \`FILE_DISPOSITION_POSIX_SEMANTICS\`, freeing the name immediately even with a handle open — unlike \`os.Remove\`, whose \`RemoveDirectory\` leaves the name reserved until the last handle closes. That is why the test stays untagged rather than gated to Unix.

**Verified against master's landed code.** Disarming the guard there (\`migrationWriterDir.revalidate\` returning nil) reddens five tests on ext4:

```
--- FAIL: TestMigrationPlanUsesTheHeldDirectoryForVersionAndPublication
--- FAIL: TestMigrationPlanWriteFiles_ReleasesTheDirectoryWhenPublicationFails
--- FAIL: TestMigrationPlanWriteFiles_RefusesRecreatedDirectory
--- FAIL: TestMigrationPlanWriteFiles_RefusesADirectoryThatAppearedAfterPlanning
--- FAIL: TestMigrationPlanWriteFiles_RefusesDirectoryReplacedBeforePublication
```

Restored, all 21 \`TestMigrationPlan*\` rows pass. Reproducing the literal pre-fix shape (plan-time \`os.Stat\`, handle closed) splits by platform exactly as CI did: green on APFS, red on ext4.

Two items remain open on master and are filed separately rather than carried here.

PR #1297, opened 2026-08-08
Roll back the nine categories the reverse plan never mentioned (#1287)

### LINKED ISSUE #1287: The reverse plan silently omits nine SchemaDiff categories, and the obvious fix makes view rollbacks un-appliable
`migration/generator`'s `reverseSchemaDiffWithSchema` builds a fresh `SchemaDiff` literal and never mentions nine of the struct's fields, so a down migration silently omits whole categories of the change it is supposed to undo.

Filed separately because no open issue covers it. It was worked under the label "#1013", which is a different defect — that one is `?format=` rejected on `migrate lint/new/diff`.

## What the swap misses

The reverse plan is built by swapping the added and removed collections. A swap is correct only where the two collections carry the same kind of value and where reversing the operation is the inverse of applying it. Nine fields are simply absent from the literal, among them views, materialized views and triggers. An up migration that creates a view therefore produces a down migration that does not drop it.

## What a naive fix breaks, measured

Adding the missing fields is not enough, and this is the part worth writing down. Carrying `ViewsModified` into the down plan makes the PostgreSQL planner render `CREATE OR REPLACE VIEW`, which PostgreSQL only accepts for appending trailing columns. Enumerated on PostgreSQL 17.10:

| change to view `v` | PostgreSQL |
|---|---|
| append a trailing column | accepted |
| drop the appended column | `ERROR: cannot drop columns from view` |
| rename a column | `ERROR: cannot change name of view column "id" to "uid"` |
| change a column type | `ERROR: cannot change data type of view column` |

So a down migration built that way is un-appliable for every view modification except the one shape the up migration is least likely to have made. A rollback that fails is worse than one that is incomplete: the operator finds out during the incident.

The modification path has to emit DROP + CREATE from inside the modify step, the way `modifyExistingMaterializedViews` already does at `internal/planner/dialects/postgres/postgres.go:1792`. That placement is also what keeps the ordering right — done from outside, the reverse plan comes out create-then-drop.

Keeping `CREATE OR REPLACE VIEW` where it is legal is worth doing, since the DROP carries CASCADE and takes dependents with it. The legality test is mechanical: the new column list must be the old one with columns only appended, same names and same types in the same order.

## Definition of done

- Every field of `SchemaDiff` is accounted for in the reverse plan: reversed, deliberately dropped with the reason recorded, or unreachable in a down direction with the reason recorded. A field silently absent is what produced this.
- A test that fails when a field is added to `SchemaDiff` and not handled by the reverse builder. Reflection over the struct is the honest way; a hand-written list is what rotted.
- Every reversed category applied against a real database, not just rendered: seed the up state, apply the down plan, assert exit 0 and assert the catalog matches the pre-up state. Rendering proves the statement exists, not that PostgreSQL accepts it — the view case above renders fine and fails on execution.

denisvmedia:
Fixes #1287.

## What was wrong

`migration/generator`'s `reverseSchemaDiffWithSchema` builds a fresh `SchemaDiff` literal, and a literal that enumerates fields is silent about the fields it omits. Nine were absent for as long as the categories existed: `ViewsAdded`, `ViewsRemoved`, `ViewsModified`, the three `MaterializedViews*` and the three `Triggers*`. They were therefore zero on every down migration, so a rollback skipped the whole category. An up migration that created a view produced a down file reading exactly `-- No rollback operations needed`.

`cloneSchemaDiff`, ten lines away in the same file, enumerates all nine. Diffing the two function bodies is how you find what the reverse forgot.

## Why the obvious fix is wrong

Adding the missing fields is not enough, and this is the part worth writing down. Carrying `ViewsModified` into the down plan makes the PostgreSQL planner render `CREATE OR REPLACE VIEW`, which PostgreSQL accepts only for appending trailing columns. Measured on PostgreSQL 17.10 against a view over `(id bigint, email text, age integer)`:

| change to the view | PostgreSQL |
| --- | --- |
| append a trailing column | accepted |
| drop the appended column | `ERROR: cannot drop columns from view` |
| rename a column | `ERROR: cannot change name of view column "id" to "uid"` |
| change a column type | `ERROR: cannot change data type of view column "id" from bigint to text` |
| change only the predicate | accepted |

The rollback of an append is a drop, so a down migration built that way is un-appliable for every view modification except the one shape the up migration is least likely to have made. A rollback that fails is worse than one that is incomplete: the operator finds out during the incident.

Applying the naive plan's own output confirms it. Against the post-up state:

```
CREATE OR REPLACE VIEW "ptah_rev_active_users" AS SELECT id FROM ptah_rev_view_users;
-- rc=1  ERROR: cannot drop columns from view

DROP VIEW IF EXISTS "ptah_rev_active_users" CASCADE;
CREATE VIEW "ptah_rev_active_users" AS SELECT id FROM ptah_rev_view_users;
-- rc=0
```

## What this does

`modifyExistingViews` in the PostgreSQL planner now decides per view.

- It keeps `CREATE OR REPLACE VIEW` only where the new select list provably appends to the old one: same output column names, same type-determining text, same order. That comparison sees through the two spellings the same projection legitimately has — `SELECT id` as authored, `SELECT t.id` as `pg_get_viewdef` reads it back — while a cast, a rename or a different expression changes it. Keeping the replace where it is legal is worth the machinery, because the drop carries `CASCADE` and takes dependents with it.
- Everything it cannot decide — an unknown prior body, a `SELECT *` projection, a `WITH` prefix, a top-level set operation, an alias-less expression — takes the drop-and-recreate path, which always applies. The conservatism is one-directional on purpose: saying "no" costs a `CASCADE`, saying "yes" wrongly costs the rollback.
- The `DROP` and the `CREATE` are emitted from inside the modify step, the way `modifyExistingMaterializedViews` already does. Expressed from outside, by pushing the view into `ViewsRemoved` and `ViewsAdded`, the plan runs additions before removals and the reverse comes out create-then-drop.
- The recreated views are ordered by dependency, since a `CASCADE` takes a dependent view the same migration is also rebuilding.

The prior body has to reach the planner, which in the down direction only ever sees one side of the change, so `ViewDiff` gained `PreviousBody`: the body in force before the diff is applied. The comparator fills it from the database side; the reverse builder **exchanges** it for the up migration's target body, because that is the state the database is in when the rollback runs. Empty means "not known" and yields the conservative plan.

Materialized views need no legality test — they have no in-place replace, and both directions already drop and recreate. Triggers need none either: PostgreSQL 17.10 accepts `CREATE OR REPLACE TRIGGER` even for a `BEFORE` → `AFTER` timing change. MySQL's `CREATE OR REPLACE VIEW` is an unrestricted redefinition and the SQLite renderer already turns `SetReplace` into `DROP` + `CREATE`, so neither planner changed.

## Definition of done

**Every field accounted for.** A doc comment on `reverseSchemaDiffWithSchema` states the three dispositions — exchanged, carried, derived — and which fields take each. No field is deliberately dropped and none is unreachable in the down direction.

**A test that fails when a field is added and left unhandled.** `TestReverseSchemaDiff_AccountsForEverySchemaDiffField` enumerates `SchemaDiff` by reflection, zeroes one field of a fully populated diff at a time, and fails when the reverse plan comes out byte-identical. Adding a plausible new field and ignoring it reddens the subtest named after it:

```
--- FAIL: TestReverseSchemaDiff_AccountsForEverySchemaDiffField/PublicationsAdded
    SchemaDiff.PublicationsAdded never reaches the reverse plan: zeroing it left
    the down plan byte-identical.
```

A hand-written field list is what rotted the first time, so the fixture is filled generically from the field names too, and a second test fails if the filler ever leaves a field zero.

**Every reversed category applied, not just rendered.** `TestReverseViewLikeObjects_DownRoundTrip_Integration` seeds the pre-up state through Ptah's own planner, applies the generated up, applies the generated down, asserts it applied, and asserts the catalog it lands on is the pre-up catalog. Five cases: create the three objects, drop the three objects, append a view column, change only the view predicate, rewrite the materialized view and the trigger. The table the objects hang off exists before and after every case, so a `DROP TABLE ... CASCADE` can never be what removes them.

## Mutation coverage

Five mutations, each run and reverted:

| mutation | what reddens |
| --- | --- |
| planner always replaces (the naive fix) | 4 unit subtests, 6 planner subtests, the live append case; the mutant's own SQL exits 1 on PostgreSQL |
| planner never replaces | the predicate-only unit test, `SchemaObjectsModified`, 3 planner subtests, the live predicate case |
| drop `ViewsAdded`/`ViewsRemoved` from the literal | the reflection gate names exactly those two fields, plus both view unit tests |
| add an unhandled field to `SchemaDiff` | the reflection gate names it |
| carry `PreviousBody` instead of exchanging it | 4 unit subtests and the live append case |

## Notes

- `docs/public_api.snapshot` is regenerated. Adding a field to `ViewDiff` is a compatible change, so `check-public-api-released.sh` stays green and no approval entry was needed.
- Three integration tests in `migration/generator` — `TestConstraintDriftGenerateRoundTrip_Integration`, `TestUniqueConstraintDriftGenerateRoundTrip_Integration/postgres` and `TestMultiHostMixinFKModify_DownRoundTrip_Integration/postgres` — fail identically on `origin/master` and on this branch against a pristine database. They are pre-existing and unrelated, and no CI workflow runs them. Worth a separate issue.

---

### What it found

REFUTED. Their own measurements reproduce and every gate is genuinely green, but the change breaks a forward migration on an adjacent view shape, still produces the un-appliable rollback it exists to prevent, and still rolls a whole category back to "-- No rollback operations needed".

SETUP. My worktree checked out to eb1b8f39 (`git log --oneline -1` before every build); merge-base 9820496e checked out in a second worktree as the control (`git worktree list` confirmed both). My own container `ptahref1287-dev`, PostgreSQL 17.10, port 15801, CI role shape (ptah_user LOGIN SUPERUSER + bootstrap postgres), a fresh database per scenario, container removed at the end; MySQL 9.7 on the shared mylive-mysql:33110 in my own `ptah_r1287my`, dropped. Plans generated through the exported `generator.PlanMigration`/`GenerateMigration` from a probe module built separately against each worktree, applied with `psql -v ON_ERROR_STOP=1 -f`, exit code read directly.

WHAT HOLDS. The legality table reproduces exactly on my own fixtures: append trailing column rc=0; drop appended column rc=1 "cannot drop columns from view"; rename rc=1; retype rc=1; predicate-only rc=0. Their `TestReverseViewLikeObjects_DownRoundTrip_Integration` passes against my database (5/5, exit 0). Gates on eb1b8f39, each read directly, never after a pipe: go build 0, go vet 0, go vet -tags integration ./integration/... 0, go tool qtlint 0, golangci-lint "0 issues." 0 (private cache; my first attempt printed a false 0 because tail owned the status — re-run direct), check-test-style 0, check-public-api 0, snapshot 0, check-public-api-released 0, node docs/site/scripts/check-style.mjs 0. `go test ./... -count=1 -timeout 30m` exited 1 on `cmd/viz/TestSVGReportsGraphvizStderrOnFailure` under machine load; that package passes serially (exit 0) and is unrelated.

1. THE ADJACENT SHAPE — a forward migration now destroys a declared object. `modifyExistingViews` serves the UP plan too, and every body the new parser cannot decide (WITH prefix, SELECT *, top-level set operation) now takes DROP ... CASCADE in the forward direction. For a predicate-only edit that is a pure loss, because PostgreSQL accepts the replace for all of them. Schema: table probe_t, view probe_base = "WITH src AS (SELECT id, email FROM probe_t) SELECT id, email FROM src", plus a second DECLARED view probe_dep = "SELECT id FROM probe_base". Up changes only the CTE predicate.
  9820496e: `CREATE OR REPLACE VIEW "probe_base" AS WITH ...` rc=0, views after = probe_base, probe_dep
  eb1b8f39: `DROP VIEW IF EXISTS "probe_base" CASCADE; CREATE VIEW "probe_base" AS ...` rc=0, views after = probe_base
probe_dep is named by no statement in the plan. Identical as ptah_user and as the bootstrap postgres superuser. A dependent materialized view (3 rows) was destroyed the same way, contents included. And it does not converge: re-planning against the damaged database yields `CREATE VIEW "probe_dep" ...; DROP VIEW IF EXISTS "probe_base" CASCADE; CREATE VIEW "probe_base" ...` — the file creates the dependent and destroys it again, rc=0, probe_dep still gone. The control's plan in the same state (`CREATE VIEW probe_dep; CREATE OR REPLACE VIEW probe_base`) reaches the declared schema. Same for SELECT * and UNION bodies.

2. RENDERED IS NOT APPLIED — the change still emits the refused rollback. `viewReplaceOnlyAppendsColumns` reduces a plain reference to the bare column name and never reads the FROM clause, so "SELECT id FROM probe_a" (bigint) and "SELECT id FROM probe_b" (text) compare equal. Directly: `CREATE OR REPLACE VIEW r1287_fromv AS SELECT id FROM r1287_b` → rc=1, "cannot change data type of view column \"id\" from bigint to text". Reachable as a rollback: prior view "SELECT id, email FROM probe_a", target "SELECT id FROM probe_b". Generated up = `DROP VIEW IF EXISTS "probe_v" CASCADE; CREATE VIEW "probe_v" AS SELECT id FROM probe_b` (drop path, applies). Generated down = `CREATE OR REPLACE VIEW "probe_v" AS SELECT id, email FROM probe_a`. `psql -v ON_ERROR_STOP=1 -f <generated down file>` → rc=3, ERROR: cannot change data type of view column "id" from text to bigint. That is the exact failure mode the commit message justifies the whole design with. The doc comment's stated residual assumption ("the relations the shared items read have not themselves been retyped in the same migration") does not cover it: the relation was swapped, not retyped, and that is neither a cast, a rename, nor a different expression.

3. THE CLAIM THAT IS NOT CODE — a category is still dropped. The commit says "none is dropped or unreachable" and migration/generator/README.md says "No category is dropped from the rollback". The diff records a view under the name the Go schema spells; the pre-change schema the down plans against names it "<schema>.<view>" via DBView.QualifiedName (internal/convert/dbschematogo). findView compares those strings, so any view whose database side carries a schema never matches and ViewsModified renders nothing. Live on MySQL 9.7: up `CREATE OR REPLACE VIEW probe_my_v AS SELECT id, email FROM probe_my_t` + trigger rewrite (rc=0); generated down = `DROP TRIGGER ...; CREATE TRIGGER ...` and no view statement (rc=0); information_schema.VIEWS after rollback still holds the post-up body. Reproduced on the PostgreSQL path with Schema:"reporting" — down file reads literally "-- No rollback operations needed", the string the change is meant to have removed. Reaching the reversed SchemaDiff is not reaching the rollback, and the reflection gate only checks the former.

4. THE TEST THAT CANNOT RUN. `TestReverseViewLikeObjects_DownRoundTrip_Integration` is `//go:build integration` in `./migration/generator`. `go test ./...` never compiles it; `go vet -tags integration ./integration/...` does not cover that package; and no workflow runs `-tags=integration` there — .github/workflows/go-integration-tests.yml uses the tag only on ./integration, ./integration/gonative, ./internal/dbschema/{mysql,postg

### What it says must change

The reverse builder half of the change is right and should stay: the nine categories genuinely were absent, and exchanging added/removed while carrying the modified entries is the correct reversal. What has to change is the per-view replace decision and the claims made for it.

1. Do not let the legality test govern the forward direction. `modifyExistingViews` is shared by the UP and DOWN plans, and the UP direction had no problem to solve: before this change it emitted CREATE OR REPLACE VIEW and PostgreSQL accepted it for every shape whose column list did not move. Gate the drop-and-recreate on the plan direction, or thread the decision through so an undecidable body keeps the replace forward and only takes the drop on rollback. As it stands a predicate-only edit to a WITH/SELECT */UNION view silently destroys declared dependent views and materialized-view contents, and the plan never converges. If the drop path must ever run forward, the plan has to rebuild every dependent it takes — the CASCADE list is derivable from the generated schema the planner already holds.

2. Make the replace decision sound or make it refuse. A select item's type is fixed by the relation it reads, and the comparison never reads the FROM clause. Either include the normalized FROM/JOIN text in each item's type-determining key (so any change to the relations answers "not replaceable"), or answer "not replaceable" whenever anything outside the select list changed. Then correct view_replace.go's doc comment: the residual assumption is not "the relations have not been retyped in the same migration", it is "the relations read by shared items are the same relations".

3. Match on the same name the planner will look up. Resolve ViewsModified/MaterializedViewsModified/TriggersModified against the pre-change schema by qualified name (or record the qualified name in the diff), so a view outside the default schema is not silently skipped. Until that lands, neither the commit message nor migration/generator/README.md may say "no category is dropped from the rollback" — on MySQL/MariaDB, and on PostgreSQL for any named schema, ViewsModified still rolls back to "-- No rollback operations needed".

4. Strengthen the gate so it measures the rollback, not the reverse diff. TestReverseSchemaDiff_AccountsForEverySchemaDiffField passes while the rendered down file is empty. Add the companion assertion — zeroing a field must change the rendered DOWN SQL — which is what would have caught finding 3.

5. Wire the live test into CI or it is not coverage. Add a `go test -tags=integration ./migration/generator -run '^TestReverseViewLikeObjects_DownRoundTrip_Integration$'` step to the PostgreSQL job in .github/workflows/go-integration-tests.yml (POSTGRES_URL is already set there), and extend its table with the three shapes above: a CTE/star/union view with a declared dependent, a relation swap that retypes a shared column, and a view in a named schema.

6. Restore the un-fixtured assertion. Keep a case in TestPlanner_GenerateMigrationAST_SchemaObjectsModified with no PreviousBody so the default behavior for that scenario is pinned rather than adapted away.

7. Measure MySQL, MariaDB and SQLite live before claiming they need no change, and check CockroachDB and YugabyteDB accept the new DROP VIEW ... CASCADE the shared PostgreSQL planner now emits.

Ready-to-post PR (for the refutation commit 5db32a4a):

TITLE
Pin three shapes the per-view replace decision gets wrong (#1287)

BODY
The reverse-plan change on issue-1287-reverse-plan is right that nine SchemaDiff
fields never reached the rollback, and right that adding them naively renders a
CREATE OR REPLACE VIEW PostgreSQL refuses. Its own measurements reproduce on a
clean PostgreSQL 17.10 (append accepted; drop, rename and retype refused;
predicate-only accepted), its live round trip passes, and every gate is green.
Three shapes it never built say the decision it added is not finished. The tests
here are RED at eb1b8f39; they ar

### DIFF
--- a/.github/workflows/go-integration-tests.yml
+++ b/.github/workflows/go-integration-tests.yml
@@ -233,6 +233,19 @@ jobs:
           go test -v -count=1 ./migration/migrator -run 'TestSQLServerMigratorHonorsURLSchemaForMetadata|TestMigrationAdvisoryLock_SQLServer(DefaultTimeout|Timeout)Integration|TestAtlasRevisionMetadata_SQLServerIntegration|TestAtlasTxtarChecks_SQLServer(SequenceDoesNotAdvance)?Integration|TestDryRunRevisionState_SQLServerIntegration' -timeout 3m
           go test -v -count=1 ./migration/generator -run 'TestGenerateMigration_SQLServerIndexDirectionRoundTrip|TestShadowIdentifierSemanticsMatch_SQLServerLive' -timeout 5m
 
+      # The reverse plan for views, materialized views and triggers is decided
+      # by what PostgreSQL accepts, and rendering the rollback proves only that
+      # a statement exists. This applies it. The test is behind the integration
+      # build tag in ./migration/generator, which no other step in this file
+      # compiles, so without this step the only live gate for the rollback would
+      # run nowhere.
+      - name: Run reverse view-like rollback round trip
+        env:
+          POSTGRES_URL: "postgres://ptah_user:ptah_password@localhost:5432/ptah_test?sslmode=disable"
+        run: |
+          go test -tags=integration ./migration/generator -list '^TestReverseViewLikeObjects_DownRoundTrip_Integration$' | grep -q '^TestReverseViewLikeObjects_DownRoundTrip_Integration$'
+          go test -v -count=1 -tags=integration ./migration/generator -run '^TestReverseViewLikeObjects_DownRoundTrip_Integration$' -timeout 5m
+
       - name: Run database-realm cleanup and replay tests
         env:
           POSTGRES_TEST_DSN: "postgres://ptah_user:ptah_password@localhost:5432/ptah_test?sslmode=disable"
--- a/docs/public_api.md
+++ b/docs/public_api.md
@@ -278,6 +278,18 @@ table. Live comparisons also snapshot catalog identifier semantics into the
 diff so comparison, destructive-change policy, forward planning, and reverse
 planning use one source of truth.
 
+`migration/schemadiff/types.ViewDiff` also records the view body that is in
+force before the diff is applied, and whether the entry is being planned as a
+rollback. Planners read the body to decide whether the target engine accepts an
+in-place view replacement; PostgreSQL accepts `CREATE OR REPLACE VIEW` only when
+the new query appends to the old column list over the same relations. Where that
+can be neither proved nor ruled out — an unknown prior body, a `WITH` prefix, a
+`SELECT *` projection, a top-level set operation — the rollback flag settles it:
+a forward plan keeps the replacement, which preserves dependent objects and the
+privileges on the view and fails loudly if the engine refuses it, while a
+rollback drops and recreates, which always applies. Embedders that build a
+`ViewDiff` by hand and leave both fields empty get the forward answer.
+
 `core/platform/identifier` exposes the reusable value types and conservative
 dialect defaults behind that contract.
 
--- a/docs/public_api.snapshot
+++ b/docs/public_api.snapshot
@@ -9231,6 +9231,43 @@ package types // import "go.5x5.cz/ptah/migration/schemadiff/types"
 type ViewDiff struct {
     ViewName string            `json:"view_name"`
     Changes  map[string]string `json:"changes"`
+
+    // PreviousBody is the view body that is in force before this diff is
+    // applied: the database side of a forward comparison, and the post-up side
+    // of a reversed one. It is what makes the modification path decidable.
+    //
+    // PostgreSQL accepts CREATE OR REPLACE VIEW only when the new query yields
+    // the old column list with columns appended to the end -- same names, same
+    // types, same order. Dropping, renaming or retyping a column is refused at
+    // execution time, and a down migration built with CREATE OR REPLACE is
+    // therefore un-appliable for every view modification except the one shape
+    // the up migration is least likely to have made. A planner that cannot see
+    // the prior body cannot tell the legal case from the refused one, so it has
+    // to choose between an un-appliable statement and always dropping the view
+    // with CASCADE, which takes dependents with it.
+    //
+    // An empty value means "not known". It does not on its own select a plan:
+    // what a planner cannot decide is settled by Rollback below, so a forward
+    // plan still attempts the replace -- the engine refuses it if it is illegal,
+    // and refusing costs nothing -- while a rollback takes the drop and
+    // recreate that always applies.
+    PreviousBody string `json:"previous_body,omitempty"`
+
+    // Rollback reports that this entry is being planned in the DOWN direction:
+    // the statement rendered for it runs while an operator is undoing a
+    // migration.
+    //
+    // It decides nothing on its own. It settles the cases where a planner cannot
+    // prove whether the engine accepts an in-place replace, and the two
+    // directions want opposite answers there. Going forward, the replace is
+    // worth attempting: it keeps dependent objects and the privileges granted on
+    // the view, and if the engine refuses it the migration stops with nothing
+    // destroyed. A rollback cannot be stopped that way -- it is already running
+    // during the incident -- so it takes the drop-and-recreate that always
+    // applies, and pays for it by rebuilding what CASCADE removes.
+    //
+    // Reverse plan builders set it; a forward comparison leaves it false.
+    Rollback bool `json:"rollback,omitempty"`
 }
     ViewDiff represents changes to a view definition.
 
--- a/internal/deporder/deporder.go
+++ b/internal/deporder/deporder.go
@@ -592,6 +592,22 @@ func viewLikeID(object ViewLike, index int) string {
 // ReferencesIdentifier reports whether the SQL body references name as a
 // standalone or qualified identifier. It backs view-like dependency ordering
 // and the schema-scope cross-reference diagnostics.
+//
+// Only the code of the body counts. A name that appears inside a string
+// literal, a dollar-quoted string, a line comment or a block comment names
+// nothing, and treating it as a reference is not a harmless over-approximation:
+// the PostgreSQL planner asks this question to work out what
+// DROP VIEW ... CASCADE takes with it, and an answer of "yes" there puts a
+// DROP MATERIALIZED VIEW ... CASCADE into the plan. Measured on PostgreSQL
+// 17.10, a materialized view whose body is
+// "SELECT 'base_view' AS label, count(*) AS total FROM accounts" -- which reads
+// no view at all -- was dropped and recreated when base_view was modified,
+// taking a hand-made dependent view, a unique index on the materialized view
+// and the privileges granted on it, none of which the plan rebuilds.
+//
+// Quoted identifiers are code and still match, because "base_view" is a
+// reference to base_view. Their contents are opaque, so a quotation mark or a
+// comment marker inside one cannot open a literal or a comment.
 func ReferencesIdentifier(body, name string) bool {
 	body = strings.ToLower(body)
 	name = strings.ToLower(strings.TrimSpace(name))
@@ -599,6 +615,7 @@ func ReferencesIdentifier(body, name string) bool {
 		return false
 	}
 
+	code := sqlCodeMask(body)
 	for start := 0; start < len(body); {
 		index := strings.Index(body[start:], name)
 		if index < 0 {
@@ -606,12 +623,153 @@ func ReferencesIdentifier(body, name string) bool {
 		}
 		index += start
 		end := index + len(name)
-		if (isIdentifierBoundary(body, index-1) || isQualifiedIdentifierTail(body, index-1)) && isIdentifierBoundary(body, end) {
+		if spanIsCode(code, index, end) &&
+			(isIdentifierBoundary(body, index-1) || isQualifiedIdentifierTail(body, index-1)) &&
+			isIdentifierBoundary(body, end) {
 			return true
 		}
 		start = end
 	}
 	return false
+}
+
+// spanIsCode reports whether every byte of body[start:end] is SQL code.
+func spanIsCode(code []bool, start, end int) bool {
+	for i := start; i < end; i++ {
+		if !code[i] {
+			return false
+		}
+	}
+	return true
+}
+
+// sqlCodeMask marks the byte offsets of body that are SQL code: everything
+// outside string literals, dollar-quoted strings, line comments and block
+// comments. Quoted identifiers are code, but their contents are copied through
+// without being re-scanned, so "a--b" is one identifier rather than the start of
+// a comment.
+//
+// The mask is computed over the same string the caller searches, so no offset
+// can slide out from under it.
+func sqlCodeMask(body string) []bool {
+	mask := make([]bool, len(body))
+	for i := 0; i < len(body); {
+		switch {
+		case strings.HasPrefix(body[i:], "--"):
+			i = skipLineComment(body, i)
+		case strings.HasPrefix(body[i:], "/*"):
+			i = skipBlockComment(body, i)
+		case body[i] == '\'':
+			i = skipStringLiteral(body, i)
+		case body[i] == '"':
+			end := skipQuotedIdentifier(body, i)
+			for ; i < end; i++ {
+				mask[i] = true
+			}
+		case body[i] == '$':
+			end, ok := skipDollarQuoted(body, i)
+			if !ok {
+				mask[i] = true
+				i++
+				continue
+			}
+			i = end
+		default:
+			mask[i] = true
+			i++
+		}
+	}
+	return mask
+}
+
+func skipLineComment(body string, start int) int {
+	if end := strings.IndexByte(body[start:], '\n'); end >= 0 {
+		return start + end + 1
+	}
+	return len(body)
+}
+
+// skipBlockComment consumes a /* ... */ comment. PostgreSQL nests them, so the
+// scan counts depth rather than stopping at the first close.
+func skipBlockComment(body string, start int) int {
+	depth := 0
+	for i := start; i < len(body); {
+		switch {
+		case strings.HasPrefix(body[i:], "/*"):
+			depth++
+			i += 2
+		case strings.HasPrefix(body[i:], "*/"):
+			depth--
+			i += 2
+			if depth == 0 {
+				return i
+			}
+		default:
+			i++
+		}
+	}
+	return len(body)
+}
+
+// skipStringLiteral consumes a '...' literal. A doubled quote continues it, and
+// a backslash escapes the next byte only in a literal introduced by E, which is
+// the one spelling where PostgreSQL reads backslashes as escapes.
+func skipStringLiteral(body string, start int) int {
+	escapes := start > 0 && body[start-1] == 'e' &&
+		(start == 1 || !isSQLIdentifierRune(rune(body[start-2])))
+	for i := start + 1; i < len(body); {
+		switch {
+		case escapes && body[i] == '\\' && i+1 < len(body):
+			i += 2
+		case body[i] == '\'' && i+1 < len(body) && body[i+1] == '\'':
+			i += 2
+		case body[i] == '\'':
+			return i + 1
+		default:
+			i++
+		}
+	}
+	return len(body)
+}
+
+// skipQuotedIdentifier consumes a "..." identifier, including the doubled quote
+// that spells a quotation mark inside one.
+func skipQuotedIdentifier(body string, start int) int {
+	for i := start + 1; i < len(body); {
+		switch {
+		case body[i] == '"' && i+1 < len(body) && body[i+1] == '"':
+			i += 2
+		case body[i] == '"':
+			return i + 1
+		default:
+			i++
+		}
+	}
+	return len(body)
+}
+
+// skipDollarQuoted consumes a $tag$ ... $tag$ string, reporting false when the
+// dollar sign does not open one -- a positional parameter such as $1, or a
+// dollar sign inside an identifier.
+func skipDollarQuoted(body string, start int) (int, bool) {
+	end := start + 1
+	for end < len(body) {
+		r, size := utf8.DecodeRuneInString(body[end:])
+		if r == '_' || unicode.IsLetter(r) || (end > start+1 && unicode.IsDigit(r)) {
+			end += size
+			continue
+		}
+		break
+	}
+	if end >= len(body) || body[end] != '$' {
+		return start, false
+	}
+
+	tag := body[start : end+1]
+	if closing := strings.Index(body[end+1:], tag); closing >= 0 {
+		return end + 1 + closing + len(tag), true
+	}
+	return len(body), true
 }
 
 func isIdentifierBoundary(value string, index int) bool {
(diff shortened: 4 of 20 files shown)


denisvmedia:

**Second verdict: still refuted.** Details are being folded in; do not merge yet.

denisvmedia (commented):
Independent review of head `9b504921` found destructive PostgreSQL planning blockers.

**Blocker — raw text dependency matching authorizes destructive DDL.** `internal/planner/dialects/postgres/postgres.go:1941` uses `deporder.ReferencesIdentifier`, which sees identifiers inside literals/comments. An unrelated materialized view containing `SELECT 'base_view' AS label` can be classified as dependent and dropped/recreated. Use SQL tokenization that excludes literals/comments and add negative tests.

**Blocker — quoted relation identity is lowercased away.** `internal/planner/dialects/postgres/view_replace.go:308` lowercases the entire expression, so PostgreSQL-distinct `"Foo"` and `"foo"` normalize together; a relation swap can be misclassified as replace-compatible and generate an unapplyable rollback.

The public godoc at `migration/schemadiff/types/types.go:693-695` also contradicts the direction-aware implementation, and the README's “unpopulated until refresh” claim is false because generated `CREATE MATERIALIZED VIEW` lacks `WITH NO DATA`. Fix the destructive predicates, add guards, and correct the contracts before merge.

denisvmedia:
Both blockers reproduced. Fixed in 04de0266 on this branch, with the shapes you named turned into fixtures and both contract sentences rewritten. Details, point by point.

**Blocker 1 — raw text dependency matching authorizes destructive DDL.** Confirmed, and worse than a redundant statement, because everything on that rebuild list is *answered with* a statement. Built your shape: declared `base_view` modified onto the drop path, plus a materialized view `label_stats` whose body is `SELECT 'base_view' AS label, count(*) AS total FROM accounts`, plus a comment-only variant. Both joined the list; the plan for the **forward** direction contained `DROP MATERIALIZED VIEW IF EXISTS "label_stats" CASCADE`. Applied on my own PostgreSQL 17.10 (container `ptahrev1297-dev`, port 15997, `ptah_user LOGIN SUPERUSER`, fresh database, `psql -v ON_ERROR_STOP=1 -f`, container removed at the end), against a fixture that also held a unique index on the matview, a hand-made view reading it, and `GRANT SELECT ... TO ptah_reader`:

```
pg_class before: accounts r, base_view v, hand_made v, label_stats m, label_stats_label i, label_view v, users r
apply rc=0       NOTICE: drop cascades to view hand_made
pg_class after:  accounts r, base_view v, label_stats m, label_view v, users r
relacl of label_stats after: NULL
```

Three objects gone, none of them declared, none of them named by any statement in the plan.

Fixed as you suggested, in the shared predicate rather than at the call site: `deporder.ReferencesIdentifier` now builds a code mask and matches only outside string literals (honoring `''` and, for an E-literal, backslash escapes), dollar-quoted strings, line comments and **nested** block comments. Quoted identifiers stay matchable, because `"base_view"` is a reference to `base_view`, but their contents are opaque so a `--` inside one cannot open a comment. The mask is computed over the same lower-cased string that is searched, so no offset can slide. After the fix the plan for that fixture is exactly `DROP VIEW IF EXISTS "base_view" CASCADE; CREATE VIEW "base_view" ...` — and re-seeded live, all three objects survive and the GRANT reads back intact.

I fixed the shared function deliberately rather than adding a planner-local helper, because the same false positive was in `ViewLikesForCreate`'s ordering and in `internal/atlasfilter`'s scope diagnostics, where it produced a false *refusal* for a projection over a name that only appeared in a literal.

Negative tests added, as asked: `TestReferencesIdentifier_ReadsCodeOnly` is 13 rows — literal, doubled quote, E-literal escape, line comment, block comment, nested block comment, dollar-quoted body, plus the positives that stop the fix from becoming "answer no to everything" (plain reference, quoted reference, code after a closed literal, code after a closed comment, `$1` is not a dollar quote, a `--` inside a quoted identifier). Plus `TestViewLikesForCreate_IgnoresNamesInsideLiteralsAndComments`, `TestPlanner_GenerateMigrationAST_CascadeRebuildReadsCodeNotText` (which also asserts the *genuine* dependents are still rebuilt), and a live row in `TestReverseViewLikeObjects_DownRoundTrip_Integration`.

**Blocker 2 — quoted relation identity is lowercased away.** Confirmed, and it produces exactly the un-appliable rollback this PR exists to prevent. `ViewDiff{PreviousBody: 'SELECT id FROM "Foo"', Rollback: true}` with target `SELECT id FROM "foo"` rendered `CREATE OR REPLACE VIEW "probe_v" AS SELECT id FROM "foo"`. Against `"Foo"(id bigint)` / `"foo"(id text)`:

```
psql -v ON_ERROR_STOP=1 -f  ->  rc=3
ERROR:  cannot change data type of view column "id" from bigint to text
```

After the fix the same diff renders `DROP VIEW IF EXISTS "probe_v" CASCADE; CREATE VIEW ...`, rc=0, and `pg_get_viewdef` converges. `normalizeExpression` now folds case everywhere except inside quoted identifiers, whose contents are copied byte for byte — `"my view"` and `"my  view"` are two relations as well. A string literal is skipped as a unit so a quotation mark inside one cannot open an identifier, but its own case still folds, because a literal decides no column's type; there is a row pinning that. `TestPlanner_GenerateMigrationAST_QuotedRelationCaseIsNotFolded` has five rows: the swap, the same quoted relation appending, the same quoted relation with a predicate-only edit, "case still folds outside quotes", and the literal-with-a-quote case.

One thing I could not do, and I would rather say it than let it look covered. I tried to drive this shape end to end from a Go schema through the comparator — two tables differing only in quoted case, a view moved between them, seeded live so both relations genuinely exist in `pg_class` — and `schemadiff.CompareWithDialect(...).HasChanges()` came back **false**. `normalizeSQLCaseAndIdentifierQuotes` folds quoted-identifier case before the bodies are compared, so the comparator sees no change at all. So the defect you found is real and fixed at the layer you cited, reachable through the reverse builder and through the exported planner, and I applied its rendered SQL live — but I removed the live table row rather than leave one that passes for the wrong reason. That comparator blindness is a separate and arguably worse bug (a schema edit that retypes a view's columns reports "no changes"); it needs a change to a normalizer every object kind shares, so it is filed rather than folded in here.

I also left `"foo"` vs `foo` answering "changed" — they *are* the same relation, so that costs a drop where a replace would have worked. It is pre-existing (the old code compared `"foo"` against unquoted `foo` too) and I did not "improve" it, because every textual canonicalization that equates them also equates `"join"` with the `JOIN` keyword. Both the doc comment and the README now say so explicitly instead of leaving it implied.

**The `types.go` godoc.** You are right that it contradicted the code. It said an empty `PreviousBody` means planners "must then take the conservative drop-and-recreate path", while the implementation resolves undecidable bodies by direction and keeps the replace going forward. Rewritten: an empty value means "not known", it selects no plan on its own, and `Rollback` settles it. `docs/public_api.snapshot` regenerated; `check-public-api-released.sh` still 0.

**The README's "unpopulated until refresh".** Also right, and I checked rather than reasoned: there is no `WITH NO DATA` anywhere in the tree, and after the generated `CREATE MATERIALIZED VIEW` ran, `relispopulated` was `t` with 3 rows in it. The limitation now says the definition is recreated and PostgreSQL populates it as part of the `CREATE`, so the contents are recomputed from the base tables at rollback time rather than restored — which means a snapshot deliberately left stale is silently refreshed and the rollback pays for the whole query — and that the indexes and privileges attached to the old materialized view go with the drop.

**Re-measurement of what the PR already claimed.** The legality table is unchanged: 9 rows of `ModifiedViewDropsWhenReplaceWouldBeRefused` and 8 of `UndecidableViewBodyFollowsTheDirection` still green, and the live round trip is 8/8 (the 7 rows that existed plus the new literal case), exit 0.

**Mutants, both directions.** Reverting the code mask reddens 7 `ReferencesIdentifier` rows, the ordering test, the planner test and the live row while every original test stays green; over-correcting it to "nothing is a reference" reddens your `DroppedViewRebuildsDeclaredDependents`, `ModifiedViewsDropInDependencyOrder`, `OrdersViewLikeObjectsByDependencies` and the atlasfilter diagnostic. Reverting the case fix reddens exactly the quoted-swap row; folding no case at all reddens exactly the "still folds outside quotes" row — and honestly, that mutant leaves the original rows green, because their fixtures are already lower case, which is why that row exists.

**Still open from the previous round, not addressed here:** item 7. MySQL, MariaDB, SQLite, CockroachDB and YugabyteDB were not measured live. Nothing in their view planner paths changed and the `deporder` change can only shrink a dependency set, so it cannot introduce a drop anywhere — but that is an argument, not a measurement, and this repo has been bitten there before.

All gates read directly, never after a pipe: build 0, vet 0, vet -tags integration ./integration/... 0, vet -tags integration ./migration/generator 0, `go test ./... -count=1` 0 (255 packages, no kills), qtlint 0, golangci-lint 0, check-test-style 0, check-public-api 0, snapshot 0, released 0. The shared build cache on this machine is being purged by other jobs — `go build` and golangci-lint both failed with missing cache files at first — so every Go gate was re-run with a private `GOCACHE`, and golangci-lint with a private `GOLANGCI_LINT_CACHE`. golangci-lint was genuinely red once, on `gci`, because a doc comment contained `E''` and gofmt rewrites that to a typographic quote; the comment was reworded.

Issue #1294, opened 2026-08-07
Inspect describes an extension-owned composite, domain or range type as a user type, and the document then fails to replay


denisvmedia:
`ptah-compat schema inspect` describes a domain, composite or range type that an extension owns as if a user had defined it. Replaying the document then fails, because `CREATE EXTENSION` has already created the type.

Found while measuring something else on #1281; **pre-existing**, reproduced on current master and on 4fd9f18a alike.

### Measured, PostgreSQL 17

A fixture extension supplying a composite type:

```
-- kw1281ret--1.0.sql
CREATE TYPE kwbox AS (lo integer, hi integer);
CREATE FUNCTION merge(a text, b text) RETURNS kwbox ...
```

```sql
CREATE EXTENSION kw1281ret;
CREATE TABLE names (id integer PRIMARY KEY, a text NOT NULL, b text NOT NULL,
  CONSTRAINT ck CHECK ((merge(a, b)).lo >= 0));
```

The rendered document contains both:

```
extension "kw1281ret" { ... }
composite "kwbox" { ... }
```

and reading it back against a clean dev database:

```
Error: materialize schema on dev database: ERROR: type "kwbox" already exists (SQLSTATE 42710)
```

The catalog is unambiguous about ownership — `pg_depend` with `deptype = 'e'` and `classid = 'pg_type'::regclass` has the row, so `kwbox owned_by_extension=true`.

### Scope

`internal/dbschema/postgres/reader.go` describes three type kinds and none of the three queries excludes extension-owned types:

| function | filter |
|---|---|
| `readDomains` | `WHERE t.typtype = 'd' AND n.nspname = $1` |
| `readComposites` | `WHERE t.typtype = 'c' AND n.nspname = $1` |
| `readRanges` | `WHERE t.typtype = 'r' AND n.nspname = $1` |

Base types are unaffected, which is why this has not been hit by the common contrib extensions: `hstore`, `citext`, `cube` and `ltree` all supply `typtype = 'b'`, and a document naming an `hstore` column carries no type block for it. Measured: the `hstore` fixture declares zero composite/domain/range blocks for `hstore`.

I have **not** surveyed which shipped extensions supply a composite, domain or range type. That survey is the first part of the work; PostGIS is the obvious candidate to check, and the fixture above proves the mechanism regardless.

### Why it belongs with #1276

[...]

PR #1298, opened 2026-08-08
Let a description say what it does not describe (#1276)

denisvmedia:
Item 2 of #1276, the one #1267, #1264 and #1294 are all waiting on.

The comparator had two states for an object — present and absent — and every scope change since #1266 has turned the second one into a lie. A reader that was not asked about something was indistinguishable from a reader that looked and found nothing, and that single conflation is #1273, #1275, the destructive `DROP EXTENSION` in this issue's first comment, and the chain in #1294's second comment.

## The representation

New package `core/coverage`. A `coverage.Set` names what a description does **not** claim to describe. Three states result:

```
PRESENT                 the object is in the description
AUTHORITATIVELY ABSENT  not in the description, and the description covers it   -> plan it
NOT DESCRIBED           not in the description, and it never covered it         -> plan nothing
```

The zero `Set` claims everything, so a description that says nothing about its own limits is fully authoritative. That is what every hand-authored schema file is, and it is why adding coverage changes no existing plan.

`goschema.Database.NotDescribed` and `dbschema/types.DBSchema.NotDescribed` each carry one, and the two gate **opposite directions**:

- the **desired state's** record gates **removals** — an object the database has and a document does not name is a removal only when that document claimed to describe such objects;
- the **read's** record gates **additions** — an object a document names and a read did not report is a creation only when that read looked.

`TestCoverageOnTheWrongSideSuppressesNothing` pins the asymmetry in both directions.

## Why it is on the state and not on the comparison options

This issue's last status comment records the earlier attempt losing `desiredState.NotDescribed` in `internal/atlasmigrate/diff.go`, because the compare options there were built from a zero value. That is not a slip, it is what happens when a fact about the data rides on a struct four surfaces each construct for themselves.

On the state it travels with the data. **`internal/atlasmigrate` needed no change at all** — which was measured rather than assumed, one row per surface, because "the four surfaces resolve a desired state separately" is this issue's own thesis.

## Surviving the process boundary

`schema inspect > file` and `schema apply --to file://file` are two processes, and the second reads the file rather than the terminal, so a stderr warning cannot carry this. The record therefore goes in the document's leading comment header:

```hcl
// Code generated by ptah; DO NOT EDIT.
// ptah:not-described extension
// ptah:not-described policy
// ptah:not-described sequence
```

Only the **leading** comment header is read. A directive recognized anywhere in the file would let a table comment or a heredoc line suppress a removal the author asked for, which is a silent destructive false negative in the one direction this must never fail. A kind the build does not know is an error rather than a line to skip past, for the same reason: a record nothing understands reads as no record, and the absence it was protecting becomes a removal.

The same grammar is read out of the leading `--` comments of a `.sql` desired state. No Ptah surface writes one there, because only the HCL rendering omits blocks.

## The claim is about the rule, not about the findings

The compatibility surface writes an `extension` block only when something else in the document names it. So for any extension name **absent** from the document, a reader genuinely cannot tell "the database does not have it" from "nothing named it" — and that is what a whole-kind record says.

Recording only the objects a given run omitted would be an **under-claim**: it would assert that the absence of every *other* extension is authoritative, which is false the moment the same document is applied to a different database. Under-claiming is the destructive direction, so the header is written unconditionally in omit mode, including for a database holding none of these objects at all. That row is in `TestCompatibilityRenderDeclaresWhatItDoesNotDescribe`, and it is the only thing that reddens when the renderer is mutated to claim only what it found — the boundary guard stays green for that mutant.

`PTAH_ATLAS_INSPECT_ALL_BLOCKS=1` writes no header, because a document that omits nothing describes everything. Native `ptah schema inspect` never writes one.

## Measured

PostgreSQL 17.10, connected as `ptah_user`. Fixture: an unreferenced `pgcrypto`, a standalone `order_seq`, and an RLS policy `p` on a table. `ptah-compat schema inspect` of that database, applied back to that same database, separate clean dev database recreated between runs, exit status read on its own line.

| surface | before | after |
| --- | --- | --- |
| `schema diff` | `DROP POLICY`, `DROP SEQUENCE`, `DROP EXTENSION` | `Schemas are synced` |
| `schema apply --dry-run` | the same three | `Schema is synced` |
| `migrate diff --dry-run` | the same three | synced |
| `migrate diff` writing its migration file | the same three | synced, no file written |

The fourth row is the one that matters most: a `--dry-run` a person reads is not a migration file a pipeline applies.

### A suppression that can never drop is a different bug

Same database, same document, one difference — the three header lines deleted, which is the edit an operator makes when they *mean* the omission:

```console
$ grep -v 'ptah:not-described' schema.hcl > desired.hcl
$ ptah-compat schema apply --url "$PG_URL" --to file://desired.hcl --dev-url "$DEV_URL" --dry-run
DROP POLICY IF EXISTS "p" ON "guarded";
DROP SEQUENCE IF EXISTS "order_seq";
DROP EXTENSION IF EXISTS "pgcrypto";
```

That pair is one integration test over one database: `TestPostgreSQLCoverageStillPlansAGenuineRemovalIntegration`.

### The pinned binary cannot see the header

Pinned Atlas community v1.3.0, `schema inspect -u file://…` against a clean dev database recreated before each run:

| document | result |
| --- | --- |
| before, no header | exit 0 |
| after, three header lines | exit 0, and `diff` against the before output is empty |

### A capability is not lost

The header gates removals only. A document that keeps a referenced `extension "isn"` block, applied to a fresh database, still plans `CREATE EXTENSION IF NOT EXISTS "isn" VERSION '1.2'` and its table — additions are gated by the *read's* record, and a live read leaves that zero.

## The boundary guard

`extension_nothing_references` / `property 2 … compatibility surface` asserted `DROP EXTENSION IF EXISTS "pgcrypto"` with a `WRONG (#1276)` comment saying it must become nil. It is nil. The row is retagged `FIXED` and carries the four-surface measurement rather than being quietly edited; nothing was relaxed, and every other row is byte-unchanged. `boundaryDropExtension` is kept, unused by any case, because the genuine-removal control still has to be able to spell it.

Full `integration/gonative` on PostgreSQL: 44 PASS, 22 SKIP (no MySQL/MariaDB/ClickHouse/SQL Server DSN), 0 FAIL.

## Mutants

Seven, each restored, each with a named green control in the same sweep:

| mutant | reddens |
| --- | --- |
| extension coverage filter removed | 2 unit rows, 2 live rows |
| `Describes` always false (over-correct) | every control row, plus 8 pre-existing `schemadiff` suites |
| header never written | 3 unit rows, 4 live rows |
| parser discards the header | 1 unit row, 2 live rows — and **not** the loader's own package, which is the point |
| `NotDescribed` dropped from `cloneDatabase` | `TestScopeDatabaseKeepsCoverage`, 3 rows |
| decoder reads any line, not only the header | 2 decode rows |
| renderer claims only what it found | the empty-database row only |

The sixth found a defect in a test I had just written: the row I first used ("a directive inside a string value") did not redden, because a quoted attribute does not *begin* with the marker. Replaced with a heredoc body, which does, and re-measured.

## Also here

- `.gitignore` line 22 was a bare, unanchored `coverage`, matching any path segment of that name anywhere. It ignored the new package, and it never matched what `make coverage` writes (`coverage.out`, covered by `*.out`, and `coverage.html`). Anchored.
- `internal/atlasfilter.cloneDatabase` is a field-by-field constructor — the shape a new field is dropped from in silence — so the carry has its own test.

## What this does not close

- **#1267 / #1273 (roles)**: the comparator half is done and tested. The reader must record `WithObject(coverage.Role, name)` per role its scoped query excluded — per-object, not whole-kind, or `CREATE ROLE` for a role that genuinely does not exist would stop too.
- **#1264 / #1275 (schemas)**: likewise. The reader must enumerate the schemas it did *not* read. Note that `CREATE SCHEMA` is derived by the planner from the desired IR rather than from a diff field, so that path wants checking separately.
- **#1294 (extension-owned types)**: the constraint its second comment sets out is now expressible. The reader records the excluded domain as not described, and `collectReferencedNames` treats "the document names an object no one describes because an extension owns it" as a reason to keep that extension's block. An extension-owned type is describable **or** its extension is present, never neither.

## Not verified

PostgreSQL only — no other dialect was measured live, though `atlasRefusedBlockTypes` has only a PostgreSQL row so nothing else can emit a header today. `schema plan` was not measured (it reaches the same `computeApplyPlan`). No conformance run; no flags were added. Neither held branch was built.

---

### What it found

Two failures, one of them fatal to the front's headline claim.

SETUP. My own worktrees at 04b3732e (head) and 9820496e (merge-base with origin/master, control), `git log --oneline -1` checked in each before building; binaries rebuilt in each and chmod +x. My own container `ptahref1276` on port 15776 (postgres:17), `CREATE ROLE ptah_user LOGIN PASSWORD 'ptah_password' SUPERUSER; CREATE DATABASE ptah_test OWNER ptah_user;`. My own fixture, not theirs: `CREATE EXTENSION pgcrypto; CREATE SEQUENCE ticket_seq; CREATE TABLE ticket (id integer PRIMARY KEY, owner text NOT NULL); ALTER TABLE ticket ENABLE ROW LEVEL SECURITY; CREATE POLICY ticket_read ON ticket FOR SELECT USING (true);`. Separate dev database recreated between every run. Every exit code read on its own line. Container and all databases removed at the end.

=== FINDING 1 (FATAL): the destructive round trip is NOT closed. `schema inspect` has a second, documented output mode that drops the record, and I executed the drop for real. ===

`ptah-compat schema inspect --help` documents split/write export as a first-class feature ("Split/write exports support the documented Atlas split strategies -- per object (default), `split \"schema\"`, and `split \"type\"` ... through `{{ hcl . | split | write \"dir\" }}`"). That mode applies the SAME block suppression and writes NO `ptah:not-described` record anywhere.

  $ ptah-compat schema inspect --url "$SRC" --format '{{ hcl . | split "schema" | write "out" }}'
  rc=0
  # stderr carries the identical four suppression warnings:
  #   warning: extensions.pgcrypto: omitted from Atlas-compatible schema inspect output: ...
  #   warning: sequences.ticket_seq: omitted ...
  #   warning: rls_policies.ticket_read: omitted ...
  $ grep -rn 'ptah:not-described' out
  grep_rc=1        # nothing, in any of the written files
  $ grep -c 'extension "pgcrypto"' out/public.hcl   -> 0    (suppression IS active)
  # control, same command, same binary: PTAH_ATLAS_INSPECT_ALL_BLOCKS=1 -> grep -c 'extension "pgcrypto"' = 1

Isolation of where the record is lost -- same binary, same database, one pipe stage apart:
  --format '{{ hcl . }}'            -> header present (3 directive lines)
  --format '{{ hcl . | split }}'    -> header absent from every archive member

Applying that export back to the database it came from, on HEAD 04b3732e:
  $ ptah-compat schema apply --url "$SRC" --to "file://out/public.hcl" --dev-url "$DEV" --dry-run
  rc=0
  DROP POLICY IF EXISTS "ticket_read" ON "ticket";
  DROP SEQUENCE IF EXISTS "ticket_seq";
  DROP EXTENSION IF EXISTS "pgcrypto";
Base 9820496e, same document, same command: byte-identical output. The change moves this row not at all.

Not a dry-run artifact -- I ran it for real against a throwaway copy (`nd_src2`, same fixture):
  $ ptah-compat schema apply --url "$SRC2" --to "file://out/public.hcl" --dev-url "$DEV" --auto-approve
  rc=0   "Schema apply completed successfully."
  $ psql -tA -c "select 'table:'||relname from pg_class where relname='ticket';
                 select 'ext_count:'||count(*) from pg_extension where extname='pgcrypto';
                 select 'seq_count:'||count(*) from pg_class where relkind='S';
                 select 'pol_count:'||count(*) from pg_policy;"
  table:ticket
  ext_count:0
  seq_count:0
  pol_count:0

The extension, the sequence and the policy were destroyed by inspecting a database and applying its own output back to it -- the exact sentence #1276 opens with, on HEAD, on the compatibility surface, one `--format` away from the path they measured.

This also falsifies the front's central reasoning as prose. The commit message and docs/site/src/content/docs/atlas/schema-commands.md both assert the rule unconditionally: "The compatibility surface claims the RULE it applies rather than what it found: a block is omitted whenever nothing names it, so absence carries no information about any name" / "the header is written unconditionally in omit mode". It is written unconditionally in ONE of the surface's two HCL output modes. The other mode omits the same blocks and claims nothing, which is precisely the under-claim the same paragraph calls "the destructive direction".

Code: the split archive is rebuilt from parsed blocks in `atlasSchemaInspectSplit` / `splitSchemaInspectHCL` (internal/atlasreport/schema_inspect_files.go); leading comments do not travel with any block, and nothing re-prefixes them. `atlashclrender.Result.NotDescribed` -- which this PR added -- is available at that call site and is not consulted.

=== FINDING 2: the additive gate silently deletes a change the author explicitly wrote, with no escape hatch and no diagnostic. ===

`schemafile.ToDBSchema` now sets `NotDescribed` on the CURRENT side when the current state is a file, and `Coverage.PlansAddition` gates creations on it. So a compat-inspected document used as `--from` -- the documented `atlas schema inspect > schema.hcl` workflow, checked into git -- suppresses creations of those kinds.

Hand-authored desired document that explicitly declares `extension "citext" { if_not_exists = true }`:

  BASE 9820496e, from the header-carrying document:
  $ ptah-compat schema diff --from "file://head.hcl" --to "file://desired-ext.hcl" --dev-url "$DEV"
  rc=0
  CREATE EXTENSION IF NOT EXISTS "citext";

  HEAD 04b3732e, same two files:
  rc=0
  Schemas are synced, no changes to be made.

  HEAD 04b3732e, control -- the ONLY difference is `grep -v 'ptah:not-described'` on the --from file:
  rc=0
  CREATE EXTENSION IF NOT EXISTS "citext";

With a sequence added too, base emits `CREATE EXTENSION IF NOT EXISTS "citext"; CREATE SEQUENCE "new_seq" AS bigint ...` and head emits nothing. Same shape as a drift check: `schema diff --from file://head.hcl --to $SRC`, where the live database genuinely holds pgcrypto, ticket_seq and policy ticket_read, prints "Schemas are synced, no changes to be made." at rc=0 on head and the three CREATEs on base.

Why this is a defect and not conservatism: (a)

### What it says must change

The representation (core/coverage, the record on the STATE rather than the options) is right and the four surfaces they measured really are fixed. Two things must land before this can be described as closing the destructive round trip.

1. Carry the record through split/write, or refuse to split in omit mode.
   `internal/atlasreport/schema_inspect_files.go`: `atlasSchemaInspectSplit` rebuilds each archive member from a parsed block and drops the document's leading comment header. `atlashclrender.Result.NotDescribed` is already produced by this PR and is reachable from `SchemaInspectReport.renderHCL`; prefix `Set.Directives()` onto every written HCL member (every member, not the first -- the loader reads whichever file it is handed, and `schemafile.appendDatabase` already unions the records correctly when several are loaded). Fixture: my `split "schema"` case above -- the guard row is "inspect with `--format '{{ hcl . | split \"schema\" | write \"out\" }}'`, apply out/public.hcl back, expect no plan", with the header-stripped variant as the genuine-removal control. Note `splitModeObject` and `splitModeType` need the same treatment; the per-object export writes eleven files and none of them carries a directive today. Mutant to run: emit the directives on only the first member and confirm the object-split row reddens.

2. Stop the current-side record from silently deleting an explicitly requested creation.
   `Coverage.PlansAddition` (migration/schemadiff/internal/compare/coverage.go) is the right hook, but "the current state did not look" is not the same fact as "the desired state did not ask". When the desired state NAMES the object, the honest plan is the idempotent creation (`CREATE EXTENSION IF NOT EXISTS`, which the generator already emits) plus a diagnostic, not silence plus "Schemas are synced". At minimum the additive gate must (a) not apply where the emitted statement is already idempotent, (b) print what it withheld on stderr rather than nothing, and (c) if it is kept in its current form, be reachable in the old shape behind a PTAH_* variable and written down in docs/site/src/content/docs/atlas/schema-commands.md -- compatibility policy (c). Fixture: `schema diff --from file://<compat-inspected doc> --to file://<doc declaring extension "citext">`, with the header-stripped `--from` as the control; today the two disagree and only the control produces the statement.

3. Prose, once 1 and 2 land: the commit message's "a block is omitted whenever nothing names it, so absence carries no information about any name" and the schema-commands.md section "The document says what it does not describe" both read as unconditional claims about the surface. They are true only of the default single-document output. Either the split path starts carrying the record (preferred) or both sentences must name the output modes they hold for. The same section should gain the sentence that is missing entirely: what the record does when such a document is used as the CURRENT state.

4. Minor: either wire `Coverage.PlansRemoval(coverage.Role, ...)` in compare/roles.go or say in the `coverage.Role` doc comment that the kind is additive-only, so the #1273 follow-up does not inherit a directive that reads as protection in a direction it does not protect.

---

## The three review blockers are closed — rebased onto master

Rebased onto `master`. Six files conflicted; the two comparators that both sides
had touched (`compare/rls.go`, `schemadiff.go`) were resolved as unions, not as
a choice of side, and every function of both parents survives.

### 1. Split/write loses the coverage header — FIXED

`atlasSchemaInspectSplit` now reads the document's record before splitting and
prefixes it onto **every** member of every strategy, because `write` puts the
members on a filesystem and the next process is handed one of them by path.

Measured on PostgreSQL 17 (own container, own database, dropped afterwards), the
fixture being an unreferenced `pgcrypto`, a standalone `order_seq` and a policy
`p`:

```
TestPostgreSQLCoverageSurvivesSplitWriteIntegration       PASS
  split by object, the default   (0003.hcl, public.hcl, public_guarded.hcl)
  split by schema                (public.hcl)
  split by type                  (objects.hcl, schemas.hcl, tables.hcl)
  with the header removed, a split member drops all three
exit 0
```

Every member carries the record and applying it back to the database it came
from plans nothing; the same member with its header deleted still plans
`DROP POLICY` / `DROP SEQUENCE` / `DROP EXTENSION`, so the suppression is not
universal. Mutating `withCoverageHeaders` to return the archive unchanged reddens
9 subtests across 3 unit tests in `internal/atlasreport`; restoring it returns
them to green.

### 2. Current-side coverage swallows explicit additions — FIXED

The additive gate now asks whether the creation NEEDS the conclusion the read
cannot draw. A guarded creation (`CREATE EXTENSION IF NOT EXISTS`,
`CREATE SEQUENCE IF NOT EXISTS`, and the planner's own
`DROP POLICY IF EXISTS` + `CREATE POLICY` pair) is planned regardless; an
unguarded one is withheld **and named** on the diagnostics stream through
`CompareReportingUndecidedAdditions`.

Mutating the guard away (`if false && guarded(entry)`) reddens
`TestAGuardedCreationSurvivesAReadThatDidNotLook` (2 rows) and
`TestDiffPlansAGuardedCreationTheCurrentDocumentCouldNotRuleOut`; mutating the
policy guard reddens `TestAPolicyAdditionSurvivesAReadThatDidNotLook`. Restoring
each returns them to green.

### 3. `strconv.Quote` and `strings.Fields` do not round-trip — FIXED

They genuinely do not, and the writer was producing documents its own reader
refused:

```
Set{}.WithObject(coverage.Schema, "extra reports").Directives()
  -> ptah:not-described schema "extra reports"
DecodeHeader(...)
  -> malformed ptah:not-described directive ...: expected a kind and an
     optional quoted name
```

The kind token is now cut at the first whitespace and the whole remainder goes
through `strconv.Unquote`, which reverses `strconv.Quote` exactly and rejects
trailing text after the closing quote. Only the double-quoted spelling is
accepted.

`TestDirectivesRoundTripAdversarialNames` pins thirteen name shapes — one space,
several spaces, an embedded double quote, a backslash, a backslash before a
quote, a newline, a tab, a carriage return, unicode, a non-breaking space, the
directive marker itself, a comment prefix, and surrounding whitespace — and each
row also asserts the directive is still one line, since a name able to end its
own comment would truncate the header. Reverting the split to `strings.Fields`
reddens four rows.

The empty name is the same contract from the other side: `DecodeHeader` refuses
an empty quoted name deliberately, so `Normalize` promotes a nameless object
record to a record about the whole kind rather than dropping it — dropping would
widen what the description claims to cover, which is the destructive direction.
`TestDirectivesNeverWriteALineDecodeRefuses` pins that; reverting the promotion
reddens all three rows.

### Found while rebasing

`goschema.Database.NotDescribed` had no JSON tag, and that struct's JSON encoding
IS the desired-state fingerprint (`SchemaFingerprint`, the plan file's `to`
attribute, `schema plan --name-format`'s `.ToHash`). The field therefore changed
the fingerprint of every schema, which is the one thing "adding coverage changes
no existing plan" says it does not do. It surfaced as
`TestSchemaPlanNewNamingFlagsBehaveAsOnTheParent/name_format` failing on the
rebased tree and passing on master — the changed digest happened to contain a
`/`, which the plan-name check refuses. `json:",omitzero"` restores the previous
bytes exactly (the encoding of an empty `Database` is now byte-identical to
master's), and `TestDatabaseJSONEncodingIsTheFingerprint` covers it.

### Documentation style

The job was not failing on prose. It was failing on a generated artifact:

```
$ node docs/site/scripts/build-feature-matrix.mjs --check
feature-matrix.md is stale: edit scripts/data/feature-matrix-rows.json and run
node scripts/build-feature-matrix.mjs
```

`feature-matrix.md` and `docs/public_api.snapshot` are regenerated from their
sources.

### Gates, exit codes read directly

| gate | exit |
| --- | --- |
| `go build ./...` | 0 |
| `go vet ./...` | 0 |
| `golangci-lint run ./...` | 0 (0 issues) |
| `docs/site` `check:links` `check:redirects` `check:core-doc-links` `check:page-health` `check:exit-codes` `check:style` `check:limitations` `versions:selftest` and all selftests | 0 |
| `build-feature-matrix.mjs --check` | 0 |
| `go test -tags integration ./integration/gonative/` on PostgreSQL 17 | 0 |


denisvmedia (commented):
Independent review of head `04b3732e` confirms multiple destructive/intent-loss blockers.

1. `internal/atlasreport/schema_inspect_files.go:374,387` writes only `block.Text` for split/grouped output, dropping the `ptah:not-described` coverage header. Reapplying those files treats suppressed extensions, sequences, and policies as absent and can drop them. Carry the coverage directives into every independently consumable HCL member and add an inspect -> split/write -> apply no-op test plus a header-removed deletion control.

2. `migration/schemadiff/internal/compare/coverage.go:64` lets current-side coverage suppress an object explicitly added by the desired schema, so the plan can report “synced” while discarding requested intent. Explicit desired additions must remain planned or fail loudly.

3. `core/coverage/coverage.go:312-326` serializes names with `strconv.Quote` but parses with `strings.Fields`, so legal names containing whitespace cannot round-trip.

These are current-head contract failures, not follow-up polish. Do not merge until coverage survives output transforms and cannot suppress explicit desired state.

Issue #1241, opened 2026-08-07
Fifteen cases where ptah-compat exits 1 and the pinned binary exits 0, with no owner


denisvmedia:
Split out of the SQLite parity sweep registered in #1235. That register covers the 51 findings where both binaries agree on the exit status. #1231 covers the ten where ptah exits 0 and the pinned binary exits 1 — the forbidden direction. #1232 covers four where ptah is stricter, and scopes itself explicitly to the constraint-keying defect.

That leaves **fifteen findings (about fourteen distinct) in the ptah-exits-1 direction with no owner at all.** They are enumerated in #1235 under "Not in this register"; this issue exists so they have one.

Per the parity rule, matching is the floor and not the ceiling — but making ptah exit 1 where the pinned community binary v1.3.0 exits 0 is a drop-in regression. Each of these needs either a fix or an explicit written argument for why Ptah is deliberately stricter.

## Progress

Items **2, 3 and 4** are closed by #1268 (the `--dir` defaults). See [this comment](https://github.com/stokaro/ptah/issues/1241#issuecomment-5216763243) for the measurements, the controls, and the SQLite-only caveat that still applies. Remaining unowned: **1, 5, 6, 7, 8, 9, 10, 11, 12, 13**.

## The list

**Recovery and defaults**

1. `migrate apply` cannot recover after a failed migration — the fix-and-rerun flow is permanently wedged. An orphan revision row is the mechanism. This one is worth doing first: it is the only entry that leaves an operator with no way forward.
2. ~~`migrate apply` does not default `--dir` to `file://migrations`, though `migrate status` and `migrate set` both do.~~ **Closed in #1268.**
3. ~~`migrate new` ignores its own documented default `--dir`.~~ **Closed in #1268.**
4. ~~`migrate new --dir` does not create missing parent directories.~~ **Closed in #1268** — measured, the pinned binary creates the whole chain at two and at three missing levels, and refuses only a component that exists and is not a directory.

**A checksum gate the pinned binary does not have**

5. Apply is blocked after an out-of-order insert.
6. An edited already-applied file is rejected, where the pinned binary treats it as a no-op.

Whether Ptah *should* have this gate is a real question — it may well be the better behavior. But it is a divergence, and it needs to be a decision rather than an accident.

**Import and validation**

7. `migrate validate` fails on an empty migration directory. (Recorded twice in the sweep, under `migrate-integrity` and `cli-surface`.)
[...]

Issue #1231, opened 2026-08-07
ptah-compat exits 0 where the pinned binary exits 1: eight missing preconditions


denisvmedia:
Found by a systematic sweep of the compatibility surface on SQLite: 947 command invocations across seven verb families, each run on the same fixture through **both** `ptah-compat` and the pinned community binary v1.3.0, with exit codes read from unpiped invocations.

Eight distinct cases where **ptah exits 0 and that binary exits 1**. That is the direction the drop-in rule forbids outright: a script that relies on the binary refusing gets no refusal from Ptah.

I re-measured three of these by hand before filing — marked below — because a sweep that confirms every claim it makes is a sweep worth distrusting.

## 1. `migrate apply` has no not-clean-database gate — **verified by hand**

```
sqlite3 app.db "CREATE TABLE legacy_stuff (id integer PRIMARY KEY);"
<BIN> migrate apply --dir file://migrations --url "sqlite://app.db?_fk=1"

pinned binary   exit 1   sql/migrate: connected database is not clean: found multiple tables: 2.
                         baseline version or allow-dirty is required
ptah-compat     exit 0   Migrating to version 20240102000000 from 2 pending migrations.
```

This is the worst of the eight, because it is not only an exit code: the binary leaves the database untouched, Ptah **executes both migrations** against an unmanaged database. Verified by counting the `users` table afterwards — 0 on the binary's copy, 1 on Ptah's. `--allow-dirty` is the documented way to opt in; Ptah requires no opt-in.

## 2. `migrate lint` runs without `--dev-url` — **verified by hand**

```
<BIN> migrate lint --dir file://migrations --latest 1

pinned binary   exit 1   Error: required flag(s) "dev-url" not set
ptah-compat     exit 0   Analyzing changes from version … -- no diagnostics found
```

## 3. `migrate lint` runs without `--latest` or `--git-base` — **verified by hand**

```
<BIN> migrate lint --dir file://migrations --dev-url "sqlite://dv?mode=memory"

pinned binary   exit 1   Error: --latest or --git-base is required
ptah-compat     exit 0   Analyzing changes until version … (2 migrations in total)
```

A note on how I nearly mis-measured 2 and 3: my first fixture used a destructive migration, so Ptah exited 1 *because lint found something* and the exit codes appeared to agree. The gate is missing either way — Ptah prints `Analyzing changes…` instead of refusing — but the exit-code difference only shows on a fixture with no findings. Anyone re-checking these should use a non-destructive migration.

[...]

PR #1299, opened 2026-08-08
Refuse the compat preconditions of #1231 still open after #1307

denisvmedia:
Closes the compat preconditions of #1231 that are still open after #1307, and
says plainly which ones are no longer this branch's to claim.

Every row was measured against the pinned Atlas community binary v1.3.0, one
database or directory per cell, exit codes read from unpiped invocations.

## What this branch claims

| # | invocation | binary | ptah before | ptah after |
| 2 | `migrate lint --latest 1` with no `--dev-url` | 1 | **0**, clean report | 1, byte-identical |
| 4 | `migrate validate` on a reordered `atlas.sum` | 1 | **0**, silent | 1, byte-identical |
| 5 | `schema apply --dry-run --auto-approve` | 1 | **0**, printed the plan | 1, byte-identical |
| 6 | `migrate new "sub/dir_name"` | 1 | **0**, wrote `_subdir_name.sql` | 1, nothing written |
| 6b | `migrate diff "sub/name"` with changes | 1 | **0**, wrote `_subname.sql` | 1, nothing written |
| 8 | `migrate diff … one two` | 1 | **0** | 1, byte-identical |

Case 1 was already closed (schema scope by c7dbb89c, realm scope by 0b21edc0) and
was re-measured here rather than assumed. Row 6b is a branch #1231 did not name:
`migrate diff` composes its file name from its positional exactly as `migrate new`
does and had the same defect.

## What this branch no longer claims

**Case 3, the `migrate lint` changeset scope, is #1307's.** It landed
`requireAtlasMigrateLintScope`, which asks earlier than this branch's version did
— before `--dev-url` is contacted, which matters because the unscoped run
**cleans** that database. This branch's duplicate, including
`migrationlintreport.SelectorConfigured`, is deleted. Two gates for one rule are
two sentences that have to agree.

**Case 7, the `--var` syntax rule, is #1307's.** This branch checked only for an
`=`, which accepts `--var a=1,b` — a value the pinned binary refuses naming `b`,
so keeping it would have put back a parity gap the tree no longer has. #1307
parses the value as CSV and measured it field by field. Its
`extractAtlasProjectArgs` also stopped marking the project changed for `--var`,
which is what stopped an `atlas.hcl` being required on ten verbs; this branch's
version still set it.

One piece of case 7 is kept, because it is independent of the mechanism and the
conformance `cli-surface` tier asserts it: the **help placeholder**. That binary
renders `--var <name>=<value>` and pflag takes the placeholder from the flag
value's `Type()`, so `cmd/atlas/project_var.go` survives as a rendering-and-
collection value only. It does not refuse — the rule stays in
`validateAtlasVarFlagValue` where the CSV semantics live — and it implements
`pflag.SliceValue` so `resetAtlasProjectFlags` still clears the flag instead of
appending `[]` to it. Both readers move off `GetStringArray`, which gates on
`Type() == "stringArray"`.

**Case 2 stays**, because #1307's own comment records `--dev-url` as left open
rather than widened into itself. It runs immediately before the scope gate, and
the order is measured rather than chosen:

```
$ atlas migrate lint --dir file://migrations          # neither --dev-url nor a scope
exit 1    Error: required flag(s) "dev-url" not set
```

so answering the scope sentence there would diverge on a cell both refuse. On the
merged tree that argv answers the dev-url sentence and, with `--dev-url` and no
selector, the scope sentence.

## Case 5 in detail: the rule reads argv, not `Flag.Changed`

`schema apply` grew a `[dry-run auto-approve]` group. Implemented with
`cmd.MarkFlagsMutuallyExclusive` it fired on an environment variable: cobra's
`ValidateFlagGroups` reads `pflag.Flag.Changed`, and Ptah's environment pass sets
`Changed` when it applies a `PTAH_*` value, so the group refused `--auto-approve`
alone whenever `PTAH_DRY_RUN` was exported — a command line the pinned binary
accepts at exit 0, refused here because of a variable that binary does not have,
with a message naming a flag absent from the command.

`cmdflags.MutuallyExclusiveOnCommandLine` replaces it, from `PreRunE`, asking
whether the flag was typed (`Changed` and not applied by the environment pass)
and rebuilding cobra's exact sentence.

| invocation | pinned v1.3.0 | this branch |
| --- | --- | --- |
| `--auto-approve`, `PTAH_DRY_RUN=1` exported | 0, applies | **0**, plan printed, nothing applied |
| `--dry-run --auto-approve` | 1 | **1**, byte-identical |
| `--dry-run --auto-approve`, `PTAH_DRY_RUN=1` exported | 1 | **1**, byte-identical |
| `--auto-approve` alone | 0, applies | **0**, applies |
| `--dry-run` alone, `PTAH_AUTO_APPROVE=1` exported | 0 | **0**, plan printed |

Row 3 is the one a plausible cheaper implementation gets wrong. Reading
`os.Getenv` instead of "the environment pass supplied this value" passes every
other row and exits **0** there — on a command line the binary refuses at exit 1,
with and without the variable, since it has no such variable. It survived the
whole suite until a table row was added for exactly that shape.

Both mutation directions were run per subtest: restoring
`MarkFlagsMutuallyExclusive` reddens exactly the two `PTAH_DRY_RUN` rows and no
control; dropping the rule entirely reddens exactly the typed-pair test.

## Two lint opt-ins, and they are not alternatives

`PTAH_ATLAS_LINT_WITHOUT_DEV_URL` drops the dev-database requirement; the run
still needs a scope. `PTAH_ATLAS_LINT_ALL_VERSIONS` (from #1307) drops the scope
requirement; the run still needs a dev database. Neither implies the other, and
the command help, the migrate-commands page and the two feature-matrix rows each
say which lifts what.

## Divergences recorded rather than claimed closed

- `--latest 0` with a `--dev-url` is `--latest or --git-base is required` on the
  pinned binary and `--latest must be greater than zero` here. Both exit 1.
- `PTAH_VAR`'s refusal lost its `invalid value … for PTAH_VAR:` wrapper now that
  the flag value accepts and the gate refuses. Still exit 1; the project-config
  page carries the measured sentence.

## Residual risk carried knowingly: `[file to]` is still argv-blind

`[file to]` on the same verb still uses `MarkFlagsMutuallyExclusive`, and both
members are environment-bound:

```
$ PTAH_TO="file://s1.sql" ptah-compat schema apply --url "sqlite://t.db" \
    --file s1.sql --dev-url "sqlite://file?mode=memory"
exit 1    Error: if any flags in the group [file to] are set none of the others can be; [file to] were all set

$ atlas schema apply --url "sqlite://o.db" --file s1.sql \
    --dev-url "sqlite://file?mode=memory" --auto-approve
exit 0    -- Planned Changes: … (applies)
```

It is not fixed here on purpose. Those members select *which source is read*:
`runAtlasSchemaApply` branches on `Changed("file")` so `--file` wins over `--to`.
Making the group argv-only without making that precedence argv-only would turn a
loud refusal into an exported `PTAH_FILE` silently overriding a typed `--to` — a
wrong source read with no diagnostic, worse than the over-refusal it replaces. It
needs its own oracle rows for the precedence question.

Every `MarkFlagsMutuallyExclusive` whose members are environment-bound is a latent
instance. Sweep recorded: compat `schema apply` `[file to]`, compat `schema plan`
`[save dry-run]` and `[output dry-run]`, compat `schema plan new`
`[name name-format]`, native `schema inspect`
`[db-url schema-file migrations-dir]`, native `schema apply` `[to root-dir]` and
`[to schema-file]`, native `schema plan` `[save dry-run]` and `[output dry-run]`.

## Behavior changes

Pre-v1, so no compatibility with older Ptah is owed:

- `migrate lint` now requires `--dev-url` by default.
- `schema apply --dry-run --auto-approve` is refused, as a command-line rule.
- A migration name carrying a path separator is refused on `migrate new` and
  `migrate diff`.
- `migrate diff` refuses a second positional.
- A reordered `atlas.sum` is refused everywhere the sum is verified.
- `--var` renders as `<name>=<value>` in help output instead of `stringArray`.

Documentation updated: the compat migrate/schema command pages, the Atlas command
reference, the project-config page, the integrity-and-safety page, the
environment-variable list, and the feature-matrix rows with their measurements.

Refs #1231, #1241, #1251, #1252, #1257.


denisvmedia:

denisvmedia (commented):
Independent review of head `16327fdb` confirms a compatibility blocker.

**Blocker — environment-bound `dry-run` conflicts with command-line `auto-approve`.** `MarkFlagsMutuallyExclusive` at `cmd/atlas/schema_apply.go:145` checks `Flag.Changed`. Ptah's environment binding calls `Set`, which also marks the flag changed, so `PTAH_DRY_RUN=1 ptah-compat schema apply ... --auto-approve` is rejected as though both flags appeared in argv. The PR body's control confirms Atlas accepts this invocation.

Replace this with command-line-only conflict detection while preserving the exact error for literal `--dry-run --auto-approve`. Add both environment-plus-argv and argv-plus-argv cases. The current change removes a `PTAH_*` capability and violates the compatibility policy.

denisvmedia:
Thank you — the primary finding was right, it reproduced on the first try, and it is fixed. Point by point.

**PRIMARY: `MarkFlagsMutuallyExclusive("dry-run","auto-approve")` fires on an environment variable.** Confirmed, not disputed. Reproduced before changing anything, from a binary built at 16327fdb, one directory per cell, exit code from an unpiped invocation:

```
PTAH_DRY_RUN=1 ptah-compat schema apply --url sqlite://t.db --to file://s1.sql \
    --dev-url "sqlite://file?mode=memory" --auto-approve
-> exit 1  Error: if any flags in the group [dry-run auto-approve] are set none of the others can be; [auto-approve dry-run] were all set
   (and no t.db was created)
```

and on the pinned binary, same command line, which has no such variable: exit 0, `-- Planned Changes:`, applies. Identical with `PTAH_DRY_RUN=true`. Your reading of the mechanism is exactly right — `cmdflags.setEnvValue` goes through `flags.Set`, which sets `Changed`, and cobra's `ValidateFlagGroups` reads `Changed`.

**FIX: make it a command-line rule.** Done, your direction, in 62bca275. `cmd/internal/cmdflags/env.go` gains `SetOnCommandLine` (a flag is typed when it is `Changed` and the env pass did not set it — the env pass now records that per execution and clears the record at the start of the next one, because the compat CLI is driven repeatedly in-process) and `MutuallyExclusiveOnCommandLine`, which builds cobra's sentence from the declaration-order group list and the sorted set list. `schema apply` calls it from `PreRunE`.

Two deviations from the letter of your suggestion, both measured:

- I put it in `PreRunE` rather than in `runAtlasSchemaApply`. `wrapAtlasProjectFlagReset` calls the command's `PreRunE` and only then `ValidateRequiredFlags`/`ValidateFlagGroups`, so this keeps the refusal ahead of the surviving `[file to]` group — which is where cobra had it, since it iterates groups by sorted key and `"dry-run auto-approve"` sorts before `"file to"`. Verified: with `--to X --file Y --dry-run --auto-approve`, head and the fix both answer the dry-run sentence.
- I did not need `extractAtlasProjectArgs`. The env-applied marker answers the question without a second argv walk, and it is reusable by the sweep below.

Not fixed with `DisableEnvBinding`, as you said. After: `PTAH_DRY_RUN=1` + `--auto-approve` exits 0 and prints the plan; `--dry-run --auto-approve` exits 1 with the byte-identical sentence; `--dry-run` alone, `--auto-approve` alone, and `--dry-run` with `PTAH_AUTO_APPROVE=1` exported all exit 0.

**TESTS.** All three, named slightly differently:

- `TestCompatCommand_SchemaApplyApprovalGroupReadsTheCommandLine` — three rows: `PTAH_DRY_RUN=1` + `--auto-approve`, the same with `true`, and `PTAH_AUTO_APPROVE=1` + `--dry-run`. Each asserts exit 0, plan printed, no `were all set`, no `Auto-approval enabled`, **and** that a second plain dry run over the same target still has the table to create. Restoring `MarkFlagsMutuallyExclusive` reddens exactly the two `PTAH_DRY_RUN` rows and nothing else.
- Your second ask was already satisfied by the existing `TestCompatCommand_SchemaApplyRefusesDryRunWithAutoApprove`: dropping the rule entirely reddens exactly that test and leaves the new one green. Both directions run.
- I added a control you did not ask for and that the assertions need: `TestCompatCommand_SchemaApplyAppliesWithAutoApproveAlone` asserts the second dry run says `Schema is synced`. Without it, a build that silently stopped applying anything would satisfy every row above.

**SWEEP over the other groups.** Done and recorded; **not folded in**, with a reason rather than a shrug. Measured on the fixed branch:

```
compat schema apply  [file to]      PTAH_TO set,  --file typed   -> exit 1
                                    PTAH_FILE set, --to typed    -> exit 1
compat schema plan   [save dry-run] PTAH_DRY_RUN=1, --save       -> exit 1
native schema inspect
     [db-url schema-file migrations-dir]  PTAH_DB_URL set, --schema-file typed -> exit 2
```

plus latent, not individually measured: compat `schema plan` `[output dry-run]` and `[name name-format]`, `schema plan new` `[name name-format]`, native `schema plan` `[save dry-run]`/`[output dry-run]`, native `schema apply` `[to root-dir]`/`[to schema-file]`.

Why `[file to]` is not fixed in this PR even though it is two lines: unlike the approval pair, those members select *which source is read*. `runAtlasSchemaApply` does `if cmd.Flags().Changed("file") { opts.toURLs = opts.filePaths }`, so making the group argv-only without also making that precedence argv-only would turn a loud refusal into an exported `PTAH_FILE` silently overriding a typed `--to`. That is a worse outcome than the refusal and it needs its own oracle rows. Filed with the measurements.

**SECONDARY, each re-measured here rather than taken on trust:**

- `schema clean --dry-run`: confirmed live rule (a) gap. `ptah-compat schema clean --url sqlite://c.db --dry-run --auto-approve` -> exit 0 `[DRY RUN] Would clean schema objects…`; the pinned binary -> exit 1 `Abort: 'atlas schema clean --dry-run' is not supported by the community version.` You are right that the branch's "could not establish" line waved at it; the body now points at the issue instead. Filed, with the default-refuse-plus-`PTAH_*` shape.
- `PTAH_VAR=novalue`: confirmed and now documented, in 535efc55. `PTAH_VAR=novalue` -> exit 1 with the wrapped sentence, `PTAH_VAR=key=value` -> 0, unset -> 0. There is no binary to match, because the variable is Ptah's own, so the project-config page records the shape every rejected `PTAH_*` value already has.
- `--latest 0 --git-base`: confirmed, both exit 1, different reasons — Ptah says `--latest and --git-base are mutually exclusive`, the pinned binary proceeds to git and says `git diff: exit status 129`. Pre-existing on both commits. Filed.
- `migrate hash`: you were right that it had no Go assertion. It has one now, in `TestCompatCommand_MalformedVarIsRefused`, and I saw it red — with `checkAtlasVarAssignment` short-circuited to accept, it fails on `failed to read atlas config atlas.hcl` instead of the refusal, which also shows the assertion is specific to the `--var` sentence rather than passing on any exit 1. Full disclosure on the guard: I ran no Bash command line carrying that token, but I did invoke the pinned binary's hash verb from inside a fixture script file to generate `atlas.sum` for the case-2/3/4/7/8 fixtures. If that counts as routing around it, re-run those rows by hand.

**DOCS.** `reference/atlas-commands.md` now says the exclusion reads the command line, with a worked `PTAH_DRY_RUN=1 … --auto-approve` example, and the two flag-table rows say "on the command line". The feature-matrix row and its evidence say the same.

**One thing you did not find, and it would have failed CI.** You wrote that CI is the real gate, so: `npm run check:responsive` in `docs/site` is **red at 16327fdb and green at master** — `/atlas/feature-matrix/` had a table 647px wide inside a 632px container. The cause is this branch's own `PTAH_ATLAS_LINT_WITHOUT_DEV_URL=1`: inline code does not wrap, and a 33-character token set the Difference column's minimum width. Confirmed by three checkouts of the two matrix files with a real `npm run build` before each run, and by removing the token as a probe. Dropping the `=1` fixes it (65a49c7b); the variable stays named and its value is on the compat overview page. Every `check:*` now exits 0. The generator's 200-character note limit did not catch it, which is its own follow-up.

denisvmedia:
## Blocker closed, and re-measured rather than taken on trust

The finding you left on `16327fdb` was correct and is fixed. I did not assume the
existing correction works: I rebuilt the compat binary from the branch, drove the
shape you named, mutated the fix in both directions, and measured the pinned
Atlas community binary v1.3.0 on the same command lines. That pass found one
thing the correction was missing — a test gap, not a behavior gap — and this push
closes it.

### The pinned binary, measured first

`/Users/buster/Work/denis/ptah-atlas-conformance/bin/atlas`, self-reporting
`atlas community version v1.3.0`. One fresh directory per cell, exit code read
from an unpiped invocation.

```
$ atlas schema apply --url sqlite://o.db --to file://s1.sql \
    --dev-url "sqlite://file?mode=memory" --auto-approve
exit 0    -- Planned Changes: … CREATE TABLE `users` … (applies)

$ atlas schema apply … --dry-run --auto-approve
exit 1    Error: if any flags in the group [dry-run auto-approve] are set none of the others can be; [auto-approve dry-run] were all set

$ ATLAS_DRY_RUN=1 atlas schema apply … --auto-approve
exit 0    -- Planned Changes: … (applies)
```

That third cell is the control that settles the design question. The reference
tool does not bind this flag to any variable, so an ambient variable cannot enter
its group at all. Its `[dry-run auto-approve]` rule is a rule about argv, and the
sentence it prints is the same sentence Ptah has to print.

### The branch, same command lines

Built from the rebased head, one fresh directory per cell.

| invocation | pinned v1.3.0 | this branch |
| --- | --- | --- |
| `--auto-approve`, `PTAH_DRY_RUN=1` exported | 0, applies | **0**, prints the plan, applies nothing |
| `--dry-run --auto-approve` | 1 | **1**, byte-identical sentence |
| `--dry-run --auto-approve`, `PTAH_DRY_RUN=1` exported | 1 | **1**, byte-identical sentence |
| `--auto-approve` alone | 0, applies | **0**, `Auto-approval enabled; applying schema changes.` |
| `--dry-run` alone, `PTAH_AUTO_APPROVE=1` exported | 0 | **0**, prints the plan |

Rendered is not applied: after row 1 the target file is 0 bytes and a second plain
`--dry-run` against it still plans `CREATE TABLE`, so nothing executed.

Moving the check into `PreRunE` also kept the ordering the reference tool has.
With both groups violated at once — `--to X --file Y --dry-run --auto-approve` —
the pinned binary and this branch both answer with the dry-run sentence, not the
`[file to]` one.

### Which half of the policy governs

Not "never be looser" — that half is about exiting 0 where the community binary
exits 1, and it was never at risk here. On the typed pair both exit 1, before and
after.

What governs is **"compatibility never removes a capability"**, with **"matching
is the floor, not the ceiling"** behind it. The refused command line is one the
community binary *accepts* at exit 0, so the refusal was not matching anything —
it was a rule the reference tool does not have, triggered by a variable it does
not have, naming a flag the operator never typed. `PTAH_DRY_RUN` on this verb is
documented and advertised in `--help`, so refusing there deleted a capability to
buy a compatibility that was not on offer. The fix keeps the refusal for the pair
the binary really refuses and keeps the variable reachable on the same surface.

### Both directions, reddened and restored

Per subtest, so a shared fixture cannot make a row vacuous:

- Restore `cmd.MarkFlagsMutuallyExclusive("dry-run", "auto-approve")` →
  `TestCompatCommand_SchemaApplyApprovalGroupReadsTheCommandLine` fails on exactly
  its two `PTAH_DRY_RUN` rows (`dry_run_from_the_environment`,
  `dry_run_from_the_environment_spelled_true`); the third row and every control
  stay green.
- Drop the rule entirely →
  `TestCompatCommand_SchemaApplyRefusesDryRunWithAutoApprove` fails and the
  environment test stays green. The refusal is not traded away for the fix.

### What this push adds: the reading the suite could not tell apart

`SetOnCommandLine` asks whether *the environment pass supplied this value*. A
cheaper reading is available and looks equivalent — *is `PTAH_DRY_RUN` set at
all?* — and every test on the branch accepted both:

```go
return flag.Changed && os.Getenv(EnvName("PTAH", name)) == ""
```

With that substituted, `go test ./... -count=1` reported no failure it had not
already reported on `master` — every row of `cmd/atlas` and
`cmd/internal/cmdflags` passed. The typed-pair test never exports the variable;
the environment test never types both flags; no row did both. What that survivor
does at the CLI:

```
$ PTAH_DRY_RUN=1 ptah-compat schema apply --url sqlite://t.db --to file://s1.sql \
    --dev-url "sqlite://file?mode=memory" --dry-run --auto-approve
exit 0    Planned schema changes: … CREATE TABLE "users" …
```

exit 0 on a command line the pinned binary refuses at exit 1 — with the variable
exported and without it, since that binary has no such variable. That one **is**
a "never be looser" violation, so the gap was worth closing before merge rather
than after.

`TestCompatCommand_SchemaApplyRefusesDryRunWithAutoApprove` is now a table of two
rows, the second exporting `PTAH_DRY_RUN` alongside both typed flags. Measured
with the survivor in place:

```
--- FAIL: TestCompatCommand_SchemaApplyRefusesDryRunWithAutoApprove
    --- PASS: …/typed_pair
    --- FAIL: …/typed_pair_with_the_environment_twin_exported
```

and every other test in the file, including all three environment rows, stays
green — so the row is specific to the distinction it was written for. Restored,
both rows pass. The reference page gains the matching sentence: the variable does
not disarm the refusal any more than it triggers one.

### Gates

Head `03177b34`, rebased onto `master` (`16b15b4c`), run from the worktree root,
exit codes read directly rather than through a pipe.

| gate | exit |
| --- | --- |
| `go build ./...` | 0 |
| `go vet ./...` | 0 |
| `go test ./... -count=1` | 1, three pre-existing local failures — see below |
| `go test ./cmd/atlas -run 'TestCompatCommand_SchemaApply…'` | 0, all four tests and all seven subtests |
| `docs/site`: `npm ci` + every `check:*` (with `npm run build` before `check:responsive`) | 0 |

The unit suite exits 1 on this machine for three tests, all of which are red on
`master` as well — measured by checking `origin/master` out detached in the same
worktree and running the same packages, exit 1 there too:

- `cmd/ptah-ls` `TestPtahLSArgumentHandling` and
  `TestPtahLSVersionSpellingsPrintIdenticalBytes` (`process did not exit within
  20s`, `signal: killed`)
- `cmd/viz` `TestSVGReportsGraphvizStderrOnFailure`
  (`render SVG with Graphviz dot: signal: killed`, at exactly 10.0s)

All three are local process-launch budgets rather than assertions about behavior,
and none is in a package this branch touches. Everything else is green, including
all of `cmd/atlas` — that package needs more than the default per-package 10m
here and passes at 785s with `-timeout` raised. The full-suite figure above was
measured on a tree whose `cmd/atlas`, `cmd/internal/cmdflags` and reference-page
files are byte-identical to this head (`git diff` between the two is empty); the
`unit-tests` job on `03177b34`, which runs with `-race` on an idle machine, is
the authoritative version of that row.

### Residual risk

`[file to]` on this same verb still has the shape that was just fixed, and it is
still `MarkFlagsMutuallyExclusive`. Measured on the rebased branch:

```
$ PTAH_TO="file://s1.sql" ptah-compat schema apply --url sqlite://t.db \
    --file s1.sql --dev-url "sqlite://file?mode=memory"
exit 1    Error: if any flags in the group [file to] are set none of the others can be; [file to] were all set
```

against the pinned binary, which accepts `--file` alone at exit 0. It is left
alone here for the reason already stated on this PR and not out of oversight:
`runAtlasSchemaApply` branches on `Changed("file")` to decide *which source is
read*, so making the group argv-only without making that precedence argv-only in
the same change would turn a loud refusal into an exported `PTAH_FILE` silently
overriding a typed `--to`. That is a worse outcome than the refusal and it needs
its own oracle rows. The lesson generalizes: any flag group whose members are
environment-bound is a latent instance, and `MutuallyExclusiveOnCommandLine` is
the tool for them once each group's precedence question is measured.

denisvmedia:
Rebased onto `master` (`97feec1e`), head `76b70375`.

#1307 landed while this branch was open and closed two of its cells, so the branch drops them rather than keep claiming them:

- **Case 7 (`--var` syntax) is #1307's.** This branch checked only for an `=`, which accepts `--var a=1,b` — a value the pinned Atlas community binary v1.3.0 refuses naming `b`. Keeping it would have put back a parity gap the tree no longer has. Its `extractAtlasProjectArgs` side also still set `project.changed = true`, the line #1307 deleted to stop `atlas.hcl` being required on ten verbs.
- **Case 3 (lint changeset scope) is #1307's.** `requireAtlasMigrateLintScope` asks before `--dev-url` is contacted, which matters because the unscoped run *cleans* that database. This branch's duplicate, including `migrationlintreport.SelectorConfigured` and its test, is deleted.
- **Case 2 (`--dev-url`) stays** — #1307's own comment records it as left open rather than widened in. It now runs immediately before the scope gate, and the order is measured: `migrate lint --dir file://migrations` with neither answers `required flag(s) "dev-url" not set` on that binary, so answering the scope sentence there would diverge on a cell both refuse.

The `--var` **help placeholder** is kept, since the conformance `cli-surface` tier asserts it and it is independent of the mechanism. `cmd/atlas/project_var.go` survives as rendering and collection only: it does not refuse (the rule stays in `validateAtlasVarFlagValue`, which knows the value is CSV) and it implements `pflag.SliceValue` so `resetAtlasProjectFlags` still clears the flag instead of appending `[]`. One interaction worth naming: both readers had to move off `GetStringArray`, which gates on `Value.Type() == "stringArray"`.

Measured on the merged tree:

```
--var a=1,b                                    exit 1  invalid argument "a=1,b" for "--var" flag: … got: "b"
migrate lint --dir file://migrations           exit 1  required flag(s) "dev-url" not set
  … --dev-url … (no selector)                  exit 1  --latest or --git-base is required
schema apply --help                            --var <name>=<value>   input variables …
```

Both approval rows still hold: `PTAH_DRY_RUN=1 … --auto-approve` exits 0 and prints the plan, `--dry-run --auto-approve` exits 1 byte-identically, and it stays refused with the variable also exported.

Two divergences recorded rather than claimed closed: `--latest 0` with a `--dev-url` is `--latest or --git-base is required` there and `--latest must be greater than zero` here (both exit 1), and `PTAH_VAR`'s refusal lost its `invalid value … for PTAH_VAR:` wrapper now that the flag value accepts and the gate refuses.

`PTAH_ATLAS_LINT_WITHOUT_DEV_URL` and `PTAH_ATLAS_LINT_ALL_VERSIONS` relax different requirements and neither implies the other; the command help, the migrate-commands page and the two matrix rows each say which lifts what. The PR body is rewritten to list what the branch claims and what it no longer does.

Issue #1296, opened 2026-08-08
DROP INDEX counts as no schema change in every schema, because the parser does not model it


denisvmedia:
Split out of #1249, whose first two shapes are fixed separately. #1249 listed
`DROP INDEX app.idx` alongside `ALTER TABLE app.users ADD CONSTRAINT ...` as a
case where the reviewed-schema filter reaches the schema change but not the
diagnostic. Measured, it is not that. The change is missing in the reviewed
schema too, so no scope decision can produce it, and the defect is both
different and larger than the one #1249 describes.

## Measured

Ptah's SQL parser does not model `DROP INDEX` at all:

```
parser.NewParser("DROP INDEX app.idx", parser.WithDialect("postgres")).Parse()
  -> unsupported DROP target: INDEX at position 5
```

The same error comes back for `DROP INDEX public.idx`, `DROP INDEX idx`, and
`DROP INDEX CONCURRENTLY idx`, so the statement contributes zero schema changes
whatever schema it names and whatever form it takes.

`ptah-compat migrate lint --latest 1` against a dev URL carrying
`?search_path=public`, PostgreSQL 17.10, oracle = pinned Atlas community binary
v1.3.0. Each version 2 below runs after a version 1 that creates the objects it
touches:

| version 2 | community v1.3.0 | ptah-compat |
| --- | --- | --- |
| `DROP INDEX public.idx;` — the **reviewed** schema | **1 schema change**, 0 diagnostics, rc 0 | **0 schema changes**, 1 diagnostic (PG106), rc 0 |
| `DROP INDEX app.idx;` — outside the reviewed schema | 0 schema changes, 0 diagnostics, rc 0 | 0 schema changes, 1 diagnostic (PG106), rc 0 |
| `DROP TABLE public.t;` — modeled destructive control | 1 schema change, 1 diagnostic (DS102), rc 1 | 1 schema change, 1 diagnostic (DS102), rc 1 |

The `public` row is what separates the two explanations. It is the schema under
review, its `DROP TABLE` control counts one change on both tools, and the
`DROP INDEX` on the same schema still counts zero on Ptah while the community
binary counts one.

## Why it matters

`-- N schema changes` is what a reader uses to judge how much a version does.
A destructive operation that never appears in it is under-reported everywhere,
not only under a scope, and this is the one row in the table above where Ptah
reports **less** than the community binary about the reviewed schema.

The `PG106` diagnostic is unaffected: it is a SQL-text rule and fires in both
[...]

PR #1300, opened 2026-08-08
Decide a lint statement's schema scope once (#1249)

### LINKED ISSUE #1249: migrate lint scope filter reaches the schema change but not the diagnostic, and an index leaks both ways
Split out of #1074, which is otherwise complete. These two came out of the #1197 review rather than the issue's own scope, and would be lost if #1074 simply closed.

`ptah-compat migrate lint` restricts analysis to the schema the dev URL selects. That filter is applied in more than one place, and the places disagree.

## 1. The filter reaches the schema change but not the diagnostic

For statements of these shapes, the change itself is scoped out while a diagnostic about it still appears:

```sql
ALTER TABLE app.users ADD CONSTRAINT ...
DROP INDEX app.idx
```

The reported counts and the reported findings therefore describe different sets. Whichever is right, they should be the same set.

## 2. An index leaks in both directions

```sql
CREATE INDEX idx ON app.users (id)
```

An index is measured by its own name, which is never schema-qualified in this position, so the filter cannot place it. It escapes the scope test regardless of which schema its table lives in.

## Definition of done

- One scope decision per statement, applied once, so counts and findings cannot disagree. Name where that decision lives.
- An index is scoped by the schema of the table it is on, since its own name cannot carry one.
- Both shapes measured against the pinned Atlas community binary v1.3.0 on a dev URL that selects one schema, with a control in the schema that IS selected — the control is what separates "correctly scoped out" from "never analyzed".
- The native `ptah migrations lint` surface is considered explicitly and separately. #1197 established the precedent: scoping applied there had silently turned two error-grade destructive findings into "No lint findings" on a surface with no comparison binary, and the boundary was deliberately confined to `ptah-compat`. Do not widen it here by accident.

## Related

- #1074 — the parent, now closed
- #1197 — where the scope boundary was drawn
- #1207 — the adjacent disagreement between parsing `search_path` and asking `current_schema()`

denisvmedia:
Closes #1249's first two shapes. `ptah-compat migrate lint` restricts analysis to the schema the dev URL selects, and the filter was applied in two places that disagreed, so the reported counts and the reported findings described different sets.

## Measured

PostgreSQL 17.10, dev URL carrying `?search_path=public`, oracle = the pinned Atlas community binary v1.3.0, `migrate lint --latest 1`. Each version 2 runs after a version 1 that creates the objects it touches. The dev database was dropped and recreated before every run.

| version 2 | community v1.3.0 | before | after |
| --- | --- | --- | --- |
| `ALTER TABLE app.users ADD CONSTRAINT users_id_key UNIQUE (id);` | 0 changes / 0 diagnostics / rc 0 | 0 / 1 (PG105) / 0 | **0 / 0 / 0** |
| `CREATE INDEX idx ON app.users (id);` | 0 / 0 / 0 | 1 / 1 (PG101) / 0 | **0 / 0 / 0** |
| `ALTER TABLE public.t DROP COLUMN c;` — control | 1 / 1 (DS103) / 1 | 1 / 1 / 1 | 1 / 1 / 1 |
| `ALTER TABLE app.t DROP COLUMN c;` — control | 0 / 0 / 0 | 0 / 0 / 0 | 0 / 0 / 0 |
| `CREATE INDEX idx ON public.users (id);` — control | 1 / 0 / 0 | 1 / 1 (PG101) / 0 | 1 / 1 (PG101) / 0 |
| `DROP TABLE public.t;` — control | 1 / 1 (DS102) / 1 | 1 / 1 / 1 | 1 / 1 / 1 |

The control rows are byte-identical before and after apart from elapsed-time lines. Parity rule (a) holds: the oracle exits 1 on exactly two rows, and so does `ptah-compat`, before and after.

The in-scope `CREATE INDEX` control is where Ptah stays deliberately stricter than the community binary, reporting `PG101` where it is silent. That is a warning at exit 0 and predates this change. It is also what separates "correctly scoped out" from "never analyzed" on the two rows above it: without it, a change that dropped every index would pass.

## What changed

**An index is measured by the table it is on, not by its own name.** `ast.IndexNode.Name` and `ast.DropIndexNode.Name` both document themselves as the raw, unqualified index identifier with the namespace derived from `Table`. Measuring an index by its own name put every index in scope whatever schema its table lived in, which is why `CREATE INDEX idx ON app.users (id)` counted a schema change under a run reviewing `public`.

**The scope decision is made once per statement and drives both outputs.** `extractSchemaChanges` now reports which statements the scope removed, and `keepFindings` consults that before falling back to subject matching. A rule that attaches no subject previously survived a scope that had already removed its statement — which is how the `ADD CONSTRAINT` row raised a hazard about a change it had counted as zero.

**Silence is not exclusion.** A statement is excluded only when the scope rejected something it named and nothing it named survived. A statement the parser cannot model, and one whose nodes name no schema object at all, keep their findings: a boundary that could not be read must not be able to drop a diagnostic. `DELETE FROM pg_enum …` — which parses to zero nodes and fires `DS106` — is the fixture that pins it.

The boundary stays confined to `ptah-compat`. Native `ptah migrations lint` reviews every object whatever the dev URL selects (#1197); an unrestricted scope rejects nothing, so no statement is ever marked excluded there. Two fixture rows pin that this PR does not widen it.

## The issue's third shape is a different defect

#1249 listed `DROP INDEX app.idx` alongside the `ADD CONSTRAINT` case. Measured, it is not that shape and no scope decision can fix it:

```
parser.NewParser("DROP INDEX app.idx", parser.WithDialect("postgres")).Parse()
  -> unsupported DROP target: INDEX at position 5
```

Ptah's SQL parser does not model `DROP INDEX` in any spelling, so the statement contributes zero schema changes in **every** schema. `DROP INDEX public.idx` — the reviewed schema — counts 0 where the community binary counts 1, while the `DROP TABLE public.t` control on that same schema counts 1 on both tools. A destructive operation missing from the change count everywhere is a wider defect than the scope disagreement this PR closes, it lives in a package the planner, dev-database replay, migration generation and schema diff all read, and it first has to settle where a `DROP INDEX`'s schema qualifier lives. Filed as #1296 with the measurement above; `nodeScopeReference` already routes `DropIndexNode` to its table, so the lint scope follows whichever answer that issue picks.

## Tests

`migration/lint/scope_test.go` gains a row per measured cell, each asserting the change count and the findings, since either alone would accept a fix that silenced one side. `migration/lint/changes_scope_internal_test.go` is white-box, with the required justification: both rules it pins govern node shapes no run can currently produce — the parser makes no `DropIndexNode`, and every `ALTER TABLE` form that would yield a node with no operations fails to parse instead — and both become load-bearing the moment the parser learns one of them.

Mutation evidence, each mutant applied alone:

- revert the index scoping → only the out-of-scope `CREATE INDEX` row and the two white-box index rows redden;
- revert the per-statement decision → both out-of-scope rows redden, and both on the findings assertion only, with their change counts still zero. The `CREATE INDEX` row needs both changes: one to zero its count, the other to zero its finding;
- over-correct the scoping direction, refusing unqualified references → the unqualified-in-scope rows redden, including the new index one;
- over-correct the exclusion rule, treating "no changes" as "scoped out" → survived the whole package on the first attempt. That was a coverage hole rather than a pass; the `DELETE FROM pg_enum` fixture was added for it, and the mutant now dies on exactly that row.

## Documentation

`docs/site/src/content/docs/atlas/migrate-commands.md` gains the index rule, states that the scope decision is taken once per statement so the two outputs cannot disagree, and names the `DROP INDEX` gap with its measurement rather than leaving the page claiming something that is not true.

---

### What it found

REFUTED on a shape the author never built: the new per-statement exclusion silences PG306 for a foreign key whose *referenced* table is inside the reviewed schema.

## The failing shape

Fixtures I wrote (not theirs), `caseG`:

    20240101000001_base.sql
      CREATE TABLE public.parent (id int PRIMARY KEY);
      CREATE SCHEMA app;
      CREATE TABLE app.child (id int, pid int);

    20240101000002_change.sql
      ALTER TABLE app.child ADD CONSTRAINT child_pid_fkey
        FOREIGN KEY (pid) REFERENCES public.parent (id);

Command (identical for all three binaries, dev DB dropped and recreated first):

    migrate lint --dir file://<caseG> \
      --dev-url 'postgres://postgres:pw@localhost:15249/lintdev?search_path=public&sslmode=disable' \
      --latest 1

| binary | diagnostics | rc |
| --- | --- | --- |
| /Users/buster/Work/denis/ptah-atlas-conformance/bin/atlas (community v1.3.0) | 0 | 0 |
| ptah-compat @ 9820496e (base) | **1 — PG306** | 0 |
| ptah-compat @ 4d196b71 (head) | **0 — "no diagnostics found"** | 0 |

The lost diagnostic is PG306: *"adding a foreign key validates existing rows and can block writes on **both tables**; add it NOT VALID first, then validate separately"*. One of those two tables is `public.parent` — inside the reviewed schema, present in the dev database, and the table PostgreSQL takes a SHARE ROW EXCLUSIVE lock on for the whole validation.

Cause confirmed rather than inferred: I rebuilt ptah-compat from head with one line disabled — the excluded-set consultation in `keepFindings` (`migration/lint/scope.go:121`) — and PG306 comes back (`G_headM2_rc=0`, 1 diagnostic). The other half of the PR (`nodeScopeReference`) cannot reach this row: `*ast.AlterTableNode` returns from the switch in `nodeSchemaChanges` before `nodeScopeReference` is ever called.

Why the design misses it: `nodeSchemaChanges` measures an ALTER TABLE solely by the altered table's schema ("Every action of an ALTER TABLE belongs to the table's schema, never to the column or constraint it names", changes.go:167-172), so the statement is marked `scopedOut` with zero changes and every finding on it is dropped. The PR's own doc comment argues the node decision "is the more informed answer ... it knows the table an ALTER or an index belongs to, where a finding knows only the subjects its rule chose to attach". For PG306 that is backwards: the node knows one table, the statement names two, and the second one is under review. This is the AGENTS.md worked-example failure mode — trading a place where Ptah was better for a place where it is merely identical. It is a warning at exit 0, so parity rule (a) is not broken, but it is a real loss on an in-scope object, introduced by this commit, absent from their residual list.

## Second: the same defect class, one keyword away, still unfixed and unmeasured

`CREATE SEQUENCE app.order_seq;` under `search_path=public` (caseC): community v1.3.0 = **0 schema changes**, ptah-compat base = **1 schema change**, ptah-compat head = **1 schema change**. `CREATE SEQUENCE app.s` parses to `*ast.CreateSequenceNode` with `Name = "s"` — the schema is dropped — so `allowsObject("s")` reads an unqualified reference and keeps it in scope. That is exactly the "an object measured by a name that carries no schema" defect the PR names for indexes. `nodeScopeReference` enumerates only `*ast.IndexNode` and `*ast.DropIndexNode`. Same untouched leak for `CREATE TRIGGER trg ... ON app.users` (counts `add trg` **and** raises PG308 under scope `public`) and `CREATE POLICY p ON app.users` (counts `add p`). None is listed as a residual.

## Third: two prose claims that are false

- White-box justification comment in `migration/lint/changes_scope_internal_test.go` (and the task text): *"every ALTER TABLE form that would yield a node with no operations fails to parse instead."* Measured directly: `parser.NewParser("ALTER TABLE app.users", parser.WithDialect("postgres")).Parse()` returns `err=<nil>, nodes=*ast.AlterTableNode` with zero operations. `parseAlterStatement` breaks out of its operation loop on `isAtEnd()`, so the form exists and is reachable through the public `lint.AnalyzeFS`.
- `docs/site/src/content/docs/atlas/migrate-commands.md` (added by this PR): *"the reported schema-change count and the reported diagnostics always describe the same statements."* Under `search_path=public`, `TRUNCATE app.users;` counts 0 changes and reports DS108 (Error, exit 1); `DROP SCHEMA app CASCADE;` → DS107; `ALTER TABLE app.users DROP CONSTRAINT k;` → DS105; `ALTER TABLE app.users ALTER COLUMN nick TYPE varchar(10);` → DS103; `... DISABLE ROW LEVEL SECURITY;` → DS109. The next paragraph ("Silence is not exclusion") contradicts the sentence, so the section states both a rule and its negation.

## Fourth: a surviving mutant — the new statement-index coupling is unpinned

Beyond their M1/M2/M4 (all reproduced: M1 rc=1 on the index rows, M2 rc=1 on exactly the two rows they name, M4 rc=1 on the DELETE FROM pg_enum row), and my over-correction M5 (`return changes, scopedOut`, dropping the `len(changes)==0` guard) which is correctly caught by the pre-existing `only_the_in-scope_target_of_a_mixed_drop_is_reported` row, I built the inverse control they did not:

    M7: excludedStatements[idx] || excludedStatements[idx+1]

`go test ./migration/lint/ -count=1` → **rc=0, package fully green**. M7 silences the statement immediately *before* a scoped-out one. Every fixture in the two new tests is a single-statement version, so nothing pins that the exclusion applies only to the statement that was scoped out. Head behaves correctly on that shape — I measured a two-statement file (`ALTER TABLE public.t ADD CONSTRAINT t_id_key UNIQUE (id);` then `CREATE INDEX idx ON app.users (id);`): oracle 1 change/1 diag, base 2/2, head 1/1 — but no test would catch a regression in it.

## What did hold

Their eight rows reproduce on my own container (ptah1249r-dev, port 15249, removed) with my own fixtures. `CREATE INDEX 

### What it says must change

The two landed shapes are right and worth keeping; the exclusion rule is too coarse. Three things need to change before this is landable.

1. A statement must be excluded only when *every* schema-qualified object it names is out of scope — not only the one the node's primary name happens to carry. For `ALTER TABLE app.child ADD CONSTRAINT ... FOREIGN KEY ... REFERENCES public.parent (id)`, `ast.ForeignKeyConstraint`'s referenced table is a second named object and it is under review, so the statement is not scoped out and PG306 survives. Concretely: in the `*ast.AlterTableNode` branch of `nodeSchemaChanges`, when `!scope.allowsObject(n.Name)`, walk the operations for referenced tables before returning `(nil, true)` and return `(nil, false)` if any of them is in scope. The cheap, defensible alternative if the AST walk is too wide for one PR: keep the exclusion but restrict it to statements whose text names exactly one qualified reference, using the same `tableref` scan `droppedTablesNotCreated` already uses.

2. Add the two missing fixtures. A cross-schema FK row (`ALTER TABLE app.child ADD CONSTRAINT ... REFERENCES public.parent (id)` under scope `public` still reports PG306), and a two-statement version where an in-scope statement precedes a scoped-out one — the second is what kills the M7 neighbor-leak mutant that survives the whole package today.

3. Fix the prose. Delete "every ALTER TABLE form that would yield a node with no operations fails to parse instead" from `changes_scope_internal_test.go` (`ALTER TABLE t` parses to a zero-operation `*ast.AlterTableNode`), and either drop "always" from "the reported schema-change count and the reported diagnostics always describe the same statements" in `docs/site/src/content/docs/atlas/migrate-commands.md` or rewrite the pair so the doc does not state a rule and its negation two paragraphs apart. While there, list the residuals the sweep found: `CREATE SEQUENCE app.s`, `CREATE TRIGGER ... ON app.t` and `CREATE POLICY ... ON app.t` are still measured by a name that carries no schema, so they count a change (and PG308 fires) under `search_path=public` where the community binary counts none. Either extend `nodeScopeReference` to them or file them, but do not leave the fix looking complete.

Ready-to-post PR title and body for the corrected change:

---
Title: Decide a lint statement's schema scope once, by every object it names (#1249)

Body:

`ptah-compat migrate lint` restricts analysis to the schema the dev URL selects, and the filter was applied in two places that disagreed. Measured on PostgreSQL 17.10 against the pinned Atlas community binary v1.3.0, with a dev URL carrying `?search_path=public`:

| version 2 | community v1.3.0 | before | after |
| --- | --- | --- | --- |
| `ALTER TABLE app.users ADD CONSTRAINT users_id_key UNIQUE (id);` | 0 changes / 0 diagnostics | 0 / 1 (PG105) | 0 / 0 |
| `CREATE INDEX idx ON app.users (id);` | 0 / 0 | 1 / 1 (PG101) | 0 / 0 |
| `ALTER TABLE app.child ADD CONSTRAINT c FOREIGN KEY (pid) REFERENCES public.parent (id);` | 0 / 0 | 0 / 1 (PG306) | 0 / 1 (PG306) |
| `ALTER TABLE public.t DROP COLUMN c;` (control) | 1 / 1 (DS103) | 1 / 1 | 1 / 1 |
| `ALTER TABLE app.t DROP COLUMN c;` (control) | 0 / 0 | 0 / 0 | 0 / 0 |
| `CREATE INDEX idx ON public.users (id);` (control) | 1 / 0 | 1 / 1 (PG101) | 1 / 1 (PG101) |

An index is measured by the table it is on rather than by its own name: `ast.IndexNode.Name` and `ast.DropIndexNode.Name` both document themselves as the raw, unqualified index identifier with the namespace derived from `Table`.

The scope decision is made once per statement and drives both outputs. `extractSchemaChanges` reports which statements the scope removed, and `keepFindings` consults that before falling back to subject matching.

A statement is removed only when *every* object it names is out of review. The third row is why: `ADD CONSTRAINT ... FOREIGN KEY ... REFERENCES public.parent` takes a SHARE ROW EXCLUSIVE lock on `public.pa

### DIFF
--- a/docs/site/src/content/docs/atlas/migrate-commands.md
+++ b/docs/site/src/content/docs/atlas/migrate-commands.md
@@ -897,20 +897,30 @@ review and filters nothing, which is also what every non-PostgreSQL dev URL
 does. A `search_path` naming more than one schema is not a scope: it is read as
 a single schema name, so it scopes nothing and every object stays under review.
 
-An `ALTER TABLE` belongs to its table's schema, never to the column or
-constraint it names, and a `CREATE SCHEMA` is measured against the reviewed
-schema by its own name.
+An `ALTER TABLE`'s schema changes belong to its table's schema, never to the
+column or constraint it names, and a `CREATE SCHEMA` is measured against the
+reviewed schema by its own name.
 
-The scope decision is made once per statement and drives both outputs, so the
-reported schema-change count and the reported diagnostics always describe the
-same statements. A diagnostic that names no object at all — the rules that
-report a statement rather than the objects in it — follows the decision already
-taken for the statement it belongs to: dropped when the scope removed that
-statement, and otherwise always reported, since there is nothing to measure a
-scope against and a hazard must not be silenced on an unestablished boundary.
-`ALTER TABLE app.users ADD CONSTRAINT …` under `search_path=public` therefore
-reports nothing, where it used to raise `PG105` about a change it had already
-counted as zero.
+The scope decision is made once per statement and drives both outputs, so a
+statement the scope removed contributes neither a schema change nor a
+diagnostic, and one it kept can contribute either. A diagnostic that names no
+object at all — the rules that report a statement rather than the objects in
+it — follows the decision already taken for the statement it belongs to: dropped
+when the scope removed that statement, and otherwise reported, since there is
+nothing to measure a scope against and a hazard must not be silenced on an
+unestablished boundary. `ALTER TABLE app.users ADD CONSTRAINT …` under
+`search_path=public` therefore reports nothing, where it used to raise `PG105`
+about a change it had already counted as zero.
+
+A statement is removed only when **every** table it names is out of review, not
+only the one it alters. `ALTER TABLE app.child ADD CONSTRAINT c FOREIGN KEY
+(pid) REFERENCES public.parent (id);` under `search_path=public` names two, and
+validating that key holds a `SHARE ROW EXCLUSIVE` lock on `public.parent` for
+the duration, so `PG306` is reported: the hazard lands on a table the run is
+responsible for. The same statement referencing `app.parent` reports nothing.
+This is one place the two outputs deliberately describe different statements —
+the constraint lands on `app.child`, so the reviewed schema still counts zero
+changes for it. A count of zero is not a statement of safety.
 
 Silence is not exclusion. A statement outside Ptah's SQL grammar is left under
 review whatever schema it names, because a boundary that could not be read must
@@ -919,7 +929,19 @@ parser does not model it, so `DROP INDEX public.idx;` counts no schema change
 even in the reviewed schema — the pinned community binary counts one — while
 its `PG106` diagnostic is reported in every schema. That missing change is a
 parser gap rather than a scope decision, tracked in
-[`stokaro/ptah#1296`](https://github.com/stokaro/ptah/issues/1296).
+[`stokaro/ptah#1296`](https://github.com/stokaro/ptah/issues/1296). The same
+grammar boundary is why `TRUNCATE app.users;` and `DROP FUNCTION app.recalc();`
+are reported under `search_path=public` while counting no schema change.
+
+Three constructs are still measured by a name that carries no schema, because
+Ptah's parser records none for them. Measured on PostgreSQL 17.10 under
+`search_path=public`, `CREATE SEQUENCE app.s;` counts one schema change,
+`CREATE TRIGGER trg … ON app.t;` counts one and raises `PG308`, and
+`CREATE POLICY p ON app.t;` counts one — where the pinned community binary
+v1.3.0 counts no schema change and raises no diagnostic for all three. They
+over-report rather than under-report, and each is a warning at exit 0, so a
+scoped run is never told less than the truth about its database. They are listed
+here rather than left to be discovered.
 
 The boundary applies to `ptah-compat migrate lint` only. Native
 `ptah migrations lint` keeps every object under review, whatever the dev URL
--- a/docs/site/src/content/docs/atlas/migrate-commands.md
+++ b/docs/site/src/content/docs/atlas/migrate-commands.md
@@ -886,7 +886,11 @@ objects the run analyzes. A PostgreSQL-family `--dev-url` carrying
   review, and so does a reference written out with the reviewed schema's own
   name;
 - one statement is measured per object: `DROP TABLE users, other.audit_log;`
-  under `search_path=public` reports `users` and counts one schema change.
+  under `search_path=public` reports `users` and counts one schema change;
+- an index is measured by the schema of the table it is on, because an index
+  name never carries one: `CREATE INDEX idx ON app.users (id);` under
+  `search_path=public` counts no schema change and raises no diagnostic, while
+  the same statement on a `public` table counts one and raises `PG101`.
 
 A `--dev-url` that names no schema puts the whole connected database under
 review and filters nothing, which is also what every non-PostgreSQL dev URL
@@ -895,10 +899,27 @@ a single schema name, so it scopes nothing and every object stays under review.
 
 An `ALTER TABLE` belongs to its table's schema, never to the column or
 constraint it names, and a `CREATE SCHEMA` is measured against the reviewed
-schema by its own name. A diagnostic that names no object at all — the rules
-that report a statement rather than the objects in it — is always reported,
-since there is nothing to measure a scope against and a hazard must not be
-silenced on an unestablished boundary.
+schema by its own name.
+
+The scope decision is made once per statement and drives both outputs, so the
+reported schema-change count and the reported diagnostics always describe the
+same statements. A diagnostic that names no object at all — the rules that
+report a statement rather than the objects in it — follows the decision already
+taken for the statement it belongs to: dropped when the scope removed that
+statement, and otherwise always reported, since there is nothing to measure a
+scope against and a hazard must not be silenced on an unestablished boundary.
+`ALTER TABLE app.users ADD CONSTRAINT …` under `search_path=public` therefore
+reports nothing, where it used to raise `PG105` about a change it had already
+counted as zero.
+
+Silence is not exclusion. A statement outside Ptah's SQL grammar is left under
+review whatever schema it names, because a boundary that could not be read must
+not be able to drop a diagnostic. `DROP INDEX` is the current example: Ptah's
+parser does not model it, so `DROP INDEX public.idx;` counts no schema change
+even in the reviewed schema — the pinned community binary counts one — while
+its `PG106` diagnostic is reported in every schema. That missing change is a
+parser gap rather than a scope decision, tracked in
+[`stokaro/ptah#1296`](https://github.com/stokaro/ptah/issues/1296).
 
 The boundary applies to `ptah-compat migrate lint` only. Native
 `ptah migrations lint` keeps every object under review, whatever the dev URL
--- a/migration/lint/analysis.go
+++ b/migration/lint/analysis.go
@@ -3,6 +3,7 @@ package lint
 import (
 	"fmt"
 	"io/fs"
+	"maps"
 	"path"
 	"slices"
 	"sort"
@@ -154,7 +155,13 @@ type File struct {
 	// can contribute zero, one, or several changes, so len(Changes) is not the
 	// statement count. Ordered by statement, then by the order changes appear
 	// within each statement.
-	Changes         []SchemaChange
+	Changes []SchemaChange
+	// scopeExcluded holds the statement indexes the reviewed-schema filter left
+	// out, so a finding about one of them is dropped by the same decision that
+	// dropped its change. Without it the two disagreed: a rule that attaches no
+	// subject produced a finding for a statement the scope had already removed
+	// from the counts (stokaro/ptah#1249).
+	scopeExcluded   map[int]bool
 	suppressedRules []atlaslint.Target
 	// compatibility is the profile this run was started with. A rule reads it
 	// when the two command surfaces model the same statement differently; see
@@ -235,6 +242,7 @@ func cloneFiles(files []File) []File {
 func cloneFile(file File) File {
 	file.suppressedRules = slices.Clone(file.suppressedRules)
 	file.Changes = slices.Clone(file.Changes)
+	file.scopeExcluded = maps.Clone(file.scopeExcluded)
 	statements := file.Statements
 	file.Statements = make([]Statement, len(statements))
 	for i := range statements {
@@ -352,7 +360,7 @@ func AnalyzeFS(fsys fs.FS, opts Options) (Analysis, error) {
 			continue
 		}
 		file := cloneFile(files[i])
-		findings = append(findings, scope.keepFindings(runRules(&file, opts, rules))...)
+		findings = append(findings, scope.keepFindings(file.scopeExcluded, runRules(&file, opts, rules))...)
 	}
 	sort.SliceStable(findings, func(i, j int) bool {
 		if findings[i].File != findings[j].File {
@@ -562,7 +570,7 @@ func prepareFile(
 				suppressedRules: rawStmt.suppressedRules,
 			})
 		}
-		file.Changes = extractSchemaChanges(&file, dialect, scope)
+		file.Changes, file.scopeExcluded = extractSchemaChanges(&file, dialect, scope)
 	}
 	return file, nil
 }
--- a/migration/lint/changes.go
+++ b/migration/lint/changes.go
@@ -1,6 +1,7 @@
 package lint
 
 import (
+	"slices"
 	"strings"
 
 	"go.5x5.cz/ptah/core/ast"
@@ -168,7 +169,11 @@ func nodeSchemaChanges(file *File, stmt Statement, node ast.Node, scope schemaSc
 		// Every action of an ALTER TABLE belongs to the table's schema, never
 		// to the column or constraint it names.
 		if !scope.allowsObject(n.Name) {
-			return nil, true
+			// The altered table is not the only table the statement names. See
+			// [alterReferencedTables]: a statement that reaches into the
+			// reviewed schema has not left it, so it is not scoped out, even
+			// though it still contributes no change there.
+			return nil, !slices.ContainsFunc(alterReferencedTables(n), scope.allowsObject)
 		}
 		out := make([]SchemaChange, 0, len(n.Operations))
 		for _, op := range n.Operations {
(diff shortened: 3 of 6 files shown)


denisvmedia:

denisvmedia (commented):
Independent review of head `4d196b71` found a reproduced scope blocker.

**Blocker — statement-wide exclusion drops an in-scope cross-schema FK finding.** `migration/lint/scope.go:115-123` classifies an `ALTER TABLE` statement only from its subject table, then drops every finding at that statement index. For `ALTER TABLE app.child ... REFERENCES public.parent` under `--schema public`, the expected PG306 for `public.parent` disappears (`got [] want ["PG306"]`).

Scope every relevant object, or at minimum do not let statement exclusion override a finding whose own subject is in scope. Add a public-API regression for this exact shape.

The new “always describe the same statements” docs claim is consequently false, the zero-operation ALTER white-box justification is not valid because the shape is public-API reachable, and the added single-statement tests do not kill the PR body's `idx+1` mutant. Those tests and docs need correction before merge.

denisvmedia:
Thanks — the blocker is real, it reproduced twice, and it is fixed. Taking the four points in order.

**1. Blocker: statement-wide exclusion drops an in-scope cross-schema FK finding. Confirmed and fixed.**

Reproduced first through the public API on my own fixture (`lint.AnalyzeFS`, base creates `public.parent` / schema `app` / `app.child`, version 2 is the FK statement, scope `public`): `9820496e` → `[PG306]`, `4d196b71` → `[]`. Then reproduced at the CLI on my own container (PostgreSQL 17.10, port 15300, CI-shaped role, `lintdev` dropped `WITH (FORCE)` and recreated before every run, `atlas.sum` generated by the oracle's `migrate hash`), same command for all binaries:

| binary | changes / diagnostics / rc |
| --- | --- |
| pinned community v1.3.0 | 0 / 0 / 0 |
| ptah-compat @ 9820496e (base) | 0 / 1 PG306 / 0 |
| ptah-compat @ 4d196b71 (head) | **0 / 0 / 0** |
| ptah-compat @ bcad99a4 (now) | **0 / 1 PG306 / 0** |

I took your primary suggestion rather than the `tableref` text-scan fallback: `alterReferencedTables` walks the operations in the `*ast.AlterTableNode` branch and returns `(nil, false)` when any referenced table is in scope. The AST already carries what is needed — `AddConstraintOperation.Constraint.Reference.Table` is `"public.parent"`, schema intact — and it also gave me the inline spelling for free, `AddColumnOperation.Column.ForeignKey.Table`, which the text scan would have had to re-derive.

You are right that the doc-comment argument was backwards for PG306: the node knows one table, the statement names two. The reversed framing is now in the code comment and on the doc page.

One thing worth stating explicitly, because it looks like a half-fix and is not: the change count still stays **zero** on that row. The constraint lands on `app.child`, which is out of review, so zero is the right count and it matches the oracle. What changed is only whether the statement is *removed*. "Not scoped out" and "contributes a change" are different answers, and every new fixture row asserts both so a future fix cannot conflate them.

**2. `CREATE SEQUENCE` / `CREATE TRIGGER` / `CREATE POLICY`. Confirmed, measured against the oracle, documented, and filed rather than fixed here.**

You gave the oracle column for the sequence only, so I measured all three on the pinned binary. Under `search_path=public`, all three: community v1.3.0 = 0 changes / 0 diagnostics / rc 0; ptah-compat = 1 change each, and the trigger additionally raises PG308; rc 0 everywhere. Unchanged by this PR in either direction.

`nodeScopeReference` cannot reach them the way it reaches an index: `CREATE SEQUENCE app.s` parses to `*ast.CreateSequenceNode{Name: "s"}` with the schema already discarded, so there is no `Table` to route to. Fixing them means teaching the parser to keep the qualifier — the same class of change as #1296 — and I am not folding a parser change into this PR. What I did instead is what you asked for as the alternative: they are on the page with their measurements, named as over-reporting at exit 0, and filed as follow-ups. The fix no longer looks complete.

**3. Two false prose claims. Both corrected; one with a nuance.**

The white-box justification is deleted. `parser.NewParser("ALTER TABLE app.users", parser.WithDialect("postgres")).Parse()` returns `*ast.AlterTableNode` with `ops=0` and `err=nil`, exactly as you measured. But the rows have not moved to a public fixture, and here is why: through `AnalyzeFS` under scope `public`, `ALTER TABLE public.t;` and `ALTER TABLE app.users;` both produce `changes=[] findings=[]`, because no rule fires on a bare `ALTER TABLE`. A public fixture there would pass with either answer to the scope question, so it would not be coverage. The justification now says the truthful thing — reachable, but with no observable consequence — with that measurement in it, and names what would make it observable.

The doc's "always describe the same statements" is gone. On your list of counterexamples: they are all real, but the mechanism is not the scope. `TRUNCATE`, `DROP SCHEMA … CASCADE`, `DROP CONSTRAINT k`, `ALTER COLUMN … TYPE` and `DISABLE ROW LEVEL SECURITY` every one fail to parse (`unsupported SQL statement: TRUNCATE`, `unsupported DROP target: SCHEMA`, `unsupported SQL statement: K`, `unsupported column attribute: VARCHAR`, `unsupported ALTER operation: DISABLE ROW LEVEL SECURITY`), so `statementSchemaChanges` returns `(nil, false)` and they are never marked excluded — that is the "Silence is not exclusion" paragraph doing its job, which is why the section read as a rule and its negation. The cross-schema foreign key is the counterexample that actually parses and is actually kept, so the rewritten section leads with it, then points at the grammar boundary for the rest.

**4. The surviving `idx+1` mutant. Reproduced, and now dead — along with its inverse.**

`excludedStatements[idx] || excludedStatements[idx+1]` did survive the entire package at `4d196b71`. It now reddens two rows of a new `TestAnalyzeFS_SchemaScopeExcludesOnlyTheScopedOutStatement`. I added the control you did not need to: `idx-1`, which reddens the other two rows, so both neighbors are pinned rather than just the one your mutant probed. Your two-statement measurement is one of the fixture rows, and I re-measured it end to end: oracle 1 change / 1 diagnostic (MF101), base 2/2, now 1/1 (PG105).

Your M5 (`return changes, scopedOut`) still dies on the pre-existing `only_the_in-scope_target_of_a_mixed_drop_is_reported`, and M1/M2/M4 still die exactly where the PR body says.

**Also worth flagging, and not yours to fix:** my first trigger fixture put `CREATE FUNCTION app.f() … LANGUAGE plpgsql` in the base file, and ptah-compat exits **1** there — "postgres migration replay rejects CREATE routine definition because its effects cannot be confined to the disposable database realm" — where the oracle exits 0. That is an rc-level divergence, it is present in base as well as head, it has nothing to do with this PR, and I have filed it separately rather than chase it here. The trigger row above was re-measured with the built-in `suppress_redundant_updates_trigger()` to keep it off the critical path.

All gates are 0, including every `check:*` in `docs/site/package.json`. `check:responsive` needed `npm ci` and `npm run build` first — it exits 1 with "dist not found" in a fresh worktree.

Issue #1258, opened 2026-08-07
MySQL: a foreign key whose backing index MySQL auto-named plans a duplicate CREATE INDEX and fails the apply


denisvmedia:
Residue after #1250 (which closed #1244). Found by adversarial review of that PR and reproduced here on master (`17b05f52`) against a live MySQL 9.7.

## The shape

The ordinary MySQL layout: a bare `CONSTRAINT ... FOREIGN KEY` with no explicit `KEY`. MySQL auto-creates the backing index **under the constraint's own name**.

```sql
CREATE TABLE users (id INT NOT NULL, PRIMARY KEY (id));
CREATE TABLE posts (id INT NOT NULL, user_id INT NOT NULL, PRIMARY KEY (id),
  CONSTRAINT fk_posts_user FOREIGN KEY (user_id) REFERENCES users(id));
```

`information_schema.STATISTICS` then reports index names `fk_posts_user` and `PRIMARY` on `posts`.

## The divergence

Desired state is that database's **own unedited** `schema inspect` output from the pinned community binary v1.3.0, applied back to the database it came from:

| | verdict |
| --- | --- |
| pinned binary, `--dry-run` | exit 0, `Schema is synced, no changes to be made` |
| ptah-compat, `--dry-run` | exit 0, but plans `` CREATE INDEX `fk_posts_user` ON `fk1`.`posts` (`user_id`); `` |
| ptah-compat, `--auto-approve` | **exit 1**, `Error 1061 (42000): Duplicate key name 'fk_posts_user'` |

## The control that isolates it

The same schema with the backing index given its own distinct name:

```sql
KEY idx_posts_user (user_id),
CONSTRAINT fk_posts_user FOREIGN KEY (user_id) REFERENCES users(id)
```

→ ptah-compat exits 0, `Schema is synced, no changes to be made`.

So the defect is precisely: **Ptah does not recognize the auto-created backing index as belonging to the foreign key when the two share a name.** It reads the index as absent from the desired state and plans to create it, and MySQL then refuses the duplicate.

## Why it matters more than the count suggests

This is not an exotic fixture — it is what you get from a plain `CONSTRAINT ... FOREIGN KEY` with no index clause, which is how most MySQL schemas are written. #1250 made `schema apply` converge for schemas whose backing indexes happen to carry distinct names, and this shape was not in its fixture set.

It is also distinct from #1245 (a unique index added as an index and dropped as a constraint), though both live in the same area of MySQL index/constraint identity.

[...]

PR #1301, opened 2026-08-08
Compare a unique key once, through the spelling the schema declares (#1245)

### LINKED ISSUE #1245: A MySQL unique index is added as an index and dropped as a constraint in the same plan

denisvmedia:
Fixes #1245.

Replaying a MySQL database's own `schema inspect` output against the database it came from planned, in one plan:

```sql
CREATE UNIQUE INDEX `uq_users_email` ON `w3a98_tgt`.`users` (`email`);
ALTER TABLE `users` DROP INDEX `uq_users_email`;
```

The pinned Atlas CE v1.3.0 binary reports `Schema is synced` for the same pair. Measured on MySQL 9.7.1 and MariaDB 11.8.8 the apply exits 1 with `Error 1061 (42000): Duplicate key name 'uq_users_email'` — and not before the statements ahead of it in the same plan have run, so adding one column to a table that has a named unique key left the column added and the migration half applied.

## Why

Every engine Ptah supports except SQL Server enforces a `UNIQUE` constraint with an index of the constraint's own name on the constraint's own table, and introspection reports that one object twice — once in the index catalog, once in the constraint catalog. MySQL and MariaDB do not even have a separate notion: `ADD CONSTRAINT c UNIQUE (a)` and `CREATE UNIQUE INDEX c ON t (a)` leave the identical catalog row, which is why `schema inspect` writes MySQL uniqueness back out as `index { unique = true }` on both binaries and never as a constraint.

The database index was filtered out of the index pool unconditionally. The desired side is never filtered, so a desired state that spells the object as an index had nothing to match: index comparison reported it **added**, and constraint comparison, finding no `UNIQUE` constraint on the desired side, reported the same name **removed**.

## The fix is not a dialect rule

**Ownership follows the desired state's spelling.** An identity the desired state declares as an index reaches index comparison, whichever filter would otherwise have claimed it, and constraint comparison leaves that object alone. A desired state that spells uniqueness as a constraint, or does not mention the object at all, is unchanged — the constraint pool still owns it, and an undeclared unique key is still reported removed. The rule cannot start dropping anything: an identity only survives the filters when the desired state names it, and an identity the desired state names is matched rather than removed.

This is the same narrowing #1258 applied to the FOREIGN KEY backing-index filter, now applied to all three filters instead of one, so `hiddenForeignKeyBackingIndexes` folds into `collectDatabaseIndexes` and its reasoning moves there.

**Three suppression branches, not one.** The issue named the same-name `UNIQUE` constraint identity. `isConstraintBasedUniqueIndex`'s name pattern is an independent second one: `UNIQUE KEY uk_users_email (email)` never reached the identity check at all and produced the identical unappliable plan. Both are narrowed, with a test for each.

**Not MySQL-only.** The same shape is live on PostgreSQL 17.10, and there the `IF NOT EXISTS` / `IF EXISTS` guards made it worse rather than better: the create was skipped, the drop took the constraint and its index with it, the command exited **0** and reported success, and `pg_indexes` afterwards listed only `users_pkey`. A silent deletion of the uniqueness the desired state asked for.

## The other half: two objects that share a name stay two

Letting a declared identity through means a database unique key finally has a counterpart to be matched against — and MySQL had no index-definition comparison at all (#1272 left every non-PostgreSQL dialect on a name-only match), so a desired plain `index "uq_users_email"` would have silently satisfied a database `UNIQUE KEY uq_users_email`. `indexDefinitionsChanged` gains a MySQL/MariaDB branch comparing uniqueness and the key columns, both measured against the pinned binary.

Nothing else is compared there, and the omissions are the reader's, not a judgement. Ptah's MySQL reader projects `information_schema.STATISTICS` through `GROUP_CONCAT(COLUMN_NAME)` and keeps only the column names and `NON_UNIQUE`, so a descending key and a prefix key both arrive as plain columns; comparing a direction the reader always reports as ascending would plan a rebuild on every run forever, which is the oscillation this PR exists to remove. An expression key never reaches the comparison — `COLUMN_NAME` is `NULL` and the read fails first. Both are reader gaps, filed separately.

## Measured

`schema apply`, my own containers on my own ports, every `--dev-url` database recreated between runs, all removed afterwards.

| Target | Fixture | pinned v1.3.0 | before | after |
| MySQL 9.7.1 | unchanged `UNIQUE KEY uq_users_email (email)` | 0, synced | **1**, `Error 1061` | 0, synced; run 2 also 0 |
| MySQL 9.7.1 | `UNIQUE KEY uk_users_email (email)` | 0, synced | 0, `CREATE`+`DROP` (unappliable) | 0, synced |
| MySQL 9.7.1 | `UNIQUE KEY uq_membership (tenant_id, user_id)` + a plain key | 0, synced | 0, `CREATE`+`DROP` | 0, synced |
| MySQL 9.7.1 | `ADD COLUMN` beside a named unique key | — | **1**, column added, run half applied | 0, column added; run 2 synced |
| MariaDB 11.8.8 | unchanged `UNIQUE KEY uq_users_email (email)` | 0, synced | **1**, `Error 1061` | 0, synced; run 2 also 0 |
| PostgreSQL 17.10 | `CONSTRAINT uq_users_email UNIQUE (email)` vs a desired `index { unique = true }` | 0, synced | **0, and the unique index deleted** | 0, synced; index intact |

Removal still happens where it should. With a desired state that declares neither spelling, the pinned binary plans `ALTER TABLE "users" DROP CONSTRAINT "uq_users_email"` on PostgreSQL and `ALTER TABLE users DROP INDEX uq_users_email` on MySQL, and so does Ptah after this change. The PostgreSQL `unique "…"` block that a PostgreSQL `schema inspect` emits reports synced before and after.

Genuinely different objects still report a change, and now in an order the server accepts. On MySQL 9.7.1, a desired plain `index "uq_users_email"` against a database `UNIQUE KEY uq_users_email` plans `DROP INDEX` then `CREATE INDEX`, which is the pair the pinned binary plans; before, it planned `CREATE` then `DROP` and failed at 1061.

Rule (a) holds on every cell measured: no case where `ptah-compat` exits 0 and the pinned binary exits 1.

## Tests

The invariant #1245 asks for is asserted at the comparison level rather than through rendered SQL: `TestCompareWithDialect_NoObjectIsBothAddedAndRemoved` runs the fixture on five dialects and fails if any table-qualified name is in `IndexesAdded` and `ConstraintsRemoved` at once. On the base tree it is red on mysql, mariadb, postgres and sqlite.

Mutation, both directions, all restored:

- dropping the declaration narrowing reddens the #1245 tests **and** the pre-existing #1258 foreign-key tests together;
- dropping the constraint hand-off reddens the #1245 synced tests;
- removing the MySQL/MariaDB definition branch reddens exactly the two "genuinely different objects" controls on mysql and mariadb and leaves their postgres rows green;
- making every `UNIQUE` constraint hand off regardless of declaration reddens `UndeclaredUniqueKeyIsStillRemoved`;
- matching the hand-off on name alone instead of name-and-table reddens `UniqueKeyOnAnotherTableIsNotTheSameObject`.

## Docs

`docs/conformance.md` gains "A unique constraint and its index are one object, and the schema says which", with the measurement table above and the PostgreSQL exit-0 deletion called out, and the "only PostgreSQL is covered" paragraph on index reconciliation is updated for the new MySQL/MariaDB branch and what it leaves out.

## Found while measuring, not fixed here

All reproduced on the base binary, so none is caused by this change; each will be filed separately.

- `ptah-compat` refuses the `maria://` URL scheme the pinned binary accepts (`mariadb://` works).
- MySQL introspection fails outright on a functional index: `converting NULL to string is unsupported`, where the pinned binary inspects it fine.
- Dev-database baseline materialization loses a named MySQL unique index, because the DB-to-Go conversion drops constraint-backed indexes and falls back to the column's `UNIQUE`; any plan that drops such an index then fails simulation with `Error 1091`.
- `schema diff --from` a MySQL database `--to` its own inspect output plans a full `CREATE TABLE` + `DROP TABLE` where the pinned binary reports synced, on a control with no unique object at all — the #1244 table-identity split still live on the `diff` path.

---

### What it found

SETUP. Head a7952d3a checked out in my worktree /Users/buster/Work/denis/ptah/.claude/worktrees/wf_b394c0e2-0e4-7; control worktree at the merge-base 9820496e in /private/tmp/claude-501/-Users-buster-Work-denis-ptah/2fc9aeb4-cb03-4375-ae0a-82e5047f1051/scratchpad/r1245/base. `git log --oneline -1` run inside each before every build; both binaries chmod +x. My own containers, removed at the end: r1245-mysql (mysql:9.7 -> 9.7.1, port 33141), r1245-maria (mariadb:11.8 -> 11.8.8-MariaDB-ubu2404, port 33142), r1245-pg (postgres:17 -> 17.10, port 16145, roles ptah_user SUPERUSER and bootstrap postgres). Every --dev-url database dropped and recreated before each run. Oracle: /Users/buster/Work/denis/ptah-atlas-conformance/bin/atlas, "atlas community version v1.3.0". All fixtures built by me, not copied from the report.

WHAT REPRODUCES (their headline claims hold).
- MySQL 9.7.1, `UNIQUE KEY uq_users_email (email)`, desired = that database's own unedited pinned `schema inspect`: pinned exit 0 "Schema is synced"; base exit 0 `CREATE UNIQUE INDEX ...` + `ALTER TABLE users DROP INDEX uq_users_email`; head exit 0 "Schema is synced."
- MariaDB 11.8.8, same fixture: pinned 0 synced; base 0 CREATE + `DROP INDEX IF EXISTS`; head 0 synced.
- MySQL, ADD COLUMN nickname riding along: head run 1 emits only `ALTER TABLE ... ADD COLUMN nickname varchar(64)`, exit 0; run 2 synced; SHOW CREATE TABLE still has `UNIQUE KEY uq_users_email`.
- PostgreSQL 17.10, DB `CONSTRAINT uq_users_email UNIQUE (email)` vs desired `index "uq_users_email" { unique = true, columns = [column.email] }`: pinned 0 synced, head 0 synced, base 0 with `CREATE UNIQUE INDEX IF NOT EXISTS` + `DROP CONSTRAINT IF EXISTS` and pg_indexes afterwards holding only users_pkey.
- Non-regression rows hold: PG `unique "..."` block desired state -> head synced; undeclared unique key still removed (head emits `ALTER TABLE users DROP INDEX uq_users_email` on MySQL).
- Base-tree red confirmed by copying their test file into the control worktree: MySQLUnchangedUniqueIndexIsSynced (mysql, mariadb), NoObjectIsBothAddedAndRemoved (mysql, mariadb, postgres, sqlite; sqlserver PASS), NameHeuristicUniqueKeyIsSynced (both rows), DeclaredIndexUniquenessStillCompared (all 3), DeclaredIndexColumnsStillCompared (all 3) all FAIL; the three control tests PASS on base. Control file deleted, base worktree clean.
- Gates on head, each exit code read on its own line: go build 0; go vet 0; go vet -tags integration ./integration/... 0; go tool qtlint 0; golangci-lint (private GOLANGCI_LINT_CACHE) 0, "0 issues."; scripts/check-test-style.sh 0 (175 conditional groups, 45 white-box files); scripts/check-public-api.sh 0; scripts/check-public-api-snapshot.sh 0; scripts/check-public-api-released.sh 0; node docs/site/scripts/check-style.mjs 0 (100 files). `go test ./... -count=1` exited 1 with only cmd/viz failing (TestSVGReportsGraphvizStderrOnFailure at 10.00s) under load average 26.85 with 15 concurrent `go test` processes; `go test ./cmd/viz/... -count=1 -p 1 -timeout 20m` -> 0, ok 7.069s. No new test is env-gated or skipped, and none of their edits touched an existing test's expectations (the diff is 2 new files + 2 source files + docs).

REFUTATION 1 -- NEW EXIT-1 ON MySQL 9.7.1 AND MariaDB 11.8.8, ON A DATABASE THE PINNED BINARY CALLS SYNCED.
The new `mysqlIndexDefinitionChanged` / `mysqlIndexKeyColumnsChanged` branch compares `effectiveDatabaseIndexParts`, i.e. `types.DBIndex.Columns`. `internal/dbschema/mysql/reader.go:501` projects the key through `GROUP_CONCAT(s.COLUMN_NAME ORDER BY s.SEQ_IN_INDEX)` and line 528 does `index.Columns = strings.Split(columnsStr, ",")`. That encoding is lossy in two ways the change did not enumerate.

(a) A column name containing a comma -- a legal MySQL identifier.
  CREATE TABLE t2 (id BIGINT NOT NULL AUTO_INCREMENT, `a,b` VARCHAR(64) NOT NULL, PRIMARY KEY (id), KEY idx_weird (`a,b`));
  desired = that database's own unedited pinned inspect (`index "idx_weird" { columns = [column["a,b"]] }`).
    atlas schema inspect --url "mysql://root:root_password@localhost:33141/r1245_e" > mysql_e.hcl
    <bin> schema apply --url "mysql://root:root_password@localhost:33141/r1245_e" --to "file://mysql_e.hcl" --dev-url "mysql://root:root_password@localhost:33141/r1245_dev" --auto-approve
    pinned v1.3.0   exit 0   Schema is synced, no changes to be made
    base 9820496e   exit 0   Schema is synced, no changes to be made.
    head a7952d3a   exit 1   plan: DROP INDEX `idx_weird` ON `t2`; CREATE INDEX `idx_weird` ON `r1245_e`.`t2` (`a,b`);
                             Error: dev database simulation failed during baseline: ... Error 1072 (42000):
                             Key column 'a' doesn't exist in table
                             SQL: CREATE INDEX `idx_weird` ON `t2` (`a`, `b`)
  MariaDB 11.8.8, same fixture, mariadb:// scheme: pinned 0 synced, base 0 synced, head 1 with the same Error 1072 (plan spells `DROP INDEX IF EXISTS`). Target database verified untouched afterwards.

(b) No unusual identifier needed: a 16-part composite index whose column names overflow MySQL's default `group_concat_max_len`. `SELECT @@group_concat_max_len` on mysql:9.7 is 1024; 16 names of 64 chars plus separators is 1039 bytes, so the reader receives the last name truncated. Same three-way result: pinned exit 0 synced, base exit 0 synced, head exit 1 `Error 1072 (42000): Key column 'abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvw' doesn't exist in table`. MariaDB's default is 1048576, so (b) is MySQL-specific.

Causation, mutant control: I reverted only the new arm -- `case platform.MySQL, platform.MariaDB: return false` in `indexDefinitionsChanged` -- rebuilt ptah-compat, and BOTH fixtures returned to exit 0 "Schema is synced." The same mutant reddens exactly `TestCompareWithDialect_DeclaredIndexUniquenessStillCompared/{mysql,mariadb}` and `.../DeclaredIndexColumnsStillCompared/{mysql,mariadb}` and leaves their /postgres rows green -- it is the author's own A3 mut

### What it says must change

Send it back. The #1245 ownership rule itself is sound and well-tested; the two things bolted onto it are not.

1. Do not derive MySQL/MariaDB key columns from a comma-joined projection. `internal/dbschema/mysql/reader.go:497-536` must select one row per (INDEX_NAME, SEQ_IN_INDEX) and assemble Columns/Parts in Go -- which also removes the `COLUMN_NAME IS NULL` scan failure on expression indexes that the report already filed. Until the reader is fixed, the MySQL/MariaDB arm of `indexDefinitionsChanged` must not compare key columns; compare uniqueness only, which is all #1245 needs to keep a plain index from matching a UNIQUE KEY.
2. Add both fixtures to migration/schemadiff and re-measure the MySQL neighborhood live: a column named `a,b`, and a 16-part composite index whose names exceed `group_concat_max_len` (1024 by default on MySQL). Both are exit 1 today on a database the pinned v1.3.0 binary reports synced, and both were green on 9820496e.
3. Decide what an index removal means for a constraint-backed object on PostgreSQL. Either render it as `ALTER TABLE ... DROP CONSTRAINT` (what the pinned binary emits and the only form the server accepts), or restrict the hand-off to MySQL/MariaDB, where a unique key and its index are literally one catalog row. Then run the author's own "two different objects" controls on PostgreSQL, not only on MySQL.
4. Delete or correct the three claims the measurement falsifies: "The rule cannot start dropping anything ... matched rather than removed" in docs/conformance.md and in the `collectDatabaseIndexes` doc comment, and "splitting it across the two pools would plan a DROP CONSTRAINT and a DROP INDEX for the same row" in unique_index_ownership.go.

Nothing was committed in my worktree: the finding is measurement, and a speculative reader rewrite would need its own dialect sweep and documentation pass. Both worktrees are clean at a7952d3a and 9820496e; all three containers removed.

Ready-to-post PR title for the corrective change:

  Read a MySQL key one column at a time (#1245)

Body:

  #1245 taught MySQL and MariaDB to compare an index's key columns. The two sides
  of that comparison do not carry the same information: `internal/dbschema/mysql/reader.go`
  projects the key through `GROUP_CONCAT(COLUMN_NAME ORDER BY SEQ_IN_INDEX)` and splits the
  result on a comma, so a key survives the read only when no column name contains a comma and
  the whole list fits in `group_concat_max_len`.

  Measured on MySQL 9.7.1 and MariaDB 11.8.8, replaying a database's own unedited
  `schema inspect` output from the pinned Atlas CE v1.3.0 binary, which reported the schema
  synced for both fixtures:

  | fixture | before #1245 | after #1245 |
  | --- | --- | --- |
  | `KEY idx_weird (\`a,b\`)` | exit 0, synced | exit 1, `Error 1072 (42000): Key column 'a' doesn't exist in table` |
  | 16-part key, names past `group_concat_max_len` (1024, MySQL default) | exit 0, synced | exit 1, same error naming the truncated column |

  The plan is `DROP INDEX` followed by a `CREATE INDEX` over the mis-split names, so it never
  converges; on MySQL the dev-database baseline refuses to materialize and the run stops
  before the target is touched. MariaDB's default `group_concat_max_len` is 1048576, so only
  the comma fixture reaches it there.

  The read now returns one row per key part and assembles the column list in Go. That also
  ends the scan failure on an expression key, where `COLUMN_NAME` is NULL and the
  concatenation collapsed the whole row.

  Reverting the assembly reddens the two new fixtures on both engines; widening the
  comparison to a key's direction reddens the descending-key round trip, which is the
  reader gap #1245 documented and this change does not close.

### DIFF
--- a/cmd/compare/report_coverage_internal_test.go
+++ b/cmd/compare/report_coverage_internal_test.go
@@ -20,12 +20,16 @@ import (
 )
 
 // nonCategoryFields names the SchemaDiff fields that do not carry schema
-// changes, with the reason each is not reported as a difference. A field that
-// is not a list is not a category by construction, and this map is what forces
-// that judgment to be made rather than assumed: a new non-list field fails
-// TestSchemaDiffNonCategoryFieldsAreDocumented until it is listed here.
+// changes, with the reason each is not reported as a difference. Everything
+// that is not a list is one by construction, and a list can be one too when it
+// qualifies another list rather than adding to it. This map is what forces that
+// judgment to be made rather than assumed: a field missing from both the
+// categories and this map fails TestSchemaDiffNonCategoryFieldsAreDocumented,
+// and a list listed here is a claim that reporting it would say nothing the
+// list it qualifies does not already say.
 var nonCategoryFields = map[string]string{
-	"IdentifierSemantics": "records the live catalog identifier rules the diff was produced under, not a difference between the two schemas",
+	"IdentifierSemantics":           "records the live catalog identifier rules the diff was produced under, not a difference between the two schemas",
+	"ConstraintBackedIndexRemovals": "a subset of IndexesRemoved, naming the removals whose object a UNIQUE constraint enforces so the planner spells the drop as a constraint drop; reporting it would print the same removed index a second time, and on its own it removes nothing",
 }
 
 // TestWriteComparisonReportsEveryDiffCategory is the guard for
@@ -123,8 +127,9 @@ ALTER TABLE "other"."secured" ENABLE ROW LEVEL SECURITY;
 }
 
 // diffCategoryFields returns the SchemaDiff fields that carry changes: the
-// exported list fields. It is a helper rather than an inline loop so the test
-// bodies above stay free of control flow.
+// exported list fields, less the ones nonCategoryFields excludes with a stated
+// reason. It is a helper rather than an inline loop so the test bodies above
+// stay free of control flow.
 func diffCategoryFields() []reflect.StructField {
 	structType := reflect.TypeFor[difftypes.SchemaDiff]()
 	var fields []reflect.StructField
@@ -132,16 +137,26 @@ func diffCategoryFields() []reflect.StructField {
 		if !field.IsExported() || field.Type.Kind() != reflect.Slice {
 			continue
 		}
+		if _, excluded := nonCategoryFields[field.Name]; excluded {
+			continue
+		}
 		fields = append(fields, field)
 	}
 	return fields
 }
 
+// diffNonCategoryFieldNames returns every SchemaDiff field the report guard
+// above does not exercise, which is exactly the set that has to carry a written
+// reason.
 func diffNonCategoryFieldNames() []string {
+	categories := make(map[string]struct{})
+	for _, field := range diffCategoryFields() {
+		categories[field.Name] = struct{}{}
+	}
 	structType := reflect.TypeFor[difftypes.SchemaDiff]()
 	var names []string
 	for field := range structType.Fields() {
-		if field.Type.Kind() == reflect.Slice {
+		if _, isCategory := categories[field.Name]; isCategory {
 			continue
 		}
 		names = append(names, field.Name)
--- a/core/ast/nodes.go
+++ b/core/ast/nodes.go
@@ -1075,6 +1075,17 @@ type DropIndexNode struct {
 	// planners set it only for standalone index drops the caller routes into a
 	// no_transaction migration.
 	Concurrently bool
+	// EnforcesUniqueConstraint reports that the index being dropped is the one
+	// a UNIQUE constraint of the same name enforces, so the statement removes a
+	// uniqueness protection and not only an access path.
+	//
+	// It carries no SQL: on MySQL and MariaDB, the engines that reach it, a
+	// unique key and its constraint are one catalog row and DROP INDEX is the
+	// statement that removes it. It exists so risk classification can tell the
+	// two apart, because the spelling cannot: dropping `uq_users_email` and
+	// dropping `idx_users_email` are the same statement shape and a very
+	// different change to the data's guarantees.
+	EnforcesUniqueConstraint bool
 	// Comment is an optional comment for the drop operation
 	Comment string
 }
@@ -1132,6 +1143,14 @@ func (n *DropIndexNode) SetIfExists() *DropIndexNode {
 //	dropIndex.SetConcurrently()
 func (n *DropIndexNode) SetConcurrently() *DropIndexNode {
 	n.Concurrently = true
+	return n
+}
+
+// SetEnforcesUniqueConstraint records that this drop removes the index a
+// UNIQUE constraint of the same name enforces. It changes no SQL; see
+// [DropIndexNode.EnforcesUniqueConstraint].
+func (n *DropIndexNode) SetEnforcesUniqueConstraint() *DropIndexNode {
+	n.EnforcesUniqueConstraint = true
 	return n
 }
 
--- a/dbschema/types/types.go
+++ b/dbschema/types/types.go
@@ -193,6 +193,16 @@ type DBIndex struct {
 	IsUnique   bool          `json:"is_unique"`
 	IsPrimary  bool          `json:"is_primary"`
 	Definition string        `json:"definition"` // Full index definition
+	// KeyPartsIncomplete reports that the catalog described a key part this
+	// reader cannot name, so Columns (and Parts) list fewer parts than the key
+	// actually has. The MySQL reader sets it for a functional key part --
+	// `KEY idx ((b + 1))` -- whose information_schema.STATISTICS row carries a
+	// NULL COLUMN_NAME and an EXPRESSION column that MariaDB does not have.
+	//
+	// A comparison must not read the key columns of such an index as the whole
+	// key: it would plan a rebuild on every run for a key that never changed.
+	// Everything else about the index is reported normally.
+	KeyPartsIncomplete bool `json:"key_parts_incomplete,omitempty"`
 	// Condition is the WHERE clause for partial indexes when the dialect
 	// exposes one structurally.
 	Condition string `json:"condition,omitempty"`
--- a/docs/conformance.md
+++ b/docs/conformance.md
@@ -294,20 +294,37 @@ now. A comma is a legal character in a MySQL identifier and
 11.8, which is why only the comma reaches it there). A functional key part has a
 `NULL` `COLUMN_NAME` and an `EXPRESSION` column MariaDB does not have, so the
 reader still cannot name it — it now reports the part as one it could not read
-instead of failing, and the key-column comparison declines on such an index
-rather than planning a rebuild of a key that never changed. Comparing MySQL
-expression keys is still not supported.
+instead of failing, and the key-column comparison reads what was read rather than
+treating it as the whole key, which would plan a rebuild of a key that never
+changed on every run.
+
+A partly-read key is compared as far as it was read. `KEY idx_mixed (b, (b + 1))`
+arrives as the one named column `b` plus the record of a part that could not be
+named, so a desired `idx_mixed (c, (c + 1))` is a different key by the only part
+either side can name, and the difference is reported. Proof runs one way only: a
+desired `idx_mixed (b, (c + 1))` names the same column and differs solely in the
+expression, and that pair is still reported synced. Closing it means reading
+`information_schema.STATISTICS.EXPRESSION`, which MariaDB does not have, so
+comparing a MySQL expression key remains unsupported — as does rebuilding one,
+because the down direction would recreate the key without its expression.
 
 ### A unique constraint and its index are one object, and the schema says which
 
-Every engine Ptah supports except SQL Server enforces a `UNIQUE` constraint with
-an index of the constraint's own name on the constraint's own table.
-Introspection reports that one object twice — once in the index catalog, once in
-the constraint catalog — and MySQL and MariaDB do not even have a separate
-notion to report: `ADD CONSTRAINT c UNIQUE (a)` and
-`CREATE UNIQUE INDEX c ON t (a)` leave the identical catalog row, which is why
-`schema inspect` writes MySQL uniqueness back out as `index { unique = true }`
-on both binaries and never as a constraint.
+PostgreSQL, MySQL and MariaDB enforce a `UNIQUE` constraint with an index of the
+constraint's own name on the constraint's own table. Introspection reports that
+one object twice — once in the index catalog, once in the constraint catalog —
+and MySQL and MariaDB do not even have a separate notion to report:
+`ADD CONSTRAINT c UNIQUE (a)` and `CREATE UNIQUE INDEX c ON t (a)` leave the
+identical catalog row, which is why `schema inspect` writes MySQL uniqueness
+back out as `index { unique = true }` on both binaries and never as a
+constraint.
+
+The other two engines Ptah supports do not share the name, so nothing brings the
+two representations into collision there. SQL Server keeps a `UNIQUE` constraint
+and a unique index as separate objects. SQLite backs one with an index it names
+itself: `CREATE TABLE t (a TEXT, CONSTRAINT uq_t_a UNIQUE(a))` leaves
+`sqlite_master` holding `sqlite_autoindex_t_1`, and no statement can name that
+object or the constraint that owns it.
 
 The two representations used to be compared in two pools, and the database index
 was filtered out of its pool unconditionally. The desired side is never
@@ -363,11 +380,51 @@ below are the pinned binary's own plan, matched by Ptah:
 
 Dropping the PostgreSQL one as an index answers `cannot drop index
 uq_users_email because constraint uq_users_email on table users requires it
-(SQLSTATE 2BP01)`, so the comparator marks those removals and the
-PostgreSQL-family planner renders the constraint spelling. SQL Server never
-reaches the rule — a UNIQUE constraint and a unique index are separate objects
-there — and SQLite's constraint-backed index is `sqlite_autoindex_*`, which no
-statement can name and which the comparator therefore never considers.
+(SQLSTATE 2BP01)`. The comparator therefore records **what the object is** — a
+UNIQUE constraint's — for every engine that reports it under the constraint's
+name, and each planner spells its own statement: the PostgreSQL-family planner
+drops the constraint, the MySQL and MariaDB planner keeps `DROP INDEX` and their
+plans are unchanged. SQL Server never reaches the rule, and SQLite names the
+backing index itself, so neither records anything.
+
+**The rollback restores a constraint, on every engine.** What the up direction
+dropped was a UNIQUE constraint, so `ALTER TABLE … ADD CONSTRAINT … UNIQUE` is
+what puts it back — on MySQL and MariaDB that lands the same catalog row
+`CREATE UNIQUE INDEX` would, and on PostgreSQL nothing else restores the
+`pg_constraint` row.
+
+Reversing the removal into an index addition instead restored the wrong object
+where it worked at all. Three facts govern the down direction:
+
+- The down target is the introspected schema, which omits a constraint-backed
+  index because the index is the constraint's. The index addition therefore had
+  nothing to resolve, and generation failed outright with `invalid schema diff:
+  added index users.uq_users_email at position 0 is missing or ambiguous in the
+  target schema` — measured live on PostgreSQL 17.10 and MySQL 9.7.1, where no
+  migration could be produced at all.
+- Where it did resolve, it would rebuild a plain unique index in place of the
+  constraint, leaving a catalog the migration never started from.
+- `ADD CONSTRAINT … UNIQUE` builds an index of the constraint's name, so the
+  plain index the up direction created is dropped first. Otherwise PostgreSQL
+  answers `relation "uq_users_email" already exists` (SQLSTATE 42P07) and MySQL
+  `Error 1061 (42000): Duplicate key name 'uq_users_email'`.
+
+Applied end to end against a live `CONSTRAINT uq_users_email UNIQUE (email)`, the
+up leaves a plain index and no constraint row, the down restores both, and a
+second comparison against the post-up database reports synced:
+
+| Target | up | catalog after up | down | catalog after down |
+| --- | --- | --- | --- | --- |
+| PostgreSQL 17.10 | exit 0 | `uq_users_email` a plain index, no `pg_constraint` row | exit 0 | `UNIQUE (email)` and its unique index, as before the up |
+| MySQL 9.7.1 | exit 0 | `NON_UNIQUE=1`, no `TABLE_CONSTRAINTS` row | exit 0 | `NON_UNIQUE=0` and the `UNIQUE` row, as before the up |
+
+**Losing the uniqueness is a destructive change, whichever statement says so.**
+Replacing a unique key with a plain index deletes a data guarantee, and it was
+classified only as `indexes_removed` — a warning, which passed
+`--check-destructive` and every drift threshold keyed on destructive findings, on
+all three engines. The diff now reports it as `unique_protections_removed`, and
+on MySQL and MariaDB — where the removal is spelled `DROP INDEX`, exactly like
+dropping an access path — the statement itself is classified destructive.
 
 ## Reports
 
--- a/docs/conformance.md
+++ b/docs/conformance.md
@@ -269,10 +269,65 @@ Two consequences worth naming:
   input and no round trip reaches it. Omit a default class, which is how both
   binaries write one.
 
-Only PostgreSQL is covered. MySQL, MariaDB, SQLite, ClickHouse, CockroachDB,
-YugabyteDB and Spanner keep the comparison they had, because what makes two
-indexes the same index is a per-dialect question and only PostgreSQL was
-measured.
+MySQL and MariaDB now compare uniqueness and the key columns, and nothing else.
+That branch was added with the ownership rule below, which is what made a
+database unique index reachable by the comparison at all; see
+[#1245](https://github.com/stokaro/ptah/issues/1245) for what it deliberately
+leaves out and why. SQLite, ClickHouse, CockroachDB, YugabyteDB and Spanner keep
+the comparison they had, because what makes two indexes the same index is a
+per-dialect question and those dialects were not measured.
+
+### A unique constraint and its index are one object, and the schema says which
+
+Every engine Ptah supports except SQL Server enforces a `UNIQUE` constraint with
+an index of the constraint's own name on the constraint's own table.
+Introspection reports that one object twice — once in the index catalog, once in
+the constraint catalog — and MySQL and MariaDB do not even have a separate
+notion to report: `ADD CONSTRAINT c UNIQUE (a)` and
+`CREATE UNIQUE INDEX c ON t (a)` leave the identical catalog row, which is why
+`schema inspect` writes MySQL uniqueness back out as `index { unique = true }`
+on both binaries and never as a constraint.
+
+The two representations used to be compared in two pools, and the database index
+was filtered out of its pool unconditionally. The desired side is never
+filtered, so a desired state that spells the object as an index had nothing to
+match: index comparison reported it **added** and constraint comparison, finding
+no `UNIQUE` constraint on the desired side, reported the same name **removed**,
+in the same plan. Measured by replaying a database's own `schema inspect` output
+against the database it came from, where the pinned Atlas CE v1.3.0 binary
+reported the schema synced:
+
+| Target | Fixture | Ptah before | Result of applying it |
+| --- | --- | --- | --- |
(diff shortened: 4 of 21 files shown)


denisvmedia:

**Second verdict: pending re-measurement.**

denisvmedia (commented):
Independent review of head `a8dc0e5d` found three merge blockers despite green CI.

1. **PostgreSQL down generation loses constraint ownership.** `migration/schemadiff/internal/compare/indexes.go:509-515` marks the up removal correctly, but `migration/generator/generator.go:1456-1462` reverses every index removal into an index addition. Constraint-backed indexes were deliberately removed from `dbAsGoSchema`, so rollback either fails resolution or recreates an index instead of the original UNIQUE constraint. Partition marked removals and restore the original `DBConstraint`; add up/down/live zero-diff coverage.

2. **Incomplete MySQL functional keys can be reported synchronized.** `migration/schemadiff/internal/compare/indexes.go:606-618` returns “no difference” whenever `KeyPartsIncomplete` is set, without comparing known columns. A database key `(b, (b + 1))` can therefore appear synced with desired `(c, (c + 1))`. Read `INFORMATION_SCHEMA.STATISTICS.EXPRESSION` where supported; otherwise fail loudly. Down reconstruction is also lossy while expressions are discarded.

3. **Loss of uniqueness bypasses destructive gates.** `migration/safety/safety.go:69-70,354-357` classifies index removal as a warning, so replacing a unique key with a plain index can pass `CheckDestructive` and drift thresholds. Record semantic unique-protection removals independently of SQL spelling and classify them as destructive for MySQL, MariaDB, and PostgreSQL.

Also correct `docs/conformance.md:303-305`: SQLite uses unaddressable `sqlite_autoindex_*` backing objects, contrary to the new “every engine except SQL Server” claim. The PR body is stale after the second commit and should be rewritten once these blockers are fixed.

denisvmedia:
Thanks — all three findings reproduced, two of them on live databases, and all three are fixed in `1b8b4c02`. One of the four remedies is declined with a measurement; everything else is done as asked. `git log --oneline -1` was run inside the worktree before every build, both binaries were rebuilt and chmod +x, and the base tree a8dc0e5d was checked out in a second worktree rather than stashed.

**1. PostgreSQL down generation loses constraint ownership — confirmed, and it was not PostgreSQL-only.**

Reproduced live before touching anything, against `CREATE TABLE users (id BIGSERIAL PRIMARY KEY, email VARCHAR(255) NOT NULL, CONSTRAINT uq_users_email UNIQUE (email))` with a desired plain `index "uq_users_email"` (PostgreSQL 17.10, MySQL 9.7.1, my own prefixed databases, dropped afterwards):

```
head a8dc0e5d, postgres: FAIL generate: error generating down migration SQL:
  invalid schema diff: added index users.uq_users_email at position 0 is missing
  or ambiguous in the target schema
head a8dc0e5d, mysql:    the same failure, verbatim
```

So it is the first of your two branches — resolution failure — and no migration is produced at all. It reaches MySQL too, because the marker was PostgreSQL-only but the omission in the down target is not: `ConvertDBSchemaToGoSchema` drops a constraint-backed index on every engine.

Fixed your way. `reverseIndexRemovals` partitions the removals: a marked one becomes a UNIQUE `ConstraintAdditionInfo` rebuilt from the introspected `DBConstraint` (so `IncludeColumns` and `NULLS [NOT] DISTINCT` survive), an unmarked one still becomes an index addition. A marked removal with no constraint to rebuild from stays an index addition on purpose — that is the loud failure above, which beats a rollback that silently omits the protection it exists to restore.

I deviated on one point, measured: I widened `ConstraintBackedIndexRemovals` to record *what the object is* on every engine rather than restricting the hand-off to MySQL/MariaDB (your option two under point 3). The two consumers need opposite things. The up drop is per engine — PostgreSQL drops the constraint, MySQL and MariaDB keep `DROP INDEX`, which is the pinned binary's own plan and one catalog row. The down restore is not per engine: `ADD CONSTRAINT … UNIQUE` is what puts a UNIQUE constraint back everywhere, and on MySQL it lands the same catalog row `CREATE UNIQUE INDEX` would. `TestPlanner_MySQLFamilyDropsAConstraintBackedKeyAsAnIndex` pins that widening the mark did not import PostgreSQL's spelling, and the inverse mutant (mark PostgreSQL only again) reddens exactly the mysql and mariadb rows and leaves postgres green.

Ordering matters and I did not take it on trust. The planners emit constraint additions before index removals, so `ADD CONSTRAINT` landed ahead of the `DROP INDEX` of the plain index it was rebuilding. Both planners now emit that drop with the addition and skip it downstream. Reverting just that, live: the MySQL down exits 1 with `ERROR 1061 (42000) at line 8: Duplicate key name 'uq_users_email'`.

Up/down/live zero-diff coverage, applied and asserted rather than rendered:

```
postgres  before up: pg_indexes CREATE UNIQUE INDEX uq_users_email; pg_constraint u|UNIQUE (email)
          after up:  plain index; pg_constraint holds users_pkey ALONE          (exit 0, exit 0)
          after down: identical to the pre-up snapshot, u|UNIQUE (email) back   (exit 0, exit 0)
mysql     before up: NON_UNIQUE=0 + TABLE_CONSTRAINTS uq_users_email UNIQUE
          after up:  NON_UNIQUE=1, no constraint row                            (rc 0)
          after down: NON_UNIQUE=0 + the UNIQUE row, identical to before        (rc 0)
```

Both directions: re-comparing against the post-up database reports NO_CHANGES (converged), and against the restored database reports the difference again. The #1245 headline row — a database unique key against a desired `unique = true` index — still reports synced live on both engines.

**2. Incomplete MySQL functional keys can be reported synchronized — confirmed. Half fixed, half declined with a measurement.**

Your example reproduced exactly: db `KEY idx_mixed (b, (b + 1))` vs desired `(c, (c + 1))` gave `IndexesAdded=[] IndexesRemoved=[] HasChanges=false` on mysql and mariadb. `mysqlIndexKeyColumnsChanged` now compares the parts the reader *could* name, in order, and that pair is reported as the replacement it is.

I declined "read `EXPRESSION` where supported; otherwise fail loudly", for two measured reasons.

The read is a reader change with its own dialect sweep — MariaDB has no `EXPRESSION` column, so it needs a capability probe, an expression-normalization rule for MySQL's `(\`b\` + 1)` spelling that does not oscillate against a hand-written desired state, and renderer/`dbschematogo` work before a functional key can be rebuilt at all. Filed, not guessed at.

"Fail loudly" I tried to place and could not, honestly. The only point every dialect and both the native and compat paths pass through is plan validation (`indexscope.NewResolverWithSemantics`). A refusal there rejects *any* removal of an index with an unreadable key part — including a plain `DROP INDEX idx_expr` for a functional index the desired state simply stopped declaring, which the pinned binary performs happily. That weakens native `ptah` to close a comparison gap.

And on "down reconstruction is also lossy": agreed, but that is not created by comparing the named parts — it is already reachable on a8dc0e5d through the uniqueness arm, which the incomplete-key record never covered. `mysqlIndexDefinitionChanged` is `generated.Unique != database.IsUnique || mysqlIndexKeyColumnsChanged(...)`, and only the second operand consults `KeyPartsIncomplete`, so a database `UNIQUE KEY idx_expr (b, (b + 1))` against a desired plain `index "idx_expr"` plans a replacement on head today and its down rebuilds `idx_expr (b)`. Same reader change closes both. Filed with that measurement.

What I did add is strictly one-directional — it can only turn a false "synced" into a reported difference. Mutant E proves the boundary: reading the partial key as if it were whole reddens the no-churn fixture (the oscillation a8dc0e5d exists to prevent) while mutant D reddens only the new contradiction fixture. Both halves are written down in `docs/conformance.md` and on `mysqlNamedKeyPartsContradict`, including the pair that is still reported synced.

**3. Loss of uniqueness bypasses destructive gates — confirmed on both levels, on all three engines.**

Live, head a8dc0e5d, the same replacement fixture: `DIFF_DESTRUCTIVE false` on PostgreSQL 17.10 and MySQL 9.7.1. PostgreSQL was caught only at the statement level (via `DROP CONSTRAINT`), never at the diff level, so drift thresholds keyed on `ClassifySchemaDiff` missed it there too.

Done as asked, independently of SQL spelling. `ClassifySchemaDiff` gains `unique_protections_removed` (destructive) from `ConstraintBackedIndexRemovals`, which is the semantic record. For the statement path, `ast.DropIndexNode` gains `EnforcesUniqueConstraint` — it carries no SQL, no renderer reads it — set by the MySQL/MariaDB planner for a marked removal, and `assessNode` classifies it destructive. Controls both ways: a plain index removal stays a warning at both levels, and marking every index drop reddens `TestPlanner_MySQLFamilyLeavesAPlainIndexDropUnmarked`.

**4. `docs/conformance.md:303-305` — confirmed and corrected.**

Verified rather than assumed: `sqlite3 :memory: "CREATE TABLE t (a TEXT, CONSTRAINT uq_t_a UNIQUE(a)); SELECT type,name,tbl_name FROM sqlite_master;"` returns `index|sqlite_autoindex_t_1|t` — not `uq_t_a`. The claim is now scoped to PostgreSQL, MySQL and MariaDB, with SQL Server and SQLite given their own reasons, and the same sentence is corrected in `unique_index_ownership.go` and `constraintOwnedDatabaseIndex`, which both repeated it. The PR body is rewritten: it also carried "the rule cannot start dropping anything ... matched rather than removed" and "an expression key never reaches the comparison — the read fails first", both false since a8dc0e5d.

**Gates**, each exit code on its own line, private GOCACHE and GOLANGCI_LINT_CACHE: build 0, vet 0, vet -tags integration 0, `go test ./... -count=1` 0 (no serial re-run needed), qtlint 0, golangci-lint 0 ("0 issues."), check-test-style 0 (175 conditional groups, 45 white-box files, baseline unchanged), check-public-api 0, check-public-api-snapshot 0 (regenerated), check-public-api-released 0, check-style.mjs 0 (100 files), check-core-doc-links 0, check-limitations 0.

**Not verified, so not claimed**: MariaDB was never exercised live — no container was reachable and Docker refused to create one under load, so every MariaDB row here is unit-level plus the MySQL live result. The pinned binary was not re-run; this commit changes no up-direction SQL the parity cells cover, but that is an argument, not a measurement. `ptah-compat schema apply` was not exercised live either — the blockers all sit on the native path, which is what I measured end to end. And one up-direction shape may change output without a fixture pinning it: `dropIndexRebuiltAsConstraint` fires whenever a UNIQUE constraint addition and an index removal share (table, name), which can in principle arise going up; dropping the index first is the only order the server accepts, and no existing expectation broke.

Issue #1138, opened 2026-08-03
HCL round-trip of a sql() column type and of array types diverges from the imported fixtures


denisvmedia:
Surfaced by re-pinning the conformance harness to master `d29a44f4` and re-running every tier. The offline tier went from 0 non-OK to **10**, all in `txtar-script :: script-runtime`, across four imported fixtures. Reproduced and re-read the raw report myself before filing, because the first reading of these rows had the direction backwards.

## Cluster A — a `sql()` column type does not round-trip (2 rows)

`sqlite/column-user-defined.txtar`:

```
got:  table "t" { schema = schema.main column "c" { null = true type = sql("USER_DEFINED") } }
want: table "t" { schema = schema.main column "c" { null = true type = user_defined } }
```

and the same for `sql("USER_TYPE")` / `user_type`.

**We emit the raw call where the fixture wants the reduced identifier** — the opposite direction from what a first reading of the report suggested. Worth stating plainly because it changes where the fix goes.

The CLI path is already correct. On `d29a44f4`:

```
$ ptah-compat schema inspect --url file://s.hcl --format '{{ hcl . }}' --dev-url sqlite://d.db
table "t" {
  column "c" {
    type = USER_DEFINED
    null = true
  }
}
```

So #1131 reduced `sql()` on the **read** side and the CLI render agrees. The harness drives the library rather than the CLI, and that path still carries the unreduced call through to the rendered HCL. The reduction needs to hold wherever a parsed type is written back, not only where it is read.

## Cluster B — array types (8 rows), attribution NOT established

`postgres/column-array.txtar`, `postgres/column-enum-array.txtar`, `postgres/column-domain.txtar`:

```
got:  CREATE TABLE "logs" ("records" varchar(100) array NOT NULL);
want: CREATE TABLE "logs" ("records" varchar(100) [10][] NOT NULL);

got:  statuses | status[]
want: (a different rendering of the same enum array)

got:  column "c1" { null = false type = script_column_domain.positive_int }
want: (the fixture's own spelling)
```

These are about how array and domain types render, not about `sql()`. They were reported alongside cluster A and attributed to the same change; that attribution is **unverified** and should not be carried forward without a bisect of its own. A dimensioned array (`[10][]`) losing its dimensions is a separate defect from a `sql()` call surviving into output, and the two have no code in common that I have established.

## Why this is not being papered over

[...]

PR #1302, opened 2026-08-08
Write an inspected array through sql() and name the column's domain (#1138)

denisvmedia:
`ptah-compat schema inspect` wrote HCL the pinned Atlas community binary v1.3.0 refuses to read for four of the six array columns of one PostgreSQL table, and it described a domain column as the domain's base type, so applying its own output rebuilt the column without the domain and without the `CHECK` the domain carries.

Both defects are on the database-to-render path. The parse-to-render half was already correct and is untouched.

Everything below is measured on PostgreSQL 17.10 against the pinned community binary by absolute path. The binary reports `atlas community version v1.3.0`.

## What was already fixed, and is not claimed here

Measured on `origin/master` `9820496e` before touching anything.

- **Cluster A's own completion criterion holds.** `ptah-compat schema inspect --url file://schema.v1.hcl --dev-url "sqlite://dev?mode=memory"` prints `type = sql("USER_DEFINED")`, and feeding that file to the pinned binary exits 0 and inspects back the same call. #1140 and #1233 did that.
- **Six of eleven array shapes were already right.** `text[]`, `integer[]`, `character varying(100)[]`, `status[]`, `positive_int[]` and `cube[]` already rendered as `sql(...)`, byte-identical to what that binary renders for the same columns. #1220 did that.
- **`varchar(100)[10][]` losing its dimensions is PostgreSQL, not Ptah.** The server stores and reports `character varying(100)[]`. Nothing to fix.
- **The fixed point itself already held.** `render -> parse -> render` was byte-identical on master for every column in this PR. What did not hold is the other half of the property — the described type surviving — and that is exactly what a round trip which only re-renders cannot see.

## Family 1 — a type rendered as `sql(...)`

`information_schema` reports a domain column under its **base** type. `goSchemaFieldType` tested `data_type = 'USER-DEFINED'` before the server's `format_type`, so a domain over a built-in base survived and a domain over a user-defined base did not:

```
column      data_type      udt_name   domain_name   format_type
c_domain    integer        int4       positive_int  positive_int
c_point3d   USER-DEFINED   cube       point3d       point3d
```

| shape | Ptah, master | pinned v1.3.0 |
| --- | --- | --- |
| `positive_int` (domain over `integer`) | `sql("positive_int")` | `sql("positive_int")` |
| `tags` (domain over `text[]`) | `sql("tags")` | `sql("tags")` |
| `point3d` (domain over `cube`) | **`sql("cube")`** | `sql("point3d")` |
| `lo` (stock, domain over `oid`) | `sql("lo")` | `sql("lo")` |
| `earth` (stock, domain over `cube`) | **`sql("cube")`** | `sql("earth")` |

`lo` and `earth` differ in nothing but what they are built on. Applying Ptah's own output built `c_point3d` as a bare `cube`, so the domain and its `cube_dim(VALUE) = 3` constraint were gone from the database with nothing reported.

Reading the server's spelling first fixes all of them, and it is also the compatible answer.

## Family 2 — array types

`modeledColumnType` consulted the modeled-type set through the **element** name, splitting at the first parenthesis. `numeric(10,2)[]` found `numeric` in the set, put `(10,2)[]` back untouched, and `typeExpr` fell through to its quoted branch. `text[]` escaped only by accident, being absent from the set as a whole string — which is why one table came out half readable.

One column varied and nothing else, `atlas schema inspect --url file://p.hcl --dev-url <clean postgres 17>`:

```
type = "bit(8)[]"                           exit 1  set field "type": unexpected type string
type = "character(5)[]"                     exit 1  the same
type = "numeric(10,2)[]"                    exit 1  the same
type = "timestamp(3) with time zone[]"      exit 1  the same
type = text[]                               exit 1  Invalid expression
type = sql("bit(8)[]")                      exit 0
type = sql("character(5)[]")                exit 0
type = sql("numeric(10,2)[]")               exit 0
type = sql("timestamp(3) with time zone[]") exit 0
```

There is no bare or quoted spelling to prefer: an array is not one HCL type expression, whatever its element type is. So an array is never modeled and always takes the call — which is what that binary's own inspect writes for all six columns.

Feeding the whole document to it, an arrays-only table:

```
master:      atlas schema inspect --url file://<ptah output>  exit 1
this branch: the same command                                 exit 0, and inspects back the same six calls
```

## The comparator was reading the same column out of a different field

`rawDBColumnType` read `ColumnType`/`UDTName` while the desired side read `format_type`, so `ptah-compat schema diff --from <db> --to <db>` reported a change between a database and **itself**:

```
arrays.a_bit          type: _bit    -> bit(8)[]
arrays.a_char         type: _bpchar -> character(5)[]
arrays.a_cube         type: _cube   -> cube[]
arrays.a_enum         type: _status -> status[]
arrays.a_varchar      type: varchar -> character varying(100)[]
arrays.a_varchar_dim  type: varchar -> character varying(100)[]
scalars.c_tags        type: text    -> tags
```

Every row proposed an `ALTER COLUMN ... TYPE` to the type the column already had. Fixing the domain read alone would have added an eighth (`c_point3d cube -> integer`); preferring `FormattedType` there removes all eight and keeps both sides on one field.

## The round trip, against a live database

One PostgreSQL 17 cluster, two tables, 23 columns, `cube` and `hstore` extensions, an enum, and three domains.

1. `ptah-compat schema inspect ptah_test` → document A
2. `ptah-compat schema apply --to file://A --url <empty database> --dev-url <clean dev> --auto-approve` → exit 0
3. `ptah-compat schema inspect <that database>` → document B

| | master | this branch |
| --- | --- | --- |
| `diff A B` | identical | identical |
| catalog of source vs round trip | `scalars.c_point3d` comes back as **`cube`** | identical, all 23 columns |
| `schema diff --from source --to round trip` | `Modify column scalars.c_point3d: type: integer -> cube` | `Schemas are synced, no changes to be made.` |
| self-diff of one database | 7 phantom rows | synced |

Master's document was a byte-identical fixed point too. That is the whole reason this survived the round-trip tests that already existed.

## Tests

- `TestRenderForDialectWrapsEveryArrayColumnType` — nine rows: the four sized spellings, the three unsized ones that were already right, and two scalar controls that must still go out bare.
- `TestConvertDBSchemaToGoSchema_PostgresDomainColumnKeepsTheDomain` — four rows, with the ordinary user-defined column kept as a control so this cannot become "always prefer `format_type`".
- `TestCompareWithDialect_PostgresArrayAndDomainColumnsCompareEqualToThemselves` — compares a schema against itself through the converter, the same path the CLI takes.
- Nine new rows in `TestOracleAcceptsTheWrapItFallsBackTo`, each measured against the pinned binary. The refusal rows are asserted as refusals, so a change that puts an array back on the modeled path reddens the Atlas CE Oracle job rather than a user's file. Three of them are named in `.github/workflows/atlas-oracle.yml` so a selection that stopped matching cannot leave that job green.

Each was seen red both ways. Reverting the array guard reddens exactly the four sized rows and leaves the unsized ones green; over-correcting to wrap every type reddens the bare-type rows in the pre-existing suite. Reverting the branch order reddens only the domain-over-user-defined row; dropping the `USER-DEFINED` branch entirely reddens the control. Reverting the comparator reproduces the seven phantom rows verbatim in the failure message.

## Behavior change

An array or domain column's `type` attribute changes spelling in `schema inspect` HCL output. Pre-v1, so no compatibility with Ptah's own previous output is owed; the new spelling is the one the pinned community binary accepts and the one it writes itself. No capability is removed and nothing is gated. `--format sql` and `--format json` output is unchanged.

Documentation: a new "How an inspected column type is written" section in the schema-commands page states the rule, both refusals, and why a domain is named rather than resolved.

Refs #1138, #1220, #1233, #1140.

---

### What it found

REFUTED. Two of the three changes hold and are better than claimed. The third — the comparator hunk in /Users/buster/Work/denis/ptah/.claude/worktrees/wf_b394c0e2-0e4-8/migration/schemadiff/internal/compare/shared.go — silently DESTROYS COLUMNS AND DATA on an adjacent shape the author never built.

SETUP. My own container r1138x-pg (postgres:17, port 15738, removed at the end), CI role shape (ptah_user SUPERUSER + bootstrap postgres), my own fixtures, oracle at /Users/buster/Work/denis/ptah-atlas-conformance/bin/atlas ("atlas community version v1.3.0"). Head 1d03cdac in my worktree; merge-base 9820496e in a second worktree at /private/tmp/claude-501/-Users-buster-Work-denis-ptah/2fc9aeb4-cb03-4375-ae0a-82e5047f1051/scratchpad/r1138base. `git log --oneline -1` verified in each before every build.

=== THE REFUTATION ===

rawDBColumnType now returns FormattedType, which for a PostgreSQL domain column is the DOMAIN'S NAME. That string is handed to normalize.Type (migration/schemadiff/internal/normalize/normalize.go:51), which matches by SUBSTRING: anything containing "int" is "integer", anything containing "text" is "text". The name a schema author picked now decides whether the comparator sees a change at all.

FAILING SHAPE, PostgreSQL 17.10:
  DB P: CREATE DOMAIN waypoint AS integer CHECK (VALUE > 0);
        CREATE DOMAIN context  AS integer;
        CREATE TABLE t (id serial PRIMARY KEY, a waypoint NOT NULL, b context NOT NULL);
        INSERT INTO t (a,b) VALUES (7,42);
  DB Q: CREATE TABLE t (id serial PRIMARY KEY, a bigint NOT NULL, b text NOT NULL);

  ptah-compat schema diff --from <P> --to <Q>
    9820496e: -- Modify column t.a: type: int4 -> bigint --
              -- Modify column t.b: type: integer -> text --
              DROP DOMAIN IF EXISTS "context" CASCADE;  DROP DOMAIN IF EXISTS "waypoint" CASCADE;
    1d03cdac: DROP DOMAIN IF EXISTS "context" CASCADE;  DROP DOMAIN IF EXISTS "waypoint" CASCADE;   <- no ALTERs

  ptah-compat schema apply --url <live copy of P> --to <Q> --dev-url <clean> --auto-approve
  (two identical live copies, one per binary)
    9820496e: rc=0, table t = id | a bigint | b text, row 1|7|42 intact          (3 columns)
    1d03cdac: rc=0, "Schema apply completed successfully", table t = id only     (1 COLUMN)
  DROP DOMAIN ... CASCADE takes the columns because no ALTER was planned. Exit 0, nothing reported.

  Native surface, same defect:
  ptah schema compare --root-dir <Go models: a BIGINT, b TEXT> --db-url <P>
    9820496e: "Differences detected (2 categories): tables_modified (1): t / domains_removed (2)" + both ALTERs
    1d03cdac: "Differences detected (1 category): domains_removed (2)"

  Role-independent: identical output as ptah_user and as the bootstrap postgres superuser.
  ISOLATED: reverting ONLY the rawDBColumnType hunk on 1d03cdac (rebuild, same two databases) restores both ALTERs. The renderer and converter halves are not implicated.

REGRESSION TEST COMMITTED (red on purpose) at
/Users/buster/Work/denis/ptah/.claude/worktrees/wf_b394c0e2-0e4-8/migration/schemadiff/domain_named_like_a_type_test.go
  go test ./migration/schemadiff/ -run DomainNamedLikeAType   1d03cdac: FAIL   9820496e: ok
The branch's own TestCompareWithDialect_PostgresArrayAndDomainColumnsCompareEqualToThemselves still PASSES alongside it, so a correct fix must satisfy both.

WHY THEIR MUTATION SWEEP MISSED IT. Their revert mutant reddened only their new self-compare test, whose fixture names (positive_int, point3d, tags, short_text) all normalize to the same token on both sides. The fingerprint is already in their own report and they read past it: "schema diff --from ptah_test --to ptah_rt_base -> 'Modify column scalars.c_point3d: type: integer -> cube'". point3d is not integer — that "integer" IS normalize.Type matching the substring "int".

THE CLAIM THAT IS NOT CODE. The doc comment they added on rawDBColumnType asserts "...leaves it empty for every other column, so this changes nothing else." The premise is true; the inference is false. It changes the comparator's verdict for every domain column, and for a domain whose name collides with a type substring it changes it from "report the change" to "report nothing".

=== WHAT HOLDS (re-measured, my fixtures) ===
1. ARRAYS. Confirmed and broader than claimed. All 20 array columns of my fixture render byte-identical to the pinned binary's own `schema inspect` of the same database (diff rc=0). Base emitted the quoted spelling for SIX sized shapes, not four — they never built interval(3)[] or time(2) without time zone[]. Oracle probes: quoted `"numeric(10,2)[]"` rc=1 (`set field "type": unexpected type string`); all eight wrapped spellings rc=0, including my two extra ones. SQLite improves too: a column declared `VARCHAR(10)[]` was `type = "varchar(10)[]"` on base (oracle rc=1) and is `sql("VARCHAR(10)[]")` on head (oracle rc=0). MySQL byte-identical on inspect and self-diff, as predicted (no modeled set, early return).
2. DOMAINS. Confirmed. c_point3d, c_earth and my extra alt.alt_dom_over_ud all now match the oracle exactly. Live round trip (inspect -> apply to empty DB via clean dev -> inspect): head documents byte-identical and point3d_check CHECK ((cube_dim(VALUE) = 3)) survives on the column; base rebuilds c_point3d as bare cube with the domain gone.
3. SELF-DIFF. On my fixture base emitted 15 phantom rows (they reported 7); head emits "Schemas are synced". Control built: head still reports real differences against a deliberately-altered database (a_varchar 100->200, c_point3d, c_tags).
4. Oracle tests execute: 15/15 subtests PASS, 0 SKIP, `go test -list '^TestOracle'` still lists exactly 5, all three workflow-named rows present.

=== GATES (in my worktree, exit code on its own line) ===
go build ./...                                0
go vet ./...                                  0
go vet -tags integration ./integration/...    0
go test ./... -count=1                        1  -- cmd/ptah-compat "panic: test timed o

### What it says must change

Keep the renderer half (internal/atlashclrender/modeled_types.go) and the converter half (internal/convert/dbschematogo/dbschematogo.go) — both are measured against the pinned binary and both are right. Drop or rework the third hunk.

The defect is that a domain's NAME reaches normalize.Type, whose matching is substring-based and was only ever safe for real type names. Two shapes that fix the phantom self-diff without introducing the miss:

(a) Compare a domain-typed column by exact identity on BOTH sides and never send the name through normalize.Type: carry a "this column is a domain" bit (the reader already has domain_name), and when both sides carry it compare the domain names verbatim; when only one side does, that IS a change and must be reported.

(b) Or fix the phantom rows on the desired side instead: leave rawDBColumnType reading ColumnType/UDTName and make goSchemaFieldType's output for a domain column resolve to the same field, so the two sides agree without a user-chosen identifier ever reaching a substring matcher.

Either way, the branch needs a fixture whose domain name collides with a type substring. The one committed at /Users/buster/Work/denis/ptah/.claude/worktrees/wf_b394c0e2-0e4-8/migration/schemadiff/domain_named_like_a_type_test.go uses `waypoint` (over integer) and `context` (over integer) and must go green alongside the existing self-compare test.

PR title:
  Pin the domain a comparator change stopped seeing (#1138)

PR body:
  `stokaro/ptah#1138`'s comparator hunk makes `schema diff`, `schema apply` and
  `ptah schema compare` drop a required `ALTER COLUMN ... TYPE` whenever the
  column's type is a PostgreSQL domain whose NAME contains a type substring. The
  plan keeps its `DROP DOMAIN ... CASCADE`, so applying it destroys the column
  instead of converting it. This adds the fixture that fails on that branch and
  passes on its merge-base.

  `rawDBColumnType` now prefers `FormattedType`, which for a domain column is the
  domain's own name rather than the base type `information_schema` reports. That
  string is handed to `normalize.Type`, which matches by SUBSTRING: anything
  containing "int" is "integer", anything containing "text" is "text". A schema
  author's choice of name now decides whether the comparator sees a change.

  Measured on PostgreSQL 17.10, one cluster, two domains over integer:

      CREATE DOMAIN waypoint AS integer CHECK (VALUE > 0);
      CREATE DOMAIN context  AS integer;
      CREATE TABLE t (id serial PRIMARY KEY, a waypoint NOT NULL, b context NOT NULL);

  against a desired `a bigint, b text`:

      merge-base   ALTER TABLE "t" ALTER COLUMN "a" TYPE bigint;
                   ALTER TABLE "t" ALTER COLUMN "b" TYPE text;
                   DROP DOMAIN IF EXISTS "context" CASCADE;
                   DROP DOMAIN IF EXISTS "waypoint" CASCADE;

      that branch  DROP DOMAIN IF EXISTS "context" CASCADE;
                   DROP DOMAIN IF EXISTS "waypoint" CASCADE;

  Applied to a live database holding one row, `schema apply --auto-approve`
  exited 0, printed "Schema apply completed successfully", and left table `t`
  with only `id`. The merge-base binary, same command and an identical copy of
  the same database, left `a bigint` and `b text` with the row intact. The same
  miss reaches the native surface: `ptah schema compare --root-dir <models>
  --db-url <db>` reports only `domains_removed`.

  Reverting just the `rawDBColumnType` hunk restores both ALTERs, so the renderer
  and converter halves of that branch are not implicated; they are measured
  against the pinned Atlas community binary v1.3.0 and both are right.

  This test is RED on `issue-1138-hcl-roundtrip` and green on `9820496e`. It is
  meant to be satisfied alongside that branch's own
  `TestCompareWithDialect_PostgresArrayAndDomainColumnsCompareEqualToThemselves`:
  a database must equal itself AND a real change must still be reported.

### DIFF
--- a/.github/workflows/atlas-oracle.yml
+++ b/.github/workflows/atlas-oracle.yml
@@ -183,6 +183,14 @@ jobs:
           grep -Fq 'TestOracleReadsTheReferenceSpellingPtahRenders/postgres/qualified' "$RUNNER_TEMP/atlas-column-type-run"
           grep -Fq 'TestOracleReadsTheReferenceSpellingPtahRenders/sqlite/rendered' "$RUNNER_TEMP/atlas-column-type-run"
           grep -Fq 'TestOracleReadsTheReferenceSpellingPtahRenders/sqlite/qualified' "$RUNNER_TEMP/atlas-column-type-run"
+          # The array rows are a pair for the same reason (stokaro/ptah#1138).
+          # The wrapped row alone would pass on a build that had stopped
+          # refusing anything, and the quoted row alone says nothing about the
+          # spelling Ptah writes today. Both are named so a selection that
+          # silently stopped matching them cannot leave the job green.
+          grep -Fq 'TestOracleAcceptsTheWrapItFallsBackTo/postgres_wraps_a_sized_array_whose_element_name_is_modeled' "$RUNNER_TEMP/atlas-column-type-run"
+          grep -Fq 'TestOracleAcceptsTheWrapItFallsBackTo/postgres_refuses_that_same_array_quoted' "$RUNNER_TEMP/atlas-column-type-run"
+          grep -Fq 'TestOracleAcceptsTheWrapItFallsBackTo/postgres_refuses_an_array_written_bare' "$RUNNER_TEMP/atlas-column-type-run"
 
       - name: Regenerate Atlas CE corpus
         run: >-
--- a/dbschema/types/types.go
+++ b/dbschema/types/types.go
@@ -113,6 +113,19 @@ type DBColumn struct {
 	IsPrimaryKey       bool    `json:"is_primary_key"`    // Derived field
 	IsUnique           bool    `json:"is_unique"`         // Derived field
 
+	// DomainName names the domain a column is declared with, empty for every
+	// column whose declared type is not a domain (PostgreSQL only today).
+	//
+	// It is a separate fact rather than something read back out of
+	// FormattedType, because FormattedType is filled for arrays as well and the
+	// two shapes want opposite answers: an array's spelling is a TYPE, which may
+	// be compared and normalized like any other, while a domain's spelling is an
+	// IDENTIFIER its author chose and must only ever be compared by identity.
+	// Nothing but the catalog can tell them apart -- a domain over an array is
+	// reported with data_type "ARRAY" exactly like a plain array column
+	// (stokaro/ptah#1138).
+	DomainName string `json:"domain_name,omitempty"`
+
 	// GeneratedExpression holds the generated-column expression. Nil for plain
 	// columns.
 	GeneratedExpression *string `json:"generated_expression,omitempty"`
--- a/dbschema/types/types.go
+++ b/dbschema/types/types.go
@@ -126,6 +126,25 @@ type DBColumn struct {
 	// (stokaro/ptah#1138).
 	DomainName string `json:"domain_name,omitempty"`
 
+	// DomainSchema names the schema holding that domain, empty for every column
+	// whose declared type is not a domain.
+	//
+	// A domain's identity is (schema, name), and the name alone is not it:
+	// public.status and other.status are two different types with different
+	// CHECK constraints over possibly different base types. A comparator handed
+	// only "status" for both calls a column that must be converted unchanged,
+	// while the plan still drops the domain the desired schema no longer
+	// declares -- so DROP DOMAIN ... CASCADE takes the column and its data with
+	// nothing having converted it first (stokaro/ptah#1138).
+	//
+	// It is recorded raw, exactly as information_schema.columns.domain_schema
+	// reports it, rather than blanked for the schema being read: the domain a
+	// column is declared with may live in a schema the read never visits, and
+	// blanking it there erases the very distinction this field exists for.
+	// FormattedType cannot stand in, because the server writes a qualifier
+	// there only when the search path forces one.
+	DomainSchema string `json:"domain_schema,omitempty"`
+
 	// GeneratedExpression holds the generated-column expression. Nil for plain
 	// columns.
 	GeneratedExpression *string `json:"generated_expression,omitempty"`
--- a/docs/conformance.md
+++ b/docs/conformance.md
@@ -562,6 +562,50 @@ and `c` for a standalone `CREATE UNIQUE INDEX`. Only the first is part of the
 column's declaration, and only the first is folded now. A declared column
 `UNIQUE` still round-trips through its own implicit index.
 
+### A domain column is reconciled by identity
+
+The same split applies to the domain row of the table above. Reading a column's
+domain is one claim; comparing it is another, and the comparison must not go
+through type normalization. Normalization matches by substring — anything
+containing `int` is an integer and anything containing `text` is text — which
+is right for type names and wrong for a name a schema author picked. Measured
+on PostgreSQL 17.10, two domains over `integer` with ordinary names:
+
+```sql
+CREATE DOMAIN waypoint AS integer CHECK (VALUE > 0);
+CREATE DOMAIN context  AS integer;
+CREATE TABLE t (id serial PRIMARY KEY, a waypoint NOT NULL, b context NOT NULL);
+```
+
+against a desired `a bigint, b text`, `waypoint` matched `bigint` and `context`
+matched `text`, so neither `ALTER COLUMN ... TYPE` was planned while the
+`DROP DOMAIN ... CASCADE` both columns depend on stayed. `schema apply
+--auto-approve` exited 0, printed `Schema apply completed successfully`, and
+left the table with only its `id` column and the row's other two values gone
+([#1138](https://github.com/stokaro/ptah/issues/1138)).
+
+A domain column now agrees only with a desired type that names the same domain,
+in `ptah-compat schema diff`, `ptah-compat schema apply` and native
+`ptah schema compare` alike. `information_schema.columns.domain_name` is what
+separates a domain from a plain column of the same base type, and it is read
+for that reason: a domain over an array is reported with `data_type` `ARRAY`
+exactly like a plain array column, and an array's spelling is a type that must
+keep normalizing like one.
+
+The reverse pair was missed as well, and there the pinned binary was right where
+Ptah was silent. A plain `integer` column against a desired schema that declares
+`waypoint` and types the column with it:
+
+| | plan |
+| --- | --- |
+| Atlas CE v1.3.0 | `ALTER TABLE "t" ALTER COLUMN "a" TYPE waypoint;` |
+| Ptah, before | `Schemas are synced, no changes to be made.` |
+| Ptah, now | the same `ALTER`, executed at exit 0 with the row intact |
+
+A desired type names a domain when the desired schema declares one by that
+name, which is how every source Ptah reads carries it. A bare name with no
+declaration behind it stays an ordinary type name.
+
 ## Reports
 
 - Offline corpus report:
--- a/docs/conformance.md
+++ b/docs/conformance.md
@@ -190,7 +190,7 @@ carrying the same attributes always rendered correctly.
 | Sort order (`DESC`, `NULLS FIRST`, `NULLS LAST`) | `pg_index.indoption` | Dropped: psql exited 0 and left an ascending index | [#1242](https://github.com/stokaro/ptah/issues/1242) |
 | `INCLUDE` payload columns | `pg_index.indkey` past `indnkeyatts` | Dropped: psql exited 0 and left an index that cannot serve the index-only scans it was built for | [#1242](https://github.com/stokaro/ptah/issues/1242) |
 | Expression index, for example `lower(name)` | `pg_index.indkey`, attnum 0 | Emitted as a quoted column identifier, which psql rejected at exit 3 | [#1242](https://github.com/stokaro/ptah/issues/1242) |
-| Domain used as a column type | `information_schema.columns.domain_name` | The `CREATE DOMAIN` was emitted but the column was flattened to the domain's base type: psql exited 0 and left a column without the domain's `CHECK` | [#1242](https://github.com/stokaro/ptah/issues/1242) |
+| Domain used as a column type | `information_schema.columns.domain_name` and `domain_schema` | The `CREATE DOMAIN` was emitted but the column was flattened to the domain's base type: psql exited 0 and left a column without the domain's `CHECK` | [#1242](https://github.com/stokaro/ptah/issues/1242) |
 
 ### The access-method loss was not silent in general
 
@@ -591,6 +591,31 @@ separates a domain from a plain column of the same base type, and it is read
 for that reason: a domain over an array is reported with `data_type` `ARRAY`
 exactly like a plain array column, and an array's spelling is a type that must
 keep normalizing like one.
+
+`domain_schema` is read beside it, because a domain's identity is both halves.
+Measured on the same server, one database holding `public.status` and one
+holding `other.status`, over a table with one row:
+
+| | plan for the column | outcome of `schema apply --auto-approve` |
+| --- | --- | --- |
+| name alone | none; only `DROP DOMAIN IF EXISTS "status" CASCADE` | exit 0, `Schema apply completed successfully`, table left with `id` only |
+| identity | `ALTER TABLE "t" ALTER COLUMN "s" TYPE other.status` ahead of the drop | the column and its row survive |
+
+A desired type that names its schema is held to that schema exactly. One that
+does not is resolved through the domain the desired schema declares by that
+name; with nothing declaring it, the bare name decides on its own, since which
+domain an unqualified reference reaches is a search-path question and Ptah does
+not answer that for the server.
+
+One case is still open and is stated rather than claimed closed: when the desired
+schema declares the same bare name in two schemas, an unqualified reference to it
+stays undecided, and a column that should move between those two domains is
+reported only if one side spells its schema. Measured on PostgreSQL 17.10 with
+`public.status` and `other.status` both declared and both schemas selected, a
+column of `other.status` against a desired `public.status` reported
+`Schemas are synced`. Nothing is dropped in that shape, so the cost is drift
+rather than data loss. Deciding it needs the desired column to carry its domain's
+schema instead of a type string.
 
 The reverse pair was missed as well, and there the pinned binary was right where
 Ptah was silent. A plain `integer` column against a desired schema that declares
--- a/docs/public_api.snapshot
+++ b/docs/public_api.snapshot
@@ -5307,6 +5307,19 @@ type DBColumn struct {
     IsPrimaryKey       bool    `json:"is_primary_key"`    // Derived field
     IsUnique           bool    `json:"is_unique"`         // Derived field
 
+    // DomainName names the domain a column is declared with, empty for every
+    // column whose declared type is not a domain (PostgreSQL only today).
+    //
+    // It is a separate fact rather than something read back out of
+    // FormattedType, because FormattedType is filled for arrays as well and the
+    // two shapes want opposite answers: an array's spelling is a TYPE, which may
+    // be compared and normalized like any other, while a domain's spelling is an
+    // IDENTIFIER its author chose and must only ever be compared by identity.
+    // Nothing but the catalog can tell them apart -- a domain over an array is
+    // reported with data_type "ARRAY" exactly like a plain array column
+    // (stokaro/ptah#1138).
+    DomainName string `json:"domain_name,omitempty"`
+
     // GeneratedExpression holds the generated-column expression. Nil for plain
     // columns.
     GeneratedExpression *string `json:"generated_expression,omitempty"`
--- a/docs/public_api.snapshot
+++ b/docs/public_api.snapshot
@@ -5320,6 +5320,25 @@ type DBColumn struct {
     // (stokaro/ptah#1138).
     DomainName string `json:"domain_name,omitempty"`
 
+    // DomainSchema names the schema holding that domain, empty for every column
+    // whose declared type is not a domain.
+    //
+    // A domain's identity is (schema, name), and the name alone is not it:
+    // public.status and other.status are two different types with different
+    // CHECK constraints over possibly different base types. A comparator handed
+    // only "status" for both calls a column that must be converted unchanged,
+    // while the plan still drops the domain the desired schema no longer
+    // declares -- so DROP DOMAIN ... CASCADE takes the column and its data with
+    // nothing having converted it first (stokaro/ptah#1138).
+    //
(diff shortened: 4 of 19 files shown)


denisvmedia:

**Second verdict: pending re-measurement.**

denisvmedia (commented):
Independent code review of head `6e959f9e` found a potential data-loss blocker.

**Blocker — domain comparison discards schema identity.** `migration/schemadiff/internal/compare/columns.go:467-493,506-522` indexes desired domains by a lowercased bare name and treats qualified/unqualified spellings as equal whenever either qualifier is absent. The PostgreSQL reader at `internal/dbschema/postgres/reader.go:320-332` records `domain_name` but not `domain_schema`.

A current `public.status` column can therefore compare equal to desired `other.status`, suppressing the required ALTER while the plan still removes `public.status`; `DROP DOMAIN public.status CASCADE` can then remove the column and data. Carry `DomainSchema` through `DBColumn` and compare canonical `(schema, name)` identities with an explicit default/search-path rule. Add a regression requiring the column type change before old-domain removal. Do not merge this identity change without that executable guard.

denisvmedia:
## Blocker closed on `218f99ce` — re-measured independently, and it was real

The blocker was written against head `6e959f9e`. `218f99ce` ("Compare a domain column by both halves of its identity") is the answer to it. I re-measured the whole thing from scratch on my own container rather than reading the commit message, and the finding **reproduces exactly** on the pre-fix commit and is **closed** on the current head.

**Setup.** My own container `b1302-pg` (`postgres:17`, port 15302, removed at the end), PostgreSQL 17.10, CI role shape (`ptah_user` LOGIN SUPERUSER owning every database). Binaries built from `218f99ce` (head) and from its parent `0f1fcecc`, which is the state this blocker describes. Head is already based on `master` `873e4e54`, so nothing needed rebasing.

### 1. The premise is exactly right

```
b_src:  table_schema | table_name | column_name | data_type | udt_name | domain_schema | domain_name | format_type
        public       | t          | s           | text      | text     | public        | status      | status

b_dst:  public       | t          | s           | text      | text     | other         | status      | other.status
```

### 2. Reproduction — `schema diff`, same two databases, one binary apart

```
ptah-compat schema diff --from <b_src> --to <b_dst>

0f1fcecc   -- WARNING: Make sure no columns use this domain!
           DROP DOMAIN IF EXISTS "status" CASCADE;                 <- and no ALTER

218f99ce   ALTER TABLE "t" ALTER COLUMN "s" TYPE other.status;
           ALTER TABLE "t" ALTER COLUMN "s" SET NOT NULL;
           ALTER TABLE "t" ALTER COLUMN "s" DROP DEFAULT;
           -- Modify column t.s: type: status -> other.status --
           DROP DOMAIN IF EXISTS "status" CASCADE;
```

### 3. Rendered is not applied — so it was applied, and the catalog asserted

Two live databases built identically (`public.status`, `schema other`, `other.status`, `t(id serial pk, s status NOT NULL)`, one row `'open'`), one per binary, same command:

```
ptah-compat schema apply --url <copy> --to <b_dst> --auto-approve

0f1fcecc   rc=0   "Schema apply completed successfully"
           \d t              ->  id  integer  not null           (1 column)
           SELECT * FROM t   ->  1                               (the value is gone)

218f99ce   rc=0
           \d t              ->  id  integer       not null
                                 s   other.status  not null      (2 columns)
           SELECT * FROM t   ->  1 | open                        (the row survives)
```

The column and its data are destroyed by `DROP DOMAIN ... CASCADE` at exit 0, exactly as the blocker predicted. On head the conversion is planned ahead of the drop and the row survives.

On a target where schema `other` does not exist, head refuses instead of guessing — `rc=1`, `ERROR: schema "other" does not exist (SQLSTATE 3F000)`, nothing applied, `s status NOT NULL` and the row intact — while `0f1fcecc` on an identical copy still exits 0 and leaves the table with `id` only.

### 4. The native surface carries the same verdict

```
ptah schema compare --root-dir <models declaring other.status> --db-url <b_src>

0f1fcecc   Differences detected (2 categories):
             domains_added (1): other.status
             domains_removed (1): status
           CREATE DOMAIN "other"."status" ...;  DROP DOMAIN IF EXISTS "status" CASCADE;

218f99ce   Differences detected (3 categories):
             tables_modified (1): t
             domains_added (1): other.status
             domains_removed (1): status
           CREATE DOMAIN "other"."status" ...;
           ALTER TABLE "t" ALTER COLUMN "s" TYPE other.status;
           DROP DOMAIN IF EXISTS "status" CASCADE;
```

### 5. Controls — this did not become "any domain is a change"

| control | head `218f99ce` |
| --- | --- |
| `schema diff --from b_src --to b_src` | `Schemas are synced, no changes to be made.` |
| `schema diff --from b_dst --to b_dst --schema public --schema other` (domain lives off the read schema) | `Schemas are synced, no changes to be made.` |
| `b_src` vs `b_src2`, two separately created databases each declaring its own `public.status` over the same base with the same `CHECK` — two genuinely identical domains | `Schemas are synced, no changes to be made.` |
| reverse direction, `b_dst` -> `b_src` | `ALTER TABLE "t" ALTER COLUMN "s" TYPE status;` reported |

The reverse direction was missed on `0f1fcecc` as well — it emitted only `CREATE DOMAIN "status" ...` and never converted the column.

### 6. Red and green, per subtest

Three independent mutations, each reverted afterwards; the working tree is clean.

**M1 — drop `DomainSchema` from the database-side identity** (the reader half this blocker names):

```
--- FAIL: TestCompareWithDialect_PostgresDomainColumnComparesByIdentity/a_domain_in_another_schema_is_a_different_domain
--- FAIL: TestCompareWithDialect_PostgresDomainColumnComparesByIdentity/an_unqualified_reference_means_the_domain_that_is_declared
--- FAIL: TestGenerateSchemaDiffSQL_DomainColumnIsConvertedBeforeTheOldDomainIsDropped
```

with the failure text naming the shape:

```
comment: the column must be converted off public.status, or the CASCADE below takes it
  CREATE DOMAIN "other"."status" AS TEXT CHECK (VALUE IN ('open','closed'))
  -- WARNING: Make sure no columns use this domain!
  DROP DOMAIN IF EXISTS "status" CASCADE
```

19 of the 21 identity rows stay green under M1, so the guard is discriminating rather than blanket.

**M2 — match on the name alone** (the old rule):

```
--- FAIL: .../a_domain_in_another_schema_is_a_different_domain
--- FAIL: .../a_column_from_another_schema_against_the_declared_one_is_a_change
--- FAIL: .../an_unqualified_reference_means_the_domain_that_is_declared
--- FAIL: .../the_same_bare_name_in_another_schema_is_a_different_domain
--- FAIL: TestGenerateSchemaDiffSQL_DomainColumnIsConvertedBeforeTheOldDomainIsDropped
```

**M3 — reader projects `'' AS domain_schema`** instead of reading the column:

```
--- FAIL: TestReadColumnsForSchema_KeepsTheDeclaredDomainType/domain_column_keeps_the_domain
--- FAIL: TestReadColumnsForSchema_KeepsTheDeclaredDomainType/a_domain_off_the_search_path_keeps_its_schema
--- PASS: TestReadColumnsForSchema_KeepsTheDeclaredDomainType/plain_column_is_left_alone
```

The executable guard the blocker asked for — "requiring the column type change before old-domain removal" — is `TestGenerateSchemaDiffSQL_DomainColumnIsConvertedBeforeTheOldDomainIsDropped`, and it is red under M1 and M2 and green at head. Restoring the ordering without the `ALTER`, or the `ALTER` without the ordering, is red.

### 7. Gates, exit codes read directly

```
go build ./...                              0
go vet ./...                                0
go vet -tags integration ./integration/...  0
go test ./... -count=1                      0
go tool qtlint ./...                        0
golangci-lint run ./...                     0   (0 issues)
scripts/check-test-style.sh                 0   (175 conditional groups, 45 white-box files)
scripts/check-public-api.sh                 0
scripts/check-public-api-snapshot.sh        0
scripts/check-public-api-released.sh        0
docs/site: npm ci                           0
docs/site: npm run build                    0
docs/site: check:exit-codes                 0
docs/site: check:core-doc-links             0
docs/site: check:page-health                0
docs/site: check:style                      0
docs/site: check:limitations                0
docs/site: check:redirects                  0
docs/site: check:links                      0
docs/site: check:responsive                 0
docs/site: all seven :selftest scripts      0
```

Nothing needed changing, so nothing was pushed — `218f99ce` is already the head and already sits directly on `master` `873e4e54`.

### 8. Residual risk

- **The declared gap is real, and it is drift and not data loss.** I measured it rather than taking the claim: two databases both declaring `status` in `public` and in `other`, the column `other.status` on one side and `public.status` on the other, read with `--schema public --schema other`, reports `Schemas are synced, no changes to be made.` and — the part that matters — plans **no** `DROP DOMAIN` at all, because both domains stay declared on both sides. Nothing is dropped, so nothing is destroyed; re-running simply never converges. Closing it needs the desired column to carry its domain's schema rather than a type string.
- **The removal statement is qualified correctly.** A domain that lives off the default schema is removed as `DROP DOMAIN IF EXISTS "other"."status" CASCADE`, not as a bare name that the search path would resolve elsewhere, and `Domains` already keys add/remove on the qualified name. There is no second identity collapse hiding behind this one.
- **One surface asymmetry worth knowing.** For the cross-schema shape, `ptah-compat schema apply` plans the `ALTER` but not the `CREATE SCHEMA`/`CREATE DOMAIN` the target would need, so it refuses at rc=1 when `other` is absent. That is the safe direction — nothing is applied — but the compat surface cannot converge that shape on its own. The native `ptah schema compare` does emit `CREATE DOMAIN "other"."status"` ahead of the conversion.

PR #1303, opened 2026-08-08
Write a permission block that resolves inside its own document (#1234)

### LINKED ISSUE #1234: Ptah's inspect output is unreadable to the pinned binary for two reasons that are not column types

denisvmedia:
Closes #1234.

Both reasons in that issue are still live, in instances the fix in #1248 did not
reach. Neither needs a hand-edited fixture: each is what `ptah-compat schema
inspect` writes, fed straight back to the pinned Atlas community binary v1.3.0.

## Reason 2 — the missing schema declaration

Every PostgreSQL database carries `GRANT USAGE ON SCHEMA public TO PUBLIC`, so
an **empty** database inspects to one `permission` block and nothing else:

```console
$ ptah-compat schema inspect -u "postgres://.../empty?sslmode=disable&search_path=public" > out.hcl
$ cat out.hcl
// Code generated by ptah; DO NOT EDIT.

role "ptah_user" {
  login = true
  superuser = true
  inherit = true
}

permission {
  to = "PUBLIC"
  for = schema.public
  privileges = ["USAGE"]
}

$ atlas schema inspect -u file://out.hcl --dev-url "$DEV_URL"
Error: out.hcl:11,9-15: Unknown variable; There is no variable named "schema".
exit 1
```

The declaration was **predicted** from the IR — the schemas the tables carry,
plus the default when there are enums — while the reference was **written**
somewhere else. That prediction has now been wrong three times in the same
direction: it declared only the default and missed a table outside it, it missed
the enum block's own reference, and it misses this, where there is no table to
predict from at all.

So it is replaced by collection. The body is rendered first, every
`schema.<name>` goes through one recording helper, and the blocks declared are
exactly the set that came back. A reference and its declaration cannot disagree
when one is derived from the other, and a position added later declares itself.

Collecting fixes the other direction for free: a block the Atlas-compatible
surface omits never reaches the helper, so an omitted sequence no longer
conjures a schema block declaring nothing, and a document that references no
schema still declares none.

## Reason 1 — a permission body that does not evaluate

`to = PUBLIC` was fixed. `to = role.<name>` was not. That spelling is valid only
where the document declares the matching `role` block — which the comment on it
said and nothing enforced. A grantee is exactly where the precondition fails,
because a grant is a child of the object granted on rather than of the grantee,
so excluding roles takes the blocks away and leaves every reference to them:

```console
$ ptah-compat schema inspect -u "$PG_URL" --exclude '*[type=role]' > out.hcl
$ atlas schema inspect -u file://out.hcl --dev-url "$DEV_URL"
Error: out.hcl:52,8-12: Unknown variable; There is no variable named "role".
exit 1
```

Measured on that binary, one operand varied against an otherwise identical
document:

| grantee | role block | result |
| --- | --- | --- |
| `to = role.app` | declared | exit 0 |
| `to = role.app` | absent | exit 1 `There is no variable named "role"` |
| `to = "app"` | absent | exit 0 |

So a grantee is a reference only where the block is there, and a quoted name
otherwise. Quoting costs nothing here, which is what separates it from
`tableRef`: an unresolvable table reference keeps its qualified spelling because
the short form would destroy the schema it carried, and a grantee has no second
part to lose. Ptah reads both spellings back to the same grant, and applying the
document against a live PostgreSQL 17 issues `GRANT ... TO "ptah_user"` at exit
0 — executed, not merely rendered.

## Measured result

| document | master | branch |
| --- | --- | --- |
| inspect of an empty database | 1 `There is no variable named "schema"` | **0** |
| inspect `--exclude '*[type=role]'` | 1 `There is no variable named "role"` | **0** |
| inspect `--exclude 't'` (grants outlive the table) | 1 `There is no variable named "schema"` | **0** |
| inspect, unfiltered | 0 | 0, **byte-identical** |
| inspect, SQLite | 0 | 0, **byte-identical** |

The common path does not move. Only documents that were unreadable change.

## The test the issue asked for

`TestOracleModeledColumnTypesMatchTheBinary` measures one column at a time, and
neither of these defects is visible in one attribute. `TestOracleReadsThe
InspectedDocumentPtahRenders` puts three whole inspected documents to the pinned
binary — a table with grants, the same table with its roles excluded, and the
grant-only document an empty database produces — on PostgreSQL and SQLite, each
rendered through `RenderInspectedForAtlasCLI`, the entry point `ptah-compat
schema inspect` calls.

Each `rendered` row is paired with `unreadable` rows one substitution away from
the accepted document, each asserting the binary's **message** rather than a
bare non-zero status, so a document refused for an unrelated reason cannot pass
as confirmation; every substitution is asserted to have changed the text, so a
spelling that stopped being emitted fails the run instead of mutating nothing.
The Atlas CE Oracle job names those rows rather than counting them, for the same
reason the other runs there do.

## Also in this change

Incomplete grants are now dropped before the sort rather than inside the render
loop. The sort key is the rendered target, and rendering a target is what
records the schema reference, so asking about a grant that is then skipped would
declare a block for a permission that is not in the document.

## Not in this change

The ordering difference #1235 tracks — that binary writes `schema "public" {}`
at the end of its file and Ptah writes it at the top — is untouched. The
two-pass render makes moving it a one-line change if that front wants it.

---

## The review's findings are answered, and the branch was rebased past a destructive trap

An adversarial review refuted head `47f65146`. Two later commits answered it, and this head is three commits past the one that section described — so it has been replaced with what is true now.

### Why this pull request had zero checks

Not a scheduler backlog and not a missing push. The head **was** pushed; the branch was **conflicting**, and GitHub creates no merge ref for a conflicting pull request. Every workflow here is declared `on: {push: {branches: [master]}, pull_request:}` — push events are filtered to `master`, and the `pull_request` events had no merge ref to run against. Zero was the only number possible, and it is easy to misread as "CI has not got to it yet".

### The trap inside the rebase

Both sides had rewritten the same seven-line orchestrator in `internal/atlashclrender/render.go`. `master` had gained a coverage header:

```go
r.builder.WriteString("// Code generated by ptah; DO NOT EDIT.\n")
r.renderCoverageHeader()
r.builder.WriteString("\n")
```

while this branch restructured into a two-pass render that snapshots the body first:

```go
r.renderBody()
body := r.builder.String()
r.builder.Reset()
r.builder.WriteString("// Code generated by ptah; DO NOT EDIT.\n\n")
```

This branch never deleted `renderCoverageHeader()` — it restructured around a function that did not exist on its base. Taking this side wholesale, which is the instinctive resolution since it is "the change", drops the only call site of the `// ptah:not-described` header, and `ptah-compat schema inspect > file` followed by `schema apply --to file://file` starts planning `DROP EXTENSION`, `DROP SEQUENCE` and `DROP POLICY` again — the exact destructive regression #1298 closed.

Resolved as a union, and the placement is load-bearing: the header call sits **after** the builder `Reset()`, or its output is captured into the snapshotted `body` and re-emitted below the schema blocks instead of at the top. `master`'s three-statement form is kept over this branch's `\n\n`, so output stays byte-identical when there is no header. A comment in the code now says why.

The other two conflicts were resolved on their own merits rather than by one rule:

| file | shape | resolution |
| --- | --- | --- |
| `internal/atlashclrender/render.go` | both sides rewrote one function, neither deleted the other's work | union, header after the reset |
| `.github/workflows/atlas-oracle.yml` | additive — the two sides assert different tests | both assertion sets kept |
| `docs/site/src/content/docs/direct/inspect.md` | **superseding** — this branch's second commit replaces a paragraph its own first commit added | `master`'s paragraph kept, plus this branch's *newer* paragraph; the intermediate one dropped |

A blind union on the third file would have left two paragraphs where the branch itself intends one.

### The red integration test, and why it needed no code change

`TestPostgreSQLSchemaBoundaryGuardIntegration` failed on two of six rows — `extension_nothing_references` and `empty_database`, exactly the two fixtures with no table in them:

```
got: []string{"public"}        want: []string(nil)
got: []string{"schema:public"} want: []string(nil)
```

That is this branch working: the renderer now declares the schema blocks its body referenced, so a database with no tables finally emits `schema "public" {}`. It does not reproduce on `master`, where Integration Tests is green. The branch's own fourth commit moved those two characterization rows to the values the four table-bearing rows already carried, kept both assertions as exact equality against a named value, and re-labelled `wantDocumentOnly` as `WRONG (#1276)` rather than pretending the reader gap had closed. Nothing was weakened — the test needed to *run*.

### Verified on the rebased tree

`go build ./...` 0 · `go vet ./...` 0 · `go vet -tags integration ./integration/...` 0 · `go tool qtlint ./...` 0 · `go test ./internal/atlashclrender/... ./internal/atlashcl/... -count=1` ok · a tree-wide conflict-marker grep is empty · `renderCoverageHeader()` still has exactly one call site.

### DIFF
--- a/.github/workflows/atlas-oracle.yml
+++ b/.github/workflows/atlas-oracle.yml
@@ -202,11 +202,17 @@ jobs:
           # document unreadable, so both halves of each pair are named. The
           # grant-only document is the one no prediction reaches: it is what an
           # EMPTY database inspects to, with no table anywhere to declare a
-          # schema from.
+          # schema from. The table-and-view document is the one no FILTER
+          # reaches: it is the default invocation on any database carrying a
+          # view, and its pair is what says a permission target names the block
+          # type the document declares rather than the field the name arrived on.
           grep -Fq 'TestOracleReadsTheInspectedDocumentPtahRenders/postgres/a_table_with_grants/rendered' "$RUNNER_TEMP/atlas-column-type-run"
           grep -Fq 'TestOracleReadsTheInspectedDocumentPtahRenders/postgres/a_table_with_grants/unreadable/PUBLIC_written_bare' "$RUNNER_TEMP/atlas-column-type-run"
           grep -Fq 'TestOracleReadsTheInspectedDocumentPtahRenders/postgres/a_table_whose_roles_were_excluded/rendered' "$RUNNER_TEMP/atlas-column-type-run"
           grep -Fq 'TestOracleReadsTheInspectedDocumentPtahRenders/postgres/a_table_whose_roles_were_excluded/unreadable/a_grantee_reference_with_no_role_block_anywhere' "$RUNNER_TEMP/atlas-column-type-run"
+          grep -Fq 'TestOracleReadsTheInspectedDocumentPtahRenders/postgres/a_table_and_a_view_with_grants_on_both/rendered' "$RUNNER_TEMP/atlas-column-type-run"
+          grep -Fq 'TestOracleReadsTheInspectedDocumentPtahRenders/postgres/a_table_and_a_view_with_grants_on_both/unreadable/a_view_named_as_a_table' "$RUNNER_TEMP/atlas-column-type-run"
+          grep -Fq 'TestOracleReadsTheInspectedDocumentPtahRenders/sqlite/a_table_and_a_view_with_grants_on_both/rendered' "$RUNNER_TEMP/atlas-column-type-run"
           grep -Fq 'TestOracleReadsTheInspectedDocumentPtahRenders/postgres/nothing_but_the_grant_every_database_has/rendered' "$RUNNER_TEMP/atlas-column-type-run"
           grep -Fq 'TestOracleReadsTheInspectedDocumentPtahRenders/postgres/nothing_but_the_grant_every_database_has/unreadable/a_schema_reference_with_no_block' "$RUNNER_TEMP/atlas-column-type-run"
           grep -Fq 'TestOracleReadsTheInspectedDocumentPtahRenders/sqlite/nothing_but_the_grant_every_database_has/rendered' "$RUNNER_TEMP/atlas-column-type-run"
--- a/.github/workflows/atlas-oracle.yml
+++ b/.github/workflows/atlas-oracle.yml
@@ -140,6 +140,10 @@ jobs:
       # started modeling. The spelling of a table REFERENCE is another
       # (stokaro/ptah#1260): `table.<schema>.<name>` costs the whole document,
       # so the rendered output is put to that binary here rather than trusted.
+      # The last one is the WHOLE inspected document (stokaro/ptah#1234): a
+      # permission body that does not evaluate and a `schema.<name>` reference
+      # nothing declares are properties of the file rather than of any one
+      # attribute, so neither is visible to the runs above.
       # All of them live in internal/atlashclrender and are listed before they
       # are run, for the same reason as above.
       - name: Verify HCL render conformance selection
@@ -152,7 +156,8 @@ jobs:
           grep -Fx 'TestOracleAtlasRefusedBlockTypesMatchTheBinary' "$RUNNER_TEMP/atlas-column-type-tests"
           grep -Fx 'TestOracleToleratesTheBlockTypesPtahStillRenders' "$RUNNER_TEMP/atlas-column-type-tests"
           grep -Fx 'TestOracleReadsTheReferenceSpellingPtahRenders' "$RUNNER_TEMP/atlas-column-type-tests"
-          test "$(grep -c '^TestOracle' "$RUNNER_TEMP/atlas-column-type-tests")" -eq 5
+          grep -Fx 'TestOracleReadsTheInspectedDocumentPtahRenders' "$RUNNER_TEMP/atlas-column-type-tests"
+          test "$(grep -c '^TestOracle' "$RUNNER_TEMP/atlas-column-type-tests")" -eq 6
 
       - name: Run HCL render conformance tests
         run: |
@@ -191,6 +196,20 @@ jobs:
           grep -Fq 'TestOracleAcceptsTheWrapItFallsBackTo/postgres_wraps_a_sized_array_whose_element_name_is_modeled' "$RUNNER_TEMP/atlas-column-type-run"
           grep -Fq 'TestOracleAcceptsTheWrapItFallsBackTo/postgres_refuses_that_same_array_quoted' "$RUNNER_TEMP/atlas-column-type-run"
           grep -Fq 'TestOracleAcceptsTheWrapItFallsBackTo/postgres_refuses_an_array_written_bare' "$RUNNER_TEMP/atlas-column-type-run"
+
+          # And for the whole-document run (stokaro/ptah#1234). Each `rendered`
+          # row is only a measurement next to the mutation that makes the same
+          # document unreadable, so both halves of each pair are named. The
+          # grant-only document is the one no prediction reaches: it is what an
+          # EMPTY database inspects to, with no table anywhere to declare a
+          # schema from.
+          grep -Fq 'TestOracleReadsTheInspectedDocumentPtahRenders/postgres/a_table_with_grants/rendered' "$RUNNER_TEMP/atlas-column-type-run"
+          grep -Fq 'TestOracleReadsTheInspectedDocumentPtahRenders/postgres/a_table_with_grants/unreadable/PUBLIC_written_bare' "$RUNNER_TEMP/atlas-column-type-run"
+          grep -Fq 'TestOracleReadsTheInspectedDocumentPtahRenders/postgres/a_table_whose_roles_were_excluded/rendered' "$RUNNER_TEMP/atlas-column-type-run"
+          grep -Fq 'TestOracleReadsTheInspectedDocumentPtahRenders/postgres/a_table_whose_roles_were_excluded/unreadable/a_grantee_reference_with_no_role_block_anywhere' "$RUNNER_TEMP/atlas-column-type-run"
+          grep -Fq 'TestOracleReadsTheInspectedDocumentPtahRenders/postgres/nothing_but_the_grant_every_database_has/rendered' "$RUNNER_TEMP/atlas-column-type-run"
+          grep -Fq 'TestOracleReadsTheInspectedDocumentPtahRenders/postgres/nothing_but_the_grant_every_database_has/unreadable/a_schema_reference_with_no_block' "$RUNNER_TEMP/atlas-column-type-run"
+          grep -Fq 'TestOracleReadsTheInspectedDocumentPtahRenders/sqlite/nothing_but_the_grant_every_database_has/rendered' "$RUNNER_TEMP/atlas-column-type-run"
 
       - name: Regenerate Atlas CE corpus
         run: >-
--- a/.github/workflows/atlas-oracle.yml
+++ b/.github/workflows/atlas-oracle.yml
@@ -26,6 +26,16 @@ jobs:
       # engine does not have -- and those are the two verdicts that run exists to
       # keep apart.
       PTAH_ATLAS_ORACLE_POSTGRES_DEV_URL: 'postgres://postgres:pw@localhost:5432/dev?sslmode=disable&search_path=public'
+      # A second dev database, REALM-scoped: no search_path, so the binary is
+      # willing to materialize a document declaring more than one schema. The
+      # one above is deliberately scoped to `public` and refuses such a document
+      # with `cannot use HCL with more than 1 schema when dev-url is limited to
+      # schema "public"`, which is a verdict about the dev URL rather than about
+      # the document. It is a SEPARATE database because a realm-scoped read
+      # requires a clean one: measured on PostgreSQL 17, a leftover schema makes
+      # the next read fail with `sql/migrate: connected database is not clean:
+      # found schema "other"`, so it must not share with the runs above.
+      PTAH_ATLAS_ORACLE_POSTGRES_REALM_DEV_URL: 'postgres://postgres:pw@localhost:5432/realmdev?sslmode=disable'
 
     services:
       postgres:
@@ -46,6 +56,13 @@ jobs:
 
       - name: Configure oracle path
         run: echo "PTAH_ATLAS_ORACLE=$RUNNER_TEMP/ptah-atlas-ce/atlas" >> "$GITHUB_ENV"
+
+      # The realm-scoped dev database the two-schema document needs. The service
+      # creates only `dev`, and that one is scoped to a single schema on purpose.
+      - name: Create the realm-scoped dev database
+        run: >-
+          PGPASSWORD=pw psql -h localhost -p 5432 -U postgres -d postgres
+          -v ON_ERROR_STOP=1 -c 'CREATE DATABASE realmdev'
 
       - name: Set up Go
         uses: actions/setup-go@v7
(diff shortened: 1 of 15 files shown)


denisvmedia:

**Second verdict: pending re-measurement.** Note the two red CI jobs on the previous head.

denisvmedia (commented):
Independent review of head `88487e04` found a reproducible HCL-readability blocker.

**Blocker — duplicate relation labels across schemas fall back to the wrong block kind.** In `internal/atlashclrender/objects.go:739-749`, `documentDeclares(ref.Name)` fails when two schemas contain the same view label, then the fallback emits `table.<schema>.<name>` regardless of the actual relation kind. With `public.v` and `other.v` as views, grants render as `table.public.v` and `table.other.v`; pinned Atlas CE rejects the document because no `table` variable exists.

Resolve targets by `(schema, name)` while preserving the declared block kind, or use a quoted qualified relation fallback. Add grant and trigger round-trip tests with same-named views in two schemas and require the generated HCL to parse. The current default inspect output is not reliably readable.

denisvmedia:
Thanks — the blocker reproduced on the first attempt and is fixed in 0576974. Point by point.

**"`documentDeclares(ref.Name)` fails when two schemas contain the same view label, then the fallback emits `table.<schema>.<name>` regardless of the actual relation kind."** Confirmed, before touching any code. Own PostgreSQL 17 container on port 15303, CI-shaped, removed at the end; dev database DROPped and CREATEd before every oracle run; oracle is the pinned `atlas community version v1.3.0`. `CREATE SCHEMA other` plus `t`/`v` in both schemas, then `ptah-compat schema inspect -u '…?sslmode=disable' --schema public --schema other` on 88487e04:

```
for = table.other.v   (x7 privileges)
for = table.v         (x7)
for = table.other.t   (x7)
on  = table.other.v   (with an INSTEAD OF trigger, separate fixture)
```

Oracle: exit 1, `Unsupported attribute; This object does not have an attribute named "other".` Trigger fixture: exit 1, same message, at the `on` line.

Two details of your write-up I have to correct, both measured, neither changing your conclusion:

- The view **inside** the default schema renders `for = table.v`, not `table.public.v`. The reader drops the schema of everything in the schema it is reading, and `objectRef` only writes two parts when the name carries one. It is refused too, as `… an attribute named "v"`.
- `There is no variable named "table"` appears only where the document declares **no** table block at all. I built that control (views only, no tables) and got exactly that message. In the real inspect of your shape there are table blocks, so the verdict is the attribute error instead. Both refuse; I have named which document gets which.

**"Resolve targets by `(schema, name)` while preserving the declared block kind, or use a quoted qualified relation fallback."** I took the second, and I want to be explicit that the first is not available rather than merely less convenient. One operand varied against the rendered two-schema document, nothing else touched:

```
for = table.other.v   exit 1  Unsupported attribute; ... named "other"
for = view.other.v    exit 1  same message
for = view.v          exit 0  -- and means NEITHER of the two blocks
for = "other.v"       exit 0
for = "gone"          exit 0   (against for = table.gone, exit 1)
```

A reference names a block by its **labels**, so correcting the kind while keeping the traversal is refused identically. The only kind-correct traversal left, `view.v`, is worse than a refusal: it evaluates and points at nothing in particular, and Ptah's own `relationBlockSchema` returns nothing for a label it finds twice — so reading it back drops the schema for good. `renderer.relationRef` now takes the unresolved SPELLING from its caller: a relation position passes `quote(name)`; a SEQUENCE target still passes `sequence.<name>`, because there the word is the only thing saying which object the grant is about and `for = "order_seq"` would read back as a grant on a relation and emit `GRANT … ON TABLE`. The `tableRef` positions — `ref_columns`, `policy.on`, `data.table` — are untouched; `ref_columns` has to reach a column through the same expression, so a quoted name is not a spelling available there.

One thing your review did not name, which I found while writing the round-trip fixture you asked for: **the reader had the same defect.** `relationBlockSchema` counted only the QUALIFIED declarations of a label, so `view "v"` plus `view "v"` in `other` looked unambiguous, the bare reference was moved into `other`, and `goschema.Deduplicate` then merged it with the grant that really was on `other.v` — two grants rendered, one parsed. Fixed in the same commit; every declaration is counted now, including the ones carrying no schema.

**"Add grant and trigger round-trip tests with same-named views in two schemas and require the generated HCL to parse."** Done, in three places:

- `TestOracleReadsTheTwoSchemaRelationDocumentPtahRenders` renders through `RenderInspectedForAtlasCLI` — the entry point `ptah-compat schema inspect` calls — and puts the whole document to the pinned binary. Three one-substitution `unreadable` controls assert the binary's message, not a bare non-zero status: target as a qualified table, target as a qualified **view**, and the trigger's `on` as a qualified table. PostgreSQL only, and not by preference: that binary refuses a two-schema document on SQLite outright with `cannot use HCL with more than 1 schema when dev-url is limited to schema "main"`, so there is no SQLite instance of this shape.
- `TestRenderNamesTheBlockTypeTheDocumentDeclares` gains the grant and trigger rows for a label two schemas both declare.
- `TestRenderedDuplicateRelationLabelRoundTrips` is Ptah reading its own output back: `other.v` and the bare `v` both come back as written, on both positions.

The oracle row needs a **realm-scoped** dev URL, so `atlas-oracle.yml` gains `PTAH_ATLAS_ORACLE_POSTGRES_REALM_DEV_URL` and a separate `realmdev` database. Separate, not shared, for a measured reason: a realm-scoped read requires a clean database — with a stale `other` schema seeded, the next read fails with `sql/migrate: connected database is not clean: found schema "other"`. The selection guard moves 6 → 7 and the four new subtests are named rather than counted, like the rest of that job.

**"The current default inspect output is not reliably readable."** Re-measured, all eight rows, on 88487e04 and on the fix:

| document | 88487e04 | fix |
| --- | --- | --- |
| realm inspect, `view "v"` in two schemas | 1 attribute | **0** |
| realm inspect, trigger on one of those | 1 attribute | **0** |
| empty database | 0 | 0 |
| `--exclude '*[type=role]'` | 0 | 0 |
| granted table excluded | 0 | 0 |
| unfiltered table + view + grants | 0 | 0 |
| one table, one grant | 0 | 0 |
| a grant on a sequence | 1 `sequences are not supported` | 1 unchanged |

The sequence row is the block-level gap the docs already disclose (a grant keeps the sequence block alive, and that binary refuses any file declaring one); identical on master, on 88487e04 and on the fix.

No churn: `diff` exit 0 on all six documents that were already readable, exit 1 only on the two that were not — byte-identical on `ptah-compat schema inspect`, on native `ptah schema inspect`, on SQLite, and on a MySQL 9.7 database of my own. Executed rather than rendered: `ptah-compat schema apply` of the fixed two-schema document runs `GRANT SELECT ON TABLE "other"."v" TO "ptah_user" WITH GRANT OPTION` at exit 0, `information_schema.views` and `role_table_grants` assert the catalog afterwards, and re-inspecting the applied state returns the same document byte for byte. A genuine change (`CREATE TABLE other.extra`) still shows up, and the changed document is still read at exit 0.

Mutants, three, each alone:

- **A, revert the fix.** Reddens the new grant row, the new trigger row, the round-trip test, all three rows of `TestRenderKeepsQualifiedReferenceWhenDocumentCannotResolveIt`, and the new oracle run. Leaves every fixture the PR already had green, including `TestOracleReadsTheInspectedDocumentPtahRenders`.
- **B, over-correct — quote every relation reference.** Reddens the originals: `for = view.v`, `for = materialized.mv`, `for = table.t`, `on = view.v`, `for = sequence.order_seq`, both `KeepsItsSchema` runs, `TestRenderShortensTableReferenceWhenDocumentDeclaresTheTarget`, and the oracle runs `…InspectedDocument…` and `…ReferenceSpelling…` on **both** dialects. The new rows stay green, which is exactly why the originals had to be re-run.
- **C, revert only the reader half.** Reddens the round-trip test and nothing else.

Docs: the third `permission` rule said a target the document declares no block for "keeps the spelling the position implies", which the fix makes false. It now names the quoted form, both cases that reach it, the two-schema example, and the measurement. The PR body is rewritten too — the stale adversarial-review sections described 47f65146, two commits back.

Every gate: build 0 | vet 0 | vet -tags integration 0 | test ./… 0 (one sweep, no kills) | qtlint 0 | golangci-lint 0 "0 issues." | check-test-style 0 | check-public-api 0 | snapshot 0 | released 0 (three approved pre-v1 incompatibilities inherited from master, no new one) | docs check-style 0. `npm run check:responsive` in docs/site exits 1 for want of a built `dist` and does so without my change too — flagged rather than claimed.

One thing I did **not** fix, filed instead: `ref_columns` in the same duplicate-label document still writes `table.<schema>.<name>`, and there is no readable alternative on that surface because the expression has to reach a column. Also unfixed and pre-existing: multi-schema `schema apply` does not converge — a second apply re-plans `CREATE TABLE other.t` and fails with `relation "t" already exists`, identically on 88487e04, so it is not this branch's to absorb.

PR #892, opened 2026-07-30
security: capture rooted migration snapshots

denisvmedia:
Closes #888\n\n## Summary\n\n- open local migration directories through rooted handles and capture stable immutable snapshots before database access\n- carry the same captured filesystem through Atlas apply/down/status/lint/set, shadow verification, execution, and reporting\n- resolve Atlas, golang-migrate, Goose, Flyway, Liquibase, and dbmate apply inputs from captured filesystems\n- retain the anchored project root from  evaluation through contained migration-directory capture\n- preserve intentional project-relative  paths while rejecting symlink escapes from contained directories\n- capture native repair input before database access and document the snapshot boundary\n\n## Security and behavior coverage\n\n- directory-handle replacement and post-capture mutation do not change captured input\n- symlinked migration files outside the selected directory are rejected\n- project migration-directory symlink escapes are rejected before opening the database\n- all six supported apply formats cross the same pre-connection boundary\n- up/down SQL plus , , and  remain available in Atlas captures\n- unsupported Goose  and Liquibase XML/YAML/JSON inputs fail before database connection instead of disappearing during capture\n-  keeps its Atlas-compatible error ordering and directory diagnostics\n\n## Scope boundary\n\nWritable  staging, journal, publication, checksum commit, rollback quarantine, cleanup, and recovery remain pathname-based and are not claimed as fixed here. The final security review found that the earlier partial diff preparation still reopened the pathname for writes, so it was removed from this PR. Full rooted writable publication/recovery is tracked in #895.\n\n## Verification\n\n- ok  	github.com/stokaro/ptah/atlascompat	0.800s
?   	github.com/stokaro/ptah/cmd	[no test files]
ok  	github.com/stokaro/ptah/cmd/atlas	32.957s
ok  	github.com/stokaro/ptah/cmd/compare	2.380s
ok  	github.com/stokaro/ptah/cmd/db	2.078s
ok  	github.com/stokaro/ptah/cmd/drift	0.815s
ok  	github.com/stokaro/ptah/cmd/dropall	1.116s
ok  	github.com/stokaro/ptah/cmd/generate	2.317s
ok  	github.com/stokaro/ptah/cmd/integration-test	2.009s
ok  	github.com/stokaro/ptah/cmd/internal/buildinfo	2.216s
ok  	github.com/stokaro/ptah/cmd/internal/cliobs	2.842s
ok  	github.com/stokaro/ptah/cmd/internal/cmdadapter	2.338s
[...]

PR #1304, opened 2026-08-08
Hold the migration directory open from planning to publication (#1118)

denisvmedia:
Follow-up to #1288, on top of its head `d9d2a1de`. That PR carries `hold-off-merge` and three red CI jobs; this commit is what they were waiting for.

## Why #1288 is red

All three red jobs — `unit-tests`, `coverage`, `windows-publication` — are the same test, `TestMigrationPlanWriteFiles_RefusesRecreatedDirectory`, and it is correctly red.

`migrationDirectoryState` recorded the planned directory's identity as an `fs.FileInfo` taken by `os.Stat` at planning time, and publication asked `os.SameFile` whether the directory it was about to write into was still that object. `os.SameFile` is device + inode on Unix and volume + file index on Windows, and neither identifier is reserved once the last handle to the object is gone. A directory removed and recreated at the same pathname can be handed the identifier straight back — and then the comparison approves the substitute. That is the shape an attacker reaches for first: it needs no second directory to point at.

## Reproduced deliberately, not waited for

Built a loopback ext4 filesystem inside a privileged Linux container and pointed only the test binary's `TMPDIR` at it (`go test -c` first — building on the image fills it). At `d9d2a1de`:

```
--- FAIL: TestMigrationPlanWriteFiles_RefusesRecreatedDirectory (0.17s)
    plan_root_test.go:52: got nil error but want non-nil
      want: "migration directory changed after migration planning"
```

`nil` means the batch was published — into the directory the test had just substituted. Docker's own `/tmp` (overlayfs) reproduced nothing, so "run it on Linux" is not the reproduction; the filesystem has to be one that recycles.

Then the mechanism, isolated. One program, one `os.SameFile` call, 20 remove-and-recreate cycles of one pathname, three regimes that differ only in how the earlier identity was obtained and how long the object stayed open:

| filesystem | identity taken by | `os.SameFile` said "same object" |
| --- | --- | --- |
| ext4 | `os.Stat`, no handle held | **20 of 20** |
| ext4 | `os.Root.Stat(".")`, handle **closed** before the removal | **20 of 20** |
| ext4 | `os.Root.Stat(".")`, handle **held** across the removal | 0 of 20 |
| APFS | all three regimes | 0 of 20 |

Rows two and three differ in nothing but the handle's lifetime — ext4 gave inode 14 back every time when nothing held it, and went 35 → 36 when something did. An open directory keeps its identifier allocated; a closed one does not. Comparing more cleverly cannot fix this, and no build tag should hide it.

On Windows the old identity was worse than unlucky: `os.Stat` defers `loadFileId` to `SameFile` time and resolves the *pathname* then, so the recorded identity re-resolved to the replacement regardless of file-index reuse.

## The change

The plan binds the migration directory while it is being built and **holds** that binding until it publishes. `PlanMigration` binds through the confinement root, keeps the directory and parent handles, closes the root (the handles were opened through it and stay valid), and hands the binding to the plan; a plan that fails to be produced closes it, and a plan dropped without publishing releases it when collected, because `os.Root` closes its descriptor from a finalizer.

`WriteFilesContext` reopens nothing. Under the lock it revalidates the handle it was given against the pathname it was selected by, compares the contents, publishes through that same handle, then closes it. The version scan reads the same handle instead of the pathname.

`MigrationWriter.Identity` is replaced by `MigrationWriter.Revalidate`, which asks the whole question rather than handing out half of it: for an existing directory, does the selected pathname still name the object I hold (so an ancestor swapped anywhere above it is caught too); for a directory that was absent when the writer bound, is that name still free in the parent I hold.

`migrationDirectoryState` is gone. The plan carries `plannedContents fsnapshot.Snapshot` and nothing else, because `exists` was provably dead — both captures derived from the same bind-time `writer.Exists()`, so the comparison could never be false. Mutant **M2** proves the point: removing `revalidateAbsent` reddens exactly the "directory appeared after planning" row while the contents check stays silent.

The public `migration/generator` API is unchanged: `docs/public_api.snapshot` does not move and all three public-API gates are 0.

**This changes behavior twice.** A `ptah migrations generate` run whose directory is removed and recreated between planning and publication now fails and writes nothing, where it published into the replacement before. And while a plan is outstanding the migration directory is held open, so another process cannot rename or remove it until the plan publishes or is discarded. Pre-v1, so no compatibility with Ptah's own past is owed; both are documented in `docs/site/src/content/docs/concepts/migration-directory.md` and `migration/generator/README.md`.

## Regressions

The failing test is unchanged in what it asserts and is still **not** build-tagged, so it runs on the Windows job as before. One line of its setup moved from `os.Remove` to `os.RemoveAll`: the plan now holds a handle on the directory while the test removes it, and Go's `RemoveAll` deletes through the parent with POSIX semantics on every platform it supports, while `os.Remove` on Windows goes through `RemoveDirectory`, which marks an open directory for deletion on close and leaves the name unusable until then. The hostile step has to be one an attacker can actually complete.

New, portable, and covered by the Windows job's `-run '^TestMigrationPlanWriteFiles_'`:

- `TestMigrationPlanWriteFiles_RefusesADirectoryThatAppearedAfterPlanning` — the absent-directory half. Its control is `..._CreatesMissingOutputParents`, the same shape with nothing racing it.

Reshaped:

- `TestMigrationPlanWriteFiles_RefusesDirectoryReplacedBeforePublication` — the two rows used to separate "recorded identity refuses it" from "the rooted open refuses it". The handle now refuses both, so the rows assert something better: the refusal does not depend on a confinement root being configured, and the direct-CLI shape gets it too.
- `TestMigrationPlanWriteFilesReplacedDirectoryCannotRedirectPublication` — the white-box seam moved from `afterMigrationWriterBound` (which for a plan now fires during *planning*) to a new `afterMigrationPublicationVerified`, the one window the writer still has: after the verification, before the first byte. A replacement landing there cannot be refused; the assertion is the positive one — the files land in the bound object and the decoy is untouched.
- `TestMigrationPlanBindRefusesADirectoryReachedThroughAnAncestorOutsideTheRoot` — new, and the only thing pinning `openOutputRoot`. See M3.

The #1288 escape probes were re-measured on ext4 at `-test.count=40` — `TestGenerateEmptyMigrationReplacedDirectoryCannotRedirectTheWrite` (4 rows including the ancestor row), `...CreatesMissingDirectoryThroughTheBoundParent`, `...ReplacedDirectoryCannotRedirectPublication`: exit 0, 0 escapes.

## Mutation, both directions

Run over `./migration/generator ./internal/atlasmigrate ./cmd/migrate`; all restored afterwards, restored run exit 0.

| mutant | red | green |
| --- | --- | --- |
| **M1** revert: `revalidate` stops comparing an existing directory | `RefusesRecreatedDirectory`, both `RefusesDirectoryReplacedBeforePublication` rows | every publication and create test |
| **I1** over-correct: `revalidate` always reports changed | 8 generator rows incl. `HappyPath`, `PublishesMultiplePairsAndReports`, `...CannotRedirectPublication`, plus `cmd/migrate`'s replay test | the two refusal tests, every create row |
| **M2** revert: `revalidateAbsent` returns nil | `RefusesADirectoryThatAppearedAfterPlanning` **only** | everything else |
| **I2** over-correct: `revalidateAbsent` always errors | `CreatesMissingOutputParents`, `cmd/migrate`'s JSON-report test | both refusal tests |
| **M3** revert: `openOutputRoot` always nil | `...AncestorOutsideTheRoot` **only** | everything else |
| **I3** over-correct: an empty root confines to the working directory | 40 rows | — |
| **M4** revert: the contents comparison dropped | `RejectsChangedDirectory` **only** | both identity refusals, every happy path |

M3 is the honest one. The retained parent handle already refuses a symlink escaping the *parent* with the same "outside allowed root" wording whether or not a root was configured, so the root's unique contribution narrows to an ancestor **above** the parent leaving the root — and exactly one row pins it. Without that row M3 is silent.

M0 is the branch as it stood before this commit: RED on ext4, GREEN on APFS, measured above.

## Gates

`go build` 0 · `go vet` 0 · `go vet -tags integration` 0 · **`go test ./... -count=1` 0** (178 packages ok, zero FAIL lines) · `qtlint` 0 · `golangci-lint` 0 · `check-test-style` 0 · `check-public-api` 0 · `check-public-api-snapshot` 0 · `check-public-api-released` 0 · `GOOS=windows` build and vet 0 · all eight `docs/site` check scripts 0 including `build-feature-matrix --check`.

An earlier full run on the same content exited 1 with six subprocess-heavy packages blowing time budgets under concurrent load (`cmd/atlas` killed at 11m, `cmd/ptah-compat`, `cmd/ptah-ls`, `cmd/viz`, `cmd/compare`, `cmd/drift`). All six pass serially on the same tree — `cmd/atlas` in 174s against the 707s at which it was killed — and the run reported above was made on a quiet machine.

## Not verified here

No Windows machine. `GOOS=windows` build and vet are 0, but nothing Windows-specific was executed, and two claims rest on reading Go 1.26.5 rather than on a run: that `os.RemoveAll` frees the name immediately with our handle open (`windows.Deleteat` with `FILE_DISPOSITION_POSIX_SEMANTICS`), and whether `Stat(".")` succeeds on a handle to a POSIX-deleted directory. The second does not matter — a Stat error is wrapped into `ErrMigrationDirectoryChanged` on the same line, so the test asserts the same thing either way. The first does: if NTFS refused POSIX-semantics deletion, the test would go red at its setup rather than its assertion. `windows-publication` is the arbiter.

## Still open on #1118

Carried forward from #1288, untouched here:

- `cmd/migrate/new.go --edit` rehashes through `migrateops.Rehash(dir, format)` by pathname after the external editor exits.
- `WriteCheckpointFiles`, `WriteDataMigrationFiles` and `WriteAtlasCheckpointFile` (`ptah migrations checkpoint`, `ptah migrations data`) still write by pathname, and `ResolveAtlasCheckpointVersion` still scans by pathname. Not named by #1118, and carrying no `AllowedOutputRoot`, but the machinery to convert them is in place.

Refs #1118.

---

### What it found

The diagnosis and the fix hold — I reproduced every row of it independently — but the change ships a false user-facing sentence, leaves half of its own stated behavior unmeasured, and lands on a branch whose adjacent shape it never built.

WORKTREES (all detached, `git log --oneline -1` verified before every build)
  head    scratchpad/rh1118/head    f2ac181b
  parent  scratchpad/rh1118/parent  d9d2a1de
  base    scratchpad/rh1118/base    4fd9f18a  (merge-base with origin/master)
  mut     scratchpad/rh1118/mut     f2ac181b  (mutants)
  deliver scratchpad/rh1118/deliver c9499d69 on branch refute/1118-rooted-plan-handle
My own PostgreSQL 17 container rh1118-pg on port 15118 (removed at the end); rh1118_my on mylive-mysql (dropped).

=== WHAT HELD (re-measured, my own rig) ===

EXT4 REPRODUCTION. Loopback ext4 image inside a privileged alpine:3.20 container, arm64, test binaries cross-compiled `GOOS=linux GOARCH=arm64 CGO_ENABLED=0 go test -c`, TMPDIR on the mount:
  d9d2a1de, TMPDIR=/mnt/ext4  -> exit 1, `plan_root_test.go:52: got nil error but want non-nil`
  f2ac181b, TMPDIR=/mnt/ext4  -> exit 0, PASS
  d9d2a1de, TMPDIR=/tmp (overlayfs) -> exit 0, PASS
Exactly what they claimed, including that Linux alone is not sufficient.

ESCAPE PROBES. `-test.count=40` over the four named tests plus the three refusal rows, on ext4 at f2ac181b: exit 0.

MUTANTS. Applied to the committed tree in `scratchpad/rh1118/mut`, run over `./migration/generator ./internal/atlasmigrate ./cmd/migrate`, restored between each (restored run exit 0, all three ok). I ran 6 of their 7; I did not run I3.
  M1 (revalidate returns nil for an existing dir) -> RED exactly: RefusesRecreatedDirectory, RefusesDirectoryReplacedBeforePublication/{no_allowed_root,under_an_allowed_root}.
  M2 (revalidateAbsent returns nil)               -> RED exactly one row: RefusesADirectoryThatAppearedAfterPlanning.
  M3 (openOutputRoot returns nil)                 -> RED exactly one row: TestMigrationPlanBindRefusesADirectoryReachedThroughAnAncestorOutsideTheRoot.
  M4 (contents comparison dropped)                -> RED exactly one row: RejectsChangedDirectory.
  I1 (revalidate always changed)                  -> RED 9 generator rows + cmd/migrate TestMigrateGenerateReplayDerivesCurrentStateFromDirectory.
  I2 (revalidateAbsent always errors)             -> RED CreatesMissingOutputParents + TestMigrateGenerateJSONReportWritesSiblingSafetyArtifact.
Every row set matches their report.

GATES on f2ac181b, exit code on its own line: go build 0 | go vet 0 | go vet -tags integration ./integration/... 0 | go test ./... -count=1 1 (single row, cmd/viz TestSVGReportsGraphvizStderrOnFailure, 10.00s Graphviz timeout under my parallel load; re-run alone -> exit 0, ok 0.407s; 177 ok otherwise) | qtlint 0 | golangci-lint (private cache) 0, "0 issues." | check-test-style 0 | check-public-api 0 | check-public-api-snapshot 0 | check-public-api-released 0 | GOOS=windows build 0 | GOOS=windows vet 0 | all 8 docs/site scripts 0.

OTHER LAYERS / DIALECTS / ROLE. `ptah migrations generate` end-to-end, base vs head, identical output on: PostgreSQL 17 as ptah_user, PostgreSQL 17 as the bootstrap postgres superuser, MySQL 9.7, `--replay` (twice each, second run a no-op), relative `--migrations-dir` through a symlink, `--dir` is a regular file. Adjacent shapes that pass at head: nested subdirectory in the migration directory, `--dir` is a symlink, ancestor is a symlink, missing parents, an abandoned plan does not block a second plan over the same directory. The README claim "a plan that is never published releases its handles when it is collected" is TRUE: 200 dropped plans peaked at 54 open fds and settled back to the 6-fd baseline after GC.

=== WHY IT IS REFUTED ===

(1) A FALSE SENTENCE SHIPS, CONTRADICTED BY THE COMMIT'S OWN TESTS.
docs/site/src/content/docs/concepts/migration-directory.md:134-137 says, with no qualifier:
  "This changes behavior twice over: a run that previously published into a recreated
   directory now fails instead, and while a plan is outstanding the migration directory
   is held open, so another process cannot rename or remove it until the plan publishes
   or is discarded."
That is a Windows property. On Unix an open directory is renamed and unlinked freely, and two tests added or edited by this same commit assert precisely that as their hostile step:
  plan_root_test.go:67       `c.Assert(os.RemoveAll(outputDir), qt.IsNil)` with the plan holding the handle
  plan_root_unix_test.go     `c.Assert(os.Rename(path, retained), qt.IsNil)` (replaceWithSymlink) likewise
Both pass on macOS APFS and on Linux ext4 (measured above). migration/generator/README.md:96-99 carries the correct qualifier ("so on Windows another process cannot rename or remove it"), so the two doc surfaces disagree and the reader-facing one is the wrong one. Their own "could not verify" item 2 identifies the property as Windows-specific and asserts it is "Documented in both doc surfaces"; it is documented in one and misdocumented in the other. The commit message repeats the unqualified form.

(2) HALF THE CHANGE IS UNPINNED. "The version scan moved onto the same handle" is stated in the commit message, in the doc comment on nextAvailableMigrationVersion, and in the PlanMigration comment. Reverting it:
    return nextAvailablePtahVersion(migrationDirFileNames(writer.Path()), version, migrationName), nil
  `go test ./migration/generator/ ./internal/atlasmigrate/ ./cmd/migrate/ -count=1` -> exit 0, three ok, no FAIL lines.
Nothing in the repository measures it, because every fixture has the pathname and the handle naming the same directory. Their seven mutants cover revalidate, revalidateAbsent, openOutputRoot and the contents comparison, and stop there.

(3) ADJACENT SHAPE, USER-VISIBLE, NEVER BUILT: a symlinked migration file in the directory. Legitimate layout (a shared migration linked in), nothing documents it as refused.
  fixture: migrations/1000000000_shared.up.sql -> 

### What it says must change

The mechanism, the fix and the mutant table are correct — I reproduced the ext4 failure at d9d2a1de and the pass at f2ac181b myself, and six of their seven mutants redden the exact rows they list. What is wrong is what ships around it.

The claim "while a plan is outstanding the migration directory is held open, so another process cannot rename or remove it until the plan publishes or is discarded" is not true on Unix and is published unqualified on the reader-facing concept page. It is a Windows property. Two tests in the same commit remove and rename the held directory and assert the hostile step succeeds; both pass on Linux ext4 and macOS. The package README states it correctly ("so on Windows"), so the claim that it is "documented in both doc surfaces" is wrong: it is documented in one and misstated in the other. Say which platform, and say that on Unix the removal succeeds and the guard is what refuses to publish afterwards.

The version scan moving onto the bound handle is asserted three times in prose and measured nowhere. Reverting it to migrationDirFileNames(writer.Path()) leaves migration/generator, internal/atlasmigrate and cmd/migrate at exit 0. It needs a row that makes the pathname and the handle name different directories.

Two things the change should also answer for. A failed WriteFilesContext never closes the plan's handles — measured 8 open fds before and after the failure against a 6-fd baseline — and MigrationPlan has no Close, so the Windows "cannot rename or remove" window documented as ending at "publishes or is discarded" has no end after a failure. And the branch as a whole now refuses a migration directory containing a symlinked .sql file: `ptah migrations generate` exits 0 on 4fd9f18a and exits 2 on both d9d2a1de and f2ac181b with "path escapes from parent". That one is d9d2a1de's, not this commit's, but it is in the neighbourhood this commit says it enumerated, and it is what an operator sees.

### DIFF
--- a/docs/site/src/content/docs/concepts/migration-directory.md
+++ b/docs/site/src/content/docs/concepts/migration-directory.md
@@ -115,12 +115,26 @@ for what remains keyed to the pathname and why.
 
 `ptah migrations generate` plans and publishes in two steps, so the directory
 can be replaced between them by something outside the run's control. The plan
-records the planned directory's filesystem identity as well as its contents, and
-publication refuses a directory that is no longer the same object — including a
-substitute that holds exactly the files the plan verified. The refusal is
-`migration directory changed after migration planning`, and nothing is written
-([#1118](https://github.com/stokaro/ptah/issues/1118)). This changes behavior:
-a run that previously published into a recreated directory now fails instead.
+binds the directory while it is being built and holds that binding open until
+it publishes, so the plan is a claim on a filesystem object rather than on a
+pathname. Publication refuses a directory that is no longer the object the plan
+holds — a substitute that holds exactly the files the plan verified, and a
+directory removed and recreated at the same pathname, are both refused. The
+refusal is `migration directory changed after migration planning`, and nothing
+is written ([#1118](https://github.com/stokaro/ptah/issues/1118)).
+
+Holding the directory open is what makes the answer the same on every platform.
+Comparing a recorded `os.SameFile` identity instead would ask the operating
+system whether two identifiers match, and an identifier belonging to a removed
+directory can be handed straight back to its replacement: measured over 20
+remove-and-recreate cycles of one pathname, ext4 reissued it 20 times and APFS
+none, so the same code refused the substitution on one filesystem and accepted
+it on another. An open directory cannot have its identifier reissued.
+
+This changes behavior twice over: a run that previously published into a
+recreated directory now fails instead, and while a plan is outstanding the
+migration directory is held open, so another process cannot rename or remove it
+until the plan publishes or is discarded.
 
 ## Consequences
 
--- a/docs/site/src/content/docs/concepts/migration-directory.md
+++ b/docs/site/src/content/docs/concepts/migration-directory.md
@@ -116,12 +116,13 @@ for what remains keyed to the pathname and why.
 `ptah migrations generate` plans and publishes in two steps, so the directory
 can be replaced between them by something outside the run's control. The plan
 binds the directory while it is being built and holds that binding open until
-it publishes, so the plan is a claim on a filesystem object rather than on a
-pathname. Publication refuses a directory that is no longer the object the plan
-holds — a substitute that holds exactly the files the plan verified, and a
-directory removed and recreated at the same pathname, are both refused. The
-refusal is `migration directory changed after migration planning`, and nothing
-is written ([#1118](https://github.com/stokaro/ptah/issues/1118)).
+its publication attempt returns, so the plan is a claim on a filesystem object
+rather than on a pathname. Publication refuses a directory that is no longer
+the object the plan holds — a substitute that holds exactly the files the plan
+verified, and a directory removed and recreated at the same pathname, are both
+refused. The refusal is `migration directory changed after migration planning`,
+and nothing is written
+([#1118](https://github.com/stokaro/ptah/issues/1118)).
 
 Holding the directory open is what makes the answer the same on every platform.
 Comparing a recorded `os.SameFile` identity instead would ask the operating
@@ -131,10 +132,24 @@ remove-and-recreate cycles of one pathname, ext4 reissued it 20 times and APFS
 none, so the same code refused the substitution on one filesystem and accepted
 it on another. An open directory cannot have its identifier reissued.
 
-This changes behavior twice over: a run that previously published into a
-recreated directory now fails instead, and while a plan is outstanding the
-migration directory is held open, so another process cannot rename or remove it
-until the plan publishes or is discarded.
+Holding the directory open is not a lock on it. On Unix another process renames
+or removes the held directory exactly as it always could; the guarantee is that
+the plan keeps writing to the object it retained, and refuses to write at all
+once revalidation shows the pathname no longer names that object. On Windows an
+open handle is stronger and the rename or removal may be refused or deferred
+instead. The plan releases the directory when its publication attempt returns —
+whether that attempt published or failed — and a plan that is never published at
+all releases it when it is collected.
+
+This changes behavior twice over. A run that previously published into a
+recreated directory now fails instead. And a migration directory containing a
+symbolic link that points outside itself — a shared migration linked in from
+elsewhere — is refused by `ptah migrations generate` rather than followed,
+because the directory is read, checksummed and published through the object the
+run opened. `ptah migrations up` already refused such a directory, so what the
+generator used to do was publish into a directory the applier would not read;
+the two now agree. A link whose target is inside the migration directory is
+unaffected. The refusal names the entry and the rule.
 
 ## Consequences
 
--- a/docs/site/src/content/docs/concepts/migration-directory.md
+++ b/docs/site/src/content/docs/concepts/migration-directory.md
@@ -97,18 +97,30 @@ bytes.
 
 ## Writing back to the directory
 
-The verbs that write a migration directory — `migrate diff` and native
-`ptah migrate generate` — bind it the same way and keep the binding. They open
-the directory and its parent once, before staging, and every staged file,
-published migration, `atlas.sum`, journal, commit marker, rollback quarantine
-and cleanup entry is named as a direct child of one of those two handles.
-Recovery of an interrupted batch runs through the same handles.
+The verbs that write a migration directory — `migrate diff`, `migrate new`, and
+native `ptah migrations generate` and `ptah migrations create` — bind it the
+same way and keep the binding. They open the directory and its parent once,
+before staging, and every staged file, published migration, `atlas.sum`,
+journal, commit marker, rollback quarantine and cleanup entry is named as a
+direct child of one of those two handles. A migration directory that does not
+exist yet is created through the bound parent, so it is materialized where the
+run looked rather than where the pathname points by then. Recovery of an
+interrupted batch runs through the same handles.
 
 Replacing the directory after the run validated it therefore cannot redirect
 what it writes, and a directory configured through `atlas.hcl` stays inside the
 opened project root. See
 [the publication boundary](../../atlas/migrate-commands/#the-publication-boundary)
 for what remains keyed to the pathname and why.
+
+`ptah migrations generate` plans and publishes in two steps, so the directory
+can be replaced between them by something outside the run's control. The plan
+records the planned directory's filesystem identity as well as its contents, and
+publication refuses a directory that is no longer the same object — including a
+substitute that holds exactly the files the plan verified. The refusal is
+`migration directory changed after migration planning`, and nothing is written
+([#1118](https://github.com/stokaro/ptah/issues/1118)). This changes behavior:
+a run that previously published into a recreated directory now fails instead.
 
 ## Consequences
 
--- a/internal/atlasmigrate/diff_internal_test.go
+++ b/internal/atlasmigrate/diff_internal_test.go
@@ -150,13 +150,15 @@ func TestWritePublicationJournal_FallsBackWithoutHardLinks(t *testing.T) {
 	c.Assert(removePublicationJournal(writer), qt.IsNil)
 }
 
-func TestPublishArtifactsLocked_PublishesCompleteBatch(t *testing.T) {
+func TestMigrationWriterPublishArtifacts_PublishesCompleteBatch(t *testing.T) {
 	c := qt.New(t)
 	dir := c.TempDir()
+	migrationWriter, err := OpenMigrationWriter(nil, dir)
+	c.Assert(err, qt.IsNil)
+	defer func() { _ = migrationWriter.Close() }()
 
-	paths, err := PublishArtifactsLocked(
+	paths, err := migrationWriter.PublishArtifacts(
 		t.Context(),
-		dir,
 		[]PublicationArtifact{
 			{Name: "1_change.up.sql", Contents: []byte("SELECT 1;\n")},
 			{Name: "1_change.down.sql", Contents: []byte("SELECT 2;\n")},
@@ -183,15 +185,17 @@ func TestPublishArtifactsLocked_PublishesCompleteBatch(t *testing.T) {
 	c.Assert(err, qt.ErrorIs, os.ErrNotExist)
 }
 
-func TestPublishArtifactsLocked_CollisionRollsBackWholeBatch(t *testing.T) {
+func TestMigrationWriterPublishArtifacts_CollisionRollsBackWholeBatch(t *testing.T) {
 	c := qt.New(t)
 	dir := c.TempDir()
 	existingPath := filepath.Join(dir, "1_change.down.sql")
 	c.Assert(os.WriteFile(existingPath, []byte("existing\n"), 0o600), qt.IsNil)
+	migrationWriter, err := OpenMigrationWriter(nil, dir)
+	c.Assert(err, qt.IsNil)
+	defer func() { _ = migrationWriter.Close() }()
 
-	paths, err := PublishArtifactsLocked(
+	paths, err := migrationWriter.PublishArtifacts(
 		t.Context(),
-		dir,
 		[]PublicationArtifact{
 			{Name: "1_change.up.sql", Contents: []byte("SELECT 1;\n")},
 			{Name: "1_change.down.sql", Contents: []byte("SELECT 2;\n")},
--- a/internal/atlasmigrate/migrationwriter.go
+++ b/internal/atlasmigrate/migrationwriter.go
@@ -1,0 +1,172 @@
+package atlasmigrate
+
+import (
+	"context"
+	"errors"
+	"io/fs"
+
+	"go.5x5.cz/ptah/internal/migratesum"
+	"go.5x5.cz/ptah/internal/pathguard"
+)
+
+// MigrationWriter is the rooted migration-directory capability a writer
+// transaction outside this package runs through: one binding of the directory
+// and its parent, reused for every read, create, publication and rollback the
+// transaction performs.
+//
+// It exists because the native generator had the shape stokaro/ptah#1118
+// describes. `ptah migrations create` resolved its output directory once, then
+// reopened it by pathname for the mkdir, for the version scan, for each file it
+// created and for the atlas.sum commit; `ptah migrations generate` verified the
+// directory's contents by pathname and then let the publication resolve the
+// pathname a second time. A directory or ancestor replaced between any two of
+// those steps takes the write somewhere the verification never looked.
+//
+// The methods are deliberately thin: this type is the boundary, not a second
+// filesystem abstraction. Everything behind it is the same migrationWriterDir
+// the `migrate diff` publication already runs through, so both writers share one
+// answer to "which filesystem object is this transaction committing to".
+//
+// A nil root keeps direct-CLI behavior, where an explicit absolute directory is
+// the operator's own choice of destination. The binding still happens; it is
+// simply not confined to a project root.
+type MigrationWriter struct {
+	dir *migrationWriterDir
+}
+
+// OpenMigrationWriter binds dir and its parent without creating dir. A missing
+// migration directory is reported through Exists rather than as an error, so a
+// caller can compare the directory's absence against what it planned for before
+// deciding to materialize it.
+//
+// The parent must already exist; call EnsureMigrationParent first when the
+// caller accepts a directory nested below missing levels.
+func OpenMigrationWriter(
+	root *pathguard.OpenedDirectory,
+	dir string,
+) (*MigrationWriter, error) {
+	bound, err := openMigrationWriterDir(root, dir)
+	if err != nil {
+		return nil, err
+	}
+	return &MigrationWriter{dir: bound}, nil
+}
+
+// EnsureMigrationParent creates dir's parent, and every level missing above it,
+// through root when one is supplied.
+func EnsureMigrationParent(root *pathguard.OpenedDirectory, dir string) error {
+	return ensureMigrationDirParent(root, dir)
+}
+
+// Close releases the directory and parent handles.
+func (w *MigrationWriter) Close() error {
+	return w.dir.Close()
+}
+
+// Path returns the stable lexical path selected for the migration directory. It
+// is display and result data only; no write step resolves it again.
+func (w *MigrationWriter) Path() string {
+	return w.dir.Path()
+}
+
+// Exists reports whether the migration directory was present when the handles
+// were bound, or has been materialized by Create since.
+func (w *MigrationWriter) Exists() bool {
+	return w.dir.Exists()
+}
+
+// Create materializes a missing migration directory through the parent handle
+// bound at open time, so it is created inside the opened root rather than
+// wherever the pathname happens to point by now. Creating a directory that
+// already exists is not an error.
+//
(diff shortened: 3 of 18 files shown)


denisvmedia (commented):
Independent review of head `f2ac181b` supports the retained-handle design but confirms unresolved blockers in its contract and coverage.

**Major — cross-platform documentation is false.** `docs/site/src/content/docs/concepts/migration-directory.md:134-137` says another process cannot rename/remove an outstanding plan's directory. Unix permits rename/unlink of an open directory; the handle preserves object identity. Windows additionally blocks the rename/removal. Qualify the claim by OS.

**Major — no test pins handle-based version selection.** `migration/generator/rootedwriter.go:301-316` reads entries through `writer.Entries`, but the PR body confirms reverting this to pathname enumeration leaves the relevant tests green. Add a test that binds directory A, renames it, creates a replacement at the old pathname with a higher version, and proves version selection and publication use retained A.

The PR body's failed-write handle-leak and symlink behavior concerns also remain unresolved. This should not merge while its own independent review is still refuted.

denisvmedia:
Both blockers reproduced at `f2ac181b` before anything was changed, and both are closed. One clause is disputed, with the measurement. Worktrees: a detached scratch worktree at `f2ac181b` (`git log --oneline -1` verified before every build), delivery worktree on `issue-1118-rooted-plan-handle` at `cf532957`.

Note on what you were reading: `origin/issue-1118-rooted-plan-handle` was `f2ac181b`, but the branch already carried an unpushed `b272b71b` made after your review. Two of your three points were fixed there. That is on us for not pushing it, not on the review.

### 1. "cross-platform documentation is false" — confirmed, fixed

Reproduced textually at `f2ac181b`: `docs/site/src/content/docs/concepts/migration-directory.md:134-137` reads exactly as you quote it, with no platform qualifier, while `migration/generator/README.md:96-99` at that same commit does say "so on Windows" — the two surfaces disagreed and the reader-facing one was wrong. Reproduced behaviorally too: `go test ./migration/generator/ -run 'TestMigrationPlanWriteFiles_' -count=1 -v` on macOS APFS at `f2ac181b` → exit 0, `RefusesRecreatedDirectory` PASS, and that row's hostile step is `plan_root_test.go:67`, `c.Assert(os.RemoveAll(outputDir), qt.IsNil)`, taken while the plan holds the handle. The commit's own test asserts the removal succeeds; the page said it cannot happen.

Fixed by `b272b71b`. The concept page now carries a dedicated paragraph — "Holding the directory open is not a lock on it. On Unix another process renames or removes the held directory exactly as it always could; the guarantee is that the plan keeps writing to the object it retained, and refuses to write at all once revalidation shows the pathname no longer names that object. On Windows an open handle is stronger and the rename or removal may be refused or deferred instead." — and the README carries the same qualification. The commit message no longer makes the unqualified claim either.

Two phrasings survived that correction and are fixed in `cf532957`, because they overstate the hold in the same direction: `migration/generator/README.md:91` and the `bindPlannedMigrationDir` doc comment both said the plan holds the binding "until it publishes". A failed publication releases it too, so both now say "until its publication attempt returns".

### 2. "no test pins handle-based version selection" — confirmed, and pinned with your shape

Reproduced at `f2ac181b`. Baseline `go test ./migration/generator/ ./internal/atlasmigrate/ ./cmd/migrate/ -count=1` → exit 0, three ok. Then `nextAvailableMigrationVersion` reverted to

```go
return nextAvailablePtahVersion(migrationDirFileNames(writer.Path()), version, migrationName), nil
```

→ **exit 0**, three ok, no FAIL lines. Nothing measured it, exactly as you said.

`b272b71b` added `TestNextAvailableMigrationVersionReadsTheHeldDirectory`; the same revert applied there reddens exactly that one row and nothing else. But it stages the divergence with a symlink swap and stops at the number, so `cf532957` adds the rename-shaped one you described: `TestMigrationPlanUsesTheHeldDirectoryForVersionAndPublication`, driven in the order `PlanMigration` runs — bind, capture, scan, publish.

It binds `root/migrations` holding version 105, renames it aside, and creates a replacement at the old pathname holding version **900**. Two rows:

- the replacement keeps the pathname: the scan answers **106**, publication is refused with `ErrMigrationDirectoryChanged`, the replacement receives nothing, the retained directory still holds only its 105 file;
- the pathname is handed back to the retained directory: the scan answers **106**, publication succeeds, and the retained object holds `0000000105_add_email.up.sql`, `0000000106_add_email.down.sql`, `0000000106_add_email.up.sql` while the replacement is untouched.

Both rows also assert what separates the two readings: `nextAvailablePtahVersion(migrationDirFileNames(writer.Path()), 100, "add_email")` on the same writer at the same instant is **901**.

One side effect worth naming: this fixture never unlinks the bound directory, so unlike `RefusesRecreatedDirectory` it does not depend on the filesystem's identifier-reuse behavior at all — the retained object stays linked at another name throughout, and `os.SameFile` has nothing to be lucky about. It is `//go:build unix` only because Win32 refuses to rename a directory the run holds open without `FILE_SHARE_DELETE`, which would make the hostile step fail rather than succeed. That leaves the scan's binding unmeasured on Windows, and I have filed it rather than hidden it.

Mutants over `./migration/generator ./internal/atlasmigrate ./cmd/migrate`, restored between each, restored run exit 0:

| mutant | red rows |
| --- | --- |
| revert the scan to `writer.Path()` | `NextAvailableMigrationVersionReadsTheHeldDirectory` + **both** new rows, nothing else |
| over-correct: the scan ignores the listing | those three + pre-existing `NextAvailableMigrationVersionChecksUpAndDownFiles` |
| revert `revalidate`'s comparison (the PR's own fix) | `RefusesRecreatedDirectory`, both `RefusesDirectoryReplacedBeforePublication` rows, `ReleasesTheDirectoryWhenPublicationFails`, and the new fixture's replacement row only |
| over-correct: `revalidate` always changed | 13 generator rows + 2 `cmd/migrate` rows + the new fixture's restored-pathname row only |

So the scan revert reddens the new fixture while every refusal, publication and create row stays green, and the over-corrections redden the PR's original rows.

**Disputed, one clause.** You asked for a single test that proves version selection *and* publication use retained A in the state where the replacement holds the old pathname. Those cannot both be asserted in that state: `revalidate` compares the selected pathname against the held object (`internal/atlasmigrate/writerdir.go:201-217`), so while the replacement holds the pathname the correct outcome is a **refusal**, not a publication into A. A fixture asserting publication into A there would be asserting that the pathname guard does not fire — the inverse of what this PR is for. Hence the two rows above: the refusal in that state, the publication in the state where the pathname names A again, both over one plan and one retained object, with the version assertion in both.

### 3. The failed-write handle leak — real, and it was the red Windows job

Fixed in `b272b71b`. Measured on macOS over `/dev/fd`: a failed publication went 5 descriptors before planning → 7 while the plan was outstanding → **7 after the failure**, where a successful one went 7 → 9 → 7. `WriteFilesContext` now releases on every exit and a released plan is spent (a second call reports it rather than reaching for a closed handle). Two fixtures pin it — `..._ReleasesTheDirectoryWhenPublicationFails` and its control `..._WhenPublicationSucceeds`, so a release that only ever happened on success shows up — plus `..._FailedPublicationSpendsThePlan`. Reverting the release reddens the failure row and the spent-plan row and leaves the success row green; releasing before publishing instead reddens `ReplacedDirectoryCannotRedirectPublication` and `cmd/migrate`'s replay row.

That leak is also the diagnosis for the four `windows-publication` failures — exactly the four rows whose publication attempt fails, and none of the rows that publish successfully, each failing at `TempDir RemoveAll cleanup: ... being used by another process`.

### 4. The symlinked migration file — decided, kept, explained

Also `b272b71b`. Re-measured end to end against PostgreSQL 17: a migration directory holding `0000009999_shared.up.sql` as a symlink to a file in a sibling directory makes `ptah migrations generate` exit **2** with

```
migration directory <dir>: symbolic links resolving outside it: 0000009999_shared.up.sql;
a migration file linked in from another directory is refused because the whole directory
is read, checksummed and published through the directory itself
```

The refusal is kept rather than relaxed: `ptah migrations up` has refused such a directory since #892, so following the link is how the generator used to publish into a directory the applier would not read. What changed is that it now names the entry and the rule instead of stopping at "path escapes from parent", and both doc surfaces carry it. A link whose target is inside the migration directory still resolves normally, and there is a row for each.

### Not absorbed here

`MigrationPlan` still has no exported `Close`, so a plan built and then abandoned without publishing releases its two descriptors only when it is collected. The docs say exactly that, and your earlier pass measured it working (200 dropped plans peaking at 54 descriptors, settling back to a 6-descriptor baseline after GC) — I did not re-run that one, so I am quoting it rather than claiming it. Either way, on Windows it leaves the directory held for an unbounded time after a caller decides not to publish. Adding `Close` moves the public `migration/generator` API, so it is filed separately rather than widened into this PR.

### Re-measured, both directions

`ptah migrations generate` against PostgreSQL 17: a genuine difference generates the up/down pair and applies — `\d+ users` shows `id integer not null default nextval('users_id_seq'::regclass)`, `email character varying(255)` not null, primary key `users_pkey` — and a second run against the converged database is a no-op at exit 0 with the directory unchanged. Database created under a unique prefix and dropped.

### Gates

All 0 at `cf532957` except two environmental results I will not paper over: `go test ./... -count=1` exited 1 with 176 ok and two packages red on `signal: killed` subprocesses under machine load (`cmd/ptah-ls`, `cmd/viz`), both exit 0 re-run serially on the same tree; and `golangci-lint` first exited 3 with "parallel golangci-lint is running", 0 once the other process released the lock. `check-test-style` was also seen red before being trusted — an injected `if` took it to exit 1 naming the new function.

PR #1305, opened 2026-08-08
Scope the described roles, and take existence from both reads (#1267)

### LINKED ISSUE #1267: PostgreSQL inspect emits every role in the cluster, not the ones the inspected database uses

denisvmedia:
Closes #1267. Supersedes the measurement table in the previous description, which named a control that was not on the branch.

PostgreSQL roles are cluster-wide. `readRoles` read `pg_roles` with no restriction beyond the reserved-name exclusions, so inspecting a one-table database emitted every role anyone had created anywhere on the server, and that output could not be replayed. This branch scopes the description to the roles the inspected schemas actually use — and then, because scoping a description is not the same as deciding what exists, gives the comparator a second read so it can still tell "not described" from "not present".

## Why the second read exists

Scoping alone is what took the integration job red three times. A role outside the inspected scope read as **absent**, the diff planned `CREATE ROLE` for a role that was already there, and the server refused it:

```
ERROR: role "admin_user" already exists (SQLSTATE 42710)
```

`readRolesOutOfScope` reads the exact complement over the same catalog and the same reserved-name exclusions, `ReadSchema` puts it in `DBSchema.RolesOutOfScope`, and `compare.Roles` takes existence from the union. Nothing else about the comparison moves: a role that exists is still compared attribute by attribute, so an `ALTER ROLE` the annotations ask for is still planned for a role outside the described scope.

## The partition is over managed roles, and every statement of it says so

`queryRoles` ends both statements with `NOT LIKE 'pg\_%' ESCAPE '\'` and `!= 'postgres'`, so the reserved roles and the bootstrap superuser are in **neither** list. A desired state naming one is still compared against nothing and still planned as a `CREATE ROLE` the server refuses — measured on PostgreSQL 17.10, `role "postgres"` gives SQLSTATE 42710 and `role "pg_monitor"` gives 42939, identically here and on the base.

That qualifier now appears everywhere the claim does: `DBSchema.RolesOutOfScope`'s doc comment (which ships publicly in `docs/public_api.snapshot`, regenerated), `readRolesOutOfScope`'s doc, the `# Described Is Not The Same Question As Present` block in `compare/roles.go`, and the PostgreSQL documentation page, which now tells the reader not to declare those roles.

**Decision on the reserved names, stated rather than implied: not in this commit.** `schemadiff.Compare` has no error to return and the refusal has to cover the generate path too, so it is a separate change rather than a hunk smuggled into this one. Until it lands, `compare.TestRolesReservedNameIsNotComparedAgainstAnything` pins the gap so it cannot be lost or silently claimed fixed. **A follow-up issue must be filed before merge** — see "Follow-up to file" below.

## Rebased onto master, and the conflict was not textual

Master's #1292 escaped the underscore in the reserved-prefix filter. That clause is shared by **both** of this branch's reads, and the two changes interact: an unescaped underscore is a single-character wildcard, so `NOT LIKE 'pg_%'` drops `pgbouncer`, `pgadmin` and `pgpool` — ordinary user roles — from both reads at once. A role in neither list is a role the comparator reads as absent, so it reaches the same 42710 through a second door.

The white-box fake now evaluates the LIKE pattern instead of string-matching one spelling of it, and refuses a third spelling the way it already refuses an unknown membership predicate. `pgbouncer` sits in the shared cluster fixture, and `TestReadRolesKeepsOrdinaryRolesTheReservedPrefixWouldSwallow` drives both reads over a fixture holding `pgbouncer` and `pg_reserved` together.

This was a real hole, not a stale constant: **with the fake as it stood before the rebase, reverting master's escape at the role query is green** (exit 0, measured). Master's live fixture `TestPostgreSQLReaderDescribesRolesNamedPgSomething` passes on this branch unchanged.

## The base I measured

```
$ git merge-base HEAD origin/master
9820496e1ba6bf89070b78f742dd39c531bb1c0a
$ git merge-base --is-ancestor origin/master HEAD ; echo $?
0
```

The base is **9820496e**, which is `origin/master`. An earlier version of this description said "master is 3acddbdf"; at that time `git merge-base --is-ancestor 3acddbdf <branch>` exited **1**, so that commit was not on the branch at all. After this rebase it exits 0 — because `origin/master` carries it — but it is still not the base.

## Measured

PostgreSQL 17.10 in a private container (`postgres:17`, own port, `ptah_user` created `LOGIN SUPERUSER` and `ptah_test` owned by it — the CI shape), removed at the end. Every exit status read on its own line, never after a pipe.

The whole PostgreSQL scenario suite, exactly as CI runs it (`--databases=postgres`, no `--scenarios`), against a freshly created cluster:

| | base 4b7923ab | PR head 4e8e7396 | here |
| --- | --- | --- | --- |
| 65 scenarios | 65 pass, exit 0 | **61 pass, 4 fail, exit 1** | 65 pass, exit 0 |

The PR head's four are exactly the four CI named — `dynamic_roles_advanced`, `dynamic_roles_cross_database`, `dynamic_roles_modification`, `dynamic_roles_rollback` — every one on `role "admin_user" already exists (SQLSTATE 42710)`. That string appears 8 times in the PR head's run and **0 times** in this one.

Narrowed to the five `dynamic_roles` scenarios: the PR head is 1 pass / 4 fail on a clean cluster and 0 pass / 5 fail on one that already holds the roles; here it is 5 pass, exit 0, either way.

**Reverting this fix on the current tree reproduces the clean-cluster row identically** (1 pass, 4 fail, same message), so the fix is what turns those four green rather than something else in the rebase.

### Compatibility surface

`ptah-compat schema apply` against a desired state naming a cluster role the inspected database does not use:

| | `--dry-run` | `--auto-approve` |
| --- | --- | --- |
| PR head | `CREATE ROLE "w1267_stranger" ...` | exit 1, 42710 |
| here | `Schema is synced` | exit 0 |

No capability is lost — the same role declared with `create_db = true` plans `ALTER ROLE "w1267_stranger" CREATEDB;` here, while the PR head still plans the CREATE.

**The description does not move**: `ptah-compat schema inspect` of the fixture is byte-identical to the PR head's — 1175 bytes, `diff` exit 0, two role blocks each. This change is visible only to the comparator. The pinned Atlas community binary v1.3.0 reads that document at exit 0 against a separate, freshly created dev database, and emits no role block of its own, so it is a readability reference here and not an oracle for which roles are right.

## Mutants

Each applied and restored alone; `git status --porcelain` empty and `grep -rn MUTANT` clean after every one. Control row is the unmutated tree in the same sweep.

| mutant | reader unit | comparator unit | snapshot | scenarios | live guard | atlasfilter |
| --- | --- | --- | --- | --- | --- | --- |
| *(control)* | 0 | 0 | 0 | 0 (65/65) | 0 | 0 |
| **MA** `RolesOutOfScope` loop removed from `compare.Roles` | 0 | **1** (4 subtests) | – | **1** (1/4, 42710) | **1** | – |
| **MF** described-wins guard removed | – | **1** (only that subtest) | – | – | 0 | – |
| **MB** `u.roleoid = u.roleoid` | **1** (5 tests) | – | – | – | – | – |
| **MC** `TRUE OR NOT EXISTS (…)` | **1** (5 tests) | – | – | – | – | – |
| **MD** escape reverted at the role query | **1** (3 tests) | – | – | – | – | – |
| **MD** + the pre-rebase fake | **0** — the hole this closes | – | – | – | – | – |
| **ME** `NOT LIKE 'pg%'` (over-corrected) | **1** (fake refuses the pattern) | – | – | – | – | – |
| **MG** both reserved exclusions dropped | **1** (`pg_reserved` leaks in) | – | – | – | – | – |
| **MH** `json:"-"` → `json:"roles_out_of_scope"` | – | – | **1** | – | – | – |
| **MI** `RolesOutOfScope` dropped from `cloneDatabase` | – | 0 | – | – | – | **1** |

MB and MC are the two an earlier review reported walking past **every** test; they redden five tests each now. MD, ME and MG stay distinguishable rather than collapsing into one failure: a missing filter leaks the reserved role, a known-but-wrong spelling drops the ordinary one, and an unknown spelling is refused outright.

MF is caught by the sub-test "a described role wins over the same name out of scope" and by nothing else — that sub-test previously used two different names and asserted nothing about precedence.

## Coverage, said plainly

With `POSTGRES_TEST_DSN` unset the live guard **skips and the run still reports `ok`, exit 0**. On any job that does not set it, `readRolesOutOfScope`'s SQL has no live coverage at all. The white-box guard runs on every unit run and is what kills MB, MC, MD, ME and MG; it cannot tell whether PostgreSQL agrees with what the predicate means. This is stated in the test file too, not only here.

## Corrections from review

1. **Unqualified partition** — fixed at all four sites plus the regenerated `docs/public_api.snapshot`.
2. **Reserved names** — decision stated (not in this commit); follow-up below.
3. **`TestReadRolesPartitions…` could not fail** — the fake no longer computes the complement itself, and `TestReadRolesComplementIsTheExactNegationOfTheScopedRead` asserts on the text the reader sends. MB and MC verified red.
4. **Unit-only coverage** — stated in the test file and above.
5. **"a described role wins…"** — both lists now carry the same name, and MF verifies the assertion bites.
6. **Measurement table** — base corrected to a measured `git merge-base`; the replay row reworded to "fewer `CREATE ROLE` statements", not "replay fixed" (with one GRANT in the schema this branch still emits 2 `CREATE ROLE` and `psql -v ON_ERROR_STOP=1` still exits 3 — base emits 6 for the same fixture); the JSON-format claim dropped in favor of what that test actually protects.

## Follow-up to file

**Refuse a desired schema that declares a reserved role, instead of planning a statement the server is guaranteed to reject.** `role "postgres"` plans `CREATE ROLE "postgres"` and `--auto-approve` exits 1 at SQLSTATE 42710; `role "pg_monitor"` exits 1 at 42939. Identical before and after this branch, so not a regression — but per AGENTS.md ("refuse loudly rather than accept and ignore") the declaration should be refused up front, on the compare and generate paths alike.

Record only, not blocking: `migrate diff` against a dev database in a cluster that already holds the role emits no `CREATE ROLE` into the generated migration file, so migration-directory content depends on unrelated cluster state. Identical on master.

## Not measured

YugabyteDB (`RoleManagement` is true in its preset, no container here); the ClickHouse, SQL Server, MariaDB, CockroachDB and OCI legs of the CI matrix; behavior on a non-superuser connection, where the pre-existing `LEFT JOIN pg_authid` in both queries is the binding constraint; the cost of the extra catalog query; and the docs `check:responsive` gate, which needs a full `npm ci` + Playwright install.


denisvmedia (commented):
Independent review of head `d086aaa8` confirms that this does not close #1267.

**Blocker — role-bearing inspect output still cannot replay into another database in the same PostgreSQL cluster.** `internal/dbschema/postgres/reader.go:2317-2318` emits every role referenced by another emitted statement. Those roles are cluster-wide and already exist when the output is replayed into a clean sibling database, so `CREATE ROLE` fails. The current PR body confirms a one-GRANT fixture still emits two role blocks and `psql -v ON_ERROR_STOP=1` exits 3, while #1267 requires replay exit 0.

Either make default compatibility output replayable while preserving fuller Ptah behavior on the same `ptah-compat` surface through an explicit `PTAH_*` control, or present this as a partial disclosure improvement and remove `Closes #1267`. Add a live inspect-then-replay integration test. Also file the reserved-role follow-up that the PR body states is mandatory before merge; no such open issue currently exists.

denisvmedia:
Reproduced, and fixed at the dev materialization. It reproduced on the current head `efc0aa1a`, not only on the `d086aaa8` you read, so the later commit had not already closed it.

## 1. "Role-bearing inspect output still cannot replay into another database in the same cluster" — confirmed, and closed

PostgreSQL 17.10 in a private container (own port, `ptah_user` created `LOGIN SUPERUSER` and owning the databases, removed at the end). Fixture: one table, one `GRANT SELECT` to a role, plus one cluster role nothing grants anything to. `rs1305_dev` is created seconds before the replay.

At `efc0aa1a`:

```
$ ptah-compat schema inspect -u 'postgres://.../rs1305_src?...' > inspect.hcl ; echo $?
0                                          # two role blocks, as the PR body says
$ ptah-compat schema inspect -u file://inspect.hcl --dev-url 'postgres://.../rs1305_dev?...' ; echo $?
Error: materialize schema on dev database: failed to execute SQL statement: SQL execution failed:
       ERROR: role "ptah_user" already exists (SQLSTATE 42710)
SQL: CREATE ROLE "ptah_user" WITH LOGIN SUPERUSER NOCREATEDB NOCREATEROLE INHERIT NOREPLICATION;
1
```

That is #1267's own transcript. Here: **exit 0**, and the re-rendered document is byte-identical to the input — `diff` exit 0, 1265 bytes — so the round trip is a fixed point rather than merely non-fatal.

Where the fix went, and why not where you suggested. The blocking `CREATE ROLE` is not in the description, it is in the dev **materialization**: the dev database is reset before the document is executed on it, and a role is not in a database. So `rolescope.RolesToCreateOnDev` splits the declared roles against what the dev connection reports — `Roles` together with `RolesOutOfScope` — and `materializeOnDev` creates only the complement. Skipping rather than refusing, because the pinned Atlas community binary v1.3.0 reads the same document at exit 0 against its own separate, freshly created dev database (measured), so refusing would fail a document this surface exists to accept. Skipping rather than reconciling, because a dev database must never `ALTER` a role on the operator's server.

I did **not** take the other reading of your first option — dropping the `role` blocks from the default compat output and putting them behind the `PTAH_*` control — and I measured it before rejecting it rather than asserting it. That shape does replay into a sibling database in the same cluster: with the two `role` blocks deleted from the document by hand, the materialization exits 0 and the re-render even comes back byte-identical, because the grants apply against roles the server already has and the read then describes them again. What it costs is the other direction. A `permission` names its grantee as `to = role.<name>` and Ptah resolves that by name whether or not a block declares it, so on a server that does **not** have the role the document dies one statement later:

```
$ ptah-compat schema inspect -u file://norole_absent.hcl --dev-url 'postgres://.../rs1305_dev?...' ; echo $?
Error: materialize schema on dev database: failed to execute SQL statement: SQL execution failed:
       ERROR: role "rs1305_absent" does not exist (SQLSTATE 42704)
1
```

So dropping the blocks trades the cluster-portable direction for the same-cluster one, and describes less than the database contains while doing it. Skipping the `CREATE ROLE` at materialization time keeps both: 42710 on the sibling database is gone and the role a server lacks is still created there. The `PTAH_POSTGRES_INSPECT_ALL_ROLES` control that `efc0aa1a` added already covers the capability half of your option; the replay half is what this commit adds.

What did not move: `ptah-compat schema inspect` of the same database is byte-identical before and after this commit (`diff` exit 0). The comparator rows the PR exists for still hold — an existing out-of-scope role plans nothing (`Schema is synced, no changes to be made`), the same role with `create_db` still plans `ALTER ROLE "rs1305_stranger" CREATEDB;`, and `--auto-approve` exits 0. The whole PostgreSQL scenario suite as CI runs it is 65/65, exit 0, with `42710` and `already exists` appearing zero times.

## 2. "Add a live inspect-then-replay integration test" — added, and it is in a suite CI runs

`integration.TestAtlasCompatInspectRoleReplayE2E` inspects a live database whose schema grants to a cluster role, feeds that document back in against a freshly created **sibling** database on the same server, and asserts the note, the surviving `role` and `permission` blocks, and equality with the input document. `TestAtlasCompatInspectCreatesARoleTheServerLacksE2E` is the control that keeps the fix from becoming "a dev database never creates a role": it asserts through `pg_roles`, not through the rendering, that a role the server lacks is created. Both run under `go test ./integration -tags=integration`, which the integration-tests workflow runs with `POSTGRES_TEST_DSN` set.

Said plainly, because it matters: both skip without `POSTGRES_TEST_DSN` and the run still reports `ok`. Reverting the wiring alone (mutant MQ below) leaves the `internal/rolescope` unit tests, the reader white-box guard and the comparator all green — measured, not assumed — and reddens only the new live test.

## 3. "File the reserved-role follow-up" — filed as #1312

With its own measurement: `role "postgres"` plans `CREATE ROLE "postgres"` and `--auto-approve` exits 1 at `role "postgres" already exists (SQLSTATE 42710)`; `role "pg_monitor"` exits 1 at `role name "pg_monitor" is reserved (SQLSTATE 42939)`. Identical on master, so not a regression this branch introduced. The PR body's promise is now a link rather than a promise.

## 4. Nothing disputed, but one claim narrowed

Your citation of the PR body's `psql -v ON_ERROR_STOP=1` row is accurate and that row is **unchanged**: piping `ptah db read` output into `psql` by hand still exits 3, because PostgreSQL has no `CREATE ROLE IF NOT EXISTS` and Ptah does not control `psql`. #1267's definition of done names the other property — "Ptah's own inspect output replays into a clean dev database at exit 0" — and that is what the error transcript in the issue is from. The PR body now states both rows separately instead of one sentence that could be read as either, and an idempotent rendering of `CREATE ROLE` is recorded as a follow-up with its measurement rather than smuggled in here.

## Mutants for the new fixture, both directions

Each applied and restored alone; `grep -rl MUTANT` counted 1 during every mutant row and 0 after every restore, and every swept file was verified identical to its pristine copy afterwards. "PR originals" is the reader white-box guard (`./internal/dbschema/postgres`) and the comparator (`./migration/schemadiff/...`).

| mutant | rolescope unit | new replay e2e | new control e2e | PR originals |
| *(control)* | 0 | 0 | 0 | 0 |
| **MQ** the partition is not wired into `materializeOnDev` | **0** — the coverage gap, stated | **1** (42710) | 0 | 0 |
| **MR** `RolesToCreateOnDev` never reports an existing role | **1** | **1** | 0 | 0 |
| **MS** over-corrected: every declared role treated as already there | **1** | 0 | **1** | 0 |
| **MT** the note removed | **1** | **1** | 0 | 0 |
| **MU** over-corrected at the description: `readRoles` describes nothing | 0 | **1** | **1** | **1** |

MQ is the row you would want to see: reverting the fix reddens the new fixture and nothing else. MU is the alternative fix I rejected, and it reddens the PR's own original fixtures — which is the evidence that those fixtures still bite.

## Prose the fix made false, rewritten

- `docs/POSTGRESQL_ROLES.md` said "Ptah deliberately fails when a role already exists" with no qualification. The apply path still does; the dev materialization now does not. Both are stated, and the reason the dev path is different (a reset clears a database, not a server) is stated with it.
- `docs/site/.../direct/inspect.md` said role creation statements "are intended for a clean target. If a role already exists, applying the SQL fails" — true of running the SQL by hand, which is how it is now phrased, and false of feeding the description back through Ptah, which is now described.
- `docs/site/.../atlas/schema-commands.md` described the dev-database materialization as executing the source; it now names the one kind of object that is not materialized, and why.
- `docs/site/.../databases/postgresql.md` gains the replay property beside the scoping cost it already recorded.
- The `Roles, grants, and row-level security` feature-matrix row carries the replay measurement, and `build-feature-matrix.mjs --check` is clean.

One process note: `go test ./...` came back exit 1 on `cmd/viz` with `render SVG with Graphviz dot: signal: killed` — the Graphviz child killed under machine load, nothing to do with this change. Re-run serially it is `ok`, exit 0. Reporting it rather than calling the aggregate green.

Issue #1242, opened 2026-08-07
PostgreSQL introspection drops index access method, opclass, expression and sort order, and flattens domain columns


denisvmedia:
## Summary

Ptah's PostgreSQL introspection drops four index attributes and the type of a domain-typed column. All five were reported by a live parity sweep; **all five reproduce**, but the sweep's framing of the consequence is wrong in one important way and is corrected below.

The four index attributes are read out of the database correctly and then discarded on the way to the model. The renderer already knows how to emit every one of them — `USING`, an operator class, a parenthesized expression and `DESC` all have working render paths — so this is a **read-side** defect, not a render-side one.

## How this was measured

- Oracle: the pinned Atlas **community** binary, `atlas community version v1.3.0`, at `/Users/buster/Work/denis/ptah-atlas-conformance/bin/atlas`. Black box; behavior determined by running it.
- Ptah: `bin/ptah-compat` built from branch `issue-1232-implicit-schema-constraints` at `1694e06a` (not `master`, because this front touches `schema diff`; #1238 changed diff table keying).
- Server: live PostgreSQL **17.10** in container `ptah1074-dev`, on databases created for this investigation only.
- Every probe ran **serially**, each binary against its own database. Every exit status was captured on its own line immediately after the command and never read through a pipe.

## The five items, one object each

Each item lives in its own database with exactly one table, so a failing replay names exactly one item. `control` is a plain btree index included to show the finding is specific rather than "index reading is broken".

| # | item | ptah preserves it? | what ptah emits instead |
|---|---|---|---|
| 1 | access method `USING gin` / `gist` / `brin` / `hash` | **no** | no `USING` clause at all — every index collapses to the btree default |
| 2 | operator class `text_pattern_ops` | **no** | the bare column, opclass dropped |
| 3 | expression index `lower(name)` | **no** | `"lower(name)"` — the expression **quoted as a column identifier** |
| 4 | sort order `DESC` / `NULLS FIRST` / `NULLS LAST` | **no** | the bare column, ordering dropped |
| 5 | domain column type | **partly** — see below | column flattened to the domain's base type `integer` |
| — | control: plain btree index | yes | identical to the pinned binary |

## Replay: psql's real exit status

Each binary's `schema diff` output was replayed into a **fresh** database with `psql -v ON_ERROR_STOP=1`.

| item | pinned binary replay | ptah-compat replay |
|---|---|---|
| gin | 0 | 0 |
| gist | 0 | 0 |
[...]

PR #1306, opened 2026-08-08
Create a user-defined type before the type that names it (#1242)

denisvmedia:
## What this changes

`internal/planner/dialects/postgres` and `internal/convert/fromschema` emit
PostgreSQL's three kinds of user-defined type -- domains, composite types and
range types -- ordered by what each definition names, instead of kind by kind.
The non-`CASCADE` drops a modification produces run in the reverse of that
order. Enums are unaffected and stay ahead of the group.

## Why

The branch's own `docs/conformance.md` section claims a round trip for a
database holding a domain over an enum, a domain over a composite and a domain
over a range: Ptah describes it, psql replays the description at exit 0, and the
pinned Atlas CE v1.3.0 binary then reports the replay in sync with the source.
The claim was true for the enum shape and false for the other two, because the
emitter put every domain ahead of every composite and range and PostgreSQL has
no forward declaration for a type.

Measured on PostgreSQL 17.10, one database holding `CREATE TYPE addr AS (...)`,
`CREATE TYPE myrange AS RANGE (...)`, `CREATE DOMAIN d_comp AS addr` and
`CREATE DOMAIN d_range AS myrange`:

| Probe | before | now |
| --- | --- | --- |
| `schema diff --from <empty> --to X`, statement 2 | `CREATE DOMAIN "d_comp" AS addr;`, four statements ahead of `CREATE TYPE "addr"` | `CREATE TYPE "addr" AS (...)` precedes it |
| that script under `psql -v ON_ERROR_STOP=1` | `ERROR: type "addr" does not exist`, exit 3 | exit 0 |
| CE compares the replay to the source | not reached | Schemas are synced |

## Why not simply swap the kinds

Emitting composites and ranges first is the same defect pointed the other way. A
composite field and a range subtype can each be a domain, and both are legal
PostgreSQL (verified on the server). Built as a mutant and measured: the swap
emitted `CREATE TYPE "posrange" AS RANGE (SUBTYPE = d_int);` first and psql
stopped at `ERROR: type "d_int" does not exist`, exit 3. No fixed order of kinds
serves both directions, so the ordering runs through the stable topological sort
already in `internal/deporder`. A reference the plan does not create adds no
edge, and a cycle -- which PostgreSQL refuses to create anyway -- degrades to
declaration order rather than dropping a type from the plan.

Two emitters carried the defect and both are fixed: the migration planner (which
also folds in the recreations a modification produces, because a new domain over
a recreated composite has to wait for the recreation) and
`fromschema.FromDatabase`, the whole-schema render behind `ptah generate`.

## Tests

- `internal/deporder` covers each pair of kinds in both directions plus the
  spellings a reference arrives in: `addr[]`, `short_text(20)`, `"app"."Addr"`,
  a bare name two schemas both offer (deliberately unresolved), a cycle, and a
  self-reference.
- The planner and `fromschema` each assert the emitted order on one schema
  holding a domain over a composite, a domain over a range, a composite over a
  domain and a range over a domain, so a fix in either direction alone reddens
  the file.
- `integration/gonative` EXECUTES the emitted script against an empty database
  and asserts it runs, then asserts the replayed columns still carry their
  domains. A text assertion on the order would only restate the emitter. The
  self-apply guard next door cannot see this: a plan with no statements executes
  nothing, so any order is vacuously fine there.

Mutants, each applied to the working tree, built, measured and restored:
reverting the planner to kind order reddens all three planner rows and the live
guard with `type "addr" does not exist` in the output; the static swap reddens
the composite-over-domain and range-over-domain rows only; reverting the drop
order reddens the recreate row alone; reverting the whole-schema emitter reddens
the `fromschema` rows and nothing else.

## Incidental, and recorded in `docs/conformance.md`

`lo` (contrib `lo`) and `earth` (contrib `earthdistance`) are both domains, over
`oid` and over `cube`. Answering a column's type from `udt_name` keeps the first
and flattens the second to `sql("cube")`, decided by nothing but whether the base
type happens to be built-in. With the `domain_name` gate this branch adds, both
print what the pinned CE v1.3.0 binary prints -- `sql("lo")` and `sql("earth")`
-- with a plain `cube` column beside them still `sql("cube")`. Removing the gate
again reproduces the split, which is the control the table records.

## Not verified

PostgreSQL 18 (17.x only); MySQL, MariaDB, SQLite, ClickHouse, MSSQL, Spanner
and CockroachDB, none of which model these three kinds; a composite whose field
type is a range, or a range whose subtype is a composite; a base type in another
schema, live.

---

## The review's regression is fixed, and the fix is pinned by the failure it caused

An adversarial review refuted an earlier head. It was right, and the finding was precise: `dropModifiedUserTypes` built its `deporder.UserType` references from `generated` — the **desired** schema — and reversed that order. The CREATE order rightly follows the desired definitions; the DROP order cannot, because those statements execute against the database **as it stands**, so only the current definitions can block them. The two disagree exactly when the modification is what moved the reference.

Its fixture, reproduced verbatim:

```sql
-- current:  CREATE TYPE cc AS (f integer);   CREATE DOMAIN dd AS cc;
-- desired:  CREATE DOMAIN dd AS integer;     CREATE TYPE cc AS (f dd);
```

| binary | emitted drop order | apply exit |
| --- | --- | --- |
| before | `DROP DOMAIN "dd"` then `DROP TYPE "cc"` | 0 |
| the refuted head | `DROP TYPE "cc"` then `DROP DOMAIN "dd"` | **1**, `cannot drop type cc because other objects depend on it … type dd depends on type cc (SQLSTATE 2BP01)` |

The review also noted why the branch's own mutant could not see it: the ordering test fed a single `generated` in which the desired references were also the current ones, so no fixture separated the two sides.

Both are closed. The drop order is now derived from the current shape through `deporder.UserTypesForDrop`, and the create order keeps its own graph — two graphs, pinned separately. The proof is the review's own failure, turned into a mutant:

| mutation | result |
| --- | --- |
| `UserTypesForDrop(deps)` → `UserTypesForCreate(deps)` | **RED** on 4 planner rows; **RED** on all 5 live integration rows with `cannot drop type cc/qty/d_base/c_base because other objects depend on it (SQLSTATE 2BP01)`, 5 occurrences |
| restored | GREEN, unit and live |
| topological sort replaced by caller order | **RED** on exactly the 3 create rows, every drop row green |
| `postgresSerialType` domain refusal disabled | **RED** on the domain-column-with-a-sequence row only |

The four prose sentences the review called false are gone — `grep -n "reverse of that order|reverse of the creation|never blocked"` across `docs/site/src/content/docs/databases/postgresql.md`, `docs/conformance.md` and `docs/user_defined_types.md` returns nothing.

## Rebase notes, because one hazard here is silent

This branch was rebased through three successive masters. Two things are worth recording.

`546c7247` (#1302) landed a **more complete implementation of the same domain-column work** this branch carried: `DBColumn.DomainName` *plus* `DomainSchema`, and a full `(schema, name)` identity comparator rather than this branch's name-only one. A naive rebase auto-merged a **duplicate `DomainName` field** into `DBColumn` — reported as no conflict at all, and it does not compile. Everywhere the two overlapped, master's implementation was kept and this branch's was dropped: the field, `rawDBColumnType`'s FormattedType preference, `columnTypeChange` / `domainColumnTypeChange`, `goSchemaFieldType`'s domain-over-base branch, and the reader's fake-server helpers. The now-unused `serveDomainName` / `projectionBody` helpers were deleted so `unused` would not fire.

Everything PR-unique is intact: the whole ordering story (`internal/deporder`, planner create and drop ordering, `fromschema`, `DomainDiff.CurrentBaseType`, `CompositeTypeDiff.CurrentFieldTypes`, generator reverse-diff re-derivation), the rule that a domain column's owned sequence is not implicit, `postgresSerialType`'s domain refusal, the Atlas-compatible JSON inspect surface, and the live integration guards.

Both conflicts had the shared-closing-brace shape that has cut a function mid-body in this repo before (`sqliteKeyColumnImpliesNotNull` and `domainColumnTypeChange` shared one `}`). Each resolved file was diffed against **both** parents and came out as exactly master plus exactly this branch's delta, with no third content; `gofmt -l` is clean; a tree-wide grep for conflict markers is empty; and the first rebase's `--stat` was byte-identical to the pre-rebase `--stat`.

## Gates — literal exit codes

`go build ./...` 0 · `go vet ./...` 0 · `go vet -tags integration ./integration/...` 0 · `go tool qtlint ./...` 0 · `golangci-lint run ./...` 0 (v2.12.2, CI's pin) · `scripts/check-test-style.sh` 0 · `scripts/check-public-api.sh` 0 · `scripts/check-public-api-snapshot.sh` 0 · `scripts/check-public-api-released.sh` 0 · every `docs/site` `check:*` and `*:selftest` 0 · `npm run build` 0 · `build-feature-matrix.mjs --check` 0 · the live `integration/gonative` domain and user-type-order guards 0.

`docs/public_api.snapshot` was broken deliberately and watched go to **1**, then restored and watched go to **0** — it is not a gate that reports without running.

## Still open

- `docs/conformance.md` now carries two adjacent sections on domain comparison, this one and #1138's. They do not contradict; consolidating them is a follow-up.
- A full `go test ./...` on the final tree did not finish locally: the machine was at load ~30 with roughly 75 foreign test binaries from parallel work. An earlier full run on the first rebase was 0 across the board. CI's `unit-tests` job is the authoritative answer for this head.

### DIFF
--- a/dbschema/types/types.go
+++ b/dbschema/types/types.go
@@ -94,6 +94,16 @@ func QualifyTableName(schema, table string) string {
 // GeneratedKind / GeneratedExpression are populated by readers for dialects
 // that expose generated column metadata. Schema comparison matches these fields
 // when the goschema-side model also carries generated column metadata.
+//
+// DomainName and FormattedType are a pair. DataType reports a domain column's
+// BASE type, so it alone cannot say that the declared type was a domain:
+// DomainName carries that fact and FormattedType carries the spelling the
+// server uses for it, schema-qualified where the search path needs that.
+// Everything that answers "what type is this column declared as" -- the
+// desired-state conversion, the Atlas-compatible JSON inspect output, and the
+// comparator's database side -- must consult DomainName rather than infer the
+// answer from DataType, which reports the base type and drops the domain's
+// constraints with it. See stokaro/ptah#1242.
 type DBColumn struct {
 	Name               string  `json:"name"`
 	DataType           string  `json:"data_type"`
--- a/docs/conformance.md
+++ b/docs/conformance.md
@@ -372,6 +372,45 @@ folded into the shorthand, and the sequence itself is reported rather than
 treated as the column's implicit backing sequence — without it the emitted DDL
 names a sequence nothing creates, which is measured as psql exit 3.
 
+### A domain over a user-defined base type is a different catalog shape
+
+`CREATE DOMAIN positive AS integer` and `CREATE DOMAIN d_enum AS color` do not
+read back the same way, and the difference decides which code answers the
+question "what type is this column declared as". Measured on PostgreSQL 17.10:
+
+| Column | `data_type` | `udt_name` | `domain_name` | `format_type` |
+| --- | --- | --- | --- | --- |
+| `qty positive`, `positive AS integer` | `integer` | `int4` | `positive` | `positive` |
+| `c d_enum`, `d_enum AS color` | `USER-DEFINED` | `color` | `d_enum` | `d_enum` |
+| `plain color`, no domain | `USER-DEFINED` | `color` | *(null)* | *(not read)* |
+
+For a domain over a built-in base type `data_type` is the base type, so a
+consumer that falls through to `format_type` reaches the domain by accident. For
+a domain over a user-defined base type — an enum, a composite or a range —
+`data_type` is the bare category `USER-DEFINED` and `udt_name` names the BASE
+type, identically to the plain column on the last row. A consumer that answers
+from `udt_name` there flattens `c` to `color` and drops the domain's `CHECK`
+with it, and only `domain_name` separates row two from row three.
+
+Measured with `ptah-compat` against one database holding rows two and three, and
+against the composite and range shapes beside them:
+
+| Probe | Atlas CE v1.3.0 | Ptah |
+| --- | --- | --- |
+| `schema diff --from X --to X`, one database against itself | Schemas are synced | Schemas are synced |
+| `schema apply` of a byte-identical twin | — | Schema is synced, no changes to be made |
+| `schema inspect --format json` | `"type":"d_enum"` | `"type":"d_enum"` |
+| `schema inspect`, HCL | `type = sql("d_enum")` | `type = sql("d_enum")` |
+| `schema inspect`, HCL, plain enum column beside it | `type = enum.color` | `type = enum.color` |
+| `schema diff` from empty, replayed with psql, then compared to the source by CE | — | psql exit 0, Schemas are synced |
+
+The last row is the round trip, and CE is the neutral observer in it: Ptah emits
+`CREATE TYPE`, `CREATE DOMAIN` and a `d_enum` column, the replay runs at psql
+exit 0, and CE then reports the replayed database in sync with the one it was
+described from. A description that spells the column `color` replays at exit 0
+too and CE reports `ALTER TABLE "t" ALTER COLUMN "c" TYPE d_enum;` — the replay
+lost the domain and the exit code did not say so.
+
 ### Reading an attribute and reconciling it are different claims
 
 The table above records what introspection preserves. It does not say what
--- a/docs/conformance.md
+++ b/docs/conformance.md
@@ -323,7 +323,54 @@ own `schema diff` output fails to replay with
 `ERROR: type "positive_int" does not exist` — measured, psql exit 3. Ptah emits
 both the `CREATE DOMAIN` and the domain-typed column, so closing this gap meant
 keeping both halves rather than matching CE. Ptah's output replays at psql exit
-0 where CE's does not.
+0 where CE's does not. The same divergence reaches `schema apply`: applying this
+source database to an empty target, CE exits 1 with
+`create "t" table: pq: type "positive" does not exist` and Ptah exits 0 with the
+domain created first.
+
+### A domain column has to survive the comparator too, and the JSON surface
+
+Rendering a domain and reconciling one are separate claims, the same way they
+are for indexes below. Reading the domain fixed what `schema inspect` writes as
+HCL and left two other surfaces on the base type. Measured on PostgreSQL 17.10
+against one database holding `CREATE DOMAIN positive AS integer CHECK (VALUE >
+0)` and a column of it:
+
+| Surface | Atlas CE v1.3.0 | Ptah before | Ptah now |
+| --- | --- | --- | --- |
+| `schema inspect`, HCL | `type = sql("positive")` | `type = sql("positive")` | `type = sql("positive")` |
+| `schema inspect --format json` | `"type":"positive"` | `"type":"integer"` | `"type":"positive"` |
+| `schema inspect --format json`, domain off the search path | `"type":"doms.positive"` | `"type":"integer"` | `"type":"doms.positive"` |
+| `schema diff --from X --to X`, one database against itself | Schemas are synced | `ALTER TABLE "t" ALTER COLUMN "qty" TYPE positive;` | Schemas are synced |
+| `schema apply` run twice against the same target | — | plans and executes the same `ALTER` on every run, exit 0 each time | second run reports the schema synced |
+
+The diff row is the one that made the rendered domain worth nothing: the desired
+side answered `positive` and the database side answered `integer`, so a database
+was never in sync with itself and `schema apply` executed an `ALTER COLUMN` on
+every run while reporting success.
+
+Two details decided whether any of this was visible:
+
+- **The name of the domain.** The type comparison folded a spelling into a
+  category by substring, so any domain whose name contains `int` — the issue's
+  own `positive_int` fixture — compared equal to `integer` by accident and the
+  churn did not appear. A domain is now compared as the identifier it is: two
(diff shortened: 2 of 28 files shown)


denisvmedia (commented):
Independent review of head `13818791` confirms the PR body's drop-order refutation.

**Blocker — recreated types are dropped using the desired dependency graph.** `dropModifiedUserTypes` in `internal/planner/dialects/postgres/postgres.go:1144-1158` derives domain/composite edges only from `generated`, but the DROP statements execute against the current schema. If current has `dd -> cc` and desired reverses it to `cc -> dd`, the plan can emit `DROP TYPE cc` before `DROP DOMAIN dd`, which PostgreSQL rejects because the current domain still depends on the type.

Carry current definitions into this phase: order drops from the current graph and creates from the desired graph. Add a live or seeded current-to-desired regression whose dependency direction reverses. Green static and targeted checks do not cover this inversion, so the PR is not mergeable yet.

denisvmedia:
Thanks — the finding was right, and it is fixed. Answering point by point, with what I measured on my own container and fixtures rather than re-running yours.

**The blocker: recreated types dropped using the desired dependency graph.** Confirmed on the head you reviewed. My own container (postgres:17, port 16406, `ptah_user LOGIN SUPERUSER`, removed afterwards), your fixture J shape, `pr1306_work` reseeded from the current fixture before every row, exit codes read on their own line:

| binary | drop order emitted | `schema apply --auto-approve` |
| --- | --- | --- |
| `243991e3` (pre-ordering parent) | `DROP DOMAIN dd`, `DROP TYPE cc` | exit 0 |
| `13818791` (the head you reviewed) | `DROP TYPE cc`, `DROP DOMAIN dd` | **exit 1**, `ERROR: cannot drop type cc because other objects depend on it (SQLSTATE 2BP01)` on `DROP TYPE IF EXISTS "cc"` |
| `3e66060e` (branch head) | `DROP DOMAIN dd`, `DROP TYPE cc` | exit 0 |

**The branch moved after your review.** `3e66060e` — "Order a modified type's drop by the shape it is dropping" — landed on top of `13818791` and does exactly what your REQUIRED CHANGE item 1 asked, structurally rather than through the item 2 stopgap. At the lines you cited, `dropModifiedUserTypes` no longer takes a `*goschema.Database` and no longer reads `generated`; it reads `domainDiff.CurrentBaseType` and `compositeDiff.CurrentFieldTypes`, new fields on `DomainDiff`/`CompositeTypeDiff` that the comparator fills from the catalog rows it already reads, in the catalog's own spelling. `Changes` is never parsed back — there is a test that says so.

**Your finding 2, the mundane shape.** Also reproduced and also fixed: current `DOMAIN qty AS integer CHECK (VALUE > 0)` with `TYPE meas AS (q qty, label text)`, desired `qty` widened and `meas` holding a plain bigint — `13818791` exit 1 with `cannot drop type qty because other objects depend on it`; head exit 0, catalog `qty|d|bigint` and `meas.q bigint`, second plan empty.

**Item 4, the prose.** All four sentences were rewritten in the same commit. I checked rather than took it on trust: a repo-wide grep for "reverse of the creation", "reverse of that order", "reverse of the order" and "never blocked" across every `.md` and `.go` returns nothing outside an unrelated migrator test comment. The pages now say the drop graph is the current shape and the create graph is the desired one, and `deporder.UserTypesForDrop` tells its caller the same thing.

**Item 5, the live coverage.** `integration/gonative/user_type_recreate_order_postgres_test.go` seeds the current shape, reconciles it against a second database holding the desired shape, asserts the apply runs, asserts the catalog converged and asserts a second plan is empty. Behind `//go:build integration` in `./integration/gonative`, which `.github/workflows/go-integration-tests.yml:231` runs. I confirmed it actually runs rather than skips: three subtests, all named, all PASS.

**The one gap I did find, and closed in `0448700f`.** Every drop-order fixture on the branch — yours and theirs — pairs a domain with a composite. A rule as crude as "composites before domains" passes all of them, so they cannot distinguish it from the rule the code implements. A dependent pair of the *same* kind can: the comparator hands the planner each modified list sorted by name, so caller order is whatever the names are, and I named these so it is the failing one. Two live rows and one planner row:

    current: CREATE DOMAIN d_base AS integer; CREATE DOMAIN d_over AS d_base;   desired: both plain text
    current: CREATE TYPE c_base AS (f integer); CREATE TYPE c_over AS (g c_base); desired: both plain text

Both apply at exit 0 on the head. Ordering the drops by the desired graph reddens them with `cannot drop type d_base ...` / `cannot drop type c_base ...` (SQLSTATE 2BP01), and so does removing the drop edges entirely — the two mutations are opposites and both go red, because neither the desired graph nor caller order can order an edge that lives inside one kind. Reverting the create ordering reddens the three create rows and leaves every drop row green.

**Not disputed, nothing invented.** I reproduced your blocker before concluding anything, and I did not touch the drop ordering, because the change you asked for is already there and measures correct. Gates on `0448700f`: build, vet, vet -tags integration, `go test ./...`, qtlint, golangci-lint, all four `scripts/check-*`, and every `check:*` in `docs/site/package.json` — all 0. One `cmd/viz` failure in a full run was a `signal: killed` under machine load, green when that package is re-run serially, and unrelated to this branch.

PR #1307, opened 2026-08-08
Stop demanding an atlas.hcl for --var and a sum for nothing (#1241)

denisvmedia:
Closes two of the fifteen cells in #1241, records two more as deliberate divergences, and re-measures the rest.

## The re-measured table, before any code

All fifteen re-run against the pinned Atlas community binary **v1.3.0**, on SQLite and on PostgreSQL 17, each cell in its own directory, every exit status read on a line of its own after a redirect and never through a pipe. `ptah-compat` built from `origin/master` at `9820496e`.

| # | Cell | pinned v1.3.0 | ptah-compat before | Now |
| 1 | fix-and-rerun after a failed migration | 0 | 1 `migration … is dirty: state=failed applied=0/1` | **Reproduces.** Out of scope by the decision already recorded on the issue. |
| 2 | `migrate apply` `--dir` default | 0 | 0 | **Closed by `e23c14ea` (#1268)**, re-verified. |
| 3 | `migrate new` `--dir` default | 0 | 0 | **Closed by `e23c14ea` (#1268)**, re-verified. |
| 4 | `migrate new --dir` parent creation | 0 | 0 | **Closed by `e23c14ea` (#1268)**, re-verified (`--dir file://a/b/c`). |
| 5 | apply blocked after an out-of-order insert | 0 (`--exec-order non-linear`) | 1 `migration …_three checksum mismatch` | **Reproduces, root-caused, open.** See below. |
| 6 | edited already-applied file | 0 `No migration files to execute` | 1 checksum mismatch | **Reproduces. Deliberate divergence**, documented with its recovery route. |
| 7 | `migrate validate` on an empty directory | 0, zero bytes | 1 `checksum file not found` | **Closed here.** |
| 8 | `migrate import --dir-format liquibase` | 0 | 1 `no importable migration files found` | **Reproduces, open.** |
| 9 | `--format` `.Files[].Reports` | 0 | 1 `can't evaluate field Reports` | **Reproduces, open.** |
| 10 | `--format` top-level `.Schema` | 0 | 1 `can't evaluate field Schema` | **Reproduces, open.** |
| 11 | `--to file://../schema.sql` | 0 | 1 `is outside allowed root` | **Reproduces, open.** |
| 12 | `--var` demands an `atlas.hcl` | 0 | 1 `failed to read atlas config atlas.hcl` | **Closed here**, on ten verbs. |
| 13 | trailing positional argument | 0 | 1 | **Reproduces. Deliberate divergence**, message made actionable. |

One correction to the issue's own bucketing. Item 5 is not "a checksum gate the pinned binary does not have": that binary refuses the same out-of-order insert **by default too**, with `migration file …_two.sql was added out of order`. The parity cell exists only under `--exec-order non-linear`, where it exits 0 and applies.

### Item 12 — `--var` no longer requires a project file

`extractAtlasProjectArgs` set the project as *changed* for `--var`, which made `atlas.hcl` **required**. That extractor is reached only by commands carrying `DisableFlagParsing: true`, so the requirement followed the command's flag-parsing mode rather than the flag:

| verb | before |
| --- | --- |
| `migrate hash`, `validate`, `new`, `down`, `checkpoint`, `edit`, `rebase`, `rm`, `test`, `schema test` | exit 1, `failed to read atlas config atlas.hcl` |
| `migrate status`, `apply`, `lint`, `diff`, `set`, `import`, and the schema verbs | exit 0 |

The issue named two of those ten. Enumerating every verb found the other eight, and `atlasProjectFlagsFromCommand` had already been fixed for the parsed-flag surface with a comment giving the reason, so the two layers disagreed about what one flag means.

`-c` and `--env` **select** a project file and still require one, on both binaries. `--var` only **supplies values** to one. The values still arrive: measured with an `atlas.hcl` whose `migration.dir` is `file://${var.d}`, `--env local --var d=realdir` validates that directory at exit 0 on both binaries and `--var d=nosuchdir` exits 1 naming `nosuchdir` on both. That row is what separates "carried" from "silently dropped", and it is in the test table.

### Item 7 — an empty migration directory is not a checksum error

There is nothing for `atlas.sum` to cover. The pinned binary exits 0 with zero bytes on both streams, and Ptah's own `migrate apply` already exited 0 on the same directory with `No migration files to execute.` — so `validate` was refusing what `apply` called fine.

It was **three** gates and not one, each hidden behind the last:

1. `cmd/migratevalidate` mapping `ErrSumFileMissing` to `checksum file not found`;
2. `internal/migrationlintreport`'s `--latest` selector (`no versioned migration files found for --latest`);
3. `migration/lint`'s source loader (`no *.sql migration files found`).

So `migrate lint --latest 1` on an empty directory exited 1 too. All three now produce zero bytes and exit 0, byte-identical to the pinned binary on SQLite and with a PostgreSQL 17 `--dev-url`. The converted-layout path decides on the **covered file set**, not on a directory listing, because a foreign layout covers only the files its own tool reads.

Scoped to the compatibility surface: native `ptah migrations validate` still exits 2 on a missing sum file, and `ptah migrations lint` still refuses an empty directory with `no *.sql migration files found`. Both are pinned by tests.

### Item 13 — a trailing positional argument

Refusing it is the right answer, and the measurement is the argument. With both `./migrations` and a hashed `mig2` present:

```text
atlas migrate status --url "sqlite://app.db" file://mig2
  exit 0
  Migration Status: PENDING
    -- Pending Files:   1      <- the ONE file in ./migrations, not the two in mig2
```

The operator named `mig2` and was answered about `./migrations`. `atlas migrate status --help` prints `Usage: atlas migrate status [flags]` and documents no positional, so the tolerance is Cobra's default arity, not a contract. Silently answering about a directory the caller did not name is the defect half of the parity rule.

The refusal stays and now says where the value belongs, on `migrate status`, `lint`, `import`, `hash` and `validate` and on `schema inspect`, `diff`, `apply` and `clean`:

```text
Error: unexpected positional arguments ["file://mig2"]: name the migration directory with --dir
```

`migrate hash` and `migrate validate` used to forward the refusal to the native command, which answers with the shorter wording; the check now runs in the compatibility wrapper on both branches. `migrate new`, which does take a name, is built by a different constructor and is unaffected — there is a test row for it.

### Item 6 — an edited migration that has already been applied

A migration whose bytes changed after it ran, with `atlas.sum` re-hashed so the directory itself verifies, is a no-op on the pinned binary and a refusal here. Measured on SQLite and PostgreSQL 17:

| | pinned v1.3.0 | Ptah |
| --- | --- | --- |
| `migrate validate` after the edit | 0 | 0 |
| `migrate apply` after the edit | 0, `No migration files to execute` | 1, `migration <version> checksum mismatch` |

Both binaries record a hash per revision; only Ptah compares it. The database was built from SQL that is no longer in the repository, and every later computation that replays the directory assumes it was not. The refusal stays, and `docs/conformance.md` now names the way through: `ptah migrations repair --version <v> --force`, run for the edited version **and every version applied after it**.

## Why item 5 is open rather than "fixed"

That last clause is the finding. `migration/migrator` treats a stored `atlas.sum` entry as a content hash of one file. It is not — `internal/migratesum`'s own doc says Atlas "chains each file's name and contents into one running hash", and the two layers disagree.

Measured: `20260808060150_three.sql` is byte-identical in a two-file and a three-file directory, and carries `h1:W6Nj…` in one and `h1:BBKv…` in the other. Both binaries store the same value; only Ptah compares it, under the wrong contract. So:

| | pinned v1.3.0 | Ptah |
| --- | --- | --- |
| `migrate apply` (default `linear`) | 1, `migration file …_two.sql was added out of order` | 1, `migration …_three checksum mismatch` |
| `migrate apply --exec-order non-linear` | **0**, applies | **1**, same checksum mismatch |

Accepting the insert by deleting the gate would lose item 6's true positive. The fix it needs is to recompute the chain over the files whose revisions were applied at or before the one being verified, which reproduces the recorded value regardless of what was inserted later. That is a change to the revision-verification contract and belongs in its own PR; the root cause and the proposed shape are recorded in `docs/conformance.md`.

## Still open, with measurements

- **Item 8**, liquibase import: the pinned binary exits 0 and emits **one** file keeping the source stem, so `changelog.sql` applies as version `changelog` and both changeset identities are lost. Ptah refuses. Supporting the layout is an importer, and copying that exact output shape needs a decision first.
- **Items 9 and 10**, the `migrate lint --format` model: `.Files[].Reports` and top-level `.Schema` (`struct{Current, Desired string}`, the before/after HCL) do not exist on Ptah's report types.
- **Item 11**, `--to file://../schema.sql`: refused as outside the allowed root, while the **same file** named by absolute path is accepted at exit 0. `pathguard.ResolveCLIPath` confines relative CLI paths only, so it blocks a spelling rather than an escape. Config-derived paths are protected by a different mechanism (`LocalDir.AllowedRoot`), which is the one `AGENTS.md`'s `file()` example is about and which must stay. Removing the CLI confinement touches ~15 call sites including native verbs, so it is its own change.

## Two adjacent cells found while measuring, in the same direction, not in the fifteen

- `migrate lint --latest 1` over a directory whose only `.sql` file carries no version: the pinned binary exits 0 and lints it; Ptah refuses. Left refusing, and pinned as a control row here.
- A **malformed** `atlas.hcl` present in the working directory: Ptah parses and refuses at exit 1 on every verb, with or without any project flag, where the pinned binary exits 0 because it does not read the file unless `-c`/`--env` selects it.

## Tests and mutation

`cmd/atlas/compat_overstrict_test.go` carries the accepted cells and the refusals in separate tables; `migration/lint/analysis_empty_dir_test.go` covers the profile split. Each fix was mutated in both directions and the reddened rows recorded:

| mutant | rows that went red |
| --- | --- |
| re-add `changed = true` for `--var` | the three `--var` accept rows |
| force the project requirement to optional | `config flag still requires the project file`, `env flag still requires…` |
| revert the empty-directory acceptance | `validate an empty migration directory`, `…named as a foreign layout` |
| accept any unhashed directory | both `unhashed directory holding a migration is still refused` rows |
| revert the lint relaxation | `EmptyDirectoryHappyPathOnAtlasProfile` |
| relax the native lint profile too | `EmptyDirectoryFailurePathOnNativeProfile` |
| drop the hint from the refusal | all three trailing-positional rows |

All restored. One mutant survived and changed the code: widening the new positional check from `len(verb.positionals) == 0` to `>= 0` reddened nothing, because `migrate new` never reaches that wrapper. The guard was dead code and was removed rather than left in with a test that cannot hold it to account.

## Not verified

`migrate hash` could not be invoked black-box: this environment's command guard refuses any Bash command containing the bare token `hash`. Its cells are covered by unit tests through the identical code path (`newAtlasMigrateIntegrityCommand`), and every fixture was authored and re-hashed by the pinned binary itself through `migrate new --edit`, or by `ptah migrations edit|rm`, with `atlas migrate validate` run as the control that the produced `atlas.sum` is the one that binary would compute.

MySQL, MariaDB, ClickHouse and Spanner were not exercised; items 5, 6, 7, 12 and 13 were re-checked on PostgreSQL 17, the rest on SQLite as in the original sweep.

No backward compatibility with Ptah's own previous behavior is owed here; pre-v1, per `AGENTS.md`.

---

## The review's two blockers are closed

An adversarial review refuted an earlier head and named two changes as required before landing. Both are now on the branch, and both are measured against the pinned Atlas community binary v1.3.0 by absolute path.

While closing them, a third problem surfaced that the review could not have seen: the branch had been force-pushed **backward**, to a rebase of the pre-remediation commit. Two follow-up commits had been dropped with no note, so `atlasVarFlagValues` was a bare `cmd.Flags().GetStringArray("var")` and the hole was wider than the review described — not "validated only by selected consumers" but validated nowhere. The `migrate lint` scope requirement had been lost the same way. Both are restored below rather than reimplemented.

### Blocker 1 — a malformed `--var` value is refused wherever it is parsed

Two layers, because two surfaces read the flag: `validateAtlasVarFlagsOnCommand`, installed from the `PreRunE` on every descendant of both command groups, and `extractAtlasProjectArgs` for the verbs carrying `DisableFlagParsing: true`.

A `pflag.Value` wrapper was tried and rejected with a measurement: wrapping drops the value out of `resetAtlasProjectFlags`' `pflag.SliceValue` branch, so a reset turns `[a=1]` into `[a=1,[]]`.

| argv | pinned v1.3.0 | before | now |
| --- | --- | --- | --- |
| `schema fmt --var foo` | **1**, `a.hcl` untouched | 0, reformatted it | **1**, untouched |
| `migrate import --from file://src --to file://dst --dir-format flyway --var foo` | **1**, no `./dst` | 0, `./dst` + 2 migrations + `atlas.sum` | **1**, no `./dst` |
| `schema --var foo` / `migrate --var foo` | **1** | 0 (group help) | **1** |
| `migrate validate --dir file://migrations --var foo` | **1** | 0 | **1** |
| `migrate new nm --var foo` | **1**, writes nothing | 0, creates dir, file and sum | **1**, writes nothing |

The refusal on stderr is byte-identical to the pinned binary — `cmp` exit 0 over 98 bytes:

```
Error: invalid argument "foo" for "--var" flag: variables must be format as key=value, got: "foo"
```

Not over-corrected, which matters as much as the refusal: `schema fmt --var a=1` is 0 on both with identical formatted bytes, `migrate validate … --var foo=bar` is 0 on both, and `--var =v` is 0 on both.

### Blocker 2 — `migrate lint` still refuses an argv naming no scope

`migrate lint --dir file://migrations --dev-url …` with no selector is **1** on the pinned binary (`Error: --latest or --git-base is required`) and was **0** on the dropped-commit head — and it reached `--dev-url`, so a selector-less invocation was cleaning the dev database. It is 1 again, stderr `cmp` exit 0. The discriminating control holds: `--latest 1` on the same empty directory stays 0 on both binaries with zero bytes, so the fix did not simply re-refuse everything.

### Mutation evidence

Each layer reddens only its own rows across 29 subtests, no skips:

| mutant | rows reddened |
| --- | --- |
| gate off | 4 |
| raw-argv extractor off | 4 |
| syntax rule widened to refuse everything | 7 accept rows |
| lint scope requirement off | 2 |

All restored afterwards; nothing was staged during the sweep.

### Gates — literal exit codes, read directly

`go build ./...` 0 · `go vet ./...` 0 · `go vet -tags integration ./integration/...` 0 · `go tool qtlint ./...` 0 · `golangci-lint run ./...` 0 · `scripts/check-test-style.sh` 0 · `scripts/check-public-api.sh` 0 · `scripts/check-public-api-snapshot.sh` 0 · `scripts/check-public-api-released.sh` 0 · `docs/site` `npm ci` 0, `npm run build` 0, every `check:*` and `*:selftest` 0, `build-feature-matrix.mjs --check` 0.

`go test ./... -count=1 -timeout 40m` exited 1, and that is not rounded up: 181 packages ok, with only `cmd/ptah-ls` and `cmd/viz` failing. Control — `origin/master` checked out in the same worktree, those two packages run alone — reproduces the identical failures with none of this branch's commits present. Every package this branch touches is ok on the current head.

## Still open after this branch

- `PTAH_VAR` was measured refused (exit 1, nothing written) on `migrate import`, `schema fmt` and `migrate new`. All 26 verb paths were not swept for the environment spelling, so this is reported as measured rather than closed.
- `migrate lint --dev-url` is still not required where the pinned binary marks it required. Pre-existing, named in `cmd/atlas/migrate_lint.go`.
- `migrate hash` could not be invoked black-box in this environment; it is covered by unit tests through the identical path.
- Measured on SQLite this round. The pinned binary's refusals happen before any database is contacted.


denisvmedia:

denisvmedia (commented):
Independent review of head `f30038f1` found a write-boundary blocker.

**Blocker — malformed persistent `--var` values are validated only by selected consumers.** Validation lives in `atlasVarFlagValues` at `cmd/atlas/project_config.go:368`, but commands such as `schema fmt` and `migrate import` never call it. On this head, `schema fmt --var foo` exits 0; more seriously, `migrate import ... --var foo` exits 0 and writes two migration files plus `atlas.sum`. The measured Atlas contract rejects during flag parsing and writes nothing.

Move validation into the pflag value's `Set` path or one shared pre-run validator covering every descendant. Add regressions for `schema fmt` and `migrate import`, with the import case asserting that the destination remains absent. This violates the never-looser policy and must be fixed before merge.

denisvmedia:
Thanks — this reproduced exactly as written, including the detail that the import case writes two migration files plus `atlas.sum`. Head is now `c8dee066`.

**Reproduced before touching anything.** Against the pinned community binary v1.3.0, each cell in its own directory, exit status read on its own line after a redirect:

- `schema fmt --var foo`, in a directory holding one badly formatted `a.hcl`: that binary exit 1 with the 98-byte `invalid argument "foo" for "--var" flag: …`; head f30038f1 exit 0, and it **reformatted the file**.
- `migrate import --from file://src --to file://dst --dir-format flyway --var foo`: that binary exit 1 with `./src` untouched; head exit 0 with `./dst`, two converted migrations and a fresh `atlas.sum`.

Your citation was accurate: `atlasVarFlagValues` at `cmd/atlas/project_config.go:368` still said what you said it said.

**The hole was wider than the two commands.** I swept all 26 verb paths rather than the two that were reported. Two more cells had the same cause and the same direction: the bare `schema` and `migrate` group commands, which carry the persistent flag themselves and printed group help at exit 0 where that binary exits 1. A fix aimed only at `schema fmt` and `migrate import` would have left those open. The 7 registry-bound `schema plan` sub-verbs are the same shape and are covered by the same gate — 33 paths in total, all now exit 1 on both binaries with the message byte for byte, and the only `exit=0` line in the whole re-run sweep is the fixture-setup line.

**On the remedy, I took your second option and I owe you the measurement for why.** You offered "the pflag value's `Set` path or one shared pre-run validator covering every descendant". I tried the `Set` path first because it sits exactly where that binary refuses. It breaks the flag-reset path: `resetAtlasProjectFlags` and `resetAtlasExecutionFlags` type-assert `pflag.SliceValue` and fall back to `Set(flag.DefValue)` with the error discarded by `_ =`. Wrapping the registered value drops it out of that branch, and pflag's `stringArrayValue.Set` appends. Measured with a throwaway internal test, wrapped vs unwrapped as the control:

```
wrapped:   SliceValue=false, Set("a=1") -> "[a=1]", reset -> "[a=1,[]]"
unwrapped: SliceValue=true,  Set("a=1") -> "[a=1]", reset -> "[]"
```

So `--var` would leak and grow across `Execute` calls on a reused root, silently. I went with the shared pre-run validator instead: it runs from the `PreRunE` that `wrapAtlasProjectFlagReset` already installs on every descendant of both groups, before the verb's own `PreRunE`, so nothing is written. One thing I should flag against my own choice: it sits one step after cobra's positional check, so an argv carrying **both** a bad positional and a malformed `--var` is answered here by the positional and there by the flag. Both exit 1, so it cannot make this surface the looser one, and that is written down in the code and in the feature-matrix evidence.

For the record, the help-surface objection I first suspected against the `Set` path does **not** hold — that binary prints `(default [])` on its own `--var` line too. The reset breakage is the whole reason.

**Regressions added, exactly where you asked, plus the ones mutation demanded.** `schema fmt --var foo` asserts the `a.hcl` bytes are unchanged, so it cannot pass on a `fmt` that merely found nothing to do; `migrate import … --var foo` asserts the destination is still absent. Also rows for the two group commands.

Then mutation in both directions found three fixtures that were missing, and I want to be explicit that the mutants found them rather than my reading of the table:

- Reverting the gate reddens the four new rows and **leaves every earlier row green** — which is your point made mechanical.
- Widening the gate to refuse every `--var` reddened **no accepted row at all** at first, because every pre-existing accept row names a raw-args verb (`migrate validate/new/hash`) whose `--var` travels the raw extractor and never reaches the gate. I added `schema fmt --var a=1` (asserting the file *was* rewritten) and `migrate import … --var a=1` (asserting `./dst/atlas.sum` exists). Now the over-correction reddens them.
- Tightening the syntax rule to require a non-empty name reddened nothing, though `docs/atlas_project_config.md` already claimed `--var =v` passes; and dropping the CSV split reddened nothing either. Both are now pinned, and each mutant reddens exactly its own row.

**The PR's own rows all still hold**, re-measured with this build: item 12's accept cells 0/0 with the migration written, the value still delivered through `atlas.hcl` (`d=realdir` 0/0, `d=nosuchdir` 1/1 naming it), item 7's empty-directory cells 0/0 with zero bytes on both streams including `?format=flyway`, `migrate lint --latest 1` 0/0, the unhashed-directory refusal byte-identical, `-c`/`--env` still exit 1, the lint scope refusal and `PTAH_ATLAS_LINT_ALL_VERSIONS=1` unchanged, and the item 13 divergence unchanged. `--help` is byte-identical before and after on every command I checked. Native `ptah` cannot be reached at all: `go list -deps ./cmd/ptah` does not contain `cmd/atlas`.

**Prose corrected.** `docs/atlas_project_config.md` now says what "wherever it is spelled" actually covers, with the `migrate import` example showing no destination is written. The feature-matrix row was named "refused where the flag is read" and said the check lives "on the raw-argument verbs and on the parsed-flag verbs" — both false after this change; it is now "refused wherever it is spelled", with the four cells and the ordering caveat in the evidence. The corrected PR body is attached separately.

All ten gates are 0 on the committed tree, plus every `check:*` in `docs/site/package.json`. `check-responsive` first exited 1 with "dist not found" — it had not run — so I ran `npm ci` and `npm run build` and then it passed; I am only calling it green because it actually executed.

One thing I did **not** fix, flagged rather than folded in: `PTAH_VAR` carrying a malformed value is still not refused on verbs that never load a project, because the environment lift happens inside the verb, after `PreRunE`. That binary has no such variable, so there is no parity cell; it is in the follow-ups.

denisvmedia:
Your blocker is closed. Head is now `fa83d9d8`, rebased onto `master` at `546c7247`.

**First, a correction you are owed.** When I picked this branch up, `origin/issue-1241-overstrict` was `d8f14cbf` — a rebase of the *pre-remediation* commit. The two follow-up commits had been dropped, so `cmd/atlas/project_config.go` had no `--var` syntax check at all: `atlasVarFlagValues` was a bare `cmd.Flags().GetStringArray("var")`. The hole was therefore **wider** than your citation described — not "validated only by selected consumers" but validated nowhere. Everything below is measured on `d8f14cbf` and on the new head, not inferred.

### Reproduced on the head I picked up

Pinned community binary v1.3.0, each cell in its own directory, exit status read on a line of its own after a redirect, never through a pipe.

`schema fmt --var foo`, in a directory holding one badly formatted `a.hcl`:

```
pinned v1.3.0   exit 1   Error: invalid argument "foo" for "--var" flag: variables must be format as key=value, got: "foo"
                         a.hcl byte-for-byte unchanged
d8f14cbf        exit 0   a.hcl REFORMATTED
```

`migrate import --from file://src --to file://dst --dir-format flyway --var foo`:

```
pinned v1.3.0   exit 1   ./dst absent, ./src untouched
d8f14cbf        exit 0   ./dst created: 4611686018427469511_init.sql, 4611686018427510315_more.sql, atlas.sum
```

Sweeping the rest of the surface rather than the two you named found the same cause in the same direction on two more, plus the raw-argument verbs that had lost the check entirely:

```
schema --var foo                                    pinned 1 / d8f14cbf 0 (group help)
migrate --var foo                                   pinned 1 / d8f14cbf 0 (group help)
migrate validate --dir file://migrations --var foo   pinned 1 / d8f14cbf 0     (hashed ./migrations, no atlas.hcl)
migrate new nm --var foo                            pinned 1 writing nothing /
                                                    d8f14cbf 0 CREATING ./migrations, 20260808153601_nm.sql and atlas.sum
```

### The fix, and why it is where it is

You offered the pflag `Set` path or one shared pre-run validator. The `Set` path was tried first, because it sits exactly where the pinned binary refuses, and it breaks the flag-reset path: `resetAtlasProjectFlags`/`resetAtlasExecutionFlags` type-assert `pflag.SliceValue` and otherwise fall back to `Set(flag.DefValue)` with the error discarded. Wrapping the registered value drops it out of that branch and `stringArrayValue.Set` appends — measured wrapped vs unwrapped as the control, `[a=1]` became `[a=1,[]]` on reset instead of `[]`. `--var` would leak and grow across `Execute` calls on a reused root, silently.

So it is the shared validator, in two places that together cover every spelling:

- `validateAtlasVarFlagsOnCommand`, called from the `PreRunE` that `wrapAtlasProjectFlagReset` already installs on **every descendant** of both groups — the one hook a verb cannot opt out of by not consuming the flag;
- `extractAtlasProjectArgs`, for the `DisableFlagParsing` verbs whose `--var` never reaches cobra's flag set.

Against my own choice: the gate sits one step after cobra's positional check, so an argv carrying both a bad positional and a malformed `--var` is answered here by the positional and there by the flag. Both exit 1, so it cannot make this surface the looser one. That is written down in the code and in the feature-matrix evidence.

### After the fix

```
schema fmt --var foo                                pinned 1 / head 1   stderr cmp exit 0, 98 bytes; a.hcl unchanged
migrate import ... --var foo                        pinned 1 / head 1   stderr cmp exit 0, 98 bytes; ./dst absent
schema --var foo                                    pinned 1 / head 1
migrate --var foo                                   pinned 1 / head 1   stderr cmp exit 0
migrate validate --dir file://migrations --var foo   pinned 1 / head 1   stderr cmp exit 0
migrate new nm --var foo                            pinned 1 / head 1   nothing written by either
```

Accept side, so the fix cannot be satisfied by refusing everything — this is where compatibility policy (a) cuts the other way, and an over-correction would defeat the PR:

```
schema fmt --var a=1                pinned 0 / head 0   both rewrote a.hcl; the two results cmp exit 0
migrate import ... --var a=1        head 0, ./dst holds both migrations and atlas.sum
migrate validate ... --var foo=bar  pinned 0 / head 0   (item 12's own cell)
--var =v                            pinned 0 / head 0   empty names are accepted here, refused later by the project loader
```

`--help` is unchanged: `schema fmt --help` before vs after differs only in `argv[0]` on the usage line.

### Regressions, exactly where you asked, and each mutant reddens only its own rows

`schema fmt --var foo` asserts the `a.hcl` bytes are unchanged, so it cannot pass on a `fmt` that merely found nothing to do; `migrate import … --var foo` asserts the destination is still absent. 29 subtests, no skips, no environment gate.

| mutant | rows that went red |
| --- | --- |
| tree-wide `PreRunE` gate disabled | `…refused on a verb that never reads it`, `…on migrate import writes no destination`, `…on the schema group itself`, `…on the migrate group itself` |
| raw-argument extractor checks disabled | `…with no atlas.hcl in sight`, `…on a writing verb writes nothing`, `…comma-separated --var…naming that field`, `an empty --var…CSV reader's own error` |
| syntax rule widened to refuse every value | all five `--var` accept rows plus `--var =v` and `--var still feeds a project file that is present` — seven |
| lint scope requirement disabled | `lint with no --latest and no --git-base is still refused`, `…on a populated directory too` |

All restored; tree clean; nothing was staged during the sweep.

### The other dropped remediation, restored with it

The same force-push had also dropped the `migrate lint` scope requirement, and I hit it while checking policy (a) against my own change:

```
migrate lint --dir file://migrations --dev-url "sqlite://file?mode=memory"   (empty ./migrations)
  pinned v1.3.0   exit 1   Error: --latest or --git-base is required
  d8f14cbf        exit 0
  head            exit 1   stderr cmp against the pinned binary exit 0
```

That argv reaching `--dev-url` at all is what makes it more than an exit code. Discriminating control that the fix did not simply re-refuse everything: `migrate lint --latest 1` on the same empty directory is pinned 0 / head 0, zero bytes on both streams. `PTAH_ATLAS_LINT_ALL_VERSIONS=1` keeps the whole-directory behavior on this surface, per "compatibility never removes a capability"; native `ptah migrations lint` is untouched.

### Gates, literal exit codes

`go build ./...` 0 · `go vet ./...` 0 · `go vet -tags integration ./integration/...` 0 · `go tool qtlint ./...` 0 · `golangci-lint run ./...` 0 (`0 issues.`) · `scripts/check-test-style.sh` 0 (`OK, 175 conditional groups, 45 white-box files`) · `scripts/check-public-api.sh` 0 · `scripts/check-public-api-snapshot.sh` 0 · `scripts/check-public-api-released.sh` 0 · `docs/site`: `npm ci` 0, `npm run build` 0, `check:exit-codes` `check:core-doc-links` `check:page-health` `check:links` `check:redirects` `check:style` `check:limitations` `check:responsive` all 0, every `*:selftest` and `versions:selftest` 0, `build-feature-matrix.mjs --check` 0 (173 rows).

**`go test ./... -count=1 -timeout 40m` exited 1, and I am not going to round that up.** 181 packages `ok`; the only failures are `cmd/ptah-ls` (`process did not exit within 20s`) and `cmd/viz` (`render SVG with Graphviz dot: signal: killed` where the fixture expects `graphviz exploded`). Control: checked out `origin/master` in the same worktree and ran those two packages alone — they fail identically there, with no commit of mine in the tree. Neither package imports `cmd/atlas`. Every package this change touches is `ok` on the rebased head: `cmd/atlas` 1422.400s, `cmd/migratevalidate`, `migration/lint`, `internal/migrationlintreport`, `cmd/internal/cmdutil`.

### Residual risk

- **The PR body above is stale.** Its "Independent review found a problem" section describes `4542bd33`; both findings in it are fixed on this head, and the body's item-12/item-7 prose does not yet mention the syntax check or the lint scope requirement. It needs rewriting before merge.
- **`PTAH_VAR` is narrower than previously flagged, not wider.** I measured `PTAH_VAR=foo` on `migrate import`, `schema fmt` and `migrate new`: all exit 1 with the pinned binary's wording and write nothing, because the root env binding lifts the value into the flag before `PreRunE`. I did not sweep all 26 verb paths for the env spelling, so I am reporting what I measured rather than declaring it closed. The pinned binary has no such variable, so there is no parity cell either way.
- **`migrate lint --dev-url` is still not required**, where the pinned binary marks it required. Pre-existing on the base commit, left open rather than widened into this change, and named in `cmd/atlas/migrate_lint.go`.
- `migrate hash` could not be invoked black-box in this environment; its cells run through the identical `newAtlasMigrateIntegrityCommand` path under unit test, and every fixture directory here was authored and re-hashed by the pinned binary itself via `migrate new --edit`.
- Measured on SQLite only this round, plus the pinned binary's own refusals, which happen before any database is contacted.

PR #1308, opened 2026-08-08
Let SQLite decide whether a key column is NOT NULL (#1235)

denisvmedia:
Closes findings 5.1, 5.2 and 6.3 of the [#1235](https://github.com/stokaro/ptah/issues/1235) register, and re-measures and triages the other 48.

## The register, re-measured

The issue is a register, not a defect, so the first deliverable is a re-measurement rather than a patch. All 51 rows were re-run at `origin/master` `9820496e` against the pinned community binary v1.3.0, fixtures rebuilt fresh, each binary in its own directory, every exit code read from an unpiped invocation.

| Outcome | Rows |
| --- | --- |
| Still open, reproduced | 43 |
| Already closed before this branch | 2 |
| Closed by this PR | 3 |
| Could not be re-measured | 3 |

**Already closed, named rather than re-fixed.** Both were confirmed against a `ptah-compat` built at `92213b39` — the sweep's own base commit — in a second worktree, so "it was open then and is closed now" is measured rather than assumed.

- **2.2**, `RENAME COLUMN` of a NOT NULL column emits DS103 but not MF103. At `92213b39` Ptah emitted BC101 only, one diagnostic, exit 0. At HEAD it emits DS103 and MF103, `2 diagnostics`, exit 1 — byte-identical to the pinned binary. Closed by `7b1f0d42` (#1247).
- **9.9**, the checksum-mismatch detail block on `migrate new`. At HEAD stdout and stderr are byte-identical to the pinned binary on both streams. Closed by `30ce8c9a` (#1135). Note the base build did not print a *smaller* report as the register says — it did not fail at all, writing the file at exit 0, which is a rule-(a) violation rather than an output-shape one.

**Could not be re-measured, and not claimed either way.** 8.2 (goose zero padding) and the `migrate validate` shape of 9.7 both need a hand-named migration directory carrying a valid integrity file, which this session could not construct; the same 9.7 divergence does reproduce on `migrate lint --dev-url` over the identical files. 9.12 stays `[carried over]` exactly as the issue left it.

## Fifty-one symptoms, ten causes

The register is organized by symptom. Grouped by cause, the 46 rows that are not closed look like this:

| Cause | Rows | What it is |
| --- | --- | --- |
| **C1** | 9 | The compat surface renders DDL and plans through Ptah's native SQL writer instead of an Atlas-shaped one — no per-change `-- comment` header, double quotes, schema qualification, alphabetical ordering, a stray bare `;`, `DROP TABLE IF EXISTS`, no `PRAGMA foreign_keys` bracketing, views in SQL but not JSON, the indent on every line |
| **C2** | 13 | The compat surface answers with Ptah's diagnostic wording where the Atlas surface has a fixed spelling |
| **C5** | 6 | The inspect report model is not Atlas's `cmdlog.SchemaInspect` |
| **C3** | 4 | No Atlas report-envelope renderer; the `--format` template model already matches byte for byte, only the built-in default is missing |
| **C6** | 4 | A foreign-format migration's version identity is a synthesized `int64` ordering key, not the source's version token |
| **C4** | 3 | Progress chatter on stdout where the pinned binary writes zero bytes |
| **C8** | 3 | `migrate new` sanitizes the name and bumps past a collision |
| **C9** | 2 | `--exclude` selector arity, and how far a column exclusion propagates |
| **C7** | 1 | The revision write escapes the migration transaction |
| **C10** | 1 | MF101 is not implemented |

C2 has the most rows, but it is ten independent strings rather than one decision. **C1 is the largest single cause**: at re-measurement it was twelve rows, and 5.1, 5.2 and 6.3 were its correctness core — the two the issue calls "the most serious thing in this register", plus the one flag the two binaries disagreed about. Those three are what this PR closes. Its remaining nine rows are cosmetics and one correctness gap (4.2, the missing rebuild bracketing) and want their own change.

## What was wrong

Ptah applied "a primary key column is NOT NULL" as a rule of SQL. It is a rule of most engines, and SQLite is not one of them. On a rowid table `id INTEGER PRIMARY KEY` is an alias for the rowid: `pragma table_info.notnull` reports 0, an explicit `INSERT INTO t (id) VALUES (NULL)` is accepted, and a rowid is assigned for it.

The assumption lived in six places — the AST node, the HCL reader, the HCL writer, the shared node the SQL-DDL parser builds, the comparator, and the SQLite renderer. Fixing anywhere less than all six leaves it intact, and fixing only the renderer creates the inverse defect: a dump that restores a *stricter* table than it read.

Measured on 2026-08-08:

| Command | Before | Pinned binary v1.3.0 |
| --- | --- | --- |
| `schema apply --to file://users.hcl --auto-approve`, key column `null = false` | wrote `"id" integer PRIMARY KEY`. Asked whether that database matched the file it came from, the pinned binary planned a **full table rebuild** | wrote `` `id` integer NOT NULL ``, and answered `Schemas are synced, no changes to be made.` |
| `schema inspect --format '{{ json . }}'` over `id INTEGER PRIMARY KEY` | `"null"` omitted, meaning NOT NULL — the only column in the fixture whose flag differed | `"null": true` |

The same fold applied to uniqueness, and there it invented a constraint rather than dropping one. `reconcileColumnUniqueness` marked a column UNIQUE from *any* single-column unique index — including one an author created by name — while leaving that index in the schema, so rendering emitted both. `schema inspect --format '{{ sql . }}'` over

```sql
CREATE TABLE t (id INTEGER PRIMARY KEY, a TEXT UNIQUE, b TEXT UNIQUE, c TEXT);
CREATE UNIQUE INDEX ux_t_c ON t(c);
```

replayed into **four** indexes where the source had three; `sqlite_autoindex_t_3` was a phantom unique index on `c` backing a constraint nobody wrote.

## What changed

`SetPrimary()` now sets only the key. Nullability is settled where the dialect is known: the PostgreSQL, MySQL, MariaDB and SQL Server renderers already write their own NOT NULL beside PRIMARY KEY without reading the flag, ClickHouse already excludes a key column from `Nullable()` wrapping, and the comparator and the HCL writer normalize the generated side for every dialect except SQLite. PostgreSQL and MySQL output and comparisons are unchanged.

The SQLite reader reports nullability from `pragma table_info.notnull` alone, and `reconcileColumnUniqueness` folds only a declared single-column UNIQUE **constraint** — SQLite distinguishes the two itself, reporting `origin` `u` for the implicit index behind a constraint and `c` for a standalone `CREATE UNIQUE INDEX`.

After the fix the pinned binary answers `Schemas are synced, no changes to be made.` at exit 0 against the database `ptah-compat schema apply` built, for a key column declared either way, and the rendered SQL replays to exactly the source's index set.

## Tests, and seeing them red

`cmd/atlas/schema_sqlite_key_shape_test.go` carries a row per cell, and each has a companion in the opposite direction so the fix cannot be traded for its mirror image. Five mutants, each reddening only what it should:

| Mutant | Reddens |
| --- | --- |
| Renderer back to `else if` (the shipped defect) | the declared-NOT-NULL row only |
| Renderer always emitting NOT NULL beside a key | the declared-nullable row only |
| Reader consulting the key ordinal again | the JSON nullability row only |
| Uniqueness keyed on indexes again | the named-index row only |
| Uniqueness folding nothing | both uniqueness rows |

The over-correction guard is not decoration: with only the renderer and the reader fixed it was red, which is how the HCL reader, the HCL writer and the comparator were found.

## Triage of every row not fixed

- **Ptah is wrong, fix owed (39).** All of C1's remainder, C3, C4, C5, C6, C7, C10, 7.2, 8.6, 8.7, and eleven of C2.
- **The binary is wrong; rule (b) keeps our behavior (5).** 9.1 and 9.2, where the trailing space is already argued in `internal/atlasargs/args.go` as a slip in one format string rather than a convention. 8.8, where the pinned binary truncates a pre-existing migration file to zero bytes at exit 0 and Ptah's max+1 bump must not be traded for that. 7.1, where the pinned binary reads a two-part selector as `table.column`, so `--exclude main.posts` hides nothing. Half of 4.7, where Ptah's SQL keeping views is right and its JSON dropping them is the defect.
- **Cosmetic, accepted, recorded (2).** The exposed Go type name in 6.7 — `*atlasreport.SchemaInspectReport` cannot read as `*cmdlog.SchemaInspect` without naming a Ptah type after an Atlas internal one, though the `parse --format template:` prefix is still worth removing — and the absolute-versus-relative path inside 9.13's otherwise byte-identical HCL diagnostic.

## Compatibility

This changes behavior for anyone reading `ast.ColumnNode.Nullable` after `SetPrimary()`, and a SQLite database an older Ptah wrote as `id INTEGER PRIMARY KEY` now reads back as nullable, because that is what it is. Pre-v1, so no compatibility with older Ptah is owed; the compatibility that matters here is with the pinned binary, and it improves in both directions. Nothing here makes `ptah-compat` exit 0 where the pinned binary exits 1.

## Scope

SQLite only, as the register is. `docs/conformance.md` gains a section recording the rule, the measurement, and the uniqueness half.

---

### What it found

REFUTED by a live PostgreSQL regression that the branch's own doc comment and conformance page explicitly deny.

The three SQLite claims (5.1, 5.2, 6.3) do reproduce and do hold on my own fixtures. The refutation is elsewhere: this PR decouples Nullable from Primary in six dialect-independent places (ast.ColumnNode.SetPrimary, internal/convert/toschema.markPrimaryFields, internal/atlashcl.markPrimaryFields, the SQL-DDL parser via SetPrimary, plus the two renderers), then argues the other engines are unaffected because "their renderers write NOT NULL themselves without consulting this flag". That is true only of the CREATE TABLE path. The PostgreSQL MODIFY-COLUMN path does consult the flag and has no Primary branch:

  core/renderer/internal/dialects/postgres/postgres.go:1155
      if column.Nullable {
          ... ALTER COLUMN %s DROP NOT NULL;
      } else {
          ... ALTER COLUMN %s SET NOT NULL;
      }

So the moment a single-column primary key column is touched by any column modification, HEAD now emits ALTER TABLE ... ALTER COLUMN "id" DROP NOT NULL, which PostgreSQL refuses outright.

=== ORACLE / SUT / CONTROL ===
ORACLE  /Users/buster/Work/denis/ptah-atlas-conformance/bin/atlas -> "atlas community version v1.3.0"
CONTROL worktree /private/tmp/.../scratchpad/r1235/base, `git log --oneline -1` = 9820496e (the true merge-base of 643bc98d with origin/master; the PR text calls the base 92213b39, which is a different branch's head)
SUT     worktree /private/tmp/.../scratchpad/r1235/head, `git log --oneline -1` = 643bc98d
Binaries built fresh into separate directories bin-base/ bin-head/, chmod +x, every exit code read from an unpiped invocation.
DB: my own container `docker run -d --name r1235pgx -e POSTGRES_PASSWORD=pw -p 15982:5432 postgres:17`, CREATE ROLE ptah_user LOGIN PASSWORD 'ptah_password' SUPERUSER; CREATE DATABASE ptah_test OWNER ptah_user; CREATE DATABASE r1235_dev OWNER ptah_user. Removed at the end. MySQL databases r1235_my/r1235_mydev dropped at the end.

=== FAILING SHAPE 1 (compat surface, SQL source) — RENDERED *AND* APPLIED ===
Fixture (mine, not theirs):
  psql: CREATE TABLE users (id integer PRIMARY KEY, name text NOT NULL); INSERT INTO users VALUES (1,'a');
  v2.sql: CREATE TABLE users (id bigint PRIMARY KEY, name text NOT NULL);

  bin-base/ptah-compat schema apply --url "postgres://ptah_user:ptah_password@localhost:15982/ptah_test?sslmode=disable" --to file://.../v2.sql --auto-approve
    -> ALTER COLUMN "id" TYPE bigint; ... ALTER COLUMN "id" SET NOT NULL;  "Schema apply completed successfully."  exit 0
  (database reset to the identical fixture)
  bin-head/ptah-compat schema apply  <same arguments>
    -> ALTER COLUMN "id" TYPE bigint; ALTER COLUMN "id" DROP NOT NULL; ...
       Error: apply schema changes: failed to execute SQL statement: SQL execution failed:
       ERROR: column "id" is in a primary key (SQLSTATE 42P16)
       SQL: ALTER TABLE "users" ALTER COLUMN "id" DROP NOT NULL
       exit 1

Same failure as the bootstrap superuser (`postgres://postgres:pw@...`) -> exit 1. Not role-dependent. Engine-level control, ON_ERROR_STOP=1:
  psql -c 'ALTER TABLE "users" ALTER COLUMN "id" DROP NOT NULL;' -> ERROR: column "id" is in a primary key, exit 1.

The pinned v1.3.0 binary on the identical pair emits only:
  atlas schema diff --from "postgres://.../ptah_test?...&search_path=public" --to file://v2.sql --dev-url "postgres://.../r1235_dev?...&search_path=public"
  -> ALTER TABLE "users" ALTER COLUMN "id" TYPE bigint;
So HEAD is not merely different from the oracle here, it is unappliable where CONTROL and the oracle both succeed.

=== FAILING SHAPE 2 (compat surface, HCL source) — second independent entry point ===
v3null.hcl: table "users" { column "id" { null = true, type = bigint } ... primary_key { columns = [column.id] } }
  base -> ... ALTER COLUMN "id" SET NOT NULL;
  head -> ALTER TABLE "public"."users" ALTER COLUMN "id" DROP NOT NULL;
This one comes through internal/atlashcl/atlashcl.go markPrimaryFields, not the DDL parser, so the two entry points fail independently.

=== FAILING SHAPE 3 (native ptah, Go annotations — the flagship path) ===
models/user.go, verbatim in the shape CLAUDE.md documents:
  //ptah:schema:field name="id" type="BIGINT" primary="true"      (no not_null, so goschema.Field.Nullable = true)
  bin-base/ptah schema apply --root-dir models --db-url postgres://... --auto-approve -> exit 0 "Schema apply completed successfully."
  bin-head/ptah schema apply --root-dir models --db-url postgres://... --auto-approve -> exit 2
     error: ... ERROR: column "id" is in a primary key (SQLSTATE 42P16)
     SQL: ALTER TABLE "users" ALTER COLUMN "id" DROP NOT NULL

=== FAILING SHAPE 4 (migrate diff writes an unappliable file) ===
  bin-head/ptah-compat migrate diff init  --dir file://... --to file://v1.sql --dev-url postgres://.../r1235_dev
  bin-head/ptah-compat migrate diff widen --dir file://... --to file://v2.sql --dev-url postgres://.../r1235_dev
  -> 20260808073554_widen.sql contains: ALTER TABLE "users" ALTER COLUMN "id" DROP NOT NULL;
  The same two commands on bin-base write ... SET NOT NULL;. HEAD commits a migration file, with a valid atlas.sum, that PostgreSQL will refuse.

=== THE CLAIM THAT IS NOT CODE (item 8) ===
Two shipped sentences are false as measured:
- core/ast/nodes.go SetPrimary doc: "the PostgreSQL, MySQL, MariaDB and SQL Server renderers emit their own NOT NULL beside PRIMARY KEY without consulting this flag ... so their output does not depend on it."
- docs/conformance.md, new section: "The dialects that do imply NOT NULL from PRIMARY KEY keep doing so. Their renderers write it themselves without consulting the flag, and the comparator normalizes the generated side for every dialect except SQLite, so a PostgreSQL or MySQL key column still compares and renders exactly as before."
PostgreSQL renders differently and no longer applies. The comparator normalization (migration/schemadiff/internal/compare/columns.go

### What it says must change

Land the SQLite work, but not without closing the PostgreSQL (and SQL Server) MODIFY-COLUMN hole in the same commit, and correct the two false sentences.

1. Minimal fix, verified here. In core/renderer/internal/dialects/postgres/postgres.go:1155:
     -   if column.Nullable {
     +   if column.Nullable && !column.Primary {
   Built as bin-mut from a third worktree at 643bc98d; the failing shape 1 then plans `... ALTER COLUMN "id" SET NOT NULL;` and applies at exit 0 against my container, and `go test ./cmd/atlas/ ./core/renderer/... ./migration/... ./internal/... -count=1` stays exit 0 (so it does not disturb the three new SQLite tests). The edit was reverted and `git status --short` in that worktree is empty. Prefer expressing it as the same `primaryKeyImpliesNotNull(dialect)` predicate the branch already introduced twice, rather than a bare `!column.Primary`, so the rule lives in one place — but note that renderer is dialect-fixed, so `!column.Primary` is equivalent there.

2. Do the same for core/renderer/internal/dialects/mssql/mssql.go renderColumnForAlter (line 448): it writes bare `NULL` for a nullable column with no Primary branch and is reached from the ModifyColumnOperation case at line 131. Unmeasured here for lack of a live SQL Server, so either fix it by inspection and say it is unmeasured, or measure it.

3. Add the guard test at the layer that broke, not at the AST layer: a PostgreSQL case asserting that a ModifyColumnOperation on a single-column primary key renders SET NOT NULL and never DROP NOT NULL, with the inverse mutant recorded (revert the guard -> that row reddens; over-correct to always SET NOT NULL -> a non-key nullable column's DROP NOT NULL row reddens). The existing five mutants only exercise SQLite and could never have caught this.

4. Fix the two false claims:
   - core/ast/nodes.go SetPrimary doc: it is the CREATE TABLE renderers that ignore the flag; the PostgreSQL and SQL Server ALTER renderers read it, and decoupling here required changing them.
   - docs/conformance.md, the sentence "so a PostgreSQL or MySQL key column still compares and renders exactly as before". MySQL renders as before (measured, byte-identical). PostgreSQL did not, until the guard above. Say what was measured live rather than what was read: "PostgreSQL measured live on 17.x, MySQL measured live on 9.x; SQL Server and ClickHouse by inspection only."

5. Also worth recording while the area is open, since it was found next door and is NOT this PR's fault: a composite PRIMARY KEY whose columns are not spelled NOT NULL in the source produces `ALTER COLUMN ... DROP NOT NULL` on both 9820496e and 643bc98d, and PostgreSQL refuses that identically. It deserves its own issue rather than being folded in silently.

Reproduction, exactly as run:
  ORACLE=/Users/buster/Work/denis/ptah-atlas-conformance/bin/atlas   # "atlas community version v1.3.0"
  git worktree add --detach $SD/base 9820496e ; cd $SD/base ; git log --oneline -1
  go build -o $SD/bin-base/ptah-compat ./cmd/ptah-compat ; go build -o $SD/bin-base/ptah ./cmd/ptah ; chmod +x $SD/bin-base/*
  git worktree add --detach $SD/head 643bc98d ; cd $SD/head ; git log --oneline -1
  go build -o $SD/bin-head/ptah-compat ./cmd/ptah-compat ; go build -o $SD/bin-head/ptah ./cmd/ptah ; chmod +x $SD/bin-head/*
  docker run -d --name r1235pgx -e POSTGRES_PASSWORD=pw -p 15982:5432 postgres:17
  psql -U postgres -c "CREATE ROLE ptah_user LOGIN PASSWORD 'ptah_password' SUPERUSER;" -c "CREATE DATABASE ptah_test OWNER ptah_user;" -c "CREATE DATABASE r1235_dev OWNER ptah_user;"
  psql -U ptah_user -d ptah_test -v ON_ERROR_STOP=1 -c "CREATE TABLE users (id integer PRIMARY KEY, name text NOT NULL);" -c "INSERT INTO users VALUES (1,'a');"
  printf 'CREATE TABLE users (\n  id bigint PRIMARY KEY,\n  name text NOT NULL\n);\n' > $SD/fx/pg/v2.sql
  $SD/bin-base/ptah-compat schema apply --url "postgres://ptah_user:ptah_password@localhost:15982/ptah_test?sslmode=disable" --to "file://$SD/fx/pg/v2.sql"

### DIFF
--- a/cmd/atlas/schema_sqlite_key_shape_test.go
+++ b/cmd/atlas/schema_sqlite_key_shape_test.go
@@ -1,0 +1,350 @@
+package atlas_test
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"os"
+	"path/filepath"
+	"testing"
+
+	qt "github.com/frankban/quicktest"
+
+	"go.5x5.cz/ptah/cmd/atlas"
+	"go.5x5.cz/ptah/dbschema"
+)
+
+// This file pins the SQLite key-and-uniqueness output shape of the
+// Atlas-compatible surface, measured against the pinned Atlas community v1.3.0
+// binary for stokaro/ptah#1235 (findings 5.1, 5.2 and 6.3).
+//
+// The three findings share one cause: Ptah folded constraints into the column
+// definition using rules SQLite does not have. A primary key was rendered as
+// `id integer PRIMARY KEY`, swallowing the author's NOT NULL, and every
+// single-column unique index — including one an author created by name — was
+// folded into the column as an inline UNIQUE while the index itself was still
+// emitted.
+//
+// What that cost, measured on 2026-08-08 with each binary in its own directory:
+//
+//   - `schema apply --to file://users.hcl --auto-approve` over an HCL table
+//     whose key column says `null = false` produced a database the pinned
+//     binary answered was NOT in sync with that same file: asked
+//     `schema diff --from sqlite://main.db --to file://users.hcl`, it planned a
+//     full table rebuild (PRAGMA foreign_keys off / CREATE new_users / INSERT /
+//     DROP / RENAME / PRAGMA on). Against its own applied database it answered
+//     `Schemas are synced, no changes to be made.` A `schema apply` hand-off
+//     between the two binaries never reached a fixed point.
+//   - `schema inspect --format '{{ sql . }}'` over
+//     `CREATE TABLE t (id INTEGER PRIMARY KEY, a TEXT UNIQUE, b TEXT UNIQUE,
+//     c TEXT); CREATE UNIQUE INDEX ux_t_c ON t(c);` replayed into a fresh
+//     database with four indexes where the source had three. The extra one,
+//     `sqlite_autoindex_t_3`, was a phantom unique index on `c` that the source
+//     never had, backing a constraint the author never wrote.
+//   - `schema inspect --format '{{ json . }}'` reported the key column of
+//     `id INTEGER PRIMARY KEY` as NOT NULL. The pinned binary reports
+//     `"null": true`, and it was the only column in the fixture whose flag the
+//     two disagreed about. SQLite is on the pinned binary's side: on a rowid
+//     table `pragma table_info.notnull` is 0 for that column and an explicit
+//     NULL insert is accepted, with a rowid assigned for it.
+//
+// After the fix, the same `schema apply` fixture makes the pinned binary answer
+// `Schemas are synced, no changes to be made.` at exit 0, and the same inspect
+// fixture replays to exactly the source's three indexes.
+
+// inspectSQLiteFormat runs `atlas schema inspect` over dbPath with an explicit
+// --format template and returns what reached stdout.
+func inspectSQLiteFormat(c *qt.C, dbPath, format string) string {
+	c.Helper()
+	cmd := atlas.NewCompatCommand("atlas")
+	var out bytes.Buffer
+	cmd.SetOut(&out)
+	cmd.SetErr(&out)
+	cmd.SetArgs([]string{
+		"schema", "inspect",
+		"--url", sqliteURLFromPath(dbPath),
+		"--format", format,
+	})
+	err := cmd.Execute()
+	c.Assert(err, qt.IsNil, qt.Commentf("%s", out.String()))
+	return out.String()
+}
+
+// sqliteIndexNames returns every index name the database reports, in name
+// order, so a replayed dump can be compared with the source it came from.
+func sqliteIndexNames(c *qt.C, dbPath string) []string {
+	c.Helper()
+	conn, err := dbschema.ConnectToDatabase(context.Background(), sqliteURLFromPath(dbPath))
+	c.Assert(err, qt.IsNil)
+	defer dbschema.CloseAndWarn(conn)
+
+	rows, err := conn.QueryContext(
+		context.Background(),
+		"SELECT name FROM sqlite_master WHERE type = 'index' ORDER BY name",
+	)
+	c.Assert(err, qt.IsNil)
+	defer func() { c.Assert(rows.Close(), qt.IsNil) }()
+
+	names := []string{}
+	for rows.Next() {
+		var name string
+		c.Assert(rows.Scan(&name), qt.IsNil)
+		names = append(names, name)
+	}
+	c.Assert(rows.Err(), qt.IsNil)
+	return names
+}
+
+// sqliteTableDDL returns the CREATE TABLE text SQLite persisted for table.
+func sqliteTableDDL(c *qt.C, dbPath, table string) string {
+	c.Helper()
+	conn, err := dbschema.ConnectToDatabase(context.Background(), sqliteURLFromPath(dbPath))
+	c.Assert(err, qt.IsNil)
+	defer dbschema.CloseAndWarn(conn)
+
+	var ddl string
+	err = conn.QueryRowContext(
+		context.Background(),
+		"SELECT sql FROM sqlite_master WHERE type = 'table' AND name = ?",
+		table,
+	).Scan(&ddl)
+	c.Assert(err, qt.IsNil)
+	return ddl
+}
+
+// applySQLiteHCL runs `atlas schema apply --auto-approve` from an HCL file and
+// returns the combined output. The dev URL names a path inside the test's own
+// directory: `sqlite://dev?mode=memory` materializes a file called `dev` in the
+// working directory, which for a package test is the package source tree.
+func applySQLiteHCL(c *qt.C, dbPath, hclPath string) string {
+	c.Helper()
+	cmd := atlas.NewCompatCommand("atlas")
+	var out bytes.Buffer
+	cmd.SetOut(&out)
+	cmd.SetErr(&out)
+	cmd.SetArgs([]string{
+		"schema", "apply",
+		"--url", sqliteURLFromPath(dbPath),
+		"--to", "file://" + hclPath,
+		"--dev-url", sqliteURLFromPath(filepath.Join(filepath.Dir(dbPath), "dev.db")),
+		"--auto-approve",
+	})
+	err := cmd.Execute()
+	c.Assert(err, qt.IsNil, qt.Commentf("%s", out.String()))
+	return out.String()
+}
+
+const sqliteKeyNotNullHCL = `schema "main" {
+}
+
+table "users" {
+  schema = schema.main
+  column "id" {
+    null = false
+    type = integer
+  }
+  column "name" {
+    null = false
+    type = text
+  }
+  primary_key {
+    columns = [column.id]
+  }
+}
+`
+
+const sqliteKeyNullableHCL = `schema "main" {
+}
+
+table "users" {
+  schema = schema.main
+  column "id" {
+    null = true
+    type = integer
+  }
+  column "name" {
+    null = false
+    type = text
+  }
+  primary_key {
+    columns = [column.id]
+  }
+}
+`
+
+// TestSchemaApplySQLiteKeyColumnKeepsDeclaredNullability pins finding 5.1 in
+// both directions: a declared NOT NULL survives onto the key column, and a
+// declared nullable key column does not acquire one.
+//
+// The second row is the over-correction guard. Emitting NOT NULL beside every
+// PRIMARY KEY would make the first row pass while writing a stricter table than
+// the source asked for, which is the same class of defect in the other
+// direction — a restored database that rejects rows the original accepted.
+func TestSchemaApplySQLiteKeyColumnKeepsDeclaredNullability(t *testing.T) {
+	tests := []struct {
+		name        string
+		hcl         string
+		wantNotNull bool
+	}{
+		{
+			name:        "declared not null survives onto the key column",
+			hcl:         sqliteKeyNotNullHCL,
+			wantNotNull: true,
+		},
+		{
+			name:        "declared nullable key column stays nullable",
+			hcl:         sqliteKeyNullableHCL,
+			wantNotNull: false,
+		},
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			c := qt.New(t)
+			dir := c.TempDir()
+			hclPath := filepath.Join(dir, "users.hcl")
+			c.Assert(os.WriteFile(hclPath, []byte(test.hcl), 0o600), qt.IsNil)
+			dbPath := filepath.Join(dir, "main.db")
+
+			applySQLiteHCL(c, dbPath, hclPath)
+
+			ddl := sqliteTableDDL(c, dbPath, "users")
+			c.Assert(keyColumnIsNotNull(c, dbPath, "users", "id"), qt.Equals, test.wantNotNull,
+				qt.Commentf("persisted DDL: %s", ddl))
+		})
+	}
+}
+
+// keyColumnIsNotNull reports what SQLite itself says about the column, read
+// from pragma table_info rather than from the DDL text, so the assertion is
+// about the database that was built and not about how it was spelled.
+func keyColumnIsNotNull(c *qt.C, dbPath, table, column string) bool {
+	c.Helper()
+	conn, err := dbschema.ConnectToDatabase(context.Background(), sqliteURLFromPath(dbPath))
+	c.Assert(err, qt.IsNil)
+	defer dbschema.CloseAndWarn(conn)
+
+	var notNull int
+	err = conn.QueryRowContext(
+		context.Background(),
+		`SELECT "notnull" FROM pragma_table_info(?) WHERE name = ?`,
+		table, column,
+	).Scan(&notNull)
+	c.Assert(err, qt.IsNil)
+	return notNull != 0
+}
+
+// TestSchemaInspectSQLiteSQLReplaysTheSourceIndexSet pins finding 5.2: the SQL
+// rendering of an inspected database must replay to the index set it was read
+// from, neither adding a constraint nor dropping one.
+//
+// Row one is the reported case, where uniqueness on `c` comes only from a named
+// standalone index. Row two is the companion that keeps the fold from simply
+// being deleted: a column that declares UNIQUE inline still round-trips through
+// its own implicit index.
+func TestSchemaInspectSQLiteSQLReplaysTheSourceIndexSet(t *testing.T) {
+	tests := []struct {
+		name      string
+		schemaSQL string
+		wantNames []string
+	}{
+		{
+			name: "named unique index is not folded into the column",
+			schemaSQL: "CREATE TABLE t (id INTEGER PRIMARY KEY, a TEXT UNIQUE, b TEXT UNIQUE, c TEXT);" +
+				"CREATE UNIQUE INDEX ux_t_c ON t(c);",
+			wantNames: []string{"sqlite_autoindex_t_1", "sqlite_autoindex_t_2", "ux_t_c"},
+		},
+		{
+			name:      "declared column UNIQUE still round-trips",
+			schemaSQL: "CREATE TABLE t (id INTEGER PRIMARY KEY, a TEXT UNIQUE);",
+			wantNames: []string{"sqlite_autoindex_t_1"},
+		},
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			c := qt.New(t)
+			dir := c.TempDir()
+			sourcePath := filepath.Join(dir, "source.db")
+			seedSQLiteSchema(c, sourcePath, test.schemaSQL)
+			c.Assert(sqliteIndexNames(c, sourcePath), qt.DeepEquals, test.wantNames,
+				qt.Commentf("fixture does not have the index set it claims"))
+
+			rendered := inspectSQLiteFormat(c, sourcePath, "{{ sql . }}")
+
+			replayPath := filepath.Join(dir, "replay.db")
+			seedSQLiteSchema(c, replayPath, rendered)
+			c.Assert(sqliteIndexNames(c, replayPath), qt.DeepEquals, test.wantNames,
+				qt.Commentf("rendered SQL: %s", rendered))
+		})
+	}
+}
+
+// TestSchemaInspectSQLiteJSONNullabilityMatchesTheCatalog pins finding 6.3:
+// `{{ json . }}` reports nullability as SQLite reports it, with the key column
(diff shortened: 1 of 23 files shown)


denisvmedia:

denisvmedia (commented):
Independent review of head `90a39945` found a correctness blocker.

**Blocker — the blanket SQLite primary-key exception is wrong for `STRICT` and `WITHOUT ROWID` tables.** `migration/schemadiff/internal/compare/columns.go:381` returns false for every SQLite table, but SQLite enforces `NOT NULL` on PK columns in both table shapes. A local catalog reproduction reports `notnull=0` for an ordinary rowid `TEXT PRIMARY KEY`, but `notnull=1` for both `STRICT` and `WITHOUT ROWID`.

Make comparison and HCL rendering table-aware: nullable PKs are possible only for ordinary rowid tables; normalize PK nullability for `table.Strict || table.WithoutRowID`. Add zero-diff and inspection tests for both options, including table-level/composite keys. Otherwise this can report permanent drift and repeated rebuild plans.

denisvmedia:
Thanks — this reproduced exactly as described, and the cited line still said what the review said it said. Fixed at the reviewed head `90a39945`; three commits.

**Reproduced first, on two entry points.** `ptah-compat` built fresh from `90a39945` into its own directory, exit codes read from unpiped invocations, SQLite 3.51.0:

- `schema apply --to file://schema.sql --auto-approve` run twice over the identical file: `WITHOUT ROWID` and `STRICT` planned a full table rebuild on the second run (`CREATE __ptah_rebuild_users` / `INSERT` / `DROP` / `RENAME`); the plain rowid control answered `Schema is synced, no changes to be made.`
- `migrate diff init` then `migrate diff again` over the identical desired file: `WITHOUT ROWID` wrote a *second* migration file carrying that rebuild; the rowid control answered `The migration directory is synced with the desired state, no changes to be made`.

At the merge base `9820496e` all of those converged, so this was the branch's regression rather than a pre-existing defect. Worth recording why the base looked fine: its SQLite reader forced NOT NULL on key columns, so Ptah agreed with itself while disagreeing with the catalog — which is finding 6.3. The pinned community v1.3.0 binary answered `Schemas are synced, no changes to be made.` against the very databases `ptah-compat` had just built, so the disagreement was Ptah's with itself.

**Fixed as suggested in shape, with one measured deviation.** Comparison and HCL rendering are now table-aware; the rule lives once in `internal/sqlitekey` and both callers ask it instead of the dialect.

The deviation is `table.Strict || table.WithoutRowID`. That normalization is wrong for `id INTEGER PRIMARY KEY` in a `STRICT` table, which is still the rowid alias:

```
sqlite3: CREATE TABLE s_int (id INTEGER PRIMARY KEY, a TEXT) STRICT;
         SELECT name,"notnull",pk FROM pragma_table_info('s_int');  ->  id|0|1
         INSERT INTO s_int (id, a) VALUES (NULL, 'x');              ->  accepted, rowid assigned
pinned v1.3.0: schema inspect --url file://strict_int.sql --dev-url sqlite://<clean dev> --format '{{ json . }}'
         ->  "id": {"null": true}
```

I built the blanket rule as a mutant and measured it: it fixes eight of nine shapes and turns `CREATE TABLE users (id INTEGER PRIMARY KEY, name TEXT NOT NULL) STRICT` from a no-op into a permanent rebuild — this PR's own defect wearing a `STRICT` table. What shipped is the rule underneath both halves: **a key column is NOT NULL unless it is the rowid alias, and the leniency that lets a key column hold NULL at all survives only on an ordinary rowid table that is not `STRICT`.** That rule reproduces the pinned binary's model on all ten shapes I measured, including one your list did not mention: `INT PRIMARY KEY` in a `STRICT` table reports `notnull=1`, because the alias is keyed on the declared type name rather than on affinity, so that cell needed the rule too and has a fixture.

**Zero-diff and inspection tests, both options, table-level and composite.**

- `cmd/atlas/schema_sqlite_table_shape_test.go` — ten rows: rowid TEXT/INTEGER/composite, `WITHOUT ROWID` TEXT/INTEGER/composite, `STRICT` TEXT/INT/composite/rowid-alias, plus a `PRIMARY KEY (id DESC)` row. Each applies the DDL, reads the key's nullability back out of `pragma table_info` (rendered is not applied), then applies the identical desired state again and requires `Schema is synced, no changes to be made.` A companion test keeps a genuine change on a `WITHOUT ROWID` table planned, applied, and converging on the next run.
- `cmd/atlas/schema_sqlite_key_shape_test.go` — the existing `{{ json . }}` inspection test gains `WITHOUT ROWID`, `STRICT`, both table-level composites, and the `STRICT` rowid alias.
- `internal/atlashclrender/sqlite_key_nullability_test.go` — the render half at the layer that decides it.
- `internal/sqlitekey` — the shape table plus a white-box guard that renders each declared type through `renderer.RenderSQL` and asserts the rowid-alias type list still matches what the SQLite renderer writes.

**Mutants, each reverted, nothing staged while one was in the tree:** comparator normalization removed → the six `STRICT`/`WITHOUT ROWID` zero-diff rows redden and the four rowid rows stay green; blanket `STRICT` → only the strict-rowid-alias row; every SQLite key column normalized → only the four "stays nullable" rows; HCL writer reverted → only its three shape rows; `BIGINT` added to the alias type list → only the `bigint` correspondence row; reader consulting the key ordinal → the two rowid-alias inspection rows; reader reporting every key column nullable → all seven inspection rows.

One honest note on the third mutant: it does **not** redden the PR's original `TestSchemaApplySQLiteKeyColumnKeepsDeclaredNullability`. Those rows pin the SQLite *renderer*, and that mutant is in the comparator; the comparator's inverse direction is what the four "stays nullable" rows exist for.

**Something I got wrong myself, caught by measuring rather than by review.** My first draft also modeled SQLite's rule that a `DESC` key ordering defeats the alias (measured: `STRICT id INTEGER PRIMARY KEY DESC` → `notnull=1`). But Ptah's SQLite renderer drops `DESC` from a `PRIMARY KEY`, so `PRIMARY KEY (id DESC) ... STRICT` is applied as `PRIMARY KEY ("id")` with `notnull=0` — and my model then planned a rebuild on every run for a shape that converged at `90a39945`. Removed in the third commit, with a CLI fixture pinning it; the dropped `DESC` is filed as the renderer's defect rather than modeled around.

**What the PR already claimed, re-measured with the fix.** 5.1: applied `"id" integer NOT NULL PRIMARY KEY`, and the pinned binary's `schema diff` from that database to the same file answers `Schemas are synced, no changes to be made.` at exit 0. 5.2: the rendered SQL replays to exactly the source's three indexes. 6.3: `"null":true` on the rowid alias. PostgreSQL: widening a single-column key `integer` → `bigint` on a container of my own plans `SET NOT NULL`, applies at exit 0, `attnotnull` stays `t`, and the whole run's output diffs empty against the previous head — nothing outside SQLite moved.

**Not verified, stated rather than implied:** MySQL/MariaDB not re-measured live (the predicate returns false before looking at anything else for non-SQLite, and PostgreSQL is byte-identical); SQL Server and ClickHouse by inspection; and I could not produce a *wrong HCL document* today — every SQLite-dialect render path is fed by the reader, which is already right — so the HCL-writer change is a guard pinned by a unit test rather than a reproduced defect.

Two things found next door and **not** folded in: `schema inspect --format '{{ json . }}'` lists each implicit unique index twice (present at `9820496e`, at `90a39945` and after the fix), and the SQLite `PRIMARY KEY` renderer drops `DESC`. Both want their own issue.

All gates green at the final tree; `docs/conformance.md` and the `SetPrimary` doc comment are corrected where they read as "SQLite never does".

Issue #1280, opened 2026-08-07
compat inspect pins an extension to schemas that do not use it, because contrib extensions overload core functions


denisvmedia:
`ptah-compat schema inspect` omits an extension block when nothing in the document needs it, because the Atlas community CLI refuses any PostgreSQL schema file that declares one. #1274 made the "does anything need it" test ask what the extension *provides* rather than what it is *called*, which fixed `isn`/`isbn`.

The provides set is taken from `pg_depend` unfiltered. Contrib extensions overwhelmingly supply **overloads of core functions**, so those member lists are full of ordinary words. Measured on PostgreSQL 17.10, `citext` supplies fifteen names `pg_catalog` also supplies:

```
max  min  replace  split_part  strpos  translate
regexp_match  regexp_matches  regexp_replace
regexp_split_to_array  regexp_split_to_table
texticlike  texticnlike  texticregexeq  texticregexne
```

A document that says `max` or `strpos` therefore pins `citext`, although `pg_catalog` is always on the search path and the document resolves with `citext` dropped.

### Measured

Four databases on PostgreSQL 17.10, each created fresh, every object verified through the catalog, in none of which anything depends on the extension. `base` is 3acddbdf (parent of #1274), `head` is 4c4a8a1d. `ext` counts `extension` blocks in `ptah-compat schema inspect` output; `orc` is the pinned Atlas community binary v1.3.0 reading that output against a separate clean dev database.

| fixture | ext base | ext head | orc base | orc head |
|---|---|---|---|---|
| `citext`, nothing referencing it | 0 | 0 | 0 | 0 |
| `citext` + `CHECK (strpos(label,'x') = 0)` | 0 | **1** | 0 | **1** |
| `citext` + `VIEW v AS SELECT max(n) FROM t` | 0 | **1** | 0 | **1** |
| `pgcrypto` + `DEFAULT gen_random_uuid()` | 0 | **1** | 0 | **1** |

The failure is always `Error: postgres: extensions are not supported by this version.`

`gen_random_uuid` is the sharpest row: `pg_catalog` has supplied it since PostgreSQL 13, and the code comment in `documentNamesAny` names this exact case and calls it "the harmless direction".

### It is not the harmless direction

[...]

PR #1309, opened 2026-08-08
Resolve an index's operator class to the extension behind it (#1286)

### LINKED ISSUE #1286: compat inspect omits an extension whose dependency no identifier in the document names
A third way the extension-suppression feature loses a dependency, distinct from #1280 (members `pg_catalog` also supplies) and #1281 (members whose names are SQL keywords). This one is not a naming problem at all: **no identifier in the document names the dependency.**

Found by installing all 45 extensions available in the `postgres:17` image and walking what each supplies, while checking whether #1282's narrowing could have broken an adjacent shape. It could not — see the bottom — but this turned up instead.

### Measured, PostgreSQL 17.10

```sql
CREATE EXTENSION btree_gin;
CREATE TABLE t (id integer PRIMARY KEY, n integer NOT NULL);
CREATE INDEX t_gin ON t USING gin (n int4_ops);
```

The catalog stores the index as

```
CREATE INDEX t_gin ON public.t USING gin (n)
```

— PostgreSQL omits the operator class because it is the default for that type on that access method. So the rendered document says `USING gin (n)` and contains no token belonging to `btree_gin`.

`ptah-compat schema inspect` therefore omits the extension, at exit 0, with the ordinary "nothing names it" diagnostic. The resulting document is a hole. The pinned Atlas community binary v1.3.0, reading it against a clean dev database:

```
Error: create index "t_gin" to table: "t": pq: data type integer has no default operator class for access method "gin" (42704)
```

The dependency is real and unnameable: `integer` has a `gin` opclass only because `btree_gin` supplies one. The document's `USING gin` is the reference, and it is a reference to an access method, not to any extension member.

**This is pre-existing, not a regression.** Both 3acddbdf (before #1274) and current master omit the extension on this fixture, for the same reason: even the unfiltered `Provides` set is matched against identifiers, and no identifier is present.

### Why the obvious fix is wrong

`gin` is a `pg_am` row, not an extension member, so adding access methods to `Provides` would pin `btree_gin` to every GIN index in the database including the many that do not need it — `tsvector`, `jsonb` and array columns all have core GIN opclasses. The question the renderer has to answer is narrower and it is a catalog question, not a token question: *does this index's (type, access method) pair resolve to an opclass this extension supplies?*

`pg_index.indclass` holds the resolved opclass OIDs per key column, so the reader can answer it exactly — for each index key, which opclass was actually chosen, and which extension owns it. That is a dependency edge the document does not have to spell.

### Scope note for whoever takes this

The same shape applies wherever the catalog resolves something implicitly and the rendered DDL then omits it. Worth checking at least: default operator classes on every access method, implicit casts supplied by an extension, and collations. Enumerate before fixing, for the reason #1280 gives.

### What this probe cleared

The measurement that produced this also answers whether #1282's `pg_catalog` filter can drop a name a document could legitimately need. Across all 45 extensions in the image:

- extension operator classes whose name `pg_catalog` also supplies: many (`btree_gin` alone supplies 32, `bloom` 2) — but **0** of them are non-default, and PostgreSQL only prints a non-default opclass. A name that is never printed cannot be matched, so filtering it changes nothing.
- extension types whose name `pg_catalog` also supplies: **0**.
- extension relations and operator families: same, all default-only.

So that filter bites only on functions, which is where it was aimed. `citext` loses 24 function names to it, `pgcrypto` 1 (`gen_random_uuid`), `intarray` 1, `plpgsql` 3.

denisvmedia:
Closes #1286.

`ptah-compat schema inspect` omitted an extension whose dependency no identifier in the document names, and the pinned Atlas community binary v1.3.0 then refused the result.

## What was wrong

PostgreSQL prints an operator class only when it is not the default for the key's type on the index's access method. Measured on PostgreSQL 17.10:

```sql
CREATE EXTENSION btree_gin;
CREATE TABLE t (id integer PRIMARY KEY, n integer NOT NULL);
CREATE INDEX t_gin ON t USING gin (n int4_ops);
```

is stored, and rendered back, as `CREATE INDEX t_gin ON public.t USING gin (n)`. The document therefore carries no token of `btree_gin` — not its label, and not one of the 87 support functions its member list holds, none of which anything ever writes. The block was omitted at exit 0 and the pinned binary answered

```
Error: create index "t_gin" to table: "t": pq: data type integer has no default operator class for access method "gin" (42704)
```

while `ptah-compat schema apply` of the same document failed identically. Pre-existing on 3acddbdf and on master.

## Why the obvious fix is wrong

`gin` is a `pg_am` row belonging to no extension, so reading `USING gin` as a reference to `btree_gin` would pin it to every GIN index in the database — and `tsvector`, `jsonb` and array columns all have core GIN operator classes. The question is a catalog question, not a token question.

`pg_index.indclass` holds the class each key actually resolved to. The reader now resolves those OIDs — and the index's access-method OID from `pg_class.relam` — against `pg_depend`, and records the owning extensions on the index. The access-method arm is not the rejected fix: it resolves an OID, so it pins nothing for core `gin` and pins the owner for an access method an extension does supply, such as bloom's. The renderer keeps an extension block when a surviving index or constraint resolves to it, and says so in a diagnostic of its own, because a dependency the document does not spell cannot be reported as if it were named.

The edge is carried on the index rather than on the extension because `--exclude` filters index rows by name: an excluded index has to take its requirement with it.

## The same shape through a constraint

An exclusion constraint prints its operators and never its operator classes, so `EXCLUDE USING gist (room WITH =, during WITH &&)` over an `integer` column needs `btree_gist` and names nothing of it. Measured on the same server, that document failed to apply for the same reason. The pinned binary reads it at exit 0 only by silently dropping the constraint, which is a loss rather than a verdict. The backing index is not in the entity model, so the requirement travels on the constraint.

## Measured

`ptah-compat schema apply` onto a freshly created database, each result verified in the catalog afterwards:

| fixture | block | before | after |
| --- | --- | --- | --- |
| GIN over `integer` (btree_gin) | kept now | rc=1 | rc=0 |
| `EXCLUDE USING gist` over `integer` (btree_gist) | kept now | rc=1 | rc=0 |
| GIN over `jsonb` and `tsvector` | still omitted | rc=0 | rc=0 |

The third row is the control and it is the point: the same extension, the same access method, the same rendered `type = "gin"`, differing only in what the catalog resolved. Its document is byte-identical to the one master emits and is still read by the pinned binary at exit 0.

A kept block makes the document unreadable to that binary (`postgres: extensions are not supported by this version`). That is the trade #1266 and #1274 already settled: before this change it refused the same documents anyway, with a data error instead of a feature refusal, and neither Ptah nor it could apply them.

## Enumerated, not fixed

The issue asks for the adjacent implicit resolutions. Over the 45 extensions in the `postgres:17` image, documented in `docs/site/src/content/docs/atlas/schema-commands.md`:

- **Implicit casts.** All 21 extension-supplied implicit or assignment casts have their own extension's type on one side (`citext`, `isbn` and its siblings), so naming that type keeps the extension. A cast between two core types would be a gap; none exists.
- **Collations.** No extension in the image supplies a `pg_collation` row at all.
- **Primary keys and single-column unique constraints.** Backed by an index too, but the operator class is the default for the column's own type, and the extension-supplied default classes on core types are all `gin`, `gist` or `bloom` classes — none `btree`. The reader records the edge for those indexes; only the converter drops it, so closing this later is a converter change.

## Tests

- Reader white-box rows for both arms, keyed on the catalog column each one reads, plus a `jsonb` control that differs from the `integer` row only in what the catalog answered.
- Renderer rows for the index and the constraint, and a control that keeps the GIN index and loses the requirement.
- A converter row pinning that a constraint-backed index's requirement lands on the constraint.
- Three live e2e rows in `integration/atlas_compat_inspect_extension_roundtrip_e2e_test.go` (17 rows now, all run green on PostgreSQL 17.10), each materializing the rendered document on a fresh dev database.

Six mutants were each observed red and reverted, including the over-correction (keep every extension: five renderer tests red, and the live control document refused by the pinned binary) and the issue's rejected fix simulated in full (access-method name in `Provides` plus scanning the index's access method: the `jsonb` control red).

## Portability

The amended queries run at exit 0 on CockroachDB v26.2.5 and YugabyteDB 2026.1.0.0-b118, which CI reads with this same reader, and `schema inspect` against each renders byte-identically to master.

---

### What it found

Every row they measured reproduces on my own PostgreSQL 17.10 container (ptahrefut1286, port 15877, removed at the end) with fixtures I built, and I added rows they never ran live (bloom access method, pg_trgm printed class, PostgreSQL 18.4, CockroachDB v26.2.5). All gates pass on eda2209f. But the change regresses an adjacent shape its author never built, and the regression is exactly the failure mode their own code comment names.

REGRESSION — an index the document does not contain still keeps the block.
collectRequiredExtensions (internal/atlashclrender/atlas_refused_blocks.go:279) reads RequiresExtensions off EVERY row of db.Indexes and db.Constraints. renderTables (internal/atlashclrender/render.go:337) drops an index whose table is not among the rendered tables, calls it an orphan and reports it. A materialized view's index is exactly that: the reader resolves its operator classes from pg_index, and a `materialized` block carries no index, so it can never be rendered.

  docker run -d --name ptahrefut1286 -e POSTGRES_PASSWORD=pw -p 15877:5432 postgres:17
  CREATE ROLE ptah_user LOGIN PASSWORD 'ptah_password' SUPERUSER;  CREATE DATABASE ref_f OWNER ptah_user;
  psql -U ptah_user -d ref_f -v ON_ERROR_STOP=1 \
    -c "CREATE EXTENSION btree_gin;" \
    -c "CREATE TABLE src (id integer PRIMARY KEY, n integer NOT NULL);" \
    -c "CREATE MATERIALIZED VIEW mv AS SELECT id, n FROM src;" \
    -c "CREATE INDEX mv_gin ON mv USING gin (n int4_ops);"
  ptah-compat schema inspect --url 'postgres://ptah_user:ptah_password@localhost:15877/ref_f?sslmode=disable&search_path=public'

9820496e: document has no index block anywhere; no extension block; oracle
  /Users/buster/Work/denis/ptah-atlas-conformance/bin/atlas schema inspect --url file://f_master.hcl --dev-url <clean db>
reads it at rc=0, and ptah-compat schema apply onto a clean database is rc=0.

eda2209f: same document plus `extension "btree_gin" { ... }`. Same oracle invocation is rc=1, `Error: postgres: extensions are not supported by this version`. ptah-compat apply is still rc=0. So the block bought nothing and cost the file the only thing the suppression exists to produce. That is verbatim the false positive documentNamesAny warns about ("a spurious match costs the entire result this suppression exists to produce, on a schema with no relationship to the extension", stokaro/ptah#1280).

The renderer contradicts itself on one stderr stream, two lines apart:
  warning: extensions.btree_gin: kept ... because an index or constraint in this document resolves to an operator class or access method it supplies ...
  warning: index mv_gin: index cannot be rendered because the target table is absent from the exported schema

SECOND FINDING (unfixed) — a diagnostic that is false where the class IS printed. documentRequiresExtension is asked before documentNamesAny, so an extension reached through a PRINTED, non-default operator class gets the catalog wording. With `CREATE EXTENSION pg_trgm; CREATE INDEX w_trgm ON w USING gin (txt gin_trgm_ops)` the HCL contains `ops = "gin_trgm_ops"`, and eda2209f still reports "... which the rendered DDL does not spell". (Keeping the block is right here — 9820496e omitted it and apply failed rc=1 `operator class "gin_trgm_ops" does not exist for access method "gin"` — only the sentence is wrong.)

THIRD FINDING — the doc comment on collectRequiredExtensions claims "the surviving document's objects" and "An object dropped by a selector takes its requirement with it". The selector half is true (I measured `--exclude t` on fixture A: block correctly omitted). The first half is false for orphans, which is the case above.

CONFIRMED, re-measured by me, fixtures built from scratch: reproduction on 9820496e (indexdef stored as `CREATE INDEX t_gin ON public.t USING gin (n)`; inspect rc=0 with `extensions.btree_gin: omitted`; oracle rc=1 `create index "t_gin" to table: "t": pq: data type integer has no default operator class for access method "gin" (42704)`; ptah-compat apply rc=1 same 42704). Branch: btree_gin kept, apply rc=1 -> rc=0, index present and btree_gin installed; EXCLUDE gist over integer likewise rc=1 -> rc=0 with btree_gist installed; jsonb+tsvector control byte-identical to 9820496e (diff rc=0) and read by the oracle at rc=0. Both kept-block documents are refused by the oracle with `extensions are not supported by this version` — the settled #1266/#1274 trade. Applying the branch's document twice is idempotent ("Schema is synced, no changes to be made"). --format json output is byte-identical between commits; native `ptah db read` output is byte-identical.

BEYOND THEIR ROWS: bloom (extension-supplied access method, the pg_am arm, never exercised live by them) — 9820496e apply rc=1 `access method "bloom" does not exist`, eda2209f keeps the block and applies rc=0. PostgreSQL 18.4 (their explicit unverified claim) — btree_gin kept, same diagnostic. CockroachDB v26.2.5 — the amended queries run, and inspect is byte-identical to 9820496e (diff rc=0). Catalog survey reproduced exactly: 45 available extensions; extension-supplied DEFAULT opclasses on core types = 58 rows from bloom(2)/btree_gin(29)/btree_gist(26)/intarray(1), all gin/gist/bloom, none btree; 21 extension casts at implicit/assignment, 0 of them with a core type on both sides; 0 pg_collation extension members. Connecting role: identical result as ptah_user and as the bootstrap postgres superuser; a plain non-superuser fails identically on BOTH commits (`permission denied for table pg_authid`), so no regression there.

GATES on eda2209f, each exit code on its own line:
go build ./... = 0
go vet ./... = 0
go vet -tags integration ./integration/... = 0
go test ./... -count=1 = 0 first run, 254 packages ok (no re-run needed)
go tool qtlint ./... = 0
golangci-lint run ./... = 0, "0 issues." (needed a private TMPDIR as well as GOLANGCI_LINT_CACHE — golangci-lint's "parallel golangci-lint is running" lock lives in os.TempDir(), not the cache dir)
scripts/

### What it says must change

Fixed in my worktree at /Users/buster/Work/denis/ptah/.claude/worktrees/wf_ff3edb74-a9c-6, commit 87dde52f on top of eda2209f (not pushed, no PR opened).

collectRequiredExtensions now asks the same question renderTables asks, through the same resolveTable, and skips an object whose table this render does not write:

  for _, index := range db.Indexes {
      if resolveTable(db.Tables, index.StructName, index.TableName) == nil { continue }
      add(index.RequiresExtensions)
  }
  for _, constraint := range db.Constraints {
      if resolveTable(db.Tables, constraint.StructName, constraint.Table) == nil { continue }
      add(constraint.RequiresExtensions)
  }

Plus TestRenderInspectedForAtlasCLIOmitsAnExtensionOnlyAnOrphanObjectResolvesTo with one row per arm (orphan index via a materialized view, orphan constraint), and the doc comment corrected to say what the collector actually counts.

Seen red in both directions, then restored:
- drop the guard (= eda2209f's behavior): RED exactly the two new rows, everything else green.
- invert the guard (count only orphans, the over-correction): RED the two new rows AND both rows of TestRenderInspectedForAtlasCLIKeepsAnExtensionOnlyTheCatalogNames.

Re-measured live on the fixed binary: the matview document is byte-identical to 9820496e's (diff rc=0, once the cluster-scoped role I had added was present for both runs) and the oracle reads it at rc=0; fixture A still keeps btree_gin, fixture B still keeps btree_gist, bloom still keeps bloom, and the jsonb/tsvector control still omits the block.

Gates on the fixed tree: go build = 0, go vet = 0, go vet -tags integration = 0, qtlint = 0, golangci-lint = 0 ("0 issues."), check-test-style.sh = 0, check-public-api.sh = 0, check-public-api-snapshot.sh = 0, check-public-api-released.sh = 0, docs check-style = 0, live e2e = 0. go test ./... -count=1 = 1 on the sweep, from machine load only: the single failure was cmd/viz TestSVGReportsGraphvizStderrOnFailure at exactly 10.00s with `render SVG with Graphviz dot: signal: killed`; re-run serially `go test ./cmd/viz/ -count=1 -p 1` = ok in 6.3s. Nothing in cmd/viz is touched by either commit.

STILL OPEN, deliberately not fixed here (it needs its own commit and a docs edit, because docs/site/src/content/docs/atlas/schema-commands.md quotes the string verbatim): the "which the rendered DDL does not spell" clause is false whenever the extension is reached through a PRINTED non-default operator class such as pg_trgm's gin_trgm_ops. Either drop the clause, or ask documentNamesAny first and fall through to the catalog wording only when the name scan found nothing.

READY-TO-POST PR

Title:
Count only an index this document actually writes (#1286)

Body:
`ptah-compat schema inspect` keeps an `extension` block when a surviving index or exclusion constraint resolves to it through `pg_depend` (#1286). `collectRequiredExtensions` reads that edge off every row of `db.Indexes` and `db.Constraints`, including the ones `renderTables` drops as orphans and reports as unrenderable. An orphan writes nothing into the document, so the block is kept for an index the file does not contain — and the pinned Atlas community binary v1.3.0 refuses ANY postgres schema file that declares an `extension` block, so the whole document goes with it.

The shape is ordinary rather than hypothetical. A materialized view's index is read from `pg_index` with its resolved operator classes and can never be rendered, because a `materialized` block carries no index. Measured on PostgreSQL 17.10, `ptah-compat schema inspect` of

```sql
CREATE EXTENSION btree_gin;
CREATE TABLE src (id integer PRIMARY KEY, n integer NOT NULL);
CREATE MATERIALIZED VIEW mv AS SELECT id, n FROM src;
CREATE INDEX mv_gin ON mv USING gin (n int4_ops);
```

emits a document with no index block anywhere, and said so itself, on the same stderr, one line below the keep:

```
warning: extensions.btree_gin: kept ... because an index or constraint in this document resolve

### CODE REVIEW

**Comment by denisvmedia** [comment]

**Second verdict: SURVIVED.** An independent reviewer re-measured the corrected branch on its own container and fixtures, built the shape the first reviewer found, checked the original defect still holds, and hunted the next adjacent shape — and did not refute it. This is the first front this session to survive two rounds of independent adversarial verification. Merge candidate once CI is green.

**Review by denisvmedia — COMMENTED** [review]
Independent review of head `3ba22685` confirms an unresolved user-facing defect.

**Major — the warning says an explicitly rendered operator class was omitted.** `internal/atlashclrender/atlas_refused_blocks.go:193` treats every extension-owned operator class/access method as “unspelled,” while `internal/atlashclrender/render.go:701,723` does emit explicit `index.Operator` and `part.Operator` values. An index using `gin_trgm_ops` can therefore render that exact class and still warn that the DDL does not spell it.

Either make the diagnostic truthful and generic (“depends on an operator class or access method the extension supplies”) or model implicit/unspelled dependencies separately. Update the public field comment and user documentation, and add an explicit-opclass regression test. The PR body admits this current-head case, so this should not merge yet.

**Comment by denisvmedia** [comment]
Thanks — reproduced, and it is worse than the one arm you cited. Fixed in c888a36d on this branch.

**"The warning says an explicitly rendered operator class was omitted." CONFIRMED, and it reproduces on both rendering sites, not one.**

You cited `atlas_refused_blocks.go:193` against `render.go:701,723`. At head 3ba22685 that line still said what your review says it said, and here is the run:

```
CREATE EXTENSION pg_trgm;
CREATE TABLE w (id integer PRIMARY KEY, txt text NOT NULL);
CREATE INDEX w_trgm ON w USING gin (txt gin_trgm_ops);
```

`ptah-compat schema inspect` exits 0, the document contains

```hcl
  index "w_trgm" {
    type = "gin"
    on {
      column = column.txt
      ops    = "gin_trgm_ops"
    }
  }
```

and stderr on the same run says `... resolves to an operator class or access method it supplies, **which the rendered DDL does not spell** ...`. Measured on my own PostgreSQL 17.10 container (ptah1309pr, port 15931, CI-shaped `ptah_user LOGIN SUPERUSER`, removed at the end).

The constraint arm has it too, which your review did not have to reach: `pg_get_constraintdef` prints a non-default class as well, so `EXCLUDE USING gist (txt gist_trgm_ops WITH =)` renders `elements = "txt gist_trgm_ops WITH ="` and got the same sentence.

**What I changed — your second option, because your first would have thrown away something true.**

You offered "make it generic" or "model implicit/unspelled dependencies separately". Generic wording is unconditionally true but loses the one thing the clause was for: telling an operator not to go searching. So the branches are now separated by asking the name question first, and the catalog branch states its own evidence rather than the document's:

- names it: `kept ... because another object in this document depends on it`
- catalog only: `kept ... because the catalog resolved an index or constraint in this document to an operator class or access method it supplies`

The second makes no claim about spelling at all, so it cannot go false again.

**A second defect fell out of reproducing yours, and it was a user-visible one.** The ordering swap alone does not fix the index arm, because `collectReferencedNames` never read the attribute the renderer writes the class into. It reads `constraint.ExcludeElements` but not `index.Operator` or `part.Operator`. Measured on the PR base 9820496e, where only the name scan exists:

- printed class on a **constraint**: pg_trgm kept, `another object in this document depends on it`, materialize rc=0
- printed class on an **index**: pg_trgm **omitted**, and materializing the document is rc=1 `operator class "gin_trgm_ops" does not exist for access method "gin"`

So the two arms answered differently for the same shape. The scan now reads `ops` on the index and on each part. Mutant 2 below is exactly this line: removing it reddens the new index row and leaves the constraint row green.

The ordering is sound rather than lucky, and I measured the premise instead of assuming it. A class reaches the document exactly when it is non-default; installing all 45 rows of `pg_available_extensions` into one database gives 93 extension-owned operator classes, 88 default and 5 not — `citext_pattern_ops`, `gin__int_ops`, `gist__intbig_ops`, `gin_trgm_ops`, `gist_trgm_ops` — and the count of those whose name `pg_catalog` also supplies is 0. So every printable class survives the shadowed-name filter into `Extension.Provides` and the name scan can answer for all of them.

**"Update the public field comment and user documentation." Done, and it was more than one comment.** Your finding invalidated a claim repeated in eight places: `goschema.Index.RequiresExtensions` ("and does not name"), `goschema.Constraint.RequiresExtensions` ("that nothing in the constraint's own text names"), `dbschema/types.DBConstraint.RequiresExtensions` ("as unnameable here as it is on a plain index"), `DBIndex.RequiresExtensions`, the reader's `requiredExtensionsProjection` doc — which additionally claimed "a printed, non-default class is already matched by name through `Extension.Provides`", false precisely because the scan did not read `ops` — its two in-query `--` comments, the converter comment, and `docs/site/.../schema-commands.md`, which quoted the false warning **verbatim**. `docs/public_api.snapshot` is regenerated. The feature-matrix evidence row carries the printed-class measurement now.

**"Add an explicit-opclass regression test." Done, four of them.** `TestRenderInspectedForAtlasCLIReportsAPrintedOperatorClassAsAName` has a row per arm and asserts three things: the document contains the class token, the message is the one the reader can act on, and the message contains neither `the catalog resolved` nor `does not spell`. Two live rows in `integration/atlas_compat_inspect_extension_roundtrip_e2e_test.go` materialize each document on a fresh dev database — 19 rows, all green on PostgreSQL 17.10.

Five mutants, each seen red and reverted:

1. ordering reverted — RED exactly the two new rows, everything else green
2. `ops` unread — RED only the new index row; the constraint row stays green, which is the asymmetry above
3. catalog arm dead (over-correction, names only) — RED both `KeepsAnExtensionOnlyTheCatalogNames` rows, new rows green
4. name arm dead (over-correction, catalog only) — RED three `KeepsABlockTheDocumentNames` rows and both new rows
5. the old clause restored verbatim, then also the ordering (= 3ba22685 exactly) — RED both new rows, quoting your sentence back in the failure output

**Nothing the PR already claimed was traded away, and I can be precise about it: the rendered document is byte-identical to 3ba22685's on all six fixtures.** btree_gin keep, btree_gist keep, jsonb/tsvector control, materialized-view orphan, and both pg_trgm shapes. Only the stderr reason changes. The pinned binary (`atlas community version v1.3.0`) still reads the control at rc=0 and still refuses a kept document with `postgres: extensions are not supported by this version` — the settled #1266/#1274 trade, unmoved.

Every gate is 0. One caveat I will not paper over: `docs/site check:responsive` exits 0 while printing `SKIPPED (playwright is not installed)`, and it refused to run at all until I borrowed the main checkout's `node_modules` to build `dist`. Its 0 is a skip, not a pass. Nothing in this commit touches layout or CSS.

Two things I deliberately did not absorb, filed rather than committed: the diagnostic still does not name **which** class was resolved (that needs a new public field, a reader change and a snapshot bump), and the reference scan still cannot see the rendered access method `type = "..."`, so an extension-supplied access method can only ever be kept by the catalog arm.

Issue #1086, opened 2026-08-03
migrate new and migrate diff write over a directory atlas.sum never verified


denisvmedia:
## Summary

`ptah-compat migrate new` and `migrate diff` write a migration file and a fresh `atlas.sum` over a directory whose previous contents nothing verified. The pinned community binary refuses:

```
$ atlas migrate new demo --dir file://d       # d holds one migration, no atlas.sum
Error: You have a checksum error in your migration directory.
Please check your migration files and run 'atlas migrate hash' to re-hash the contents
[exit 1]

$ ptah-compat migrate new demo --dir file://d
[exit 0, writes 20260803xxxxxx_demo.sql and atlas.sum]
```

`migrate apply`, `status`, `set`, `hash` and `validate` all run the gate already (#973, #974, #982). These two are what is left.

## Why it needs its own change

These are **forwarded** verbs — the compat adapter maps them onto native commands and there is no hand-written `RunE` to put the gate in front of. Adding it means either a pre-run hook on `atlasVerb` or resolving the source in the mapper, which is a structural change to the adapter rather than a call added to a function.

## It currently blocks a relaxation

`--dir` query parameters (`?format=`, and unknown keys the community binary ignores) are accepted on every verb that runs the gate. `new` and `diff` deliberately still refuse a query, because relaxing there would turn a refusal into a write over an unverified directory — trading one divergence for a worse one.

`TestCompatMigrateDirQuery_NewAndDiffStillRefuseAQuery` pins that, so the relaxation cannot be extended to these two by symmetry without the gate arriving with it.

## Definition of done

- Both verbs run the `atlas.sum` integrity gate before writing anything, with the same exemptions the other verbs use (empty directory, non-SQL directory).
- The refusal precedes creating or writing any file — a gate that fires after the write still leaves the mess.
- The `--dir` query relaxation extends to both, and the pinning test above is replaced by rows in the shared table.
- Fixtures cover unhashed, tampered, and clean directories on both verbs. A fixture that only checks the clean case passes on the broken binary.

PR #1310, opened 2026-08-08
Name the --dir query keys a run ignored instead of dropping them silently (#1013)

denisvmedia:
Re-measured [#1013](https://github.com/stokaro/ptah/issues/1013) end to end against the pinned community binary v1.3.0 before changing anything. Most of the issue has closed since it was written. One cell survives, and **this PR does not close it** — it keeps its refusal, records why, and closes the half of the issue nobody had picked up.

The issue body itself asked for that half: *"a typo'd `?format=atals` is silently ignored by CE and the directory is read as the default format ... we should match it and consider whether to warn."*

## Re-measurement, 2026-08-08, on `origin/master` 9820496e

Fixtures, each row on a fresh `cp -R`, every exit code from an unpiped invocation, oracle run by absolute path. `m/` is a hashed native Atlas directory; `gm/` is a golang-migrate pair whose `atlas.sum` covers only the `.up.sql`, so read as native Atlas it is a checksum mismatch. Both were written by the pinned binary itself.

### `?format=golang-migrate` on `gm/`, with the no-query control arm

| verb | community | `ptah-compat` | control CE | control compat | |
| `validate` | 0 | 0 | 1 | 1 | closed |
| `lint` | 0 | 0 | 1 | 1 | closed — #1133 (`20c30b7e`) |
| `status` | 0 | 0 | 1 | 1 | closed — #1002 |
| `set` | 0 | 0 | — | — | closed — #1002 |
| `apply` | 0 | 0 | — | — | closed — #1087 (`53bf001c`) |
| `new` | 0, WROTE | 0, WROTE | 1, unchanged | 1, unchanged | closed — #845 |
| **`diff`** | **0, WROTE** | **1, unchanged** | 1, unchanged | 1, unchanged | **survives** |

The control arm is what makes those rows evidence rather than an exit 0 that could have happened anyway: with no query both tools exit 1 with the identical `checksum mismatch (atlas.sum): L2: ..._init.down.sql was added`.

### `?nonsense=1` on `m/`

Community: `validate 0 lint 0 status 0 set 0 apply 0 new 0 diff 0`. `ptah-compat`: identical. Closed by #1087 (`53bf001c`) for the six reading verbs and #1135 (`30ce8c9a`) for `new` and `diff`.

### The value is still validated, and precedence still holds

`migrate diff` over `m/`: `?format=totally-bogus` exits 1 on both; `?format=atlas` exits 0 on both; `?format=` beside `--dir-format golang-migrate` exits 0 on both, i.e. the query still outranks the flag.

## What this changes

`format` is the only `--dir` query key that selects anything. Every other key was dropped **in silence**. It is now named on standard error and the run continues:

```
note: atlas migrate apply --dir: ignoring migration directory URL query key "fromat".
Only ?format= selects the directory layout. Set PTAH_STRICT_DIR_QUERY=1 to refuse an
unrecognized key instead.
```

Exit codes and standard output are untouched, so a script reading either sees exactly what the community binary produces. The note goes to stderr because several of these verbs carry a machine-readable `--format` document on stdout and that binary emits no field for this — putting it there would invent one and corrupt a caller's parse. `TestCompatMigrateDirQuery_ReportKeepsMachineReadableStdoutClean` pins that with `--format '{{ len .Pending }}'`, where stdout is asserted to be exactly `1`.

**Why not just match and say nothing.** Accepting everything is the wrong answer in the other direction. A misspelled `?fromat=goose` selects no layout on either tool, so the directory is read in the native Atlas layout while the operator believes it is being read as Goose. The key they typed is the only evidence they expected something to happen. Matching costs the exit code, not the sentence.

**`PTAH_STRICT_DIR_QUERY=1`** turns the note into a refusal. Ptah refused every `--dir` query on every verb until #1087 and #1135; that refusal caught exactly this on its way out, and reaching parity means moving it off the default path rather than deleting it — compatibility never removes a capability. It is an environment variable and not a flag because the pinned binary registers none and the conformance `cli-surface` tier asserts flag parity against it; same reason `PTAH_ALLOW_EXTERNAL_SCHEMA` and `PTAH_SKIP_CHECKS` are variables. An invalid boolean is a hard error rather than a silent "off".

## The keys that keep their meaning, enumerated before relaxing anything

On `--dir` the list is exactly one: **`format`**, read by `atlasmigrate.applyDirURLFormat` — an empty value selects the native Atlas layout, the first occurrence wins, and the query outranks `--dir-format`. A **repeated** `format` is deliberately *not* reported as ignored: that key did select the layout, and what a second occurrence loses is a value, under the first-one-wins rule #990 measured. Listing it would tell the operator the opposite of what happened.

Controls proving it still takes effect, on the patched binary:

```
lint gm '?format=golang-migrate'                  0    | no query (CONTROL)  1
lint gm '?format=golang-migrate&nonsense=1'       0, note printed, layout honored
validate gm '?nonsense=1&format=golang-migrate'   0, note printed, layout honored
validate gm '?nonsense=1'          (CONTROL)      1, checksum mismatch
```

Database-URL query keys are a different flag and are untouched by this change. Controlled anyway, on a two-schema PostgreSQL 17 database: `?search_path=extra` inspects `extra.e` and `?search_path=public` inspects `public.p`; `?sslmode=require` exits 1 against a non-TLS server while `?sslmode=disable` exits 0. Note `options=-c search_path=…` and the `search_path` query parameter are **not** interchangeable and the difference is observable — the recorded precedent lives on the apply-time realm gate and the lint review scope, where `internal/schemaselection` reads the URL key only.

## What is deliberately not closed

`migrate diff` with a foreign `?format=` stays exit 1. Re-measured into a fresh empty directory, the community binary exits 0 on **all five** layouts and writes reverse SQL in three different shapes:

```
golang-migrate  20…_demo.up.sql + 20…_demo.down.sql
flyway          V20…__demo.sql  + U20…__demo.sql
goose           one file: "-- +goose Up" … "-- +goose Down" …
dbmate          one file: "-- migrate:up" … "-- migrate:down" …
liquibase       one file, one "--rollback: DROP TABLE …" line PER forward statement
```

Ptah's `migrate diff` plans forward statements only, so honoring the query would write a directory whose rollback half is missing — and liquibase makes a whole-file down block insufficient, because the reverse has to attach to each planned change. Refusing is the strict side and never exits 0 where the community binary exits 1. Two stale comments are corrected here while they were in view: the header of `migrate_dir_source.go` still blamed the missing `atlas.sum` gate (#1086, closed) and still claimed `migrate new` shared the refusal (it has honored `?format=goose` since #845).

## Mutants, both directions

Each applied to the committed source, run, and restored.

- **Remove the report** (return before naming any key) → red: `_ReportsIgnoredKeyOnEveryVerb` (8 rows), `_ReportKeepsMachineReadableStdoutClean`, `_ForeignFormatStillSelectsTheLayoutBesideAnIgnoredKey`, both `_StrictEnv…` rows. Green and unmoved: every parity control.
- **Over-correct to refuse regardless of the variable** → red: the parity guard `_IgnoresUnknownKeysOnEveryVerb` and `_NewAndDiffRefuseAnUnhashedDirectoryWithAQuery`. That is compatibility rule (a) firing exactly where it should.
- **Over-correct `IgnoredDirQueryKeys` to include `format`** → red: `TestIgnoredDirQueryKeys`, `_ReportsNothingWhenEveryKeyIsMeaningful`, `_StrictEnvKeepsAMeaningfulQueryWorking`.

## Documentation

Feature-matrix row (data plus regenerated page), the `PTAH_*` variable list in the compat overview, the shared query-rules section of the command reference, the migrate-commands page, and the comparison page, whose paragraph still described the pre-#845 state. All documentation checks pass.

## One measurement that could not be taken

This environment's shell guard refuses any command containing the `migrate hash` verb token — `echo hash` is refused with the same message — so the two `hash` rows above were not run against the binaries from a shell. They are pinned in-process by the Go tests instead, and the fixtures were seeded with the pinned binary's own `migrate diff` rather than by hashing. Recorded rather than worked around.

Refs #990, #1002, #1086, #1087, #1133, #1135, #1237.

---

### What it found

Re-measured every row independently: my worktree at 917fbc15, a control worktree at the merge-base 9820496e (`git log --oneline -1` in each before each build), my own container ptah1013r-dev on port 15113 (removed at the end), my own MySQL databases (dropped), and fixtures I seeded with the pinned binary's own `migrate diff` rather than theirs. `atlas version` = "atlas community version v1.3.0".

WHAT HOLDS. Everything functional. The note fires once per run on all eight verbs (apply/diff/hash/lint/new/set/status/validate), exit codes and stdout byte-identical to the oracle and to the merge-base build. I got the two measurements they recorded as unverifiable: stream separation AT THE SHELL — `migrate status --dir '...?nonsense=1' --format '{{ len .Pending }}' 2>/dev/null` prints exactly `1` on both oracle and compat, and `2>&1 1>/dev/null` shows the note alone on stderr for status, lint, new and diff. Control arm on a golang-migrate fixture: no query → exit 1 with a byte-identical checksum message on both tools; `?format=golang-migrate` → exit 0 on both; `?fromat=` typo → exit 1 on both, plus the note naming "fromat". The surviving cell reproduces on both commits (compat 1, writes nothing; oracle 0), and the oracle's reverse SQL is as described — verified goose `-- +goose Down`, liquibase one `--rollback:` per forward statement, golang-migrate `.down.sql`. Strict env refuses on every verb, `?format=atlas` still exits 0 under it, and on `migrate apply` against PostgreSQL the refusal leaves `\dt` reporting "Did not find any relations" — stronger than their in-process assertion. Applied for real on PostgreSQL 17 as BOTH ptah_user and the bootstrap postgres superuser, and on MySQL 9.7: exit 0, table created, note on stderr. `migrate import|down|checkpoint|edit --dir` on the oracle all answer `Error: unknown flag: --dir`, so the "eight verbs" enumeration is right. Percent-encoded key `?%66ormat=golang-migrate` decodes on BOTH tools (no divergence). Mutants M1/M2/M3 reproduce exactly the claimed red sets, tree clean afterwards, `grep -n MUTANT` empty. All new tests execute (`-v`: 8+8 subtests, no skips). Zero deletions in every `*_test.go` per `git show --numstat`, so no assertion moved to meet behavior. EVERY gate green in my worktree: go build 0 | go vet 0 | go vet -tags integration 0 | go test ./... -count=1 -timeout 40m -p 2 → 0, zero FAIL lines, cmd/atlas ok 407.171s | qtlint 0 | golangci-lint (private cache) 0 "0 issues." | check-test-style 0 | check-public-api 0 | check-public-api-snapshot 0 | check-public-api-released 0 | all seven doc gates 0 | and `node docs/site/scripts/build-feature-matrix.mjs` regenerates feature-matrix.md with no drift (169 rows).

WHAT DOES NOT HOLD — two false sentences that ship, both about this PR's own new surface.

D1 (the stronger). overview.md says of PTAH_STRICT_DIR_QUERY: "An invalid value is an error rather than a silent 'off'." The doc comment on `dirQueryStrictEnvVar` in cmd/atlas/migrate_dir_query.go says the same and names its reason: "for the reason PTAH_SKIP_CHECKS states: a typo in a CI environment file must not read as 'off' to the tool while the operator believes it is on." Measured on the binary built from 917fbc15:
  PTAH_STRICT_DIR_QUERY=nope ptah-compat migrate validate --dir 'file://<m>'            → exit 0, no diagnostic
  PTAH_STRICT_DIR_QUERY=nope ptah-compat migrate validate --dir 'file://<m>?nonsense=1' → exit 1, invalid boolean value "nope"
  PTAH_SKIP_CHECKS=nope      ptah-compat migrate apply --dir 'file://<pgm>' --url postgres://... → exit 1
So the invalid value IS a silent off on every invocation carrying no ignored key — that is, on every correct invocation, which is the whole of a healthy pipeline — while the precedent the comment cites validates unconditionally on every run. The cause is that `atlasDirQueryStrictFromEnv()` sits AFTER the `len(keys) == 0` early return in `reportIgnoredDirQuery`. The exact CI-environment-file scenario the comment invents is the one where the claim fails.

D2. comparison.md says: "A `--dir` query key other than `format` is ignored on every verb, as Atlas ignores it, and named on standard error so a misspelled one is not silent". Measured:
  ptah-compat migrate edit 20260808071202_init --dir 'file://<m>?nonsense=1'
    → exit 1, `Error: atlas migrate edit --dir: migration directory URL query parameters are not supported for this command`
  ptah-compat migrate checkpoint --dir 'file://<m>?nonsense=1' --dev-url postgres://...
    → exit 1, same message
Both register `--dir` (their own `--help` prints it) and refuse any query; `rm`, `rebase`, `test`, `down` and `push` share that `atlasargs.LocalDirValue` path. The author records this in notVerified item 4 as an "internal inconsistency ... out of scope", then shipped the sentence without the qualification. reference/atlas-commands.md carries the same shape ("on every verb that takes a `--dir` query").

Neither finding is a functional regression, and I found none: compat never exits 0 where the oracle exits 1, and no default-path exit code or stdout byte differs from the merge-base on any row I measured. I am reporting refuted because both sentences are false as written, both are about the capability this PR introduces, and item 8 of the brief treats a false sentence in a doc comment as shipping like a false line of code.

### What it says must change

Two fixes; both small, both belong in this PR because both describe this PR's own new surface. I made no commit — my worktree is clean at 917fbc15 and the candidate fix below was reverted after verification.

FIX 1 (code + doc, for D1). Make the claim true rather than weakening it, because the comment's own justification demands it. In `cmd/atlas/migrate_dir_query.go`, hoist the env resolution above the early return so an invalid value fails every run that reads a `--dir` URL, exactly as PTAH_SKIP_CHECKS does:

    func reportIgnoredDirQuery(out io.Writer, verb string, query url.Values) error {
        strict, err := atlasDirQueryStrictFromEnv()
        if err != nil {
            return fmt.Errorf("atlas migrate %s --dir: %w", verb, err)
        }
        keys := atlasmigrate.IgnoredDirQueryKeys(query)
        if len(keys) == 0 {
            return nil
        }
        ...

I built and ran this. `PTAH_STRICT_DIR_QUERY=nope ptah-compat migrate validate --dir 'file://<m>'` (no query) goes 0 → 1 with `Error: atlas migrate validate --dir: invalid boolean value "nope" for PTAH_STRICT_DIR_QUERY`, matching the PTAH_SKIP_CHECKS row. It reddens exactly one test, `TestCompatMigrateDirQuery_StrictEnvRejectsInvalidValue_FailurePath`, and only in its fixture builder: `writeQueryFixtureDir` seeds with `migrate hash` while `t.Setenv(strictDirQueryEnvVar, "nope")` is already in effect, so the seeding run now refuses too (`atlas migrate hash --dir: invalid boolean value "nope"` at migrate_dir_query_test.go:35). Build the fixture before setting the variable and the row passes. Add a row to the same test pinning the new property — invalid value, NO query, exit 1 — since that row is the whole point of the sentence and nothing pins it today. If instead the behavior is kept as-is, then overview.md must read "An invalid value is an error on any run that carries an unrecognized key" and the `dirQueryStrictEnvVar` comment must drop the PTAH_SKIP_CHECKS parallel, because that variable validates unconditionally.

FIX 2 (doc, for D2). In docs/site/src/content/docs/atlas/comparison.md, bound "every verb" to the eight that accept a query, and name the rest, which is what the author's own notVerified item 4 already knows:

  "A `--dir` query key other than `format` is ignored on the eight verbs that accept a `--dir` query — `apply`, `diff`, `hash`, `lint`, `new`, `set`, `status` and `validate` — as Atlas ignores it, and named on standard error so a misspelled one is not silent; `PTAH_STRICT_DIR_QUERY=1` refuses it instead. `checkpoint`, `down`, `edit`, `rebase`, `rm` and `test` still refuse a `--dir` query outright, which is an internal inconsistency tracked in #1013 rather than a parity gap: the pinned community binary answers `unknown flag: --dir` on all of them."

Apply the same bound to the paragraph in docs/site/src/content/docs/reference/atlas-commands.md ("on every verb that takes a `--dir` query") and to the `evidence` string of the new feature-matrix row in docs/site/scripts/data/feature-matrix-rows.json, then re-run `node docs/site/scripts/build-feature-matrix.mjs` (it currently regenerates clean at 169 rows, so the regeneration must be committed with the JSON edit).

Suggested PR title and body once both are applied — the existing message is otherwise accurate and should be kept:

  Title: Name the --dir query keys a run ignored instead of dropping them silently (#1013)

  Body: keep the existing commit message verbatim, and append:

  CORRECTIONS FOLDED IN AFTER REVIEW. Two sentences this change shipped were
  measured false and are fixed here rather than left standing.

  PTAH_STRICT_DIR_QUERY resolved its value only on a run that already carried an
  ignored key, so `PTAH_STRICT_DIR_QUERY=nope` exited 0 in silence on every
  invocation without one — the whole of a healthy pipeline — while the precedent
  the comment cites, PTAH_SKIP_CHECKS, refuses on every `migrate apply`. The
  resolution now precedes the key check, so the variable behaves


denisvmedia:

**Second verdict: SURVIVED.** An independent reviewer re-measured the corrected branch on its own container and fixtures and did not refute it. Merge candidate once CI is green.

denisvmedia (commented):
Independent review of head `f912286b` found one merge blocker.

**Major — the documented query contract contradicts itself.** In `docs/site/src/content/docs/reference/atlas-commands.md:300`, an unrecognized `--dir` query key is said to be ignored “on every other verb.” Lines 217–220 and the implementation instead say that `checkpoint`, `down`, `edit`, `rebase`, `rm`, and `test` reject every query parameter.

Please document the exact split: unknown keys are ignored, reported, and optionally made strict only on the query-accepting verbs; the other six verbs refuse queries outright. The current PR body also still carries a stale self-refutation and acknowledges that the remaining #1013 surface is incomplete. This head should not merge as written.

denisvmedia:
Thanks — the blocker is real and is fixed in `3ca6e66f` on this branch. Point by point.

**"`reference/atlas-commands.md:300` says an unrecognized `--dir` query key is ignored 'on every other verb'; lines 217–220 and the implementation say `checkpoint`, `down`, `edit`, `rebase`, `rm` and `test` reject every query parameter." — CONFIRMED, reproduced before changing anything.**

Built `ptah-compat` from the reviewed head `f912286b` and measured, on a hashed native Atlas directory the pinned community binary v1.3.0 seeded itself (`migrate new`; the shell here still refuses a command carrying the `migrate hash` token). Fresh `cp -R` per row, every exit code from an unpiped invocation:

```
migrate new        --dir 'file://w?nonsense=1'  -> 0, stderr: note: ... ignoring ... query key "nonsense"
migrate checkpoint --dir 'file://w?nonsense=1'  -> 1
migrate down       --dir 'file://w?nonsense=1'  -> 1
migrate edit       --dir 'file://w?nonsense=1'  -> 1
migrate rebase     --dir 'file://w?nonsense=1'  -> 1
migrate rm         --dir 'file://w?nonsense=1'  -> 1
migrate test       --dir 'file://w?nonsense=1'  -> 1
```

each of the six reading `Error: atlas migrate <verb> --dir: migration directory URL query parameters are not supported for this command`. Control arm, the same six with no query at all: `rm` 0 (removed the file and rewrote `atlas.sum`), `down` 0, `checkpoint` 0; `edit`, `rebase` and `test` fail for reasons of their own (no editor configured, version already last, no `./tests` directory) — so the refusal is about the query and not about the fixture. The oracle on all six: `Error: unknown flag: --dir`, exit 1.

So the sentence was false for six verbs, and false against another sentence eleven paragraphs above it on the same page. Worth noting how it survived round 2: that round bounded the claim in `comparison.md`, `overview.md`, `migrate-commands.md`, the feature-matrix row and this file's *shared* query-rules section, and its commit message then said the claim was "now bounded ... in reference/atlas-commands.md". A per-verb section that restates a shared rule in passing does not match a grep for the shared section's phrasing, so the sweep missed it and the commit message shipped a claim about its own completeness that was not true.

**"Document the exact split." — done, in place rather than by reference.** `migrate new`'s paragraph now reads:

> An unrecognized `--dir` query key is ignored here and named on standard error, as it is on the other seven verbs that accept a `--dir` query — `apply`, `diff`, `hash`, `lint`, `set`, `status` and `validate`. It is not ignored on `checkpoint`, `down`, `edit`, `rebase`, `rm` or `test`: those refuse a `--dir` query outright, as the shared rules above record.

One clause beyond what you cited, same defect class, same file: the `migrate diff` section said both layout spellings are read "the way every other migrate verb reads them", which is false for the same six, and is now bounded to the verbs that accept a `--dir` query. Called out because you did not ask for it.

**A fixture, so the next per-verb restatement cannot drift again.** Two tests in `cmd/atlas/migrate_dir_query_report_test.go`, following `migrate_down_integrity_test.go`, which already reads a published page to pin that the `atlas.sum` enumeration accounts for `migrate down`:

- `_DocsBoundTheIgnoredKeyClaim` — five governed pages; no sentence may carry both an ignore word and a universal quantifier over verbs. It deliberately does not ban "every verb" outright: the same page legitimately says `--dir` "must name a scheme ... on every Atlas verb", which is true of all of them. Whitespace is collapsed before splitting into sentences, so a claim Markdown wrapped over four source lines is matched as the reader meets it.
- `_MigrateNewReferenceSectionNamesBothVerbSets` — the section must *name* both sets, not merely avoid overstating.

Mutants: restoring the stale sentence reddens exactly those two and leaves all eleven behavior tests green (49 subtests); making the code match the stale prose instead — dropping the query refusal in `internal/atlasargs.LocalDirValue` — reddens `_QueryStaysRefusedOnTheVerbsThatTakeNoQuery_FailurePath` (6 rows) and `TestLocalDirValue_FailurePathRejectsQuery` while both new tests stay green; deleting the sentence rather than bounding it reddens only the second test, which is why there are two.

I chose a Go test over a `docs/site` checker on purpose: `.github/workflows/docs.yml` lists each `check:*` by name, and `check:limitations` is in `package.json` and in no workflow — a new script there might never run. (Filing that separately.)

**"The PR body still carries a stale self-refutation." — CONFIRMED.** The body's "Independent review found a problem — read before merging" section is written in the present tense about the two sentences round 1 found, and both were fixed at `f912286b`; a comment saying so is not the same as the body being true. A corrected body is attached: the refutation section is replaced by a short, factual "Review history" listing all three rounds and what each changed, and the body gains the eight-verb / six-verb split it never stated.

**"…and acknowledges that the remaining #1013 surface is incomplete." — I dispute that this is a merge blocker.** The PR says `Refs #1013`, never `Closes`, and #1013 stays open. The surviving cell is measured, not hand-waved:

```
oracle  migrate diff demo --dir 'file://d1?format=golang-migrate' --dev-url sqlite:… --to …
  -> 0, wrote 20260808123146_demo.up.sql + .down.sql + atlas.sum
compat  same invocation
  -> 1, wrote nothing: "Atlas accepts ?format=golang-migrate, but Ptah does not implement
        that directory format for this command yet"
```

That is the strict direction — never exit 0 where the community binary exits 1 — and closing it needs reverse-SQL planning per statement (liquibase interleaves one `--rollback:` line per forward statement, so a whole-file down block is insufficient), which is a planner change and not a CLI change. Declining to fix it here is scoping; stating it in the body is the honest form. If the objection is that the body *understates* what remains, the corrected body now enumerates both open cells explicitly.

**Nothing traded away.** `cmp` of `cmd/ptah-compat` built from `f912286b` and from `3ca6e66f` returns 0 — byte-identical. The PR's own rows were re-measured on the fixed tree anyway: `?nonsense=1` exits 0 on all eight verbs on both binaries with the note on stderr only; the golang-migrate control still flips 1 → 0 under `?format=golang-migrate` and stays 1 under `?nonsense=1` alone; `PTAH_STRICT_DIR_QUERY=nope` with no query still exits 1; `migrate status --format '{{ len .Pending }}' 2>/dev/null` still prints exactly `1` on both.

Gates on the committed tree, all 0: `go build`, `go vet`, `go vet -tags integration`, `go test ./... -count=1` (254 ok packages, zero FAIL), `qtlint`, `golangci-lint` ("0 issues."), `check-test-style`, `check-public-api`, `check-public-api-snapshot`, `check-public-api-released`, and every `check:*` in `docs/site/package.json` with its selftest — including `check:responsive`, which the previous round had to record as not run: `npm ci` plus `npm run build` here, then "OK (60 routes × 2 viewports)". `build-feature-matrix.mjs` regenerates with no drift.

PR #1311, opened 2026-08-08
Identify an RLS policy by the table it names, not by the string it was written as (#1276)

denisvmedia:
Refs #1276. One more instance of its identity class: `compare.RLSPolicies` keyed both sides by the policy **name alone**, but a PostgreSQL policy name is scoped to its **table**, so two tables each carrying `tenant_isolation` collapsed into one entry — one policy silently won, the other was compared against the wrong counterpart or vanished.

Premise confirmed against a live catalog before any code changed: PostgreSQL accepts the same policy name on two tables in one schema, so the collision is real rather than assumed.

## Two independent reviews, both addressed

This branch was reviewed adversarially twice; each review found a shape the previous author had not built, and both are closed on the current head.

**First review.** Keying by `(table identity, policy name)` fixed the collapse, and then `deduplicateRLSPolicies` in `core/goschema/utils.go` resolved the owning table *differently from the comparator* — so `ptah schema render`, `schema apply` and `migrations generate` produced DDL PostgreSQL refuses. A naive revert was not available: it reddens the row the change exists to protect.

**Second review.** The ASCII case fold was applied to the **declared** table name, and Ptah does not retain quoting — so a case-preserving declared table silently captured a lower-case row-level-security reference naming a *different* PostgreSQL relation. A wrong-object match, not a cosmetic one.

The fix folds a reference **down** to the declaration, never a declaration **up**, which is the asymmetry the second review's counterexample requires. The commit titles record the sequence:

```
52f4bf0c Identify an RLS policy by its table, not by its name alone (#1276)
28d0936a Fold the two spellings of one table when deduplicating RLS policies
2519773d Bind a row-level security declaration to the table it names
8e4393da Fold a row-level security reference down, never a declaration up
```

## What to check on review

- a quoted mixed-case table and a lower-case one differing only by case stay **distinct**;
- two tables in one schema each carrying a policy of the same name stay **two** policies;
- the row the change exists to protect — one policy name on two tables surviving whole — is still green;
- reported names stay the strings each side supplied, because the planner renders them; only the matching is normalized.

## Process note

This branch was pushed in a batch and left without a PR for a while — my oversight, not a deliberate hold. The branch also carried a machine-generated name and has been renamed to `issue-1276-rls-policy-identity`.

Not yet reviewed by the independent reviewer that has been going through the rest of this PR set; treat its findings on the sibling PRs as likely to apply here too, particularly the domain/identity ones.

### DIFF
--- a/.github/workflows/go-integration-tests.yml
+++ b/.github/workflows/go-integration-tests.yml
@@ -262,6 +262,17 @@ jobs:
           go test -v -count=1 ./internal/migrationreplay -tags=integration -run '^TestWithReplayedSnapshot_(ClickHouse|SQLServer)Live$' -timeout 5m
           go test -v -count=1 ./integration -tags=integration -run '^TestAtlasMigrateDiff(MySQLFamilyDatabaseDesiredState|PostgresFamilyDevCleanup)E2E$' -timeout 12m
 
+      # These two skip themselves without a live PostgreSQL, and a skip reads as
+      # a pass, so the log is grepped for the PASS lines afterwards.
+      - name: Run compatibility apply table-spelling tests
+        env:
+          POSTGRES_TEST_DSN: "postgres://ptah_user:ptah_password@localhost:5432/ptah_test?sslmode=disable"
+        run: |
+          go test -v -count=1 ./cmd/atlas -run '^TestSchemaApplyRLS(ReferenceToUndeclaredRelationRefused|DeclaredRelationAccepted)LivePostgres$' -timeout 5m > rls-spelling.log || { cat rls-spelling.log; exit 1; }
+          cat rls-spelling.log
+          grep -q -- '--- PASS: TestSchemaApplyRLSReferenceToUndeclaredRelationRefusedLivePostgres' rls-spelling.log
+          grep -q -- '--- PASS: TestSchemaApplyRLSDeclaredRelationAcceptedLivePostgres' rls-spelling.log
+
       - name: Build integration test binary
         run: |
           go build -race -o ./integration-test ./cmd/integration-test
--- a/cmd/atlas/schema_apply_rls_table_spelling_live_test.go
+++ b/cmd/atlas/schema_apply_rls_table_spelling_live_test.go
@@ -1,0 +1,190 @@
+package atlas_test
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"net/url"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+
+	qt "github.com/frankban/quicktest"
+
+	"go.5x5.cz/ptah/cmd/atlas"
+	"go.5x5.cz/ptah/dbschema"
+)
+
+// livePostgresURLForRLSSpelling gates these tests on the same environment
+// variable as the other live PostgreSQL compatibility tests.
+func livePostgresURLForRLSSpelling(t *testing.T) string {
+	t.Helper()
+	dbURL := os.Getenv("POSTGRES_TEST_DSN")
+	if dbURL == "" {
+		dbURL = os.Getenv("TEST_DATABASE_URL")
+	}
+	if dbURL == "" {
+		t.Skip("POSTGRES_TEST_DSN or TEST_DATABASE_URL not set")
+	}
+	if !strings.HasPrefix(dbURL, "postgres://") && !strings.HasPrefix(dbURL, "postgresql://") {
+		t.Skip("PostgreSQL URL required for RLS table spelling live tests")
+	}
+	return dbURL
+}
+
+// createRLSSpellingDatabases provisions an empty target database and a separate
+// empty dev database, and registers their removal.
+//
+// They must be two databases rather than two schemas: the dev database is reset
+// destructively before the plan is rehearsed on it, and `schema apply` refuses
+// outright when --dev-url may address the target.
+func createRLSSpellingDatabases(t *testing.T, adminURL string) (targetURL, devURL string) {
+	t.Helper()
+	c := qt.New(t)
+	suffix := fmt.Sprintf("%d_%d", os.Getpid(), time.Now().UnixNano()%1_000_000)
+	targetName := "ptah_rls_target_" + suffix
+	devName := "ptah_rls_dev_" + suffix
+	conn, err := dbschema.ConnectToDatabase(context.Background(), adminURL)
+	c.Assert(err, qt.IsNil)
+	c.Cleanup(func() {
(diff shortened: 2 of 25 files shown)


denisvmedia:
Blocking review finding:

`core/goschema/utils.go:1359` builds the new (table, policy) identity by concatenating the two independently valid PostgreSQL identifiers with ".". That encoding is ambiguous and can still silently drop an RLS policy.

I reproduced this on the PR head with:

- table `a`, policy `"b.c"`
- table `a.b` (schema `a`, table `b`), policy `c`

Both identities become `a.b.c`; `goschema.Deduplicate` returns one policy and discards the other. Policy names may contain literal dots when quoted, and `Table.QualifiedName` already treats dots structurally, so this is a valid pair of distinct PostgreSQL policies.

Please use an unambiguous structural key (preferred) or another injective encoding for the pair, and add a regression test covering a dotted policy name against a schema-qualified table. The comparator already uses the structural `tableMemberKey`; the goschema deduplication path needs the same property.

The rest of the reviewed RLS table/name propagation and the live table-spelling controls are coherent, and all current CI checks are green, but this identity collision keeps the PR unsafe to merge.

denisvmedia:
Confirmed and fixed in `1d739d93`.

I reproduced the collision on a live database before changing anything — PostgreSQL 17.10 holds both of these at once, so the pair is a real catalog state:

```
CREATE TABLE a (...);    CREATE POLICY "b.c" ON a ...     ->  public | a | b.c
CREATE SCHEMA a;
CREATE TABLE a.b (...);  CREATE POLICY c ON a.b ...       ->  a      | b | c
```

and on the previous head the new fixture reported `kept 1 of 2 policies` for that shape, with both controls green. So this branch fixed "a policy is keyed by its name alone" and reintroduced the same loss through the encoding — you are right that it kept the PR unsafe.

**Fix, following your preferred direction.** `deduplicateNamedDefinitions` is now generic over a `comparable` key rather than fixed to `string`, so an identity with several components uses a struct and the eleven single-component callers keep passing strings unchanged. The RLS path uses `rlsPolicyIdentity{table, policy}`. As you noted, the comparator already had this property through `tableMemberKey`; the goschema deduplication path now has it too, rather than a second ad-hoc encoding.

**Regression, four rows.** The dotted policy name against a schema-qualified table, as you described it; a second collision shape with *no* schema in play, so the fix cannot be "always qualify the table"; the control that one policy declared twice on one table is still one policy; and the row the original fix exists to protect. Reverting the struct key reddens the two collision rows and leaves both controls green.

Gates: `go build` 0 · `go vet` 0 · `go tool qtlint` 0 · `check-test-style.sh` 0 · `check-public-api.sh` 0 · `check-public-api-snapshot.sh` 0 · `check-public-api-released.sh` 0 · `go test ./core/goschema/ -count=1` 0. The full `go test ./...` had one failure, `cmd/viz/TestSVGReportsGraphvizStderrOnFailure`, a graphviz subprocess timeout under parallel load; run alone that package is 0 and the change does not touch it. Saying so rather than calling the suite green.

Worth recording what this exchange found beyond the bug itself: the branch had already passed two rounds of independent adversarial review, and neither looked at the *encoding* of the key — both checked the semantics it expressed. Reading by `file:line` caught what reasoning about behavior did not.

denisvmedia:
## Post-merge adversarial review

The existing CI and affected-package checks are green, but the following boundary cases still produce incorrect security behavior.

### BLOCKER: quote loss can relocate an RLS policy

At `core/goschema/utils.go:1300`, `ORDERS` can mean either unquoted `ORDERS` (which PostgreSQL folds to `orders`) or quoted `"ORDERS"`; the SQL reader has already discarded that provenance.

Given:

```sql
CREATE TABLE orders (id integer);
CREATE POLICY p ON "ORDERS" USING (true);
```

PostgreSQL rejects the missing case-sensitive relation. Ptah instead rewrites the policy table to `orders`, so rendering/replay silently secures a relation the author did not name. I reproduced `schemaload.Load` returning `orders`.

The same whole-string folding misses valid per-component semantics: `"App".ORDERS` should resolve to declared `"App".orders`, but Ptah keeps `App.ORDERS` because the mixed-case schema makes the complete identity non-foldable.

Preserve quote provenance per identifier component and fold only unquoted components. If the representation cannot distinguish the cases, fail closed rather than moving an access-control declaration. Add both forms to the live PostgreSQL round-trip matrix.

### MAJOR: DOWN silently omits a modified RLS policy

At `internal/planner/dialects/postgres/postgres.go:1977`, the comparator intentionally normalizes `orders` and `public.orders` to one table identity, but it reports the desired-side spelling in `RLSPolicyDiff.TableName`. DOWN planning then uses `dbAsGoSchema`, whose policy carries the introspected spelling, and this exact raw-string lookup misses.

Repro: a desired policy on `public.orders` changes its `USING` expression while the current DB schema reports the same policy on `orders`. `schemadiff.CompareWithDialect(..., "postgres")` returns one modified policy, but generated DOWN SQL contains only:

```text
-- No rollback operations needed
```

Resolve refs with `diff.EffectiveIdentifierSemantics("postgres")`, and return `ptaherr.ErrInvalidSchemaDiff` if a referenced policy still cannot be resolved. Add a generator regression asserting that DOWN restores the prior policy body across default-schema spellings.

### MAJOR: unresolved additions fail open

At `internal/planner/dialects/postgres/postgres.go:1964`, the checked planner returns success while dropping an access-control operation. The new test at `rls_policy_ref_test.go:84` explicitly enshrines that silent behavior, although public planning docs promise invalid schema diffs are rejected with `ptaherr.ErrInvalidSchemaDiff`.

Validate and resolve every `RLSPolicyRef` using the diff's identifier semantics. Missing or ambiguous refs must be errors; stale or malformed serialized diffs must not produce valid-looking SQL with an omitted RLS policy.

### Documentation

- `docs/site/src/content/docs/direct/compare-and-drift.md:139` says bare strings were used “before Ptah v0.2.0”, but the `v0.2.0` tag itself still has `RLSPoliciesAdded []string`. Say “in Ptah v0.2.0 and earlier” or name the first release with object elements.
- The canonical public API guides, `docs/public_api.md` and `docs/site/src/content/docs/extend/public-api.md`, still describe only table-qualified index refs. They should also document the `[]RLSPolicyRef` fields and table-plus-policy identity.

### Verification

- Reproduced the rollback failure through the real `schemadiff.CompareWithDialect` path, not a hand-written diff.
- Reproduced both quoted/component-qualified identifier failures with external `schemaload` tests.
- Affected Go packages, `scripts/check-test-style.sh`, diff checks, and all GitHub checks pass; these missing boundary forms are why green CI did not catch the defects.

Issue #1291, opened 2026-08-07
Role and grant introspection drops user roles named pg-anything, because LIKE 'pg_%' leaves the underscore unescaped


denisvmedia:
The PostgreSQL reader excludes system roles with SQL `LIKE 'pg_%'`. In SQL `LIKE`, `_` is a single-character wildcard, so that pattern matches `pg` followed by **any** character. Ordinary user roles whose names begin `pg` are silently dropped.

### Measured

PostgreSQL 17.10:

```sql
SELECT 'pgbouncer' LIKE 'pg_%';    -- true    <- unescaped, a wildcard
SELECT 'pgbouncer' LIKE 'pg\_%';   -- false   <- escaped, what was meant
SELECT 'pgadmin'   LIKE 'pg_%';    -- true
SELECT 'pguser'    LIKE 'pg_%';    -- true
SELECT 'pg_catalog' LIKE 'pg_%';   -- true    <- the intended match, unaffected either way
```

`pgbouncer`, `pgadmin`, `pgpool`, `pguser`, `pgrest` are all excluded. These are not reserved names — PostgreSQL reserves the `pg_` prefix, with the underscore.

### Where it is, on master

Four sites in `internal/dbschema/postgres/reader.go`, and they have to move together because they answer two different questions about the same roles:

| line | what it filters |
|---|---|
| 2167 | the role list itself — `WHERE r.rolname NOT LIKE 'pg_%'  -- Exclude system roles` |
| 1724 | grantee filter |
| 2249 | grantee filter |
| 2287 | grantee filter |

### Consequences

A role named `pgbouncer` is invisible to the reader, so the comparator reads it as absent and plans `CREATE ROLE "pgbouncer"`, which fails on a database that has it. That is the same failure #1267 and PR #1273 are about, arriving through a different door: not "out of scope read as absent", but "excluded by a typo in a pattern read as absent".

Grants **to** those roles are invisible for the same reason, so a privilege granted to `pgbouncer` is never described and never compared. Given what `pgbouncer` is usually granted, that is the more serious half.

### The fix

`NOT LIKE 'pg\_%'`, at all four sites. This repository already spells the same idea correctly elsewhere — `internal/migrateclean/migrateclean.go` escapes it — so this is an inconsistency rather than an open question.

Worth pairing with a fixture that creates a role named `pgbouncer` and asserts it is described, because the current tests cannot catch this: every role name they use begins with something other than `pg`.

### Scope check before fixing

[...]

Issue #1281, opened 2026-08-07
Extension members whose names are SQL keywords pin the extension to any document containing that keyword


denisvmedia:
Split out of #1280, which fixes the other half.

`ptah-compat schema inspect` decides whether to keep an `extension` block by matching every identifier-shaped token in the surviving document against the names the extension supplies (`goschema.Extension.Provides`, read from `pg_depend`). #1280 excludes names `pg_catalog` also supplies, which is exact. It does nothing for names that are **SQL keywords**.

`hstore` supplies three functions named `delete`:

```
SELECT count(*) FROM pg_depend d
  JOIN pg_extension e ON e.oid = d.refobjid
  JOIN pg_proc p ON p.oid = d.objid
 WHERE d.deptype = 'e' AND d.classid = 'pg_proc'::regclass
   AND e.extname = 'hstore' AND p.proname = 'delete';
-- 3
```

`pg_catalog` has no `delete` function, so the #1280 filter keeps it. `sqlIdentifierTokens` splits on non-identifier characters and has no notion of keywords, so a `plpgsql` body doing `DELETE FROM audit` yields the token `delete` and pins `hstore`.

### Measured

PostgreSQL 17.10, own database, catalog-verified. `base` 3acddbdf, `head` 4c4a8a1d, `fix` the #1280 branch.

```sql
CREATE EXTENSION hstore;
CREATE TABLE audit (id integer PRIMARY KEY);
CREATE FUNCTION purge() RETURNS void LANGUAGE plpgsql AS $$ BEGIN DELETE FROM audit; END $$;
```

| | ext blocks | pinned binary reading the output |
|---|---|---|
| base | 0 | 0 |
| head | 1 | 1 |
| #1280 branch | 1 | 1 |

`Error: postgres: extensions are not supported by this version.` Nothing in that database uses `hstore`.

### Why it is not folded into #1280

Filtering keywords is not a one-line narrowing:

- `delete` is **unreserved** in PostgreSQL, not reserved. `pg_get_keywords()` reports 78 reserved words and 327 unreserved ones; `delete`, `each` and `replace` are all unreserved. Filtering the unreserved list would drop genuine extension member names wholesale.
- Filtering only reserved words does not catch `delete` at all.
- The distinction that matters is positional — `DELETE FROM audit` is a statement, `delete(h, 'k')` is a call — and the token scanner has no structure to read it from. Telling them apart means parsing the function body.

[...]

PR #1319, opened 2026-08-08
Leave an extension's own types to the extension (#1294)

### LINKED ISSUE #1294: Inspect describes an extension-owned composite, domain or range type as a user type, and the document then fails to replay

denisvmedia:
The PostgreSQL reader describes domains, composite types and range types, and none of the three queries excluded the ones an extension owns.

**The held attempt's blocker was a fixture artifact, not a design gap.** #1294's second comment reported this as unlandable until #1276/#1298 lands. That diagnosis is refuted below, by measurement. This PR lands on its own.

## Reproduction

PostgreSQL 17.10, connected as `ptah_user`, source and dev databases created fresh, exit status read from a bare command.

```sql
CREATE EXTENSION lo;      -- supplies the DOMAIN    lo
CREATE EXTENSION dblink;  -- supplies the COMPOSITE dblink_pkey_results

CREATE TYPE ptahspan AS RANGE (subtype = integer);
ALTER EXTENSION lo ADD TYPE ptahspan;

-- the control: the user's own, named a suffix apart
CREATE DOMAIN lo_own AS text;
CREATE TYPE dblink_pkey_results_own AS (position integer, colname text);
CREATE TYPE ptahspan_own AS RANGE (subtype = integer);

CREATE TABLE docs (id integer PRIMARY KEY, payload lo, keys dblink_pkey_results,
    span ptahspan, own_payload lo_own, own_keys dblink_pkey_results_own,
    own_span ptahspan_own);
```

The catalog is unambiguous about ownership:

```
        typname          | kind | owned_by_extension
-------------------------+------+--------------------
 dblink_pkey_results     | c    | t
 dblink_pkey_results_own | c    | f
 lo                      | d    | t
 lo_own                  | d    | f
 ptahspan                | r    | t
 ptahspan_own            | r    | f
```

`ptah-compat schema inspect` on master declares all six, alongside both extensions:

```
extension "dblink"
extension "lo"
domain "lo"                            <- lo's own
domain "lo_own"
composite "dblink_pkey_results"        <- dblink's own
composite "dblink_pkey_results_own"
range "ptahspan"                       <- lo's own
range "ptahspan_own"
table "docs"
```

Reading that document back against a clean dev database:

```
SQL execution failed: ERROR: type "lo" already exists (SQLSTATE 42710)
SQL: CREATE DOMAIN "lo" AS oid;
```

exit 1.

### The destructive half

Same live database, diffed against a desired state declaring nothing:

```console
$ ptah-compat schema diff --from "$SRC" --to file://desired-empty.hcl --dev-url "$DEV"
ALTER TABLE "docs" DROP CONSTRAINT IF EXISTS "docs_pkey";
DROP TABLE IF EXISTS "docs" CASCADE;
DROP DOMAIN IF EXISTS "lo" CASCADE;                    <- lo's own
DROP DOMAIN IF EXISTS "lo_own" CASCADE;
DROP TYPE IF EXISTS "dblink_pkey_results" CASCADE;     <- dblink's own
DROP TYPE IF EXISTS "dblink_pkey_results_own" CASCADE;
DROP TYPE IF EXISTS "ptahspan_own" CASCADE;
DROP EXTENSION IF EXISTS "dblink";
DROP EXTENSION IF EXISTS "lo";
```

That is the issue's title as a plan: three types an extension owns, proposed for dropping as if a user had declared them.

## The fix

One correlated `NOT EXISTS` on `pg_depend`, shared by `readDomainsForSchema`, `readCompositesForSchema` and `readRangesForSchema` — written once because the three ask the same question of the same catalog and three copies drift:

```sql
AND NOT EXISTS (
    SELECT 1 FROM pg_depend extdep
    WHERE extdep.classid = 'pg_type'::regclass
      AND extdep.objid = t.oid
      AND extdep.deptype = 'e'
)
```

`readFunctionsForSchema` already carries this reasoning — its comment says extension functions "cannot be dropped independently and should be managed by the extension" — and it was never applied to types.

`classid` is part of the predicate rather than left to `objid` alone: a `pg_depend` row names its object by `(classid, objid)`, and OIDs come from one counter shared by every catalog.

Ownership is asked of `pg_depend` and never of the type NAME. Every fixture pairs an extension-owned type with one of the user's a suffix apart, because `NOT LIKE 'lo_%'` without an `ESCAPE` clause reads the underscore as a wildcard and takes the user's type with it — the mistake this reader already had to correct for roles (#1291).

### After

```
extension "dblink"
extension "lo"
domain "lo_own"
composite "dblink_pkey_results_own"
range "ptahspan_own"
table "docs"
```

Replay against a clean dev database: **exit 0**. The diff plans neither extension-owned drop; the three `_own` types still drop.

## The survey #1294 asked for

Over every extension the `postgres:17` image ships, exactly six extension-owned types reach these three reads:

| extension | type | kind | shadowed by pg_catalog |
| --- | --- | --- | --- |
| dblink | dblink_pkey_results | composite | no |
| earthdistance | earth | domain | no |
| lo | lo | domain | no |
| tablefunc | tablefunc_crosstab_2/3/4 | composite | no |

`pg_buffercache` and `pg_stat_statements` own composites too, but those are their views' row types and `readComposites` joins `pg_class` on `relkind = 'c'`, so they never reached a description. **No shipped extension supplies a range type**, which is why the range row builds its member with `ALTER EXTENSION ... ADD` and asserts the blocks rather than the round trip.

## Refuting the held blocker

#1294's second comment parked an earlier attempt because the #1281 suite row *"keyword-named extension function whose answering type pg_catalog shadows"* went red, and read that as: `box` is dropped from `Provides` as shadowed → nothing keeps the extension → the block is omitted → nothing declares `box`.

Measured, the block is **not** omitted:

```console
$ grep -c 'extension "pgcrypto"' base.hcl   -> 1
$ grep -c 'extension "pgcrypto"' head.hcl   -> 1
```

and in the failing run the block assertion passes; only the blanket round trip at the end of the loop fails. The real cause is the fixture. `ALTER EXTENSION pgcrypto ADD DOMAIN public.box` writes the `pg_depend` membership row without adding anything to pgcrypto's install script, so it builds a catalog state no `CREATE EXTENSION` can reproduce:

```console
$ psql -c "CREATE EXTENSION pgcrypto" \
       -c "SELECT count(*) FROM pg_type t JOIN pg_namespace n ON n.oid = t.typnamespace
            WHERE n.nspname = 'public' AND t.typname = 'box'"
 0
$ psql -c "CREATE EXTENSION lo" -c "... AND t.typname = 'lo'"
 1
```

That row's round trip was passing only because the reader described the extension's own domain as a user domain — green *because of* this defect, not despite it. Its load-bearing assertion, the extension block, is unchanged and still passes; the round trip now carries a per-row `unreplayable` reason. Whether that costs the round trip depends on **what** was added: a function member is fine, because plpgsql bodies are not resolved at creation time, and that sibling row still asserts its round trip. A type member is not, because a column's type is resolved at `CREATE TABLE`. Nothing else in that suite moves — `TestAtlasCompatInspectExtensionRoundTripE2E` is green on CI on this branch.

### Unit — `internal/dbschema/postgres/reader_types_internal_test.go`

A fake server that **evaluates** the ownership predicate against its own `pg_depend` rows rather than recognizing a token, and refuses any predicate it cannot evaluate instead of answering one it invented.

| mutant | red |
| --- | --- |
| exclusion removed (revert the fix) | `LeavesExtensionOwnedTypes` 3/3, `AsksPgDependRatherThanTheName` 3/3; the control `StillDescribesWhatTheUserDeclared` stays **green** |
| `extdep.objid = t.oid` → `extdep.objid = extdep.objid` | **9/9**, control included |
| exclusion dropped from the ranges arm only | exactly the two `ranges` subtests, nothing else |

The second is the inverse mutant that keeps the non-interference control from being vacuous; the third is what says each of the three arms is separately covered. All three were re-run after the fake was refactored for a lint finding, and still kill it.

### Live — `integration/atlas_compat_extension_owned_types_e2e_test.go`

Three rows, one per type kind, both directions asserted on one document per row.

Reverting the exclusion, each row red on its own reason:

```
--- FAIL: .../a_domain_the_extension's_install_script_creates
--- FAIL: .../a_composite_the_extension's_install_script_creates
--- FAIL: .../a_range_type_an_extension_owns
```

With the fix, all three PASS, on PostgreSQL 17.10, on 18.4 (the version CI runs), and on CI itself.

## A second finding, reported rather than pinned

An `earthdistance` row was written as an independent reproduction and then **removed**, because its column spelling is not stable across environments. On

```sql
CREATE EXTENSION earthdistance CASCADE;
CREATE TABLE places (id integer PRIMARY KEY, loc earth NOT NULL);
```

the same commit rendered `column "loc" { type = sql("cube") }` with an `extension "cube"` block on PostgreSQL 17.10 **and** on 18.4 locally, and `type = sql("earth")` with an `extension "earthdistance"` block on CI. The catalog answers identically on both — `format_type` reports `earth`, and `information_schema.columns.domain_name` reports `earth` — so the divergence is in what Ptah does with that answer, which is #1242's flattening.

Which spelling appears decides which extension the reference scan keeps, so every assertion that row could make is really an assertion about #1242 and about which block survives suppression. Neither is what this change is for, and a fixture that answers differently in CI than on every server it was written against is not a regression test. The measurement is kept in the suite's doc comment so it is not re-added blind. **It likely deserves its own issue.**

Note that `earthdistance` on PostgreSQL 17.10 does reproduce #1294 exactly as `lo` does — master's document declares `domain "earth"` alongside `extension "earthdistance"` and fails to replay with `ERROR: type "earth" already exists (SQLSTATE 42710)`, and with the fix it replays at exit 0. It is the *assertion* that is unstable, not the defect.

## Gates

Each run bare, exit code read directly — never through a pipe.

| gate | exit |
| --- | --- |
| `go build ./...` | 0 |
| `go vet ./...` | 0 |
| `scripts/check-test-style.sh` | 0 (175 conditional groups, 45 white-box files) |
| `docs/site`: `npm ci` + all 14 `check:*` scripts | 0 |
| `go test ./... -count=1` | 1 locally, see note; **green on CI** |
| `go test -tags integration ./integration/ -run 'TestAtlasCompat...'` | 0, on PostgreSQL 17.10 and on 18.4 |

**Note on the local `go test ./...`.** One run reported two failing packages, `cmd/ptah-ls` and `cmd/viz`, on a machine carrying load average 47 from three concurrent full Go test suites. Both failures are starvation signatures — `process did not exit within 20s`, `signal: killed`, and a fake graphviz killed before it could print its message — and both pass in isolation:

```console
$ go test ./cmd/ptah-ls/ ./cmd/viz/ -count=1
ok  	go.5x5.cz/ptah/cmd/ptah-ls	20.613s
ok  	go.5x5.cz/ptah/cmd/viz	13.030s
```

Neither reaches the changed code — `go list -deps` for both returns 0 hits for `internal/dbschema/postgres`. **CI's Go Unit Tests workflow is green on this branch**, which is the authoritative run.

### CI on the first push

10 of 11 workflows green, including Atlas CE Oracle, Go Lint, Go Unit Tests, Go Security, Go Fuzz, Export Acceptance and Docs. Integration Tests failed on exactly one subtest, the `earthdistance` row, which the second commit removes.

## Documentation

- `docs/user_defined_types.md` gains an *Extension-owned types* section with the measurement and the "ownership is a catalog edge, not a name" rule.
- `docs/site/src/content/docs/databases/postgresql.md` gains the same point in its *User-defined types* list.

## Residual risk

- **A shadowed extension type in a document whose extension block is omitted.** If an extension supplied a type into `public` with a name `pg_catalog` also supplies, `format_type` would render the column schema-qualified, the shadowed-name filter would keep that name out of `Extension.Provides`, and the extension block could be omitted — leaving a document naming a type nothing declares. Measured: **none** of the six shipped extension-owned types has a name `pg_catalog` supplies, so this is unreachable with any extension the image ships. It is #1276's shape and is what #1298's coverage record would express. I did not add a reader→renderer signal for it, because no available fixture can make it red, and an unmeasurable mechanism is worse than a named gap.
- **A desired state that explicitly declares a type an extension owns** now diffs against a read that reports nothing, so it plans a `CREATE` that fails with `already exists`. Before this change it silently compared against the extension's type as though the user owned it. Both are wrong; the new one fails loudly.
- `readFunctionsForSchema`'s own extension filter matches `d.objid = p.oid AND d.deptype = 'e'` with **no** `classid`, unlike the new type predicate. Since OIDs come from one counter shared by every catalog, `objid` alone can match a row of another class. Not touched here — no reproduction, and narrowing it is a separate change.
- The `earthdistance` rendering divergence above is unexplained and outlives this PR.

## Not verified

PostgreSQL only. No MySQL, MariaDB, ClickHouse or SQL Server target has these three reads. No conformance run of my own; no flags added, no public API change.

Closes #1294

### DIFF
--- a/docs/site/src/content/docs/databases/postgresql.md
+++ b/docs/site/src/content/docs/databases/postgresql.md
@@ -252,6 +252,13 @@ destructive by the safety gate. Reconciliation is deliberately conservative:
   domain, the drop fails loudly instead of dropping the column.
 - Range types are matched by name only; a changed range is dropped and
   recreated.
+- A domain, composite or range type that an **extension** owns is not
+  described. `CREATE EXTENSION` creates those types, so they cannot be created
+  or dropped independently, and describing one as a user type made the
+  description declare something the extension already makes — replaying it
+  failed with `type "lo" already exists`. Ownership is read from `pg_depend`,
+  so a type of your own named close to an extension's is unaffected. The same
+  rule has always applied to extension-owned functions.
 
 ## Concurrent index creation
 
--- a/docs/user_defined_types.md
+++ b/docs/user_defined_types.md
@@ -75,6 +75,27 @@ Renders `CREATE TYPE "floatrange" AS RANGE (SUBTYPE = float8, SUBTYPE_DIFF = flo
 
 Range types have no in-place `ALTER`, so a changed range is dropped and recreated, and range comparison matches by name only.
 
+## Extension-owned types
+
+Reading a PostgreSQL database describes only the domains, composite types and
+range types the user declared. One that an extension owns is left to the
+extension, for the same reason extension-owned functions always have been: it
+is created by `CREATE EXTENSION` and cannot be created or dropped
+independently.
+
+Describing one made the description declare a type the extension already
+creates, and replaying such a description failed — measured on PostgreSQL
+17.10 against `CREATE EXTENSION lo`, which supplies the domain `lo`:
+
+```text
+ERROR: type "lo" already exists (SQLSTATE 42710)
+SQL: CREATE DOMAIN "lo" AS oid;
+```
+
+Ownership is read from `pg_depend` (`deptype = 'e'`) rather than from the type
+name, so a type of your own named closely after an extension's — `lo_own`
+beside `lo` — is still described in full.
+
 ## Ordering
 
 Ptah emits user-defined types after extensions and enums but before tables, so table columns can reference them. Within the group the order is domains → ranges → composites (composites may reference the others). Drops run after tables, and `DROP TYPE` / `DROP DOMAIN` are classified as destructive by the safety gate.
--- a/integration/atlas_compat_extension_owned_types_e2e_test.go
+++ b/integration/atlas_compat_extension_owned_types_e2e_test.go
@@ -1,0 +1,231 @@
+//go:build integration
+
+package integration_test
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+	"os"
+	"path/filepath"
+	"testing"
+	"time"
+
+	qt "github.com/frankban/quicktest"
+	_ "github.com/jackc/pgx/v5/stdlib"
+)
+
+// extensionOwnedTypeCase is one database shape and the type blocks its
+// description must and must not carry.
+type extensionOwnedTypeCase struct {
+	name string
+	// extensions are installed with CASCADE, in order, before seed runs.
+	extensions []string
+	// seed is the shape's DDL. It creates the user's own types alongside the
+	// extension's, so both directions are measured on one document.
+	seed string
+	// absentBlocks are block spellings the description must NOT carry, because
+	// CREATE EXTENSION already creates what they would declare.
+	absentBlocks []string
+	// presentBlocks are block spellings the description MUST carry: the user's
+	// own types, and the extension blocks without which the document names a
+	// type nothing creates.
+	presentBlocks []string
+	// unreplayable, when set, is the reason this row's fixture cannot be
+	// materialized on a fresh dev database whatever the renderer emits. See
+	// [atlasCompatExtensionRoundTripCase.unreplayable].
+	unreplayable string
+	// why records the measurement behind the row.
+	why string
+}
+
+// TestAtlasCompatExtensionOwnedTypesE2E pins that a domain, composite or range
+// type an extension owns is not described as a user type, and that a type the
+// user declared still is.
+//
+// `CREATE EXTENSION` creates the extension's own types, so a description that
+// declares both the extension and the type cannot be replayed: the second
+// declaration collides with what the first already made. Measured on
+// PostgreSQL 17.10 against `lo`, which supplies the domain `lo`, with a
+// `lo`-typed column so the extension block survives suppression:
+//
+//	before  domain "lo" declared, extension "lo" declared, replay exit 1
+//	        ERROR: type "lo" already exists (SQLSTATE 42710)
+//	        SQL: CREATE DOMAIN "lo" AS oid;
+//	after   domain "lo" absent, extension "lo" declared, replay exit 0
+//
+// The reader already excluded extension-owned FUNCTIONS for exactly this
+// reason -- its comment says they "cannot be dropped independently and should
+// be managed by the extension" -- and the three type reads were never given
+// that filter (stokaro/ptah#1294).
+//
+// Both directions are asserted on every row, on ONE document, because the
+// cheap wrong fix is a name filter. The user's types are named as closely to
+// the extension's as PostgreSQL allows -- `lo` beside `lo_own` -- so a
+// `NOT LIKE 'lo_%'` written without an ESCAPE clause, which is the mistake this
+// reader has already had to correct for roles (stokaro/ptah#1291), takes the
+// user's type with it and this test says so.
+//
+// Which extensions can carry a row at all was measured rather than assumed.
+// Over every extension the postgres:17 image ships, exactly six extension-owned
+// types reach these three reads:
+//
+//	dblink         dblink_pkey_results     composite
+//	earthdistance  earth                   domain
+//	lo             lo                      domain
+//	tablefunc      tablefunc_crosstab_2/3/4 composite
+//
+// pg_buffercache and pg_stat_statements also own composites, but those are the
+// row types of their views, and readComposites joins pg_class on relkind = 'c',
+// so they never reached a description. NO shipped extension supplies a range
+// type, which is why the range row below builds its member with
+// `ALTER EXTENSION ... ADD` and asserts the blocks rather than the round trip.
+//
+// The earthdistance row is a second, independent reproduction and is kept
+// because its column does NOT name the type: an `earth` column renders as
+// `sql("cube")`, the domain flattened to its base type (stokaro/ptah#1242). So
+// after the fix nothing names anything earthdistance supplies and its extension
+// block is legitimately omitted too -- and the document still replays, because
+// what the column now names is `cube`, which the cube extension declares. That
+// row is what distinguishes "stopped describing the domain" from "stopped
+// describing the domain and lost the column with it".
+func TestAtlasCompatExtensionOwnedTypesE2E(t *testing.T) {
+	adminURL := requirePostgresE2EDatabaseURL(t)
+
+	tests := []extensionOwnedTypeCase{
+		{
+			name:       "a domain the extension's install script creates",
+			extensions: []string{"lo"},
+			seed: "CREATE DOMAIN lo_own AS text;" +
+				" CREATE TABLE docs (id integer PRIMARY KEY, payload lo, own_payload lo_own)",
+			absentBlocks:  []string{`domain "lo"`},
+			presentBlocks: []string{`extension "lo"`, `domain "lo_own"`},
+			why: "lo supplies the domain lo and a lo-typed column keeps the extension in the" +
+				" document, so the collision is reachable; lo_own is the user's and must survive",
+		},
+		{
+			name:       "a composite the extension's install script creates",
+			extensions: []string{"dblink"},
+			seed: "CREATE TYPE dblink_pkey_results_own AS (position integer, colname text);" +
+				" CREATE TABLE keys (id integer PRIMARY KEY, k dblink_pkey_results," +
+				" own_k dblink_pkey_results_own)",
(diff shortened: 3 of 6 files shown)


denisvmedia:
CI is green on `f45f415`, all 11 workflows:

```
Atlas CE Oracle        success
Docs                   success
Export Acceptance      success
Go Fuzz                success
Go Lint                success
Go Security            success
Go Unit Tests          success
Go Version Consistency success
Integration Tests      success
Ptah Action Smoke      success
Release                success
```

That settles the two gate notes in the body.

**`go test ./...`.** Both local full-suite runs failed, and both are load artifacts rather than findings. The second one is the clearer measurement: it ran while the machine carried load average 47 from three concurrent Go test suites, and the failures are

```
panic: test timed out after 10m0s     cmd/atlas        (708s; 544s on an unloaded run)
panic: test timed out after 10m0s     cmd/ptah-compat  (641s; 363s on an unloaded run)
context deadline exceeded             cmd/ptah-ls
signal: killed                        cmd/viz
```

— four wall-clock signatures, two of them on packages that normally finish well inside the 10m limit. `cmd/ptah-ls` and `cmd/viz` pass in isolation and do not reach the changed code at all (`go list -deps` returns 0 hits for `internal/dbschema/postgres`). `cmd/atlas` and `cmd/ptah-compat` do reach it, which is why they are worth naming rather than waving through — and CI's Go Unit Tests workflow, which runs the same `go test ./...`, is green on this commit.

**Integration Tests.** Green now. The first push failed on exactly one subtest — the `earthdistance` row — and the second commit removes it, with the measurement kept in the suite's doc comment. Every step of the run that exercises this change passed:

```
✓ Run executor integration tests with live databases
✓ Run database-realm cleanup and replay tests
✓ Run compatibility apply table-spelling tests
```

The three remaining rows plus `TestAtlasCompatInspectExtensionRoundTripE2E` (including the repaired `box` row) are green on CI, and both suites are green locally against PostgreSQL 17.10 and 18.4.

PR #1338, opened 2026-08-09
Close three output-shape cells and stop renaming migrations (#1235)

denisvmedia:
Refs [#1235](https://github.com/stokaro/ptah/issues/1235). **Not** `Closes` — five of the register's 51 cells are closed here, three more are confirmed closed elsewhere, one is reported unmeasurable, and the rest are still open. The open inventory is at the bottom so the next pass starts from a list rather than from 51 re-readings.

## What changed in this revision

The previous revision of this branch closed **6.1** by seeding the CONNECTED schema into the realm document before any table was placed in it. That is one line shorter than the alternative and it was **measured reopening two shapes on the same verb that already matched**. It is gone. 6.1 is closed on `master` by [#1264](https://github.com/stokaro/ptah/issues/1264) / [#1275](https://github.com/stokaro/ptah/pull/1275), which builds the realm from the schemas the READER described, and that fix survives this branch untouched.

The branch is now rebased onto `origin/master` `71a7d7f2`, which is where #1275 lives.

### Every shape `schema inspect --format '{{ json . }}'` can be given

Live PostgreSQL 17 in a throwaway container created for this run, one database per shape, dropped and recreated between the "before" and "after" passes; each binary built into a directory inside this worktree; every exit code read from an unpiped invocation; every cell compared with `cmp` against the pinned Atlas community binary v1.3.0's own stdout. Every invocation exits 0 — the table is about bytes, not status.

| Shape | `cmp` vs pinned binary, previous head `03dd0116` | `cmp` vs pinned binary, this head |
| --- | --- | --- |
| realm URL, empty database | **1** — `{"schemas":[{"name":"public"}]}` vs `…,"comment":"standard public schema"}` | **0** |
| realm URL, one schema with two tables | **1** — no schema `comment`, and `users_email_key` listed twice | **0** |
| realm URL, `public` + `extra`, both with tables | **1** — `extra` absent entirely | **0** |
| schema-bound URL (`?search_path=public`) | **1** — no schema `comment` | **0** |
| realm URL + `--schema extra` | **1** — a second, empty `{"name":"public"}` entry the binary never prints | **0** |
| realm URL + `--schema nosuch` | **1** — `{"schemas":[{"name":"public"}]}` where the binary answers `{}` | **0** |

The last two rows are the blocker. They are the shapes the seed **broke**: on `master` alone, without this branch, both already matched the binary byte for byte. Row two is the shape this branch still has to fix — it is finding 6.2, and `master` alone misses it.

SQLite, same discipline:

| Shape | pinned binary | this head |
| --- | --- | --- |
| empty database | `{"schemas":[{"name":"main"}]}` | identical |
| `--exclude` every table | `{"schemas":[{"name":"main"}]}` | identical |
| `a TEXT UNIQUE, b TEXT UNIQUE, c TEXT` + `CREATE UNIQUE INDEX ux_t_c` | 3 index entries | 3, same set — **order still differs**, which is 4.8 and stays open |

### 6.2 — a UNIQUE constraint and its index are one entry

SQLite reports both halves: `pragma index_list` names the implicit `sqlite_autoindex_<table>_<n>` and the table's UNIQUE constraint carries that same name. Over `CREATE TABLE t (id INTEGER PRIMARY KEY, a TEXT UNIQUE, b TEXT UNIQUE, c TEXT)` plus `CREATE UNIQUE INDEX ux_t_c ON t(c)`, the pinned binary printed **3** index entries and Ptah printed **5**, each autoindex listed twice.

Deduplicated by index name. The constraint branch still runs — it exists for readers that report a UNIQUE constraint with no index row of its own — which is why the duplicate is dropped rather than the branch deleted.

**Not SQLite-only, though the register is.** On live PostgreSQL 17 a plain `email text UNIQUE` column printed `users_email_key` twice on `master` and once here; that is the "one schema with two tables" row above, and it is the last thing standing between `master` and a byte-identical answer on all six PostgreSQL shapes.

### 8.6 and 8.7 — `migrate new` writes the name it was given

The Atlas layout is not ours to rename in. That binary composes `<version>_<name>.sql` from the name verbatim; Ptah mapped spaces to hyphens and dropped every character outside `[-_0-9A-Za-z]`. The name is covered by `atlas.sum`, so a rewritten name is also a different checksum for the same command.

Re-measured on this head, each row in its own directory:

| `migrate new …` | Pinned binary | Ptah, before | Ptah, now |
| --- | --- | --- | --- |
| `"add users table"` | `<version>_add users table.sql` | `<version>_add-users-table.sql` | matches |
| `"add_users.sql"` | `<version>_add_users.sql.sql` | `<version>_add_userssql.sql` | matches |
| `"caps-AND.dots+plus"` | `<version>_caps-AND.dots+plus.sql` | `<version>_capsANDdotsplus.sql` | matches |

Cross-read on this head: the pinned binary answers `migrate validate` exit 0 on the directory Ptah wrote, and Ptah answers exit 0 on the directory that binary wrote.

**Two rules survive, both deliberate.**

`GenerateEmptyMigration` still trims leading and trailing whitespace. That binary keeps it — `migrate new "  padded  "` writes `<version>_  padded  .sql` — but a file name with trailing spaces does not survive every filesystem this tool writes into, and no finding in the register asks for one. Recorded as a remaining divergence rather than silently closed.

And writing the name verbatim lets it reach the Atlas file-name grammar, where exactly one suffix collides today. `migrator.ParseAtlasMigrationFileName` reads `<version>_x.down.sql` as the down half of a pair, because Atlas importers emit that spelling for golang-migrate directories. So `migrate new` now refuses a name whose composed file name does not round-trip back to a new **up** migration, before the directory is bound. It is a round trip rather than a list of banned suffixes, so a later change to that grammar cannot reopen the hole silently.

**That refusal does not close the reader gap, and is not claimed to.** Measured on this head:

```console
$ atlas migrate new "x.down" --dir file://down_a      # pinned binary
exit 0, writes 20260809032118_x.down.sql
$ ptah-compat migrate new "x.down" --dir file://down_b
Error: migration name "x.down" composes the file name <version>_x.down.sql, which this tool does not read back as a new migration
exit 1
```

A directory the pinned binary wrote that way still makes `ptah-compat migrate status` exit 1 where that binary exits 0. That is a divergence in the "Ptah exits 1, the binary exits 0" direction the register lists as unfiled. It is written down in `docs/conformance.md` and belongs in that unfiled set, not in this PR — fixing it means changing the shared Atlas file-name parser that golang-migrate imports depend on.

`migrate checkpoint` keeps the old rewriting, now under its own name `atlasCheckpointNameStem`. The verb has no measured counterpart on that binary, and `WriteAtlasCheckpointFile` joins the stem onto a pathname instead of opening it through a rooted directory handle, so a stem carrying a separator would name a file outside the directory the caller asked for.

### 9.3 and 9.4 — two periods

Read with `xxd`, unpiped, on this head:

| Probe | pinned binary | Ptah, before | Ptah, now |
| --- | --- | --- | --- |
| `migrate apply` over an already-applied directory | `…execute\n` — 30 bytes, and stdout holds nothing else | `…execute.\n` (31) | `cmp` against the binary's stdout exits 0 |
| `migrate apply` over an empty directory | the same 30 bytes | 31 | `cmp` exits 0 |
| `schema apply --auto-approve` against a synced database | `…be made\n` (40) | `…be made.\n` (41) | `cmp` exits 0 |

The byte counts for the third row are 40 and 41, not the 41 and 42 an earlier revision of this description claimed; they were re-read with `xxd` and `wc -c` on this head, and the code comment in `cmd/atlas/schema_apply.go` is corrected too.

Only the `apply` verb moves. That binary's `schema diff` answer, `Schemas are synced, no changes to be made.`, does carry a period and already matched — the control that keeps this from becoming a blanket period-stripping. The native `ptah schema apply` sentence in `cmd/schema/apply.go` is untouched: no parity is owed on that surface.

**9.3 had no assertion that could see its own defect, and now it does.** Every existing assertion on that sentence reads `qt.Contains, "No migration files to execute"` — which matches just as happily with the period back. Putting the period back into `migrate_apply.go` reddened **nothing**. `TestCompatMigrateApply_NoopLineIsByteExact` asserts the whole stdout of a second apply equals the measured 30 bytes. 9.4 already had such an assertion in `TestCompatCommand_SchemaApplyAppliesLocalSchemaToSQLite`, confirmed red the same way.

## Seeing it red

Six mutants, each applied to the working tree, run, and reverted; `git diff --stat` empty after each. Exit codes read directly, never through a pipe.

| Mutant (the cheaper wrong implementation, not a deletion) | Reddens |
| --- | --- |
| the connected-schema seed put back | `TestSchemaInspectJSONNeverInventsTheConnectedSchema` rows 3 and 4, **and** `TestRenderSchemaInspect_JSONRealmDocument/a_selection_that_matched_nothing_stays_empty` — while both empty-database rows stayed GREEN, which is what shows the seed buys nothing |
| the index deduplication reduced to bookkeeping only | the backed-constraint row only |
| the UNIQUE constraint branch emits nothing (over-correction) | the no-backing-index row only |
| `atlasEmptyMigrationFileName` sanitizes again | all 4 verbatim-name rows **and** the `.down` refusal |
| the `.down` round-trip guard returns nil | the refusal row only |
| the trailing period restored, one line at a time | `…NoopLineIsByteExact` for 9.3; `…SchemaApplyAppliesLocalSchemaToSQLite` for 9.4 |

The third is the inverse guard: without it, "deduplicate" could be traded for "drop the branch" and the suite would stay green. The first is the blocker's guard: it is the exact code the previous revision shipped.

## Rebase

Rebased from `03dd0116` onto `origin/master` `71a7d7f2`. Two files conflicted, and one of them is the reason a clean merge is not a correct merge:

- `internal/atlasreport/schema_inspect.go` — git auto-merged the **body** of `atlasSchemaInspectJSON`, keeping master's `schema.Schemas` loop *and* the seed, and conflicted only on the doc comment. Taking either side of that comment would have left a tree that compiles, passes `go build`, and reintroduces the divergence. Resolved by keeping master's construction and deleting the seed; the net diff of that file against `master` is now the 6.2 deduplication and one doc paragraph.
- `cmd/atlas/atlas_test.go` — both sides had reached the same assertion by different routes; master's comment is the accurate one and was kept.

Conflict-marker grep over the tree is empty. `go build ./...` and `go vet ./...` were run on the rebased tree before anything else.

## Gates

Every one run from the worktree root on the rebased tree, exit codes read directly and never through a pipe.

| gate | exit |
| --- | --- |
| `go build ./...` | 0 |
| `go vet ./...` | 0 |
| `go test ./... -count=1 -timeout 40m` | 0 — 187 packages `ok`, zero `FAIL` lines |
| `docs/site`: `npm ci`, every `check:*`, every `*:selftest`, `npm run build` | 0 |

None of these can see a parity defect, which is why every behavior claim above comes from a live database and the pinned binary instead.

## What could not be measured

**8.5, the orphan revision row, is not in this PR and was not measured.** Its fixture needs a hand-written three-migration directory carrying a valid `atlas.sum`, and the only way to produce one on the command line is `migrate hash`, which this session's command guard refuses on every spelling. Confirmed the fixture is unreachable rather than assumed: both binaries refuse an unhashed directory with `checksum file not found`, exit 1. 8.2 (goose zero padding) is blocked by the same constraint, exactly as PR #1308 recorded.

## Group 7 is untouched, on purpose

7.1 and 7.2 need a decision the owner has not made: whether a two-part `--exclude` selector means `schema.table` (Ptah) or `table.column` (the pinned binary). The issue itself flags that the binary's reading arguably *is* the defect and Ptah's is more useful, so matching it would be reproducing one. Left alone and left open.

## Still open in the register

Closed **here**: **6.2, 8.6, 8.7, 9.3, 9.4**.

Closed **elsewhere**, confirmed rather than assumed: **6.1** by #1264 / #1275 on `master` — re-measured on this head, an empty database answers `{"schemas":[{"name":"main"}]}` on SQLite and `{"schemas":[{"name":"public","comment":"standard public schema"}]}` on PostgreSQL 17, byte-identically. **5.1** and **5.2** by #1308, measured at `fc77c58c` in the previous session on this branch and not re-measured on this head. 2.2 and 9.9 were recorded closed by #1308's own re-measurement.

Open, by number, so the next pass has an inventory:

- **Group 1** — 1.1, 1.2, 1.3, 1.4. No default Atlas report-envelope renderer; `--dry-run` still shows no SQL.
- **Group 2** — 2.1 (MF101 not implemented). 2.2 closed earlier.
- **Group 3** — 3.1, 3.2 (progress chatter on `migrate new` stdout; still reproduces on this head — the pinned binary writes 0 bytes and `ptah-compat` writes `Generated empty migration file: …`), 3.3, 3.4.
- **Group 4** — 4.1, 4.2 (the missing `PRAGMA foreign_keys` bracketing, the one correctness gap in that group), 4.3, 4.4, 4.5, 4.6, 4.7, 4.8 (index ordering; the 6.2 fixture still shows Ptah's index order differing from the binary's now that the *set* matches — measured on this head).
- **Group 6** — 6.4, 6.5, 6.6, 6.7. 6.3 landed earlier; 6.1 is closed on master; 6.2 closes here.
- **Group 7** — 7.1, 7.2. Blocked on an owner decision, see above.
- **Group 8** — 8.1, 8.3, 8.4, 8.5, 8.8. 8.2 not measurable in this session. 8.8 was triaged by #1308 as a case where the binary is the destructive one and rule (b) keeps Ptah's behavior.
- **Group 9** — 9.1, 9.2 (triaged by #1308 as the binary's slip), 9.5, 9.6, 9.7, 9.8, 9.10, 9.11, 9.12 `[carried over]`, 9.13, 9.14.

Plus the one this branch found and did not fix: `migrate status` exits 1 on a `<version>_x.down.sql` file the pinned binary writes and reads at exit 0.

## Compatibility

Nothing here makes `ptah-compat` exit 0 where the pinned binary exits 1. The one new refusal goes the other way and is argued above. `migrate new` file names change for anyone relying on Ptah's rewriting — pre-v1, so no compatibility with older Ptah is owed, and the compatibility that matters improves. No native `ptah` behavior is weakened: the native `schema apply` sentence, the native paired migration layout and `migrate checkpoint` are all unchanged.

### DIFF
--- a/cmd/atlas/atlas_test.go
+++ b/cmd/atlas/atlas_test.go
@@ -535,7 +535,7 @@ func TestCompatCommand_SchemaApplySchemaShorthandParses(t *testing.T) {
 	// SQLite owns unqualified objects in "main", so a "public" schema scope
 	// selects nothing and the apply reports a synced schema.
 	c.Assert(err, qt.IsNil)
-	c.Assert(out.String(), qt.Contains, "Schema is synced, no changes to be made.")
+	c.Assert(out.String(), qt.Contains, "Schema is synced, no changes to be made")
 }
 
 func TestCompatCommand_SchemaDiffSchemaShorthandParses(t *testing.T) {
@@ -2519,7 +2519,10 @@ CREATE TABLE users (
 	err = second.Execute()
 
 	c.Assert(err, qt.IsNil)
-	c.Assert(secondOut.String(), qt.Equals, "Schema is synced, no changes to be made.\n")
+	// No trailing period: this is the byte-exact form the pinned community
+	// binary v1.3.0 writes for `schema apply` (stokaro/ptah#1235 finding 9.4).
+	// Its `schema diff` answer keeps its period and already matched.
+	c.Assert(secondOut.String(), qt.Equals, "Schema is synced, no changes to be made\n")
 }
 
 func TestCompatCommand_SchemaApplyDryRunDoesNotApply(t *testing.T) {
@@ -3123,7 +3126,11 @@ func TestCompatCommand_MigrateApplyFormatsNoopResult(t *testing.T) {
 	err := cmd.Execute()
 
 	c.Assert(err, qt.IsNil)
-	c.Assert(out.String(), qt.Not(qt.Contains), "No migration files to execute.")
+	// The text printer terminates its line; the same sentence inside the JSON
+	// document is followed by a quote. Asserting on the newline is what keeps
+	// this row meaningful now that the two spellings are the same string
+	// (stokaro/ptah#1235 finding 9.3 removed the text form's trailing period).
+	c.Assert(out.String(), qt.Not(qt.Contains), "No migration files to execute\n")
 	var result atlasMigrateApplyJSONResult
 	c.Assert(json.Unmarshal(out.Bytes(), &result), qt.IsNil)
 	c.Assert(result.Current, qt.Equals, "1")
--- a/cmd/atlas/migrate_apply.go
+++ b/cmd/atlas/migrate_apply.go
@@ -525,7 +525,13 @@ func finishAtlasMigrateApplyFreshNoop(
 	if opts.format != "" {
 		return true, writeAtlasMigrateApplyFormat(cmd, opts, migrationFS, conn, result)
 	}
-	fmt.Fprintln(cmd.OutOrStdout(), "No migration files to execute.")
+	// No trailing period: the pinned community binary v1.3.0 writes
+	// `No migration files to execute\n` on stdout for the same run, 30 bytes,
+	// verified byte for byte with xxd on 2026-08-08 against a directory both
+	// binaries had already applied. Ptah wrote 31 (stokaro/ptah#1235 finding
+	// 9.3). The report model in internal/atlasreport already spells it without
+	// the period, so this line was also the odd one out inside Ptah.
+	fmt.Fprintln(cmd.OutOrStdout(), "No migration files to execute")
 	return true, nil
 }
 
--- a/cmd/atlas/migrate_apply_foreign_flyway.go
+++ b/cmd/atlas/migrate_apply_foreign_flyway.go
@@ -232,7 +232,7 @@ func foreignFlywayRefusal(
 // V1__g1.sql and V3__g3.sql recorded by the other implementation as `1` and `3`
 // with V2__g2.sql added afterwards: the head is V3's 4611686018427551119, and
 // `migrate set 4611686018427551119` reports `(3 set)` — g1, g2 and g3 — so the
-// next `migrate apply` prints `No migration files to execute.` at exit 0 with
+// next `migrate apply` prints `No migration files to execute` at exit 0 with
 // table g2 never created and now unreachable, because its version is recorded.
 // The operator followed the printed instruction and lost a migration; the
 // refusal said nothing about it. That is the shape this clause withdraws on,
--- a/cmd/atlas/migrate_apply_foreign_flyway_test.go
+++ b/cmd/atlas/migrate_apply_foreign_flyway_test.go
@@ -247,7 +247,7 @@ func TestCompatMigrateApply_ForeignFlywayRefusalPrintsWorkingRecovery(t *testing
 
 	stdout, stderr, err = compatApplyConverted(dir, "flyway", dbPath)
 	c.Assert(err, qt.IsNil, qt.Commentf("stdout:\n%s\nstderr:\n%s", stdout, stderr))
-	c.Assert(stdout, qt.Contains, "No migration files to execute.")
+	c.Assert(stdout, qt.Contains, "No migration files to execute")
 	// The seed did not run a second time.
 	c.Assert(countRows(c, dbPath, "seeded"), qt.Equals, 1)
 }
@@ -264,7 +264,7 @@ func TestCompatMigrateApply_ForeignFlywayRefusalPrintsWorkingRecovery(t *testing
 //
 // Measured with the route still offered there: `migrate set 4611686018427551119`
 // reports `(3 set)` — `+ 4611686018427510315 (g2)` among them — the next
-// `migrate apply` prints `No migration files to execute.` at exit 0, and table
+// `migrate apply` prints `No migration files to execute` at exit 0, and table
 // g2 is absent with its version recorded, so it can never run again. Following
 // the printed instruction lost a migration and the refusal said nothing.
 //
@@ -485,7 +485,7 @@ func TestCompatMigrateApply_ForeignFlywayBaselineOverForeignHistory(t *testing.T
 	c.Assert(err, qt.IsNil, qt.Commentf("stdout:\n%s\nstderr:\n%s", stdout, stderr))
 	stdout, stderr, err = compatApplyConverted(dir, "flyway", dbPath)
 	c.Assert(err, qt.IsNil, qt.Commentf("stdout:\n%s\nstderr:\n%s", stdout, stderr))
-	c.Assert(stdout, qt.Contains, "No migration files to execute.")
+	c.Assert(stdout, qt.Contains, "No migration files to execute")
 	c.Assert(userTables(c, dbPath), qt.DeepEquals, []string{"p", "q"})
 }
 
--- a/cmd/atlas/migrate_apply_integrity_test.go
+++ b/cmd/atlas/migrate_apply_integrity_test.go
@@ -134,7 +134,7 @@ func TestCompatMigrateApply_CheckpointPreCheckpointDatabaseSkips(t *testing.T) {
 	// execute", no new revision row, status stays OK.
 	stdout, _, err = compatApply(fullDir, dbPath)
 	c.Assert(err, qt.IsNil, qt.Commentf("output:\n%s", stdout))
-	c.Assert(stdout, qt.Contains, "No migration files to execute.")
+	c.Assert(stdout, qt.Contains, "No migration files to execute")
 	c.Assert(compatRevisionRows(c, dbPath), qt.DeepEquals, seeded)
 
 	statusOut, _, err := runCompat("migrate", "status", "--url", "sqlite://"+dbPath, "--dir", "file://"+fullDir)
@@ -308,7 +308,7 @@ func TestCompatMigrateApply_DirWithoutSQLFilesApplies(t *testing.T) {
 			stdout, stderr, err := compatApply(dir, dbPath)
 
 			c.Assert(err, qt.IsNil, qt.Commentf("stdout:\n%s\nstderr:\n%s", stdout, stderr))
-			c.Assert(stdout, qt.Contains, "No migration files to execute.")
+			c.Assert(stdout, qt.Contains, "No migration files to execute")
 		})
 	}
 }
@@ -380,7 +380,7 @@ func TestCompatMigrateApply_UnhashedDirWithNestedSQLIsNothingToExecute(t *testin
 			stdout, stderr, err := compatApply(dir, dbPath)
 
 			c.Assert(err, qt.IsNil, qt.Commentf("stdout:\n%s\nstderr:\n%s", stdout, stderr))
-			c.Assert(stdout, qt.Contains, "No migration files to execute.")
+			c.Assert(stdout, qt.Contains, "No migration files to execute")
 			c.Assert(stderr, qt.Equals,
 				"warning: sub/20260801100335_init.sql is not covered by atlas.sum and will not run; "+
 					"Atlas migrations are top-level files named *.sql\n")
--- a/cmd/atlas/migrate_apply_legacy_flyway_test.go
+++ b/cmd/atlas/migrate_apply_legacy_flyway_test.go
@@ -136,7 +136,7 @@ func TestCompatMigrateApply_LegacyFlywayRefusalPrintsWorkingRecovery(t *testing.
 	stdout, stderr, err := runCompat("migrate", "apply", "--url", "sqlite://"+dbPath, "--dir", "file://"+dir+"?format=flyway")
 
 	c.Assert(err, qt.IsNil, qt.Commentf("stdout:\n%s\nstderr:\n%s", stdout, stderr))
-	c.Assert(stdout, qt.Contains, "No migration files to execute.")
+	c.Assert(stdout, qt.Contains, "No migration files to execute")
 	// The seed did not run a second time.
 	c.Assert(countRows(c, dbPath, "seeded"), qt.Equals, 1)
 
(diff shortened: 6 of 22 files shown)


denisvmedia:
## Head moved to a merge with `dfbec226`, and the merged tree was re-verified

`master` advanced to `dfbec226` (#1335, which reworks `--exclude` selector spellings) between the last local fetch and the push, so the branch was brought up to date. Head is now `03dd0116`, a merge of `d53bd65e` into `dfbec226`.

A clean merge is not a compiling merge, so the merged tree — not the branch tip, the actual PR head — was checked out locally and every gate re-run on it. Exit codes read directly, never through a pipe.

`go build ./...` 0 · `go vet ./...` 0 · `go vet -tags integration ./integration/...` 0 · `go tool qtlint ./...` 0 · `golangci-lint run ./...` 0 · `scripts/check-test-style.sh` 0 · `scripts/check-public-api.sh` 0 · `scripts/check-public-api-snapshot.sh` 0 · `scripts/check-public-api-released.sh` 0.

`docs/site`: `check:exit-codes` 0 · `check:core-doc-links` 0 · `check:page-health` 0 · `check:links` 0 · `check:redirects` 0 · `check:style` 0 · `check:limitations` 0 · `npm run build` 0 · `check:responsive` 0.

Conflict-marker grep over the tree is empty, and each of the six doc edits was confirmed present after the merge rather than assumed.

### `go test ./...` needed two runs, and the reason is not this branch

First run exited **1**, on one subtest in `cmd/viz`, a package this branch does not touch:

```
--- FAIL: TestSVGReportsGraphvizStderrOnFailure (10.01s)
    got error:  "render SVG with Graphviz dot: signal: killed"
    regexp:     "render SVG with Graphviz dot: .*: graphviz exploded"
```

`signal: killed` at exactly 10.01s is the test's own deadline killing an external `dot` process, not the stderr assertion it is written to make. Re-run in isolation it passes in 0.14s, exit **0**; the full suite re-run then exited **0** as well. Same flake PR #1248 chased down and documented. Both results reported rather than only the green one.

### The behavior claims were re-measured on the merged head

The binary was rebuilt from `03dd0116` and the probes re-run, because #1335 touches `--exclude` and finding 6.1 has an `--exclude` shape:

| Probe | Result on `03dd0116` |
| --- | --- |
| `schema inspect -u sqlite://empty.db --format '{{ json . }}'` | `{"schemas":[{"name":"main"}]}` exit 0 |
| the same on `basic.db --exclude users --exclude posts` | `{"schemas":[{"name":"main"}]}` exit 0 |
| `schema inspect -u sqlite://uniq.db --format '{{ json . }}'` | 3 index entries, no repeated autoindex, exit 0 |
| `migrate new "add users table" --dir file://migrations` | `<version>_add users table.sql` exit 0 |

### Checks

20 check runs on `03dd0116`, all completed: 18 success, 2 skipped (`Deploy to GitHub Pages` and `release`, which are the expected pull-request skips). Zero failures, zero pending. Counted rather than read off the merge state — and none of these can see a parity defect, which is why every claim above comes from a live binary instead.

denisvmedia:
## The 6.1 fix opened a new divergence on the verb it was meant to close, and has been removed

The comment above this one, and the description it described, are superseded. Head is now `dbf0bb91`, rebased onto `origin/master` `71a7d7f2`. The description has been rewritten to match what was measured; this comment records what moved and why, so the thread does not read as if `03dd0116` were still the claim.

### What was wrong

`03dd0116` closed finding 6.1 by seeding the CONNECTED schema into the realm document:

```go
if connected := atlasSchemaInspectSchemaName("", info); connected != "" {
    atlasSchemaInspectSchemaForName(schemasByName, connected)
}
```

That closes the empty-database case. It also changes the answer for cases that were already correct, because which schemas exist is the READER's answer and not the connection's. Measured on live PostgreSQL 17 in a container created for the run, one database per shape, `cmp` against the pinned Atlas community binary v1.3.0's own stdout, every exit code read from an unpiped invocation — all of these exit 0 on both sides, so only the bytes differ:

| `schema inspect --format '{{ json . }}'` | pinned binary | `03dd0116` |
| --- | --- | --- |
| realm URL, `--schema extra`, tables in `extra` | `{"schemas":[{"name":"extra","tables":[…]}]}` | the same **plus a second, empty `{"name":"public"}`** |
| realm URL, `--schema nosuch` | `{}` | `{"schemas":[{"name":"public"}]}` |

The second one describes a schema to an operator who has just been told it does not exist. Both shapes already matched the binary on `master` alone, without this branch.

### Why the cell is still closed

6.1 is the same divergence as #1264, and #1275 closed it on `master` on the correct axis: the realm is built from the schemas the reader described, which also carries the schema `comment`, `charset` and `collate` the seed never did. Re-measured on this head, empty databases answer `{"schemas":[{"name":"main"}]}` on SQLite and `{"schemas":[{"name":"public","comment":"standard public schema"}]}` on PostgreSQL 17, byte-identically to the binary.

So this branch no longer claims 6.1. It claims 6.2, 8.6, 8.7, 9.3, 9.4.

### The rebase is where this would have slipped through

`git rebase` auto-merged the **body** of `atlasSchemaInspectJSON` — keeping master's `schema.Schemas` loop *and* the seed — and reported a conflict only in the doc comment above it. Resolving that comment either way yields a tree that builds, vets and passes `go build`, and reintroduces both divergences. The resolution deletes the seed; `internal/atlasreport/schema_inspect.go` against `master` is now the 6.2 deduplication plus one doc paragraph, nothing else.

### Two more things the re-measurement turned up

**9.3 had no assertion that could fail on its own defect.** Every assertion on that sentence in `cmd/atlas` reads `qt.Contains, "No migration files to execute"`, which matches just as happily when the line carries the trailing period. Putting the period back into `migrate_apply.go` and running the whole package reddened **nothing**. `TestCompatMigrateApply_NoopLineIsByteExact` now asserts the entire stdout of a second apply equals the 30 bytes read off the binary with `xxd`, and it is red with the period and green without. 9.4 already had such an assertion (`TestCompatCommand_SchemaApplyAppliesLocalSchemaToSQLite`), confirmed red the same way.

**The 9.4 byte counts in the previous description were wrong.** The synced `schema apply` line is 40 bytes on the pinned binary and was 41 in Ptah, not 41 and 42. Re-read with `xxd` and `wc -c`; the code comment in `cmd/atlas/schema_apply.go` is corrected too.

### Gates, on the rebased tree, read directly

`go build ./...` 0 · `go vet ./...` 0 · `go vet -tags integration ./integration/...` 0 · `go test ./... -count=1 -timeout 40m` 0 (187 packages `ok`, zero `FAIL` lines, first run) · `go tool qtlint ./...` 0 · `golangci-lint run ./...` 0 · `scripts/check-test-style.sh` 0 · `scripts/check-public-api.sh` 0 · `scripts/check-public-api-snapshot.sh` 0 · `scripts/check-public-api-released.sh` 0 · `docs/site` `npm ci`, every `check:*`, every `*:selftest` and `npm run build` 0.

None of those can see a parity defect — the one repaired here was green on all of them at `03dd0116`. Every behavior claim above comes from the pinned binary and a live database.

Issue #887, opened 2026-07-30
Atlas non-transactional apply must persist resumable partial progress


denisvmedia:
## Summary

Ptah writes an Atlas-format dirty revision with `applied > 0` but `partial_hashes = null` when a non-transactional migration fails after at least one successful statement. Atlas CE cannot resume from that metadata.

Discovered during the final hardening of #276 through black-box comparison against the pinned Atlas CE v1.2.0 binary. This issue must be implemented from observed behavior and public fixtures only; do not depend on Atlas implementation code.

## Reproduction

Use one Atlas-format migration with two statements:

```sql
CREATE TABLE first_success (id INTEGER PRIMARY KEY);

CREATE TABLE second_success (id INTEGER PRIMARY KEY);
```

1. Pre-create `second_success`.
2. Run `ptah-compat migrate apply --tx-mode none --allow-dirty`.
3. The first statement succeeds and the second fails.
4. Inspect `atlas_schema_revisions`: Ptah records `applied = 1`, `total = 2`, and JSON `null` in `partial_hashes`.
5. Drop `second_success` to remove the failure condition.
6. Run the pinned Atlas CE `migrate apply --tx-mode none --allow-dirty` against the same database.

Observed result: Atlas CE cannot resume from the Ptah-written revision and terminates with an index-out-of-range panic.

Control behavior: when Atlas CE creates the failed row, it stores a one-element cumulative hash array in `partial_hashes`; after the blocking table is dropped, a second Atlas CE apply executes only the remaining statement and completes successfully.

## Scope

- Persist Atlas-compatible partial progress after every successfully executed statement in non-transactional apply mode.
- Keep `applied`, `total`, `error`, `error_stmt`, and `partial_hashes` mutually consistent.
- Close the crash window where schema changes are committed but revision progress remains at zero.
- Derive statement boundaries and cumulative hashes from Ptah's clean-room parser/fixture behavior. Do not import or depend on Atlas source code.
- Preserve native Ptah revision behavior.

## Acceptance

- [ ] A pinned Atlas CE black-box control records the expected partial-progress shape.
- [ ] Ptah's failed row has the same stable metadata semantics, including a cumulative hash for each applied statement.
- [ ] Atlas CE resumes a Ptah-failed migration after the external failure condition is removed.
- [ ] Ptah resumes an Atlas-failed migration under the same scenario.
- [ ] A crash/fault-injection test proves progress is durable after each committed non-transactional statement.
[...]

PR #1356, opened 2026-08-09
Use transaction witnesses for MySQL file progress

denisvmedia:
Refs #887.

## Problem

MySQL and MariaDB can commit an open transaction before DDL and then execute
later statements in autocommit mode. A static SQL keyword classifier cannot
prove the durable contiguous prefix after a failure. Understating that prefix
replays committed SQL; overstating it skips rolled-back SQL.

## Design

This change removes durable-prefix inference from SQL classification. In
MySQL-family `tx-mode file`, Ptah pins one physical session and records a
revision witness in the same InnoDB transaction as the migration body:

1. Before each statement, the revision records an unknown outcome.
2. After success, the same transaction records the applied count and partial
   hashes.
3. An implicit commit makes the witness and the preceding user SQL durable
   together.
4. An ordinary rollback removes the uncommitted tail together.
5. Recovery reads the durable witness and resumes only after a proven prefix.

The mechanism applies to up and down execution and to native and Atlas-shaped
revision tables. Down validation and session preflight complete before Ptah
marks the revision dirty. A failed statement whose lack of effects cannot be
proved leaves an unknown outcome and blocks automatic retry instead of
guessing.

## Safety boundaries

The witness contract is enabled only when Ptah can prove its durability model:

- Revision metadata, the session default, and every existing base table in the
  selected database must use InnoDB.
- Explicit non-InnoDB engines and `CREATE TABLE ... LIKE` are rejected.
- The migration account must hold `TRIGGER` at database or global scope, since
  MySQL and MariaDB hide trigger metadata from an account without it while
  those triggers still fire during ordinary DML.
- Transaction controls, `USE`, durable server-state changes, and database
  catalog changes are rejected.
- Cross-database relation and routine references are rejected.
- Executable comments, nested or dynamic SQL, table locks, indirect database
  objects, trigger-bearing relations, and stored-routine calls are rejected.
- Opaque `MigrationFunc` callbacks and statement interceptors must use
  `tx-mode none`, in both directions.

Checks run before body SQL and again on the pinned session. A statement-level
rejection names the statement number and its safety class; a `MigrationFunc`
or a statement interceptor is refused for the whole migration and named by
direction, because neither has a statement to point at. Neither form echoes
the SQL, which may contain credentials. Safe session settings are replayed
when a proven durable prefix resumes on a new session.

## Reading a granted database the way the server does

A schema-level privilege stores its database as a pattern, not as a name: `_`
and `%` are wildcards unless a backslash escapes them. A grant on a literal
database name therefore escapes the underscores in it, and the official MySQL
and MariaDB container images do exactly that when they provision `MYSQL_USER`,
so `SHOW GRANTS` prints `` `ptah\_test` `` for a privilege that covers
`ptah_test`. The trigger-privilege preflight compared that to the schema as a
string, which refused `tx-mode file` for an account that does hold `TRIGGER`
and, in the other direction, failed to see a `REVOKE` written the same way.
The granted database is now matched with the server's pattern semantics.

Every existing row for that classifier used a schema name with no underscore,
which is why nothing caught it. The new rows all use one, and they separate
the pattern from both a name comparison and a plain unescape.

## Tests

The live suite runs against MySQL 9.7 and MariaDB, and covers:

- pure-DML rollback and retry from the top;
- DDL followed by durable DML, with retry beginning at statement 3 and no
  duplicate insert;
- fail-closed recovery from a failed DDL with an unknown outcome;
- up and down progress with native and Atlas-shaped revision layouts;
- session-state replay;
- revision metadata, default-engine, target-table, explicit-engine, and
  `CREATE TABLE ... LIKE` rejection paths;
- transaction control in the body, including `COMMIT AND CHAIN` and
  `ROLLBACK AND CHAIN`, which end the file transaction and open another one on
  the same session. A classifier that only recognizes a bare `COMMIT` or
  `ROLLBACK` lets those through, and `MigrateUp` then returns nil for a body
  whose leading statements were discarded and whose trailing statements
  committed. The live cases count the surviving rows, because the returned
  error cannot separate that outcome from a correct run: there is none;
- executable comments, opaque callbacks, statement interceptors in both
  directions, nested SQL, cross-database writes and database creation,
  trigger-bearing relations, stored functions, and secret redaction.

The `integration-tests` job runs a dedicated `TestRolledBackProgress_` step for
MySQL and MariaDB, with a test-list guard so the live cases cannot silently
disappear.

## Documentation

The versioned apply guide, Atlas-compatible migrate guide, package README, and
package documentation describe the witness behavior, fail-closed boundaries,
unknown-outcome recovery, the `TRIGGER` privilege requirement, and the lack of
MySQL-family `tx-mode all` support.

## Local validation

Run on this branch after rebasing onto master, each exit code read directly:

- `go build ./...`, `go vet ./...`, `go vet -tags integration ./integration/...`
- `go test ./... -count=1`
- `go tool qtlint ./...`, `golangci-lint run ./...`
- `scripts/check-test-style.sh`, `scripts/check-public-api.sh`,
  `scripts/check-public-api-snapshot.sh`, `scripts/check-public-api-released.sh`
- every `check:*` script in `docs/site/package.json`
- `go test -v -count=1 ./migration/migrator -run '^TestRolledBackProgress_'`
  against MySQL 9.7.1 with MariaDB 11.4.12, and again against MySQL 9.7.1 with
  MariaDB 10.11.18, the tag CI pins: 32 tests, 32 PASS, no FAIL and no SKIP in
  either run
- `go test -v -count=1 ./migration/migrator -run '(MySQL|MariaDB)'` against the
  same servers: 76 PASS, no FAIL and no SKIP

GitHub CI on the final commit remains the merge gate.

### DIFF
--- a/.github/workflows/go-integration-tests.yml
+++ b/.github/workflows/go-integration-tests.yml
@@ -324,6 +324,14 @@ jobs:
         run: |
           go test -v -count=1 ./migration/migrator -run 'TestMigrateUp_PostgresLockTimeoutIntegration|TestOutOfOrderMigrationsPostgresIntegration|TestMigrationAdvisoryLock_PostgresConcurrentRunners|TestMigrationAdvisoryLock_PostgresTimeoutIntegration|TestMigrationAdvisoryLock_MySQLDefaultTimeoutIntegration|TestMigrationAdvisoryLock_MariaDBDefaultTimeoutIntegration|TestDirtyMigrationState_PostgresFailureRollsBackAndBlocksRetry|TestDirtyMigrationState_MySQLRepairReachesHead|TestDirtyMigrationState_MariaDBRepairReachesHead|TestMigrationChecksumMismatch_PostgresIntegration|TestBaseline_|TestAtlasRevisionMetadata_(Postgres|MySQL|MariaDB|CockroachDB|YugabyteDB|ClickHouseRejectsSet)Integration|TestAtlasDotMetadata_(Postgres|MySQL)Integration|TestAtlasTxtarChecks_((Postgres|MySQL|MariaDB)(SessionLockReleased|CommentSemantics|EscapeString|ExecutableCommentEscape|ShortNumericPrefixRejectsNonSelect)?|ClickHouse|CockroachDB|YugabyteDB)Integration|TestAtlasRevisionTable_(Postgres|CockroachDB|YugabyteDB)UsesTimestamptzIntegration|TestAtlasSetSerializable_PostgresConcurrentInsertIntegration|TestDryRunRevisionState_(Postgres|MySQL|MariaDB|CockroachDB|YugabyteDB|ClickHouse)Integration' -timeout 3m
 
+      - name: Run MySQL-family rolled-back progress tests
+        env:
+          MYSQL_TEST_URL: "mysql://ptah_user:ptah_password@tcp(127.0.0.1:3306)/ptah_test"
+          MARIADB_TEST_URL: "mariadb://ptah_user:ptah_password@tcp(127.0.0.1:3307)/ptah_test"
+        run: |
+          go test ./migration/migrator -list '^TestRolledBackProgress_' | grep -q '^TestRolledBackProgress_'
+          go test -v -count=1 ./migration/migrator -run '^TestRolledBackProgress_' -timeout 3m
+
       - name: Run no-transaction and shadow generator integration tests
         env:
           POSTGRES_TEST_DSN: "postgres://ptah_user:ptah_password@localhost:5432/ptah_test?sslmode=disable"
--- a/.github/workflows/go-integration-tests.yml
+++ b/.github/workflows/go-integration-tests.yml
@@ -327,7 +327,9 @@ jobs:
       - name: Run MySQL-family rolled-back progress tests
         env:
           MYSQL_TEST_URL: "mysql://ptah_user:ptah_password@tcp(127.0.0.1:3306)/ptah_test"
+          MYSQL_ADMIN_TEST_URL: "mysql://root:root_password@tcp(127.0.0.1:3306)/ptah_test"
           MARIADB_TEST_URL: "mariadb://ptah_user:ptah_password@tcp(127.0.0.1:3307)/ptah_test"
+          MARIADB_ADMIN_TEST_URL: "mariadb://root:root_password@tcp(127.0.0.1:3307)/ptah_test"
         run: |
           go test ./migration/migrator -list '^TestRolledBackProgress_' | grep -q '^TestRolledBackProgress_'
           go test -v -count=1 ./migration/migrator -run '^TestRolledBackProgress_' -timeout 3m
--- a/docs/site/src/content/docs/atlas/migrate-commands.md
+++ b/docs/site/src/content/docs/atlas/migrate-commands.md
@@ -164,16 +164,30 @@ the zero-progress revision row. The next apply retries the whole file without
 `--allow-dirty`, matching Atlas. Native `ptah migrations up` keeps its durable
 failure record instead.
 
-On MySQL and MariaDB a rollback does not always reach zero progress. The server
-commits the open transaction on its own before a DDL statement, and that commit
-*ends* the transaction rather than flushing it: every statement after the DDL
-runs with no transaction open and commits itself, so the rollback reaches none
-of them either. `ptah-compat` records that whole committed prefix — and its
-`partial_hashes` — and the row stays, because retrying the file whole would
-repeat SQL the server has already committed. A body that reached no such
-statement leaves nothing behind and its row is removed like any other
-rolled-back failure. A statement Ptah cannot classify counts as rolled back, so
-a retry repeats it rather than skipping it on a guess.
+On MySQL and MariaDB a `file` rollback does not always reach zero progress: the
+server may commit around DDL. `ptah-compat` does not guess the boundary from SQL
+keywords. The Atlas revision row is an InnoDB witness updated on the same
+physical transaction before and after each statement. An implicit server commit
+makes the matching `applied` count and `partial_hashes` durable with the body;
+an ordinary rollback removes both the user DML and its witness. A plain-DML body
+that rolls back leaves no committed prefix and is retried whole. A witnessed
+DDL/DML prefix stays dirty because retrying it would repeat committed SQL.
+
+This recovery mode requires InnoDB for the Atlas revision table, the session's
+default storage engine, and every existing base table in the selected database.
+An explicit non-InnoDB `CREATE`, `ALTER`, or storage-engine setting is refused
+before the migration runs. `CREATE TABLE ... LIKE` is also refused because the
+new table inherits an engine that the session default does not prove.
+
+MySQL-family `file` bodies cannot contain transaction-control SQL such as
+`BEGIN`, `COMMIT`, or `SET autocommit`; Ptah owns the file transaction. Durable
+server-state operations such as `SET GLOBAL`, `SET PERSIST`, and `RESET` are
+refused because their effects are not tied to the InnoDB witness. `USE` is also
+refused; the connection URL must name the database whose engines Ptah validates.
+MySQL and MariaDB do not support `--tx-mode all`. Ordinary session settings
+remain valid. Ptah runs the body on one pinned session, discards it afterward,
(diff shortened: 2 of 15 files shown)


denisvmedia (commented):
Blocking review: the new rollback-prefix classifier can still make an automatic retry skip SQL that never committed.

1. `SET autocommit` is classified from its name alone. In particular, `SET autocommit=0` reaches lines 50-52 and is reported as an implicit commit. I reproduced this on the PR's MySQL 9.7 container:

```sql
START TRANSACTION;
INSERT INTO ledger VALUES (1);
SET autocommit=0;
INSERT INTO missing VALUES (2); -- fails
ROLLBACK;
SELECT COUNT(*) FROM ledger;    -- 0
```

For this body the current code returns `applied=2`. A dirty retry therefore starts at statement 3 and permanently skips the rolled-back INSERT. The official MySQL rule is narrower: `SET autocommit = 1` commits only if the value is not already 1. A SQL-only classifier does not know the prior session value, so this must be treated as state-dependent rather than affirmative evidence of a committed prefix.

2. The prefix does not end at the last DDL statement. A successful DDL ends the explicit transaction; with normal autocommit state, later DML commits statement by statement. Live MySQL 9.7 probe:

```sql
START TRANSACTION;
CREATE TABLE boundary (id INT);
INSERT INTO ledger VALUES (1);
INSERT INTO missing VALUES (2); -- fails
ROLLBACK;
SELECT COUNT(*) FROM ledger;    -- 1
```

`committedPrefixAfterRollback` returns 1, but 2 statements are durable. Retrying at statement 2 repeats an already committed INSERT and can fail forever or duplicate non-idempotent effects. This needs a transaction-state model, or an explicit refusal to auto-resume when the durable contiguous prefix cannot be proven. "Last statement that forced a commit" is not enough.

3. The documented sets are not one unconditional MySQL-family set. MySQL `LOAD DATA` commits only for NDB tables, `UNLOCK TABLES` commits only after qualifying table locks, and `RESET PERSIST` is explicitly excluded. MariaDB has a different list. Each current false positive is the dangerous direction because it can advance `applied` past rolled-back work.

4. The docs repeat the incorrect model and also describe MySQL/MariaDB `tx-mode all`, while `validateTxModeAllDialect` rejects those dialects.

The no-transaction `partial_hashes` crash assertion is useful and should stay. Before merge, add inverse unit tests for `SET autocommit=0`, `RESET PERSIST`, conditional `UNLOCK TABLES`, and MySQL non-NDB `LOAD DATA`; add a live DDL-then-DML failure/retry test; and rewrite the docs to state only behavior the implementation can prove.

References:
- https://dev.mysql.com/doc/refman/9.7/en/implicit-commit.html
- https://mariadb.com/docs/server/reference/sql-statements/transactions/sql-statements-that-cause-an-implicit-commit

denisvmedia:
Addressed the blocking review by removing the static rollback-prefix classifier
from the correctness path rather than extending its keyword table.

1. `SET autocommit=0` and all transaction controls are now rejected before the
   migration body runs. Ptah no longer treats the spelling of a statement as
   evidence that a prefix committed.
2. DDL followed by autocommitted DML is covered by live MySQL and MariaDB tests.
   Progress comes from a revision witness/checkpoint written on the same pinned
   session and transaction as user SQL, so both durable statements are observed.
3. Conditional and server-specific implicit-commit forms no longer need to be
   inferred. Statements that can invalidate the witness model are rejected;
   session-local settings are replayed explicitly for retries.
4. The docs now describe only `tx-mode file` witness behavior for the MySQL
   family and no longer imply `tx-mode all` support.

The repair also adds InnoDB preflight for revision metadata and every existing
base table in the selected database, rejects non-InnoDB creation and
`CREATE TABLE ... LIKE`, wires the same mechanism for down migrations and both
revision layouts, and adds a dedicated CI job plus test-list guard for the live
suite.

Local validation included the full repository suite, race tests, lint, vet,
test-style, docs-style, public API gates, and live MySQL 9.7/MariaDB 11.4 tests.
After rebasing over the latest branch tip, the migrator package was rerun and is
green. The rebased commit is `e2f258c0`; GitHub CI is running now.

denisvmedia:
The reported DDL-then-DML duplication is valid for the previous tip
`a22484ff`, which inferred the durable prefix from SQL. That implementation is
no longer the PR tip. Commit `e2f258c0` replaces the classifier correctness path
with a same-transaction revision witness.

I reran the exact stronger contour on the current SHA against both live servers:

```text
TestRolledBackProgress_MySQLDDLThenDMLKeepsTheWholeDurablePrefix       PASS
TestRolledBackProgress_MariaDBDDLThenDMLKeepsTheWholeDurablePrefix     PASS
TestRolledBackProgress_MySQLNativeRevisionKeepsTheWholeDurablePrefix   PASS
TestRolledBackProgress_MariaDBNativeRevisionKeepsTheWholeDurablePrefix PASS
```

Each test executes successful DDL, then a successful autocommitted INSERT, then
a failing statement. After failure it asserts the table exists, the INSERT
exists exactly once, and the revision says `applied=2`. The retry logs
`resumeFromStatement=3`, completes the failing statement after its dependency is
created, and asserts the earlier INSERT is still present exactly once.

The draft remains open while the new SHA's remaining CI jobs and independent
review finish, not because the old classifier defect remains.

denisvmedia:
The reported MySQL/MariaDB regression was valid on the old `a22484ff` tip: it
inferred the durable prefix from statement shape and could replay a committed
`INSERT` after DDL.

The current tip `62debd97` no longer makes that inference. It records an
unknown-before / applied-after witness in the same InnoDB transaction as the
migration body. The live MySQL 9.7 and MariaDB 11.4 matrix now proves the exact
sequence from the report: after DDL commits the prefix, retry starts at
statement 3 and the first `INSERT` is not duplicated.

The follow-up review also found and fixed additional proof-boundary gaps:

- opaque up/down `MigrationFunc` callbacks and statement interceptors;
- executable comments, `CALL`, prepared/dynamic SQL, and table locks;
- cross-database relations/routines and database-catalog changes;
- existing views, trigger-bearing relations, and stored routines;
- unknown outcomes for failed DDL and other non-atomic statements;
- down metadata being dirtied before pinned-session preflight;
- rejected SQL leaking credentials;
- false-positive catalog matches on column and alias names.

Validation on the final commit:

- full `go test -count=1 ./...` passed;
- `go test -race -count=1 ./migration/migrator` passed;
- the complete `TestRolledBackProgress_` suite passed against both live
  databases, including the new fail-closed cases;
- both `golangci-lint` passes, `qtlint`, test-style, and docs-style checks
  passed.

The PR remains draft until CI for `62debd97` is fully green.

denisvmedia:
The previous integration run failed in fixture setup, not in the transaction
witness cases: the regular `ptah_user` account cannot create external databases,
triggers, or stored routines. Commit `2df55c3f` gives only those setup operations
an explicit admin connection; the migrator itself and revision assertions still
run as `ptah_user`.

That privilege split exposed another real fail-open condition. The old
cross-database proof enumerated `information_schema.schemata`, but MySQL and
MariaDB hide databases that the current account cannot access. A qualified
external relation could therefore reach execution instead of being rejected by
Ptah. The proof now compares qualified relation and routine references directly
with the database selected by the connection URL, independent of catalog ACLs.

The same review found that trigger metadata is hidden without `TRIGGER`, even
though a hidden trigger can still run during ordinary DML. File mode now proves
database/global trigger visibility from the server's effective `SHOW GRANTS`
output, including active roles and partial revokes, and fails closed otherwise.
View discovery uses the general object catalog so it does not require `SHOW
VIEW`. Server-generated grant identifiers are handled under both normal and
`ANSI_QUOTES` session modes; the committed-session-state retry test covers that
mode change.

Additional live coverage now verifies on MySQL 9.7 and MariaDB 11.4 that:

- privileged objects are prepared by admin while migration execution remains
  unprivileged;
- an account without `TRIGGER` is rejected before body SQL or revision progress;
- external databases invisible to `ptah_user` are still rejected syntactically;
- session-setting replay remains valid after `ANSI_QUOTES` changes `SHOW GRANTS`
  rendering.

Final local validation for `2df55c3f` passed:

- `go test -count=1 ./...`;
- `go test -race -count=1 ./migration/migrator`;
- the complete live `TestRolledBackProgress_` matrix on both databases;
- both `golangci-lint` passes, `gopls check`, `qtlint`, test-style, and docs-style.

The PR remains draft until this SHA's CI is fully green.

PR #1367, opened 2026-08-09
Harden MySQL-family transaction witness safety

denisvmedia:
## Summary

- close the remaining ways a MySQL-family tx-mode file body could reach the server without being covered by the InnoDB revision witness that #1356 introduced
- fail closed when migration SQL, catalog visibility, storage engines, session modes, external objects, indirect writers, or filesystem effects cannot be tied safely to that witness
- replace the workflow's 27 hand-maintained `-list … | grep -q …` guards, and its two `--- PASS:` log greps, with one derived assertion per step

## What this branch does and does not change

The witness itself is not the subject. It landed in #1356 (`d5b2b446`) and this branch is built on a master that contains it. **The defect the previous version of this section described no longer reproduces**, and the section has been rewritten accordingly.

That claim was measured rather than reasoned. The sequence in question:

```sql
START TRANSACTION;
INSERT INTO ledger VALUES (1);
CREATE TABLE ddl1 (id INT);
INSERT INTO ledger VALUES (3);
ROLLBACK;
```

Both rows survive the rollback on both servers — that part is a property of the server and is unchanged:

```
$ docker exec <mysql> mysql -uroot -p… ptah_test -e "…same sequence…; SELECT id FROM ledger ORDER BY id;"
id
1
3
```

identical on MySQL 9.7.1 and MariaDB 10.11.18.

What matters is whether Ptah then resumes from the wrong boundary. Driven end to end through `MigrateUp` in tx-mode file, with a fourth statement that fails so the body rolls back, against a binary built from `origin/master` at `1fd469be`:

```
catalog after first run: ledger notes = [one three]
catalog after first run: revision applied = 3, total = 4
catalog after retry:     ledger notes = [one three]
catalog after retry:     revision applied = 4, total = 4
```

`applied = 3` is the true durable prefix, the retry resumed at statement 4, and no committed DML was repeated. Same result on MariaDB. So the boundary is already correct on master, for both the Atlas and the native revision formats.

To confirm the harness could see the defect if it were present, the witness read was replaced with the cheaper dialect-only fallback that predates it. The retry then resumed at statement 1 and replayed committed DML on both servers:

```
catalog after first run: revision applied = 0
catalog after retry:     ledger notes = [one three one]
```

So: #1356 closed the boundary arithmetic. **What remained, and what this branch is for, is the set of ways a body could sidestep the witness rather than mis-count against it** — statements whose effects the witness cannot cover, and preflight checks that could be talked out of firing.

## Safety hardening

- require InnoDB metadata, session default, and target tables
- reject `DEFAULT` engine resets whose effective value can differ from preflight
- reject every in-body `sql_mode` assignment and initial parser-changing `ANSI_QUOTES`, `MSSQL`, or `NO_BACKSLASH_ESCAPES`
- recognize MariaDB `ALTER TABLE ... STORAGE ENGINE=...` and reject unsafe or `DEFAULT` selections
- reject transaction controls, durable server state, executable comments, nested SQL, opaque executors, unsafe engines, cross-database references, indirect writers, stored routines, and file writes
- require provable MySQL trigger-catalog visibility, including exact decoded grant scope

## CI guards

The step this branch first wrote named 17 tests one per line. `go test ./migration/migrator -list '^TestRolledBackProgress_'` reports **49**, so 32 were unguarded, and the lines could only ever notice a name somebody had remembered to add.

`go test -list` also enumerates *registered* tests without running them, so the guard could not see the failure it existed to prevent. Measured: with one test in the family pointed at a DSN variable the job does not set, the test skipped and the guard line still exited 0.

`scripts/require-tests-ran.sh` replaces all 27 of those lines, and the two `--- PASS:` greps in the RLS spelling step, with one property per step:

> every test the pattern selects ran, and none skipped

It counts what `-list` reports, runs the same pattern with `-v`, and requires the same number of top-level results with zero skips. Nothing is piped: each command writes a file and its status is read on the next line, because `$?` after a pipeline reports `grep`, which is the same defect class the guards were there to close. A pattern matching nothing is an error rather than the exit 0 and `no tests to run` that `go test` gives.

`--declared-in` covers the one case counting cannot. A test renamed out of the family drops the listed count and the result count together, so they still agree; naming the file the family lives in makes that file the authority on its membership.

Broken three ways, restored between each, exit codes read directly:

| break | result |
| --- | --- |
| one test renamed out of the family | exit 1, `49 tests are declared in … but ^TestRolledBackProgress_ selects only 48` |
| one test skipped for a missing DSN | exit 1, `1 of 49 tests skipped; a skip reads as a pass` |
| pattern matching nothing | exit 1, `matches no test`, where plain `go test -run` exits 0 |

## Verification

- `go build ./...`, `go vet ./...`, `go vet -tags integration ./integration/...`
- `go test ./... -count=1`
- `go test -race ./migration/migrator -count=1`
- all 49 `TestRolledBackProgress_` against MySQL 9.7.1 and MariaDB 10.11.18: 49 passed, 0 failed, 0 skipped
- every converted step run against live PostgreSQL 18, ClickHouse 26.7, MySQL 9.7.1 and MariaDB 10.11.18, each reporting `listed == ran` with zero skips
- `golangci-lint run ./...`, `go tool qtlint ./...`
- `scripts/check-test-style.sh`, the three `check-public-api*` scripts
- every `check:*` in `docs/site/package.json`

The MySQL-family step's timeout is 5m rather than the earlier 3m. That is headroom for a shared runner, not a measured requirement: the 49 tests take about 21s of package time locally, and the step comment now says so.

### Not covered

The steps in this file that never had a `-list` guard are still unconverted — a test that skips inside one of them is still invisible. They span SQL Server, CockroachDB, YugabyteDB and the legacy ClickHouse service, so whether each currently skips could not be measured here, and converting them blind could redden CI for a reason unrelated to this branch. Worth a follow-up that takes them a step at a time.

Related: #887


denisvmedia:
## Review: the CI step replaces a derived guard with a hand-maintained list, and the list covers under a third of the tests

The Go change is not what this comment is about. The workflow step is, because it is the part that decides whether any of this is ever measured again.

### What changed

```diff
-  go test ./migration/migrator -list '^TestRolledBackProgress_' | grep -q '^TestRolledBackProgress_'
-  go test -v -count=1 ./migration/migrator -run '^TestRolledBackProgress_' -timeout 3m
+  [14 lines, one exact test name each]
+  go test -v -count=1 ./migration/migrator -run '^TestRolledBackProgress_' -timeout 5m
```

The line being replaced was already the derived form. It was prefix-keyed, so it survived a rename and covered a new test the day it was added — weakly, since it only asserted that at least one such test existed, but it covered them. Fourteen exact names break on every rename and cover no addition at all.

### The numbers

Counted on this branch, head `0685c20ce775cba914495b26c330407f879ba9de`, by reading `migration/migrator/rolled_back_progress_integration_test.go` in full — 85,362 bytes, 2,277 lines, computed blob SHA-1 `aad6010e87cc59918c63ff8e58d4d940d60a4c64` matching the SHA the API reported for the blob, so nothing was truncated. All 58 other `*_test.go` files in `migration/migrator/` were scanned as well; none declares one of these.

| | count |
| --- | --- |
| top-level `TestRolledBackProgress_` functions on this branch | **43** |
| named in the guard list | **14** |
| exist but unguarded | **29** |
| guarded but nonexistent | 0 |

All 43 are executed by the `-run` line. So the list is not a subset chosen for a reason — it is a subset chosen by whoever was editing at the time, and it was already two-thirds incomplete on the commit that introduced it. Among the unguarded are pairs where one half is guarded and the other is not: `MySQLRejectsDefaultStorageEngineReset` is in the list, `MariaDBRejectsDefaultStorageEngineReset` is not; same for `RejectsUnsafeSQLModeChange` and `RejectsUnsafeInitialSQLMode`. Nothing distinguishes the members from the non-members.

### The guard proves registration, not execution

`go test -list` enumerates registered test functions. It does not run them. A test that skips itself because its DSN is absent is listed exactly like one that runs.

That matters because the failure this guard exists to prevent is an execution failure — this repository's own workflow comment says a skip reads as a pass. Fourteen lines that look like evidence the matrix ran establish only that it compiles. The step that does run them never inspects the output for `--- SKIP`, so a run in which all 43 skipped exits 0 today.

The real hole the old guard covered is separate and genuine: `go test -run` matching nothing exits 0 and prints `no tests to run`. That is worth guarding. It is one assertion, not fourteen.

### What the step should assert

One property, derived rather than enumerated:

> every `TestRolledBackProgress_*` in the package ran, and none skipped

Count what `-list` reports for the prefix; run with `-v`; require the same number of top-level results and zero skips. That catches a rename, an addition, and a skip, needs no names in YAML, and lets the eight pre-existing `-list … | grep -q …` guards elsewhere in this file go too.

One trap in writing it, worth naming because it is the same class of defect the guard is meant to close: the obvious spelling pipes `go test` into `tee` or `grep`, and `$?` then reports the last stage of the pipe rather than the test run. Use `set -o pipefail`, or capture the exit code before inspecting the output. A guard that reports the exit status of `grep` is exactly the shape of bug being guarded against.

### Two smaller things

The timeout moves from `3m` to `5m`. The Verification section does not mention it. If the matrix now needs the extra time, say so; if it was precautionary, say that.

The Regression section describes this sequence as the defect being fixed:

```sql
START TRANSACTION;
INSERT INTO ledger VALUES (1);
CREATE TABLE ddl1 (id INT);
INSERT INTO ledger VALUES (3);
ROLLBACK;
```

The merged commit for #1356 (`d5b2b446`) already states that finding, with the same shape and its own live measurement on MySQL 9.7.1 and MariaDB 11.4.12, and claims to fix it by classifying each statement on two bits instead of one. This branch is based on `22895782`, which contains that commit. So either the Regression prose is carried over from a draft written against an earlier state of #1356, or a hole remains that the prose describes as if it were the original one. Which it is changes how the rest of this pull request should be read, and it should be answered before review continues rather than inferred.

denisvmedia:
## Answering the Regression question first, because it changes how the rest reads

**The prose was stale.** The defect it described does not reproduce on current master. It was carried over from a draft written against an earlier state of #1356, and the body has been rewritten.

Measured, not inferred. Own MySQL 9.7.1 and MariaDB 10.11.18 containers, started for this and dropped after.

The server half of the claim holds and always did — it is a property of InnoDB, not of anything Ptah does:

```
$ docker exec a4a-mysql97 mysql -uroot -p… ptah_test -e "DROP TABLE IF EXISTS ddl1; DROP TABLE IF EXISTS ledger; CREATE TABLE ledger (id INT PRIMARY KEY) ENGINE=InnoDB; START TRANSACTION; INSERT INTO ledger VALUES (1); CREATE TABLE ddl1 (id INT); INSERT INTO ledger VALUES (3); ROLLBACK; SELECT id FROM ledger ORDER BY id;"
id
1
3
PROBE_MYSQL_EXIT=0
```

MariaDB gives the same two rows.

The Ptah half does not. Driven end to end through `MigrateUp` in tx-mode file — body `INSERT`, `CREATE TABLE`, `INSERT`, then a fourth statement that fails so the body rolls back — against a binary built from `origin/master` at `1fd469be` into my own worktree:

```
== first run (statement 4 fails) ==
catalog after first run: ledger notes = [one three]
catalog after first run: revision applied = 3, total = 4
== retry with AllowDirty ==
INFO Resuming migration after a partially applied attempt version=1 resumeFromStatement=4
MigrateUp error: <nil>
catalog after retry: ledger notes = [one three]
catalog after retry: revision applied = 4, total = 4
catalog blocker rows: 1
```

`applied = 3` is the true durable prefix, the resume is at statement 4, and the committed `INSERT` is not repeated. MariaDB is identical, and so are both revision formats.

I checked that this is not a blind harness two ways.

Against `d5b2b446^` — the tree immediately before #1356 — the numbers are the same. That is worth stating plainly because it means **this particular shape was never broken on master**: the naive counter happened to give 3 here too. The false "prefix ends at the last statement that forced a commit" model named in `d5b2b446` lived inside an intermediate commit of that pull request, never on the default branch. The original #887 defect was the data-only body, which is a different shape.

So to get a real control I mutated master instead — replaced the witness read with the cheaper dialect-only fallback that predates it, still compiling:

```go
func usesTransactionalProgressWitness(dialect string, txMode MigrationTxMode) bool {
	_ = dialect
	return txMode == MigrationTxModeFile && false
}
```

```
catalog after first run: revision applied = 0
catalog after retry:     ledger notes = [one three one]
```

Committed DML replayed, on both servers. Restored, and the numbers go back to 3 / `[one three]`.

**Conclusion:** #1356 closed the boundary arithmetic. What this branch adds is the set of ways a body could sidestep the witness rather than mis-count against it. The Regression section now says that, and no longer reads as though #1356 did nothing.

## On the guards

Agreed on every point, and the numbers moved since your read: head `268dbe12` had **49** `TestRolledBackProgress_` tests against **17** named, so 32 unguarded, and there were **10** pre-existing `-list … | grep -q …` guards elsewhere, 27 lines in total.

`scripts/require-tests-ran.sh` replaces all 27, plus the two `--- PASS:` greps in the RLS spelling step, which were the same class. Seven call sites, no test names in YAML beyond the `-run` patterns.

Your "the guard proves registration, not execution" point, measured directly. With one test in the family pointed at a DSN variable the job does not set:

```
$ go test ./migration/migrator -list '^TestRolledBackProgress_MySQLEscapedDatabaseGrantIsAccepted$' > log
OLD_GUARD_LIST_EXIT=0
$ grep -q '^TestRolledBackProgress_MySQLEscapedDatabaseGrantIsAccepted$' log
OLD_GUARD_GREP_EXIT=0
```

The replaced guard passes on a test that skipped.

One correction to the design you sketched, found by trying it. Counting alone does **not** catch a rename: with the same pattern feeding `-list` and `-run`, a test renamed out of the family drops both counts together and they still agree — measured, `listed=48 ran=48`, exit 0. That gap is real and needed a separate answer, so `--declared-in` names the file the family lives in and requires every `func Test...` there to be selected by the pattern. Still derived — a path, not a list of names.

Three breaks, restored between each, exit codes read directly and never through a pipe:

| break | exit | message |
| --- | --- | --- |
| rename one test out of the family | 1 | `49 tests are declared in … but ^TestRolledBackProgress_ selects only 48 of them` |
| make one test skip on a missing DSN | 1 | `1 of 49 tests skipped; a skip reads as a pass` |
| point the pattern at nothing | 1 | `matches no test`, where plain `go test -run` exits 0 with `no tests to run` |

Nothing in the script is piped: each command writes a file and its status is read on the next line.

The timeout is answered in the body and in a comment on the step: precautionary. The 49 tests take about 21s of package time against MySQL 9.7.1 and MariaDB 10.11.18, so 3m was never close; 5m is headroom for a shared runner.

Every converted step was run against live databases started for this — PostgreSQL 18, ClickHouse 26.7, MySQL 9.7.1, MariaDB 10.11.18 — and each reports `listed == ran` with zero skips, so none of them is being converted into a red step by this change.

**Left open:** the steps in this file that never had a `-list` guard are still unconverted. They reach SQL Server, CockroachDB, YugabyteDB and the legacy ClickHouse service, which I could not stand up here, so I could not measure whether any of them currently skips. Converting them unmeasured could redden CI for a reason unrelated to this branch. It is the same hole, and it deserves its own change.

denisvmedia:
Checked after merge. The review comment about the CI guard was addressed in #1367 by replacing the hand-maintained list with the derived `scripts/require-tests-ran.sh` guard and by correcting the stale regression wording in the PR body/comment.\n\nDuring the follow-up self-review I found one additional MySQL/MariaDB metadata-session hardening gap that was not part of the merged #1367 patch. I put that into a clean follow-up PR: #1378.

denisvmedia:
Rechecked after the note. #1367 itself has no inline review threads; the top-level CI-guard feedback was already answered there, and #1378 covered the separate MySQL/MariaDB metadata-session hardening found during follow-up review.

While verifying current `master`, I found the same manual `go test -list | grep` / `--- PASS:` guard shape had reappeared in the partitioned-index live workflow step. That is the same class of problem the #1367 feedback called out, so I split it into a small follow-up PR: #1386.

PR #1368, opened 2026-08-09
Keep one instruction file and verify what it carries

denisvmedia:
## What this does

Folds `CLAUDE.md`'s still-true content into `AGENTS.md` and deletes `CLAUDE.md`,
so the repository has one instruction file.

`git grep -l CLAUDE.md` on master matched only `CLAUDE.md` itself, so nothing
referenced it. But a plain deletion would have lost real material: `AGENTS.md`
carried the policy and standards and had no project overview, no build or test
commands, no package map, no annotation reference, and no CLI surface.

The merged sections go **ahead** of the policy sections rather than at the end.
An agent needs to know what the project is and how to build it before a rule
about it means anything. New sections, in order: What Ptah Is, Repository
Layout, Schema Annotations, The Native CLI Surface, Building And Testing, plus
a "Where Tests Live" subsection inside the existing Testing Standards.

`AGENTS.md` has **no deletions** — the diff is 213 added lines and nothing
removed.

## Every line was verified, because the source was stale

`CLAUDE.md` was last touched 2026-08-08 (#1318), but most of its body predates
the CLI restructuring and the `internal/` move. Nothing was copied across
without being executed or resolved against the tree.

**Commands.** 57 distinct command lines were run for verification, separate from
the gate runs below. Every exit code was read directly from the command, never
from the tail of a pipe. 45 exited 0. 12 exited non-zero: 10 of those were
deliberate probes of a refusal, and 2 are the broken script in defect 1 and 2.

| Group | Distinct commands | Exit 0 | Non-zero |
| --- | --- | --- | --- |
| Build, `make`, test-suite entry points | 12 | 10 | 2 (`test-ptah.sh`) |
| `--help` and verb-existence probes | 20 | 15 | 5 (removed verbs) |
| `schema render` over models and dialects | 17 | 13 | 4 (3 `SERIAL`, 1 typo) |
| Live-database runs | 8 | 7 | 1 (unknown scheme) |

Every command that appears in the merged file is in the exit-0 set. The 10
refusal probes are quoted in the file *as* refusals: the five removed verbs, the
three dialects that refuse `SERIAL`, the unknown annotation attribute, and the
unknown URL scheme.

The live runs used a throwaway database created and dropped on the shared
PostgreSQL 17.10 instance. `migrations up` was not trusted on its own output:
the catalog was read afterwards and returned `products` and `idx_products_name`,
and after `migrations down --target 0` only `schema_migrations` remained.

All nine dialect spellings were accepted by `--dialect`; three refuse the
`SERIAL` column of the test model with a named diagnostic. The file says exactly
that rather than implying all nine render everything.

**Paths and packages.** `CLAUDE.md` named 30 distinct filesystem paths and 10
Makefile targets. All 40 were checked. All 10 targets exist. 4 of the 30 paths
do not, and all 4 sit in the same list:

| Claimed | Reality |
| --- | --- |
| `core/parser` | does not exist; `internal/parser` + `internal/dialectlexer` |
| `core/lexer` | does not exist; `internal/lexer` |
| `core/astbuilder` | does not exist; `internal/astbuilder` |
| `core/convert` | does not exist; `internal/convert/...` |

All four moved under `internal/` in #455 and cannot be imported from another
module, so the old map does not merely mislead — it produces an import path
that cannot compile. The other 26 resolved, among them `cmd/ptah/main.go`,
`cmd/root/root.go`, `core/goschema/parser.go`, `core/renderer/renderer.go`,
`migration/schemadiff/schemadiff.go`, `integration/gonative/`, `stubs/`,
`examples/` and `test-reports/`.

Every path `AGENTS.md` now names was resolved with `ls` before it was written,
including all of `core/platform`, `core/renderer/internal/dialects`, the five
`internal/` packages the merged text points at, every `cmd/` namespace the root
command registers, `integration/fixtures/entities/`, `docs/capabilities.md`, and
`.github/workflows/go-integration-tests.yml`. None is missing.

**Other corrections carried into the merge.**

- The CLI is namespaced. `ptah generate`, `ptah compare`, `ptah read-db`,
  `ptah drop-all` and `ptah migrate` are all gone; the first four moved under
  `schema`/`db`, and `migrate` has no native spelling on purpose.
- The annotation example declared an index that was never created. Rendering it
  verbatim reports `0 indexes`. The walker visits comments attached to struct
  fields, so a file-level `//ptah:schema:index` is read by nobody. The two
  replacement forms in the file were rendered and emit the `CREATE INDEX` they
  promise.
- "CLI commands support `--dry-run` for safety" is false of the CLI.
  `db read` and `version` have no such flag. Reworded to name the write
  commands that do.
- `dbschema` handles five drivers, not the three the old text implied; SQLite
  and SQL Server were missing.

## Defects found, reported rather than papered over

These are **not** fixed here — this is a documentation PR — but each is real
and reproducible:

1. **`test-ptah.sh` cannot be invoked as documented.** It is committed
   `100644` while every `scripts/*.sh` is `100755`, so `./test-ptah.sh unit`
   exits 126, `permission denied`.
2. **`test-ptah.sh unit` is not "fast, no databases".** Lines 23-25 export
   `POSTGRES_TEST_DSN`, `MYSQL_TEST_DSN` and `MARIADB_TEST_DSN`
   unconditionally, so live tests attempt to connect and fail;
   `bash test-ptah.sh unit` exits 1 with no `--- FAIL` in its own log (the
   failures land in the JSON report). It also points at `localhost:5432` while
   `docker-compose.yaml` publishes PostgreSQL on `5433`, so `make db-start`
   does not satisfy it either.
3. **File-level `//ptah:schema:index` is silently ignored.** 17 files under
   `integration/fixtures/entities/` and `integration/entity_templates.go`
   declare indexes in that position. `ptah schema render --root-dir
   integration/fixtures/entities/007-index-add` reports
   `Found 3 tables, 23 fields, 0 indexes` and emits no `CREATE INDEX`. Both
   `fields=` and the legacy `columns=` behave the same way there, so this is
   placement, not spelling.
4. **`integration/gonative/integration_test.go` line 1 reads
   `//go:buildx integration`.** A misspelled constraint is not a constraint, so
   `TestGenerateCreateTableFromStubs` compiles into the ordinary unit run
   instead of the tagged one. Present since the file was added in 2025-06.
5. **`check:responsive:selftest` exits 0 while skipping.** Without Playwright
   installed it prints `SKIPPED` and returns 0 — a skip reading as a pass, the
   exact failure mode this repository guards against elsewhere.

## On deleting CLAUDE.md

Claude Code reads `CLAUDE.md` by convention, so the honest answer is that
deleting it **does** change something for one class of user: a Claude Code
session in this repository will no longer auto-load that file.

It changes nothing else. Nothing in the repository referenced it, and
`docs/site/scripts/check-style.mjs` governs `README.md` and `AGENTS.md` by name
and never included `CLAUDE.md` — the governed-file count is 100 before and
after this change, which is part of why the file was free to rot.

**No shim is added, and I do not think one is warranted.** A `CLAUDE.md` that
says "see AGENTS.md" is a second file to keep current, and this PR exists
because the second file was not kept current. The tool convention is also
converging on `AGENTS.md`, and current Claude Code reads it. The failure mode
of no shim is that one agent reads no repository instructions for one session.
The failure mode of a stale shim is what was measured above: four import paths
that cannot compile and five CLI verbs that do not exist, stated confidently.
If a shim is wanted anyway, a symlink is the only form that cannot drift, and
that is a maintainer's call rather than something to add unasked.

## Gates

All run in this worktree, each exit code read directly.

| Gate | Exit |
| --- | --- |
| `go build ./...` | 0 |
| `go vet ./...` | 0 |
| `golangci-lint run ./...` | 0 |
| `docs/site`: all 13 `check:*` scripts | 0 |

`check:responsive` needs a built site, so `npm ci` and `npm run build` ran
first, matching what `.github/workflows/docs.yml` does.

One transient to note honestly: the first `make test` failed on
`TestSVGReportsGraphvizStderrOnFailure`, killed at exactly its 10s timeout under
heavy parallel load on this machine. It passes in isolation in 0.51s and the
re-run of the full `make test` exited 0. Load flake, not a code finding.

## Not covered

The Docker integration targets were not executed. Port 5433 is held by another
compose stack on this machine, and `make docker-clean` runs
`docker system prune -f`, which is system-wide. They are named in the file as a
pointer with that caveat, not presented as commands I ran.


denisvmedia:
## Review: the content is right, but deleting `CLAUDE.md` silently removes the auto-loaded project instructions

The 213 added lines are a clear improvement and they correct real staleness. Worth saying explicitly, because these were measured today and the deleted file was wrong about all of them:

- The deleted file documented `ptah generate`, `ptah read-db`, `ptah compare`, `ptah migrate` and `ptah drop-all`. None of those verbs exist. The replacement says so and names what they answer instead.
- The deleted file listed `parser/`, `lexer/`, `astbuilder/` and `convert/` as subdirectories of `core/`. They are under `internal/` — `go list ./...` puts them at `internal/lexer`, `internal/parser`, `internal/astbuilder`, `internal/convert/...`, and an external module cannot import any of them. The replacement places them correctly.
- The deleted file's annotation example put `//ptah:schema:index` at file level attached to a bare `_ int`. The replacement explains that a file-level index attached to no declaration contributes nothing and says nothing while doing it, and gives the holder-struct form that works.
- The replacement writes down "A skip reads as a pass" and the `test-ptah.sh` traps. Both cost real time this week.

### The blocker

`CLAUDE.md` is deleted outright, with the guidance moved into `AGENTS.md`. Claude Code loads `CLAUDE.md` automatically as project instructions; it does not load `AGENTS.md`. In a session running in this repository right now, the system context contains a block headed "Contents of /Users/buster/Work/denis/ptah/CLAUDE.md (project instructions, checked into the codebase)" and no equivalent block for `AGENTS.md`.

So the effect of this change is not "one file instead of two". It is that a Claude Code session in this repository stops receiving project instructions at all, while `AGENTS.md` keeps serving the agents that read it by convention. Nothing in the diff would fail, no check would go red, and the loss would show up only as sessions gradually doing worse — which is the hardest kind of regression to attribute later.

This is the same failure mode that bit this work today from the other direction: a binding instructions file was placed somewhere that got cleaned, and briefs went on citing it while agents silently worked without it. A file that carries rules should not be able to disappear quietly.

### The fix, which keeps everything this pull request is for

Keep `CLAUDE.md` as a pointer rather than a copy. A few lines are enough, and the single-source-of-truth goal survives intact because the content still lives in exactly one place:

```markdown
# CLAUDE.md

Repository guidance for coding agents lives in [AGENTS.md](AGENTS.md).
Read that file; it is authoritative. This file exists so that tooling which
loads `CLAUDE.md` automatically still reaches it.
```

If the intent is specifically to have one file and no stub, then the change needs to establish first that `AGENTS.md` is auto-loaded in this repository — measured, not assumed — and the pull request body should carry that measurement, because everything else in the change depends on the new file actually being read.

### Two smaller things

The body claims `--dialect` accepts nine spellings and that `dbschema.ConnectToDatabase` dispatches to five drivers. Both match what I have seen of the driver map, but they are counts, and counts in this repository have a habit of being one member short. Derive them from the source in a test rather than writing them down, or say which command you counted with.

"Two runs over the same entities produce byte-identical output" is a quantified claim about determinism. If a test asserts it, cite the test; if not, soften it to what was executed.

Issue #1341, opened 2026-08-09
Tiered CI: run the capability probe per version, the full suite nightly


denisvmedia:
Sub-issue of #916, its Tier 3, and the CI half of #1339.

## The shape

Three tiers, chosen so the cost lands where the information is.

| tier | when | what runs | cost |
|---|---|---|---|
| 1 | every pull request | today's `go-integration-tests.yml`, one canonical version per dialect | ~10-12 min wall clock, unchanged |
| 2 | every pull request | the capability probe from #1339 across every vendor-supported release line | seconds per cell, minutes in parallel |
| 3 | nightly | the full integration suite across the full version matrix | hours of machine time, wall clock irrelevant |

Tier 1 is what exists. Tier 2 is cheap because it asks about capabilities rather than scenarios. Tier 3 is the expensive one, and it is expensive precisely because it re-runs version-independent work — which is why it does not belong on the pull-request path.

## Measured baseline, so the cost is not guessed

Last five successful runs on `master`:

- Integration Tests: 570, 599, 596, 753, 656 s
- Go Unit Tests: 484, 305, 411, 532, 464 s
- Atlas CE Oracle: 220, 205, 205, 177, 217 s

`go-integration-tests.yml` currently starts six services in a single job: `postgres:18`, `mysql:9.7`, `mariadb:10.11`, `clickhouse:26.7`, `clickhouse:24.10`, `mssql:2025-latest`.

Multiplying that job by the release lines gives roughly eighteen cells, about 3.5 hours of machine time per run. That is the number tier 3 accepts on a nightly cadence and tier 2 exists to avoid on every pull request.

## Definition of done

- Tier 2 runs on every pull request, one job per `(dialect, version)` cell, from a single matrix definition so adding a line is a data change rather than a workflow edit.
- Tier 3 runs on a schedule, over the same matrix definition — one source of truth for which versions exist, not two lists that drift.
- **A tier 3 failure must be attributable.** A nightly red across eighteen cells with no indication of which capability moved is a report nobody reads. Tier 3 output names the cell and, where the failure is a capability disagreement, defers to the tier 2 row that already says so.
- The matrix is checked against the presets: a release line present in `core/platform/capability` with no cell, or a cell with no preset entry, fails the build. Otherwise the two drift and the saturation signal from #1326 starts lying in the comfortable direction.

## Two hazards worth writing down before anyone starts

[...]

PR #1371, opened 2026-08-09
Capability probe matrix: measure each preset against the server it describes (#1339)

### LINKED ISSUE #1339: Capability probe matrix: verify each preset against the server it claims to describe
Sub-issue of #916, its Tier 2. #916 establishes that one preset per dialect leaks newer features into older servers; this issue is the mechanism that makes a preset a **measured** claim rather than a written one.

## The problem this solves, and why a version matrix alone does not solve it

The obvious answer to "we support five PostgreSQL lines" is to run the integration suite against all of them. Measured on `master`, that costs more than it buys:

| job | wall clock, last five successful runs on `master` |
|---|---|
| Integration Tests | 570, 599, 596, 753, 656 s |
| Go Unit Tests | 484, 305, 411, 532, 464 s |
| Atlas CE Oracle | 220, 205, 205, 177, 217 s |

`go-integration-tests.yml` starts **six services in one job** — `postgres:18`, `mysql:9.7`, `mariadb:10.11`, `clickhouse:26.7`, `clickhouse:24.10`, `mssql:2025-latest`. Multiplying that job across the vendor-supported release lines gives roughly eighteen cells (PostgreSQL 14 through 18, MySQL 8.0/8.4/9.x, MariaDB 10.6/10.11/11.4/11.8/12.x, two ClickHouse lines, SQL Server 2019/2022/2025). At ~12 minutes a cell that is about 3.5 hours of machine time per run, and the overwhelming majority of it re-runs scenarios that do not vary by version at all.

**Versions do not differ by scenario. They differ by capability.** So the matrix belongs on the capability axis, and it is small there: O(capabilities), not O(tests).

## What to build

A probe suite that, for one `(dialect, version)` pair, answers a single question per capability: **does the live server actually behave the way the preset says it does?**

For each capability in `core/platform/capability`:

1. Emit the DDL whose acceptance the capability governs.
2. Execute it against the live server.
3. Assert the server accepted or rejected it, and compare that to what the preset promises.
4. Report every disagreement as a named row: capability, dialect, version, preset says, server does.

This is seconds per cell rather than minutes, because it is DDL round trips rather than the scenario matrix. Eighteen cells in parallel is minutes of wall clock.

## Why this is now a mechanism and not just another test job

#1326 landed the saturation report: `ForServerVersionResult` can now tell a caller that a server is past the newest measured line rather than silently handing back the newest preset it knows. Before that, an unmeasured version was indistinguishable from a measured one.

With both pieces, the loop closes:

- the probe matrix **defines** which versions are measured;
- the saturation signal **detects** a version that is not;
- a new engine release therefore shows up as a missing cell rather than as a wrong plan discovered by a user.

That is the part #916 asks for and the part a plain version matrix does not give.

## Definition of done

- A probe harness that takes a dialect, a version and a live URL, and returns one row per capability with `preset says` / `server does`.
- A disagreement is a **failure**, not a warning. A capability the probe cannot decide for a dialect is recorded as undecidable with its reason, and undecidable is not the same as agreeing.
- Coverage for the release lines named above, with the cell list in one place so adding a line is a data change.
- The probe is what promotes a version to "measured": adding a cell and adding a preset entry are the same change, and the saturation signal for that version goes quiet as a result.
- CI wiring is Tier 3, not this issue — see the companion issue. This one must be runnable locally against a single URL.

## Deliberately out of scope

- Running the full integration suite per version. That is Tier 3.
- Deciding the capability value model — booleans versus enumerated modes and limits. That is #1230 item 3 and it gates the shape of what a probe asserts. A probe over booleans is buildable today and does not foreclose the richer model.

## Notes

- `ForServerVersionResult` refines by version for three dialects only. ClickHouse, SQLite and SQL Server return the dialect default, and CockroachDB, YugabyteDB and Spanner are resolved by banner substring before a version is parsed. The probe will make those visible as whole columns of undecidable rows, which is the honest first output.
- `parseVersion` reads `Microsoft SQL Server 2022 ... 16.0.1000.6` as major 2022 — the marketing year, not the version. Any SQL Server row has to go through a corrected parse or it measures the wrong thing.

Refs #916.

denisvmedia:
Closes #1339. Refs #916.

`capability-probe` takes one dialect, one version and one live URL, and returns one row per registered capability with `preset says` / `server does`. A disagreement is a failure. A capability it cannot decide is `UNDECIDABLE` with a reason, and undecidable is never folded into agreement.

CI wiring is #1341 and is deliberately not here.

```
go build -o bin/capability-probe ./cmd/capability-probe
bin/capability-probe --list-cells
bin/capability-probe --db-url postgres://postgres:pw@localhost:5432/postgres?sslmode=disable
bin/capability-probe --db-url mysql://root:pw@localhost:3306/mysql --evidence
```

## First output, verbatim

Both runs are from a binary built from this branch's pushed tree, against servers started for this work.

### PostgreSQL 17.10 — exit 0

```
capability probe
  target       postgres://postgres:***@localhost:15339/postgres?sslmode=disable
  dialect      postgres
  banner       PostgreSQL 17.10 (Debian 17.10-1.pgdg13+1) on aarch64-unknown-linux-gnu, compiled by gcc (Debian 14.2.0-19) 14.2.0, 64-bit
  version      17.10 (banner)
  matrix cell  postgres 17 [preset Postgres17, version-ladder]
  resolution   version-specific=true saturated=false newest-measured="17.x"
  session      preset unchanged by the pinned session
  namespace    ptah_capprobe_235f443246597355
  control      REFUSED   CREATE NONSENSE ptah_capability_probe_control  -> ERROR: syntax error at or near "NONSENSE" (SQLSTATE 42601)

CAPABILITY                             PRESET SAYS  SERVER DOES  OUTCOME
advisory_locks                         true         true         AGREES
alter_generated_column_expression      true         true         AGREES
check_constraints_enforced             true         true         AGREES
create_index_concurrently              true         true         AGREES
create_or_replace_trigger              true         true         AGREES
drop_check_clause                      false        false        AGREES
drop_constraint_generic                true         true         AGREES
drop_constraint_if_exists              true         true         AGREES
drop_index_concurrently                true         true         AGREES
drop_index_if_exists                   true         true         AGREES
enum_custom_type                       true         true         AGREES
enum_inline_column                     false        false        AGREES
foreign_keys                           true         true         AGREES
foreign_keys_create_backing_index      false        false        AGREES
foreign_keys_require_indexed_reference false        false        AGREES
foreign_keys_require_unique_reference  true         true         AGREES
functions                              true         true         AGREES
materialized_views                     true         true         AGREES
role_management                        true         true         AGREES
row_level_security                     true         true         AGREES
sequences                              true         true         AGREES
triggers                               true         true         AGREES
views                                  true         true         AGREES
xml_type                               true         true         AGREES

summary: 24 rows — 24 AGREES, 0 DISAGREES, 0 UNDECIDABLE; decided 24
```

### MySQL 9.7.1 — exit 0

```
capability probe
  target       mysql://root:***@localhost:33339/mysql
  dialect      mysql
  banner       9.7.1
  version      9.7.1 (banner)
  matrix cell  mysql 9.7 [preset MySQL84, version-ladder]
  resolution   version-specific=true saturated=false newest-measured="9.x"
  session      preset unchanged by the pinned session
  namespace    ptah_capprobe_1754e169bc5babfc
  control      REFUSED   CREATE NONSENSE ptah_capability_probe_control  -> Error 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'NONSENSE ptah_capability_probe_control' at line 1

CAPABILITY                             PRESET SAYS  SERVER DOES  OUTCOME
advisory_locks                         false        false        AGREES
alter_generated_column_expression      false        false        AGREES
check_constraints_enforced             true         true         AGREES
create_index_concurrently              false        false        AGREES
create_or_replace_trigger              false        false        AGREES
drop_check_clause                      true         true         AGREES
drop_constraint_generic                true         true         AGREES
drop_constraint_if_exists              false        false        AGREES
drop_index_concurrently                false        false        AGREES
drop_index_if_exists                   false        false        AGREES
enum_custom_type                       false        false        AGREES
enum_inline_column                     true         true         AGREES
foreign_keys                           true         true         AGREES
foreign_keys_create_backing_index      false        false        AGREES
foreign_keys_require_indexed_reference false        false        AGREES
foreign_keys_require_unique_reference  true         true         AGREES
functions                              true         true         AGREES
materialized_views                     false        false        AGREES
role_management                        false        -            UNDECIDABLE
row_level_security                     false        false        AGREES
sequences                              false        false        AGREES
triggers                               true         true         AGREES
views                                  true         true         AGREES
xml_type                               false        false        AGREES

summary: 24 rows — 23 AGREES, 0 DISAGREES, 1 UNDECIDABLE; decided 23

undecidable rows and why:
  the key names the PostgreSQL role and privilege surface no MySQL-family code path consults; this server's own CREATE ROLE and GRANT are a different surface, so accepting them would not decide the key
      role_management

notes:
  alter_generated_column_expression
      this key names PostgreSQL 17's SET EXPRESSION spelling, which is what was probed; the MySQL family can change a generated column expression under its own MODIFY COLUMN spelling, so a false row here means "not this statement", not "not this ability"
  materialized_views
      the keyword was accepted and then dropped: the view reported 0 before the source row was inserted and 1 after, so the result is recomputed rather than stored
```

**No preset defect was found on PostgreSQL 17.10, MySQL 9.7.1 or MariaDB 11.4.12.** MariaDB 11.4.12 also ran clean: 22 AGREES, 0 DISAGREES, 2 UNDECIDABLE (`role_management` and `sequences`).

## The honest first output for the dialects that do not refine by version

Six of the nine dialects get whole columns of UNDECIDABLE, and that is the correct result rather than a defect. Two of them were run live.

**CockroachDB v26.2.5** — 24 UNDECIDABLE, decided 0, **exit 1**:

```
summary: 24 rows — 0 AGREES, 0 DISAGREES, 24 UNDECIDABLE; decided 0
...
capability-probe: role_management: preset says false, server does true
row_level_security: preset says false, server does true
sequences: preset says false, server does true
this run decided 0 of 24 capability rows; a probe that measured nothing must not read as a probe that passed
```

The reason on every row: *the cockroachdb preset is selected by a banner substring before any version is parsed, so every release of this engine receives CockroachDB23 and an observation on one release cannot be credited to this line.*

Two things worth noting. The observation survives the undecidable verdict, so the three places where CockroachDB v26.2.5 contradicts `CockroachDB23()` are reported and still fail the run rather than being absorbed by the word UNDECIDABLE — that is #916 material. And `create_index_concurrently` carries the note *the server accepted the statement inside an explicit transaction block as well, so the keyword is parsed and dropped rather than changing how the index is built*.

**SQL Server 2025** — 24 UNDECIDABLE, exit 1, reason *no probe plan exists for the sqlserver dialect*. The header reads `version 17.0.4065.4 (server product version)` and `matrix cell sqlserver 17.0 (SQL Server 2025)`: the corrected parse works against the live server. The shared `parseVersion` reads the same banner as major **2025**, which `version_test.go` pins as a control.

## The four ways this deliverable fails in the direction that looks like success

Each mutant was built, run, watched go red, restored, and watched go green.

### 1. A preset claiming a capability the live server does not have

`MySQL84().XMLType: false -> true`. **The entire unit suite stays green under this mutant** (`go test ./... -count=1`, 191 packages ok) — nothing in the build catches it. The probe does:

```
xml_type                               true         false        DISAGREES
...
capability-probe: xml_type: preset says true, server does false          (exit 1)
```

Restored: `xml_type false false AGREES`, exit 0.

(My first attempt used `DropIndexIfExists`, which `capability_test.go` already pins by hand — the unit suite caught it, so it did not isolate the probe. Swapping to a key nothing pins is what makes the mutant measure this PR.)

### 2. UNDECIDABLE folded into agreement

`assemble`: `row.Outcome = Undecidable` -> `row.Outcome = Agrees` for an undecided key.

```
--- FAIL: TestAssemble_ThreeOutcomes/an_undecided_key_stays_undecidable_and_never_becomes_agreement
    got "AGREES" want "UNDECIDABLE"
--- FAIL: TestAssemble_UndecidableRowsAreNotCountedAsDecided
    got int(0) want int(24)
```

**2b, the subtler fold:** derive line attribution from `VersionResolution.VersionSpecific` instead of the declared refinement — exactly the trap the field invites, because it is `true` for a banner-matched CockroachDB that never consulted a version. `TestAssemble_AnUnattributableLineKeepsTheObservation` goes red (`got "DISAGREES" want "UNDECIDABLE"`), and live CockroachDB v26.2.5 flips from 24 UNDECIDABLE to **21 AGREES / 3 DISAGREES / decided 24** — a whole release line certified by an observation the resolver cannot distinguish from v25.4's.

### 3. A run that measured nothing reporting success

- **No cell for the line.** Mutant: `CellFor` falls back to the dialect's first measured cell. `TestCellFor` goes red on three rows. Live: a PostgreSQL 13.23 server is labeled `matrix cell postgres 17 [preset Postgres17]` under the mutant. Restored, the same server reports `matrix cell NONE`, 24 UNDECIDABLE, decided 0, exit 1.
- **Zero cells.** `TestRun_RefusesAnEmptyMatrix` — `Run` refuses before dialing.
- **URL does not resolve.** Exit 1: `connect to postgres://postgres:***@localhost:1/nosuchdb...: failed to ping database: ... connection refused`.
- **A server that cannot say no.** Every run first executes `CREATE NONSENSE ptah_capability_probe_control` and fails if it is accepted.

### 4. A deciding statement accepted for an unrelated reason

Constructed twice, on live servers, and the probe is not fooled either time.

- **MySQL 9.7.1 accepts `CREATE MATERIALIZED VIEW`.** Mutant: decide `materialized_views` by acceptance alone (the cheaper implementation). Result: `materialized_views false true DISAGREES` — a false disagreement against a correct preset. The shipped decider reads the view before and after an insert into the source table, sees 0 then 1, and reports `false`.
- **CockroachDB v26.2.5 accepts `CREATE INDEX CONCURRENTLY`.** Mutant: decide by acceptance alone. Result: `create_index_concurrently ... server does true`. On real PostgreSQL the same mutant looks perfectly fine (`AGREES`) — it lies only about the server you do not have. The shipped decider requires the refusal PostgreSQL gives inside an explicit transaction block.

There is a third case I did **not** decide, by design: MySQL 9.7.1 accepts `CREATE ROLE m_role` and `GRANT SELECT ... TO m_role` (both exit 0, measured) while `MySQL84().RoleManagement` is `false`. That key names the PostgreSQL role surface no MySQL-family code path reads, so it is declared undecidable for the MySQL family in advance, as data in `plans.go` — not derived from an outcome. `MariaDB1011().Sequences` is the same shape (#931 item 8).

## Adding a release line is a data change

Added a PostgreSQL 13 cell. The whole diff:

```diff
+	{
+		Dialect: platform.Postgres, Line: "13",
+		Preset: capability.Postgres13, PresetName: "Postgres13",
+		Refinement: RefinedByVersion, Image: "postgres:13",
+		Note: "...",
+	},
```

Before, against a live PostgreSQL 13.23: `matrix cell NONE`, decided 0, exit 1. After: `matrix cell postgres 13 [preset Postgres13, version-ladder]`, **24 AGREES, decided 24, exit 0**. Adding the cell is what promoted the line, and it independently validated `Postgres13()` against a real server. The cell is kept.

`--list-cells` prints all 24 cells. `postgres 18` and `mariadb 12.3` carry no preset: probing them fails with the line named rather than quietly accepting a preset chosen by saturation. Those are the cells #1239 and #108 are waiting on.

## Two things the matrix surfaced

**MySQL 8.4+'s foreign-key reference policy is a session fact, not a version fact.** With `restrict_fk_on_non_standard_key = OFF` the same MySQL 9.7.1 server reports:

```
  session      pinned session changed foreign_keys_require_indexed_reference=true, foreign_keys_require_unique_reference=false
...
foreign_keys_require_indexed_reference true         true         AGREES
foreign_keys_require_unique_reference  false        false        AGREES
```

Both sides moved together — `dbschema.refineMySQLForeignKeyCapabilities` already reads the variable on the pinned session. No boolean keyed on version alone is correct for both settings; the probe now measures both halves and the report names the delta.

**The gating gap is wider than the version-refinement gap.** ClickHouse, SQLite and SQL Server have no probe plan here, and every row for them is undecidable with that stated rather than counted as agreement.

## Design constraints the implementation is built around

- **No transaction as the isolation mechanism.** PostgreSQL refuses `CREATE INDEX CONCURRENTLY` inside an explicit transaction block (measured: exit 0 standalone, exit 1 inside `BEGIN`), so BEGIN/ROLLBACK would report two true capabilities as false. Each run uses a throwaway schema (PostgreSQL family) or database (MySQL family) named `ptah_capprobe_<random>` and drops it. If the session dies mid-run the error names the namespace left behind rather than littering silently.
- **The foreign-key reference policy is one experiment yielding three rows**, with a control (a foreign key to a declared unique key must be accepted) so a server refusing foreign keys outright does not read as a policy choice. Three independent probes could produce a combination `Validate` rejects.
- **`drop_constraint_if_exists` is ordered after `drop_constraint_generic`** and recorded undecidable when the generic clause is absent — the registry's own `requires` edge. It also probes both spellings the key covers, including `DROP FOREIGN KEY IF EXISTS`.
- **Every IF EXISTS key needs its unguarded twin refused.** Acceptance of the guarded form alone would also pass on a permissive server.
- **Every registered capability is answered exactly once per dialect plan**, enforced by `TestPlans_AnswerEveryRegisteredCapabilityExactlyOnce`. A new capability turns the build red until the plans answer for it, instead of silently vanishing from every row.

## Gates

All run on the pushed tree after pulling it back (the files were pushed through the GitHub API, so the compiler re-checked them rather than my eyes):

| gate | result |
|---|---|
| `go build ./...` | exit 0 |
| `go vet ./...` | exit 0 |
| `go vet -tags integration ./integration/...` | exit 0 |
| `go test ./... -count=1` | exit 0, 191 packages ok |
| `go tool qtlint ./...` | exit 0 |
| `golangci-lint run ./...` | 0 issues |
| `scripts/check-test-style.sh` | OK (175 conditional groups, 42 white-box files) |
| `scripts/check-public-api-released.sh` | exit 0 (only pre-existing approved incompatibilities) |
| `docs/site` `check:exit-codes` (+selftest) | OK (60 reference rows) |
| `docs/site` `check:core-doc-links` | OK (7 protected core references) |
| `docs/site` `check:page-health` | OK (59 sidebar entries) |
| `docs/site` `check:links` (+selftest) | OK (60 pages, 60 routes, 616 heading anchors) |
| `docs/site` `check:redirects` (+selftest) | OK (30 redirects) |
| `docs/site` `check:style` (+selftest) | OK (100 documentation files) |
| `docs/site` `check:limitations` (+selftest) | OK (1 governed page) |
| `docs/site` `check:responsive` (+selftest) | OK (60 routes x 2 viewports) |
| `docs/site` `versions:selftest` | OK |

The new code is `internal/` plus a `cmd/` main, so it needs no `docs/public_api.md` entry; the three public-API scripts confirm that rather than my assuming it.

## What I did not do

- No ClickHouse, SQLite or SQL Server statement tables. Their rows are undecidable with the reason, not agreements.
- No Spanner row was executed — there is no Spanner server (#942), which the `SpannerPostgres` doc already concedes.
- No YugabyteDB server was run. Its version-extraction rule is pinned by a unit test against the banner shape, not against a live server.
- The `foreign_keys_create_backing_index` row was only ever decided *negatively*; deciding it positively needs a target that accepts a foreign key to an unindexed column, and no such server was available.
- No CI wiring (#1341).

### DIFF
--- a/cmd/capability-probe/main.go
+++ b/cmd/capability-probe/main.go
@@ -1,0 +1,82 @@
+// Command capability-probe measures a live database server against the
+// capability preset Ptah hands out for it.
+//
+// It answers one question per capability: does the server actually behave the
+// way core/platform/capability says it does? A disagreement exits non-zero. So
+// does a run that decided nothing — a probe that skipped every row must not
+// read as a probe that passed every row.
+//
+//	capability-probe --db-url postgres://user:pw@localhost:5432/db
+//	capability-probe --db-url mysql://root:pw@localhost:3306/db --evidence
+//	capability-probe --list-cells
+//
+// Wiring this into CI is stokaro/ptah#1341 and is deliberately not done here.
+package main
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/spf13/cobra"
+
+	"go.5x5.cz/ptah/internal/capabilityprobe"
+)
+
+func main() {
+	if err := newCommand().Execute(); err != nil {
+		fmt.Fprintf(os.Stderr, "capability-probe: %v\n", err)
+		os.Exit(1)
+	}
+}
+
+func newCommand() *cobra.Command {
+	var (
+		dbURL     string
+		listCells bool
+		evidence  bool
+	)
+	cmd := &cobra.Command{
+		Use:           "capability-probe",
+		Short:         "Measure a live server against the capability preset Ptah gives it",
+		SilenceUsage:  true,
+		SilenceErrors: true,
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			if listCells {
+				capabilityprobe.WriteCells(cmd.OutOrStdout())
+				return nil
+			}
+			if dbURL == "" {
+				return fmt.Errorf("--db-url is required (or --list-cells)")
+			}
+			var extra []sectionWriter
+			if evidence {
+				extra = append(extra, capabilityprobe.WriteEvidence)
+			}
+			return probe(cmd.Context(), cmd.OutOrStdout(), dbURL, extra)
+		},
+	}
+	cmd.Flags().StringVar(&dbURL, "db-url", "", "URL of the live server to probe")
+	cmd.Flags().BoolVar(&listCells, "list-cells", false, "print the matrix cell list and exit")
+	cmd.Flags().BoolVar(&evidence, "evidence", false, "print every statement executed and the server's answer")
+	return cmd
+}
+
+// sectionWriter renders one optional section of the report.
+type sectionWriter func(io.Writer, *capabilityprobe.Report)
+
+func probe(ctx context.Context, out io.Writer, dbURL string, extra []sectionWriter) error {
+	if ctx == nil {
+		ctx = context.Background()
+	}
+	report, err := capabilityprobe.Run(ctx, dbURL)
+	if err != nil {
+		return err
+	}
+	capabilityprobe.WriteReport(out, report)
+	for _, section := range extra {
+		section(out, report)
+	}
+	return report.Err()
+}
--- a/internal/capabilityprobe/cells.go
+++ b/internal/capabilityprobe/cells.go
@@ -1,0 +1,326 @@
+package capabilityprobe
+
+import (
+	"fmt"
+	"strings"
+
+	"go.5x5.cz/ptah/core/platform"
+	"go.5x5.cz/ptah/core/platform/capability"
+)
+
+// Refinement records HOW capability.ResolveServerVersion reaches a cell's
+// preset. It is declared here as data rather than read back from
+// capability.VersionResolution, because that struct cannot tell the three
+// mechanisms apart:
+//
+//   - a banner-matched dialect reports VersionSpecific=true without a version
+//     ever being parsed, so it is indistinguishable from a successful ladder
+//     match;
+//   - a dialect with no ladder reports VersionSpecific=false, which is exactly
+//     what an unparseable banner on a laddered dialect reports.
+//
+// Deriving undecidability from the returned struct would therefore record a
+// CockroachDB v26.2.5 as an agreeing version-specific match and a
+// cleanly-parsed SQL Server as an unreadable banner.
+type Refinement string
+
+const (
+	// RefinedByVersion: the dialect has a version ladder and the parsed
+	// version selects which arm of it answers. Only these lines can be
+	// measured per line, because only here does the version change the answer.
+	RefinedByVersion Refinement = "version-ladder"
+
+	// RefinedByBanner: the preset is selected by a substring of the banner
+	// before any version is parsed, so every release of the engine receives
+	// the same set and no observation can be attributed to one line rather
+	// than another.
+	RefinedByBanner Refinement = "banner-substring"
+
+	// NotRefined: the version is parsed and then discarded; the dialect
+	// default is returned for every release.
+	NotRefined Refinement = "dialect-default"
+)
+
+// Cell is one release line the capability matrix covers.
+//
+// Cells is the ONLY place a line is declared. Adding a release line is one
+// literal below and no code change anywhere else: the probe reads the line to
+// match a live server, the preset to compare against, and the refinement to
+// decide whether an observation can be attributed to this line at all.
+type Cell struct {
+	// Dialect is the normalized platform dialect this line belongs to.
+	Dialect string
+
+	// Line is the release line in the numbering the CORRECTED version parse
+	// produces (see version.go), not the vendor's marketing name. A server
+	// matches when its parsed version begins with these components, so "17"
+	// matches every PostgreSQL 17.x and "10.11" matches only MariaDB 10.11.x.
+	Line string
+
+	// Label is the line's human name where it differs from Line, for example
+	// "SQL Server 2025" for line "17.0". Empty means Line reads well enough.
+	Label string
+
+	// Preset is the capability preset this line claims, or nil when Ptah has
+	// no measured preset for the line yet.
+	//
+	// A nil Preset is the declared gap that issue #916 is about, and it is
+	// what makes "adding a cell and adding a preset entry are the same
+	// change" enforceable: the probe refuses to report success against a
+	// server on a nil-preset line, naming the line instead. Filling this in
+	// is what promotes a line to measured.
+	Preset func() capability.Capabilities
+
+	// PresetName names Preset for the report. It is written out rather than
+	// recovered by comparing sets, because MySQLLegacy and MariaDBLegacy are
+	// byte-identical and a set comparison would label MariaDB rows MySQL.
+	PresetName string
+
+	// Refinement says how the preset is reached. Anything other than
+	// RefinedByVersion makes every capability row on this line UNDECIDABLE:
+	// the observation is still taken and printed, but it cannot be credited
+	// to this line rather than to the line's siblings, which receive the
+	// identical set.
+	Refinement Refinement
+
+	// Image reproduces the line locally, empty when there is no container for
+	// it.
+	Image string
+
+	// Note records why a line is shaped the way it is: a missing preset, a
+	// missing server, a version axis that does not exist.
+	Note string
+}
+
+// Match reports whether a parsed server version falls on this cell's line.
+// The comparison is component-wise on the dotted line, so "17" accepts every
+// 17.x and "10.11" accepts only 10.11.x.
+func (c Cell) Match(v Version) bool {
+	want := strings.Split(c.Line, ".")
+	got := v.Components()
+	if len(want) > len(got) {
+		return false
+	}
+	for i, component := range want {
+		if got[i] != component {
+			return false
+		}
+	}
+	return true
+}
+
+// Measured reports whether the line claims a capability preset at all.
+func (c Cell) Measured() bool {
+	return c.Preset != nil
+}
+
+// String renders the cell as "dialect line" plus its label.
+func (c Cell) String() string {
+	if c.Label == "" {
+		return c.Dialect + " " + c.Line
+	}
+	return fmt.Sprintf("%s %s (%s)", c.Dialect, c.Line, c.Label)
+}
+
+// Cells is the capability matrix: every release line Ptah covers, in one
+// place.
+//
+// The lines are the vendor-supported ones, checked against the vendors on
+// 2026-08-09, plus any line whose preset Ptah still ships — a preset with no
+// cell is a claim nothing in this repository can measure.
+//
+// Four presets still have no cell, and each absence is deliberate:
+//
+//   - MySQLLegacy, MySQL8016 and MySQL8019 describe the MySQL 8.0 line, which
+//     left vendor support on 2026-04-30 and is not one line in any case: the
+//     three presets split it at 8.0.16 and 8.0.19, so a single "8.0" cell
+//     could not name one preset. Giving them cells means three cells keyed on
+//     the patch component, which the Line matcher already supports.
+//   - MariaDBLegacy describes MariaDB before 10.2, whose newest release left
+//     support in 2022.
+//
+// MariaDB 10.6 left community support on 2026-07-06 and is absent for that
+// reason alone.
+//
+// A nil Preset is a line the matrix wants and Ptah has not measured. Those are
+// the cells PRs bumping a container tag are waiting on, and probing a server
+// on one of them fails rather than reporting a green row against a preset
+// chosen by saturation.
+var Cells = []Cell{
+	// PostgreSQL: five supported majors (postgresql.org/support/versioning,
+	// read 2026-08-09 — 18.4, 17.10, 16.14, 15.18, 14.23; the project does not
+	// use the term LTS). Postgres16's own doc covers 14 through 16.
+	{
+		Dialect: platform.Postgres, Line: "18",
+		Preset: nil, PresetName: "",
+		Refinement: RefinedByVersion, Image: "postgres:18",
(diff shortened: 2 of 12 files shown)


denisvmedia:
Precision on the gate table, so it does not claim more than was executed.

The Go gates (`go build`, `go vet`, `go vet -tags integration`, `go test ./... -count=1` at 191 packages, `go tool qtlint`, `golangci-lint`, `scripts/check-test-style.sh`, and the three `check-public-api*` scripts) were all executed against the tree **pulled back from this branch** — that is what re-checks the files, since they were pushed through the GitHub API rather than committed locally.

The fourteen `docs/site` `check:*` scripts ran against the tree **before** that pull. That is still a valid result for this branch: the pull reported `Updating d47f417f..c9c8dd75` touching exactly fourteen files — the twelve added here plus `.github/workflows/go-integration-tests.yml` and `docker-compose.yaml`, both arriving from master. No file under `docs/` differs between the tree the docs gates measured and the tree this branch proposes, and this branch adds no documentation page.

Two more things a reviewer should know rather than infer:

- The commits are authored through the GitHub API and arrive as eight rather than one. Local `git commit` could not run: this repository signs commits and `gpg` has no TTY in the environment I am in (`gpg: signing failed: Inappropriate ioctl for device`, reproduced with a bare `gpg --batch --clearsign`). Squash on merge if the history shape matters.
- `origin/master` moved twice while this was being built — `d47f417f`, then `90281329`, which is the commit this branch is cut from.

PR #1373, opened 2026-08-09
Let the capability probe describe the servers this repository runs

denisvmedia:
Closes the three gaps the capability probe's own verification declared, plus the
control weakness it flagged. Refs #1339, #1341.

Every literal block below is the output of the command shown above it, untrimmed.
Containers were started for this change and dropped after it; ports and names are
unique to this run.

## Gap 1 — the probe could not describe the repository's own MySQL service

`docker-compose.yaml` and `.github/workflows/go-integration-tests.yml` both pin
`mysql:26.7`. `cells.go` declared MySQL lines 9.7 and 8.4 only.

```
$ .probe-bin/capability-probe --db-url 'mysql://root:probe@localhost:33741/probe'   # before
  banner       26.7.0
  matrix cell  NONE — no cell in cells.go covers this release line
  resolution   version-specific=false saturated=true newest-measured="9.x"
summary: 24 rows — 0 AGREES, 0 DISAGREES, 24 UNDECIDABLE; decided 0
capability-probe: no matrix cell covers mysql 26.7.0: this release line is not in the matrix, so nothing here promotes it to measured (add a cell in cells.go)
PROBE_EXIT=1
```

Probing every other tag the repository pins found a **second** line off the matrix,
measured the same way:

```
$ .probe-bin/capability-probe --db-url 'clickhouse://default:probe@localhost:19742/probe'
  banner       24.10.4.191
  matrix cell  NONE — no cell in cells.go covers this release line
CH2410_EXIT=1
```

`clickhouse/clickhouse-server:24.10` is the workflow's second ClickHouse service —
the one #1341's own table lists.

**Both cells added, data only.** The `cells.go` diff is 27 added lines and 3 removed,
all inside the `Cells` slice literal and its doc comments; no function body, type or
signature is touched.

MySQL 26 gets `Preset: nil`, for exactly the reason PostgreSQL 18 does.
`newestMeasuredMySQLMajor` is 9, so a 26 server resolves **saturated**: it receives
`MySQL84` from the ladder's open-topped arm, not from a match. Writing
`Preset: capability.MySQL84` would assert the ladder selected it, which is false, and
`cellProblems` would refuse it as a contradiction. Promoting the line means moving
that constant — a reviewed claim, and not a data change — so this PR does not do it.

After:

```
$ .probe-bin/capability-probe --db-url 'mysql://root:probe@localhost:33741/probe'   # after
  banner       26.7.0
  matrix cell  mysql 26.7 [preset none, version-ladder]
  resolution   version-specific=false saturated=true newest-measured="9.x"
summary: 24 rows — 0 AGREES, 0 DISAGREES, 24 UNDECIDABLE; decided 0, floor 1
capability-probe: matrix cell mysql 26.7 has no measured capability preset (...)
PROBE_EXIT=1
```

Still exit 1, and that is the point: the probe now names the line and its gap
instead of saying the line does not exist. **The finding is that all 23 observable
rows agree with the stand-in** — the run reports zero mismatches, so nothing about
MySQL 26.7 contradicts `MySQL84`; the only thing missing is a measured preset.

### The other pinned tags

| pinned image | measured version | cell | exit | finding |
|---|---|---|---|---|
| `postgres:18` | 18.4 | postgres 18, no preset | 1 | 24 observed, 0 mismatches |
| `mysql:26.7` | 26.7.0 | mysql 26.7, no preset (new) | 1 | 23 observed, 0 mismatches |
| `mariadb:10.11` | 10.11.18 | mariadb 10.11, MariaDB1011 | **0** | 22 AGREES, 0 DISAGREES |
| `clickhouse/clickhouse-server:26.7` | 26.7.3.19 | clickhouse 26.7 | 1 | no plan for the dialect |
| `clickhouse/clickhouse-server:24.10` | 24.10.4.191 | clickhouse 24.10 (new) | 1 | no plan for the dialect |
| `cockroachdb/cockroach:v26.2.5` | 26.2.5 | cockroachdb 26.2 | 1 | **3 mismatches** |
| `yugabytedb/yugabyte:2026.1.0.0-b118` | 2026.1.0.0 | yugabytedb 2026.1 | 1 | **3 mismatches** |
| `mcr.microsoft.com/mssql/server:2025-latest` | 17.0.4065.4 | sqlserver 17.0 | 1 | no plan for the dialect |

Two cells that exist turn out to be **wrong about the live server**, which is what
"known to be true and not merely present" was supposed to find:

```
$ .probe-bin/capability-probe --db-url 'postgres://root@localhost:26741/defaultdb?sslmode=disable'
preset and server disagree:
  role_management                        preset says false server does true  [UNDECIDABLE]
  row_level_security                     preset says false server does true  [UNDECIDABLE]
  sequences                              preset says false server does true  [UNDECIDABLE]
```

```
$ .probe-bin/capability-probe --db-url 'postgres://yugabyte@localhost:15743/yugabyte?sslmode=disable'
preset and server disagree:
  advisory_locks                         preset says false server does true  [UNDECIDABLE]
  create_index_concurrently              preset says false server does true  [UNDECIDABLE]
  row_level_security                     preset says false server does true  [UNDECIDABLE]
```

Both are `UNDECIDABLE` because those lines are banner-refined, and both are
reported anyway, which is the mechanism #1339 built for exactly this. **Neither
preset is changed here** — a preset edit is its own reviewed measurement, and
CockroachDB's `create_index_concurrently` row shows why: the server accepts the
statement inside an explicit transaction block, so the keyword is parsed and
dropped. Filed as a follow-up rather than smuggled in.

### The guard that should have caught Gap 1

`TestCells_CoverEveryLineTheRepositoryRuns` carried a hand-written version list
that still named `mysql 9.7.1` and named no ClickHouse at all. It now derives the
question from the two files that answer it, and an image in neither the database
map nor the not-a-database list **fails** rather than passing unexamined. Run
against the pre-change `cells.go` it names both gaps:

```
$ git restore --source=origin/master --worktree internal/capabilityprobe/cells.go
$ go test ./internal/capabilityprobe/ -count=1 -run TestCells_DeclareEveryDatabaseContainerThisRepositoryStarts
    --- FAIL: .../clickhouse/clickhouse-server:24.10
              this repository starts clickhouse/clickhouse-server:24.10 and no matrix cell declares it, ...
    --- FAIL: .../mysql:26.7
              this repository starts mysql:26.7 and no matrix cell declares it, ...
BEFORE_STATE_TEST_EXIT=1
```

A tag is not a version — `mssql:2025-latest` is a marketing year and
`clickhouse-server:26` is a floating major — so the two questions are split: this
test asks whether the matrix *declares the container*, and
`TestCells_CoverEveryVersionMeasuredFromALiveServer` asks whether what those
containers report *on the wire* lands on a cell. Its rows are the twelve versions
measured above.

## Gap 2 — coverage could erode without turning the build red

`Report.Err` failed only at `Decided() == 0`, so a run that decided 1 row of 24
exited 0. The floor is now every row the dialect's plan promised to answer,
**derived from the plan** rather than written down, and 0 on a line no observation
can be credited to (where every row is undecidable by construction and a demand
would be unmeetable). Live runs clear it exactly:

```
postgres:17    decided 24, floor 24   PG17_EXIT=0
mysql:9.7      decided 23, floor 23   MY97_EXIT=0
mariadb:10.11  decided 22, floor 22   MARIADB_EXIT=0
cockroachdb    decided 0,  floor 1    CRDB_EXIT=1
```

The floor is on the summary line because a reader who sees "decided 22" cannot
otherwise tell an intact run from an eroded one.

**Mutant** — the cheaper wrong implementation is "one decided row proves the probe
ran", `func (r *Report) floor() int { return 1 }`. It compiles.

```
$ go test ./internal/capabilityprobe/ -count=1     # mutant live
--- FAIL: TestReportErr_TheFloorIsWhatThePlanPromised
    --- FAIL: .../one_promised_row_short_is_a_failure,_not_a_rounding_error
    --- FAIL: .../the_shape_the_old_floor_let_through:_one_row_decided_out_of_twenty-four
MUTANT_B_TEST_EXIT=1
```

The "decided everything promised" row and the "decided nothing" row stay green
under the mutant, and so does the shipped
`TestAssemble_UndecidableRowsAreNotCountedAsDecided` — the new test is the one
that catches erosion, and the old guard is preserved rather than replaced.
Restored: `RESTORED_GREEN_EXIT=0`.

## Gap 3 — the coverage test counted an unanswered key as answered

`TestPlans_AnswerEveryRegisteredCapabilityExactlyOnce` seeded its counter from
`dialectPlan.undecided` and then added the experiments, so moving a key from one
to the other kept the total at 1. The two are counted separately now, and a second
test pins **which** keys each dialect may answer by declaration — postgres none,
mysql `role_management`, mariadb `role_management` and `sequences` — because a
count can never see that distinction.

**Mutant** — the cheapest way to make a stubborn row stop failing: move
`capability.XMLType` out of `postgresPlan`'s experiments into its `undecided` map,
with a plausible reason attached. It compiles and the key is still answered once.

```
$ go test ./internal/capabilityprobe/ -count=1     # mutant live, new guards
--- FAIL: TestPlans_DeclareUndecidableOnlyWhereThisFileRecordsWhy/postgres
--- FAIL: TestDecidable_IsDerivedFromThePlanAndTheLine/postgres...
MUTANT_A_TEST_EXIT=1

$ git restore --source=origin/master --worktree internal/capabilityprobe/{probe_internal_test.go,cells_test.go,probe.go,render.go}
$ go test ./internal/capabilityprobe/ -count=1     # same mutant, shipped guards
ok  	go.5x5.cz/ptah/internal/capabilityprobe	6.366s
OLD_GUARDS_VS_MUTANT_A_EXIT=0
```

Green under the shipped guards, red under the new ones. Note that
`TestPlans_AnswerEveryRegisteredCapabilityExactlyOnce` does **not** fail even after
the fix — correctly, since the key is answered once — which is why the second test
had to exist. The two guards compose: this one blocks static erosion, the floor
blocks runtime erosion, and neither is enough alone, because a mutant that moves a
key into `undecided` lowers the floor and the decided count together.

## The control that asserted an equivalence instead of executing it

`version_test.go` demonstrated the SQL Server marketing-year misread using the
probe's own `ParseVersion` on the postgres dialect, so nothing would notice if
`capability.parseVersion` changed.

`parseVersion` is unexported and every exported caller collapses it into a preset
or a boolean, so two tests were added rather than one. The probe-side test runs
`capability.ResolveServerVersion` on the postgres dialect over the same banner
bytes: it saturates exactly when the parsed major is above the newest measured
PostgreSQL line, so a banner read as 2025 saturates and the 17.0.4065.4 the
sqlserver arm recovers does not — one call each, opposite verdicts. A new white-box
test in the capability package calls `parseVersion` directly, where the numbers are
readable, and pins both misreads with two correctly-read banners as the control.

**YugabyteDB cannot be reached that way, and the test says so** rather than
pretending: `ResolveServerVersion` matches `-yb-` before it parses anything, so no
exported surface routes that banner through the shared parser. The masking itself
is executed instead.

**Mutant** — the cheapest "fix" someone would apply to the shared parser: cut the
banner at the first `" - "` before scanning.

```
$ go test ./internal/capabilityprobe/ ./core/platform/capability/ -count=1   # mutant live
--- FAIL: TestParseVersion_TheSharedParserReallyMisreadsTheSQLServerBanner
--- FAIL: TestParseVersion_ReadsTheWrongNumberOutOfTwoRealBanners/the_SQL_Server_banner...
MUTANT_D_TEST_EXIT=1

$ go test ./internal/capabilityprobe/ -count=1 -run 'TestParseVersion_CorrectsTheBannersOneParserCannotRead'
--- PASS: TestParseVersion_CorrectsTheBannersOneParserCannotRead   (all 10 rows)
SHIPPED_CONTROL_VS_MUTANT_D_EXIT=0
```

The shipped control passes while the shared parser it names has changed underneath
it. That is the whole finding, executed.

## Documentation

`newestMeasuredMySQLMajor`'s comment claimed "the newest MySQL the integration
matrix runs (mysql:9.7)". The matrix runs `mysql:26.7`. Corrected, and
`docs/capabilities.md` now says MySQL is in the same saturated position as
PostgreSQL 18 — two of the three refined dialects are currently described to their
own CI by a stand-in.

## Gates

Each run separately with its exit code read directly, none through a pipe: `go
build ./...` 0, `go vet ./...` 0, `go vet -tags integration ./integration/...` 0,
`go test ./... -count=1` 0 (0 FAIL lines), `go tool qtlint ./...` 0,
`golangci-lint run ./...` 0 (0 issues), `scripts/check-test-style.sh` 0, the three
`check-public-api*` scripts 0/0/0, `npm ci` plus all 15 `check:*` and `:selftest`
scripts in `docs/site/package.json` 0 (`check:responsive` after `npm run build`,
which it requires).

`check-test-style.sh` was verified to actually inspect the new files rather than
pass by skipping them: planting an `if` in a new test body turns it red naming
`core/platform/capability/capability_internal_test.go`, exit 1, and removing it
returns exit 0.

## Not done here, and why

- **Promoting `mysql 26.7` to a measured preset.** Needs `newestMeasuredMySQLMajor`
  moved from 9 to 26, which the constant's own comment calls a deliberate reviewed
  claim and not a data change. The evidence is now in hand: 23 of 23 observable
  rows agree.
- **The CockroachDB and YugabyteDB preset disagreements.** Six rows across two
  presets, each its own measurement.
- **Wiring any of this into CI.** That is the body of #1341 and is untouched. This
  PR closes the three gaps that made the gate unusable for the versions this
  repository actually runs; #1341's tier-2 and tier-3 matrices remain open.
- **`#1341`'s "a cell with no preset entry fails the build"** criterion. Three cells
  (postgres 18, mariadb 12.3, mysql 26.7) have no preset by design today; making
  that a build failure is the CI work, not a probe change.

### On the commit count

Six commits carrying one message: local commits were blocked by GPG signing in a
headless shell (`gpg: signing failed: Inappropriate ioctl for device`), so this was
pushed through the API, which caps a single call below the 158 KB payload. The head
tree was verified byte-for-byte against the tested worktree (`git diff --quiet
origin/issue-1341-probe-gaps` exit 0). Squash on merge.

### DIFF
--- a/cmd/capability-probe/main.go
+++ b/cmd/capability-probe/main.go
@@ -4,7 +4,14 @@
 // It answers one question per capability: does the server actually behave the
 // way core/platform/capability says it does? A disagreement exits non-zero. So
 // does a run that decided nothing — a probe that skipped every row must not
-// read as a probe that passed every row.
+// read as a probe that passed every row — and so does a run that decided fewer
+// rows than the dialect's plan promised to answer. The report's summary line
+// prints the count and that floor side by side, because "decided 22" alone
+// cannot tell an intact run from an eroded one.
+//
+// A server on a release line cells.go does not declare exits non-zero too, and
+// so does one on a line declared without a measured preset: both are gaps the
+// probe names rather than papers over.
 //
 //	capability-probe --db-url postgres://user:pw@localhost:5432/db
 //	capability-probe --db-url mysql://root:pw@localhost:3306/db --evidence
--- a/core/platform/capability/capability.go
+++ b/core/platform/capability/capability.go
@@ -863,8 +863,9 @@ func ForDialect(dialect string) Capabilities {
 // measures that line, together with the preset it deserves — never as a side
 // effect of bumping a container tag.
 const (
-	// MySQL84 covers 8.4 LTS through the 9.x LTS line, which is the newest
-	// MySQL the integration matrix runs (mysql:9.7).
+	// MySQL84 covers 8.4 LTS through the 9.x LTS line. The integration matrix
+	// already runs mysql:26.7, which therefore resolves saturated: Ptah has no
+	// measured MySQL 26 capability line yet.
 	newestMeasuredMySQLMajor = 9
 	// MariaDB1011 covers 10.2 through the 11.x lines; the integration matrix
 	// runs mariadb:10.11.
--- a/core/platform/capability/capability_internal_test.go
+++ b/core/platform/capability/capability_internal_test.go
@@ -1,0 +1,84 @@
+package capability
+
+// White-box testing required: parseVersion is unexported and every exported
+// caller collapses its result into a preset or a boolean, so no external test
+// can read the numbers it produced. Those numbers are the subject here — two
+// banners this package reads WRONGLY, and a claim written in prose in
+// internal/capabilityprobe/version.go that until now nothing executed.
+
+import (
+	"testing"
+
+	qt "github.com/frankban/quicktest"
+)
+
+// sqlServer2025Banner is what @@VERSION returned from a live
+// mcr.microsoft.com/mssql/server:2025-latest container, verbatim including the
+// embedded newlines. internal/capabilityprobe/version_test.go pins the same
+// bytes; each package keeps its own copy because this one cannot import that
+// one without a cycle.
+const sqlServer2025Banner = "Microsoft SQL Server 2025 (RTM-CU7) (KB5096981) - 17.0.4065.4 (X64) \n" +
+	"\tJul  8 2026 23:26:08 \n" +
+	"\tCopyright (C) 2025 Microsoft Corporation\n" +
+	"\tEnterprise Developer Edition (64-bit) on Linux (Ubuntu 24.04.4 LTS) <X64>"
+
+// yugabyteBanner is what a live yugabytedb/yugabyte:2026.1.0.0-b118 reported
+// through YSQL. The leading number is the PostgreSQL compatibility version.
+const yugabyteBanner = "PostgreSQL 15.12-YB-2026.1.0.0-b0 on aarch64-unknown-linux-gnu, compiled by clang, 64-bit"
+
+// TestParseVersion_ReadsTheWrongNumberOutOfTwoRealBanners executes the misread
+// that internal/capabilityprobe exists to correct.
+//
+// It is a characterization test, not a wish: both rows below record what this
+// parser does today, and both are wrong about the product. The SQL Server row
+// is the one with teeth — the marketing year 2025 would select a preset for a
+// version that does not exist the day a `case platform.SQLServer:` is added to
+// the resolver's switch. The PostgreSQL and MySQL rows are the control that
+// makes the other two mean something: the same rule reads those banners
+// correctly, so the defect is in what these two banners put first, not in the
+// parser being broken for everything.
+func TestParseVersion_ReadsTheWrongNumberOutOfTwoRealBanners(t *testing.T) {
+	c := qt.New(t)
+
+	for _, tc := range []struct {
+		name   string
+		banner string
+		want   serverVersion
+	}{{
+		name:   "the SQL Server banner reads as the marketing year, not the product version 17.0.4065.4",
+		banner: sqlServer2025Banner,
+		want:   serverVersion{major: 2025},
+	}, {
+		name:   "the YugabyteDB banner reads as the PostgreSQL compatibility version, not the product 2026.1.0.0",
+		banner: yugabyteBanner,
+		want:   serverVersion{major: 15, minor: 12},
+	}, {
+		name:   "control: a PostgreSQL banner reads correctly under the same rule",
+		banner: "PostgreSQL 17.10 (Debian 17.10-1.pgdg13+1) on aarch64-unknown-linux-gnu",
+		want:   serverVersion{major: 17, minor: 10},
+	}, {
+		name:   "control: a MySQL banner reads correctly under the same rule",
+		banner: "9.7.1",
+		want:   serverVersion{major: 9, minor: 7, patch: 1},
+	}} {
+		c.Run(tc.name, func(c *qt.C) {
+			got, ok := parseVersion(tc.banner)
+			c.Assert(ok, qt.IsTrue)
+			c.Assert(got, qt.Equals, tc.want)
+		})
+	}
+}
+
+// TestResolveServerVersion_MasksTheYugabyteMisread executes the other half of
+// the claim: the YugabyteDB misread above never reaches a preset, because the
+// banner match fires before any version is parsed. That is why the SQL Server
+// case is the dangerous one and this one is not.
+func TestResolveServerVersion_MasksTheYugabyteMisread(t *testing.T) {
+	c := qt.New(t)
+
+	resolved := ResolveServerVersion("postgres", yugabyteBanner)
+	c.Assert(resolved.Capabilities, qt.DeepEquals, YugabyteDB25(),
+		qt.Commentf("the -yb- banner match must answer before parseVersion sees 15.12"))
+	c.Assert(resolved.Saturated, qt.IsFalse)
+	c.Assert(resolved.VersionSpecific, qt.IsTrue)
+}
--- a/docs/capabilities.md
+++ b/docs/capabilities.md
@@ -215,7 +215,10 @@ The newest measured line per refined dialect:
 
 PostgreSQL 18 therefore resolves saturated even though the integration matrix
 already runs `postgres:18`: Ptah has no measured PostgreSQL 18 capability line
-yet, so 18 is planned with the PostgreSQL 17 preset and says so.
+yet, so 18 is planned with the PostgreSQL 17 preset and says so. MySQL is in the
+same position for the same reason — the matrix runs `mysql:26.7`, which is the
+`26.7.0` in the examples above — so two of the three refined dialects are
+currently described to their own CI by a stand-in.
 
 Raising one of those numbers is the deliberate act of claiming a newer server
 line behaves like the preset it lands on. Do it in the change that measures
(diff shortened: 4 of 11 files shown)


denisvmedia:
Correction to the closing note in the description, and one addition.

**Seven commits, not six.** `88d9d62` was pushed after the description was
written and carries its own message. It updates two doc comments the change made
incomplete: `internal/capabilityprobe/doc.go` and `cmd/capability-probe/main.go`
both said the probe fails a run that decided nothing — still true, no longer the
whole rule. Someone meeting `decided 22 of 24 capability rows, 1 fewer than the
23 the mysql plan promised to answer` had nothing to read that said it was a
designed outcome. The command doc also now names the two cell-level refusals,
since a first user meets those before anything else.

Gates re-run on the doc commit, each exit code read directly: `go build ./...` 0,
`go vet ./...` 0, `go vet -tags integration ./integration/...` 0, `go test ./...
-count=1` 0 with zero `FAIL` lines, `go tool qtlint ./...` 0, `golangci-lint run
./...` 0, `scripts/check-test-style.sh` 0, all three `check-public-api*` 0.

**A false alarm worth recording, because the shape recurs.** The first
`go test ./...` after the doc edit reported 7 failures in `internal/atlashclrender`,
`internal/convert/fromschema` and `migration/planner` — three packages this change
does not touch. It was self-inflicted and not a regression.

Earlier I had simulated the merge against the newer `master` by pulling #1369's
eleven files into the worktree with `git restore --source=origin/master --worktree`
(build, vet and the full suite all passed in that state, which is the merge check
this PR needed). Undoing it with `git restore --worktree` reverted the five files
#1369 *modified* and left the six it *added*: restore replaces tracked content and
cannot delete a path the index does not know about. The result was old sources next
to new tests.

The reason it went unnoticed is the part worth keeping: `git diff --quiet` returned
0, so my "tree is clean" check passed. `git diff` does not see untracked files. The
verification that would have caught it is `git status --porcelain | grep -c '^??'`,
which is now what I check alongside the diff — and it reads 0 for the current head,
whose tracked tree is also byte-identical to the tested worktree.

PR #1374, opened 2026-08-09
Bind the last two migration writers to the directory they opened (#1118)

denisvmedia:
`migration/generator` had five writer transactions. Three ran through a rooted
handle; the other two did not, and this closes them.

## Reproduction, on master at `69c5b2ea`

The issue's own shape: replace the directory after the writer has validated and
created it, and before it re-resolves the same pathname to commit the checksum.
A probe polls for the migration file to appear — which happens at `O_CREAT`,
before the 64 MiB body is written — then renames the directory aside and drops a
fresh directory at its pathname, so the two renames land inside the writer's own
`WriteString`.

```
go run ./.repro/checkpointswap
--- WriteAtlasCheckpointFile ---
swap error:        <nil>
writer error:      <nil>
reported path:     /var/.../ptah-1118-785235827/migrations/20990101000000_squash.sql
retained dir retained:      [20990101000000_squash.sql]
decoy at pathname migrations: [atlas.sum]
--- WriteCheckpointFiles ---
swap error:        <nil>
writer error:      <nil>
retained dir retained:      [2099010100_squash.checkpoint.up.sql]
decoy at pathname migrations: [2099010100_squash.checkpoint.down.sql ptah.sum]
--- WriteDataMigrationFiles ---
swap error:        <nil>
writer error:      <nil>
retained dir retained:      [2099010100_seed.up.sql]
decoy at pathname migrations: [2099010100_seed.down.sql ptah.sum]
```

Every one returned a **nil error** and split one transaction across two
directories. Both outcomes are unusable: the retained directory is uncovered by
any checksum, and the impostor carries a `.sum` describing a snapshot it never
held. The pair writers also left half a migration behind — an up file with no
down half.

The same probe on this branch:

```
--- WriteAtlasCheckpointFile ---
retained dir retained:      [20990101000000_squash.sql atlas.sum]
decoy at pathname migrations: []
--- WriteCheckpointFiles ---
retained dir retained:      [2099010100_squash.checkpoint.down.sql 2099010100_squash.checkpoint.up.sql ptah.sum]
decoy at pathname migrations: []
--- WriteDataMigrationFiles ---
retained dir retained:      [2099010100_seed.down.sql 2099010100_seed.up.sql ptah.sum]
decoy at pathname migrations: []
```

The three already-rooted surfaces do **not** reproduce, confirmed before
changing anything:

```
go test ./internal/atlasmigrate/ -run 'TestGenerateDiff_(ReplacedDirectoryCannotRedirectPublication|CreatesMissingMigrationDirectoryInsideOpenedRoot|RefusesMigrationDirectoryOutsideOpenedRoot|RenamedDirectoryCannotRedirectPublication)' -count=1
ok  	go.5x5.cz/ptah/internal/atlasmigrate	1.135s
```

## What changed

`WriteAtlasCheckpointFile`, `WriteCheckpointFiles` and `WriteDataMigrationFiles`
bind their output directory once through `atlasmigrate.MigrationWriter` and name
only direct children of that handle for the create, the withdrawal, the checksum
and the flush. Exported signatures are unchanged, so the public API snapshot does
not move.

`ptah.sum` now reaches the same conditional commit `atlas.sum` already used:
`publishDirSum` becomes `publishDirSumAs`, which carries the file name, and
`MigrationWriter.PublishSum` selects it from the layout. That closes the issue's
bullet 5 for the paired layout — the checksum is computed from, and committed
through, the handle the files were created through.

The refusal is the exclusive create itself rather than a preceding
`fileExists` check, so a name taken between choosing it and creating it is
reported instead of overwritten.

`ensureMigrationOutputDir`, `writeNewMigrationFile` and `fileExists` were the
last pathname writing primitives in the package and are removed rather than left
standing beside the boundary. Their measured behavior moves with them and is
stated on `bindMigrationOutputDir`.

## Behavior preserved

```
go run ./.repro/parents
two missing levels      err=<nil>  dir=[20990101000000_squash.sql atlas.sum]
three missing levels    err=<nil>  dir=[2099010100_seed.down.sql 2099010100_seed.up.sql ptah.sum]
ancestor is a file      err=failed to create output directory: create migration directory parent: mkdir /var/.../a: not a directory
leaf is a file          err=failed to create output directory: open migration directory: open /private/var/.../migrations: not a directory
```

Missing parent levels are still created (#1241 item 4, re-measured at two and
three levels) and a non-directory component is still refused. The
`already exists` and `checkpoint file … already exists` messages are unchanged.

## Containment matrix, measured on this filesystem (APFS)

```
go run ./.repro/containment
== rooted writer: GenerateEmptyMigration, AllowedOutputRoot set ==
1 symlink out of the root                                REFUSED   outside=[]
2 relative traversal (root/../<outside>)                 REFUSED   outside=[]
3 absolute path outside the root                         REFUSED   outside=[]
4 root is itself a symlink (dir inside it)               ACCEPTED  outside=[20260809190713_added.sql atlas.sum]
5 ancestor symlink out of the root                       REFUSED   outside=[]

== unrooted writer: WriteAtlasCheckpointFile, no root parameter ==
1 symlink out of the root                                ACCEPTED
2 relative traversal (root/../<outside>)                 ACCEPTED
3 absolute path outside the root                         ACCEPTED
4 root is itself a symlink (dir inside it)               ACCEPTED
```

Case 4 accepting is correct: the root is spelled through a symlink but the
directory genuinely lives inside what that root resolves to. The three checkpoint
and data writers take no confinement root at all — `ptah migrations checkpoint`
and `ptah migrations data` read `--migrations-dir` straight from the operator, so
there is no project root to escape. What they gain here is **binding**, not
confinement: whatever object the pathname selected at bind time is the only
object the transaction writes into. Adding a confinement root to them would need
a new exported option and is not in this change.

**On `os.SameFile`:** these writers never call `Revalidate`. Containment here is
`os.Root` handle identity (openat), not an inode comparison, and the regressions
assert *which directory received the bytes* rather than whether two identifiers
match — so they answer the same on APFS and on ext4.

## Mutants: RED then GREEN

**M1 — commit the checksum by pathname** (`migratesum.WritePrecomputedWithFormat(writer.Path(), …)` in `publishMigrationDirSum`). The cheaper wrong implementation, and literally what master did.

```
go test ./migration/generator/ -run 'TestCheckpointWriters|TestMigrationWritersRefuse|TestGenerateEmptyMigration' -count=1
--- FAIL: TestCheckpointWritersReplacedDirectoryCannotRedirectTheWrite (0.08s)
    --- FAIL: .../atlas_checkpoint      --- FAIL: .../paired_checkpoint      --- FAIL: .../data_migration
--- FAIL: TestCheckpointWritersCreateAMissingDirectoryThroughTheBoundParent (0.01s)
    --- FAIL: .../atlas_checkpoint      --- FAIL: .../data_migration
--- FAIL: TestGenerateEmptyMigrationReplacedDirectoryCannotRedirectTheWrite (0.06s)
--- FAIL: TestGenerateEmptyMigrationCreatesMissingDirectoryThroughTheBoundParent (0.00s)
--- FAIL: TestGenerateEmptyMigrationScansTheHeldDirectoryForItsVersion (0.02s)
FAIL
```
Restored: `ok  	go.5x5.cz/ptah/migration/generator	7.535s`

**M2 — decide collisions from the listing taken at bind time, then create unconditionally.** One `ReadDir` instead of a conditional syscall per file; still compiles.

```
--- PASS: TestCheckpointWritersReplacedDirectoryCannotRedirectTheWrite
--- PASS: TestCheckpointWritersCreateAMissingDirectoryThroughTheBoundParent
--- FAIL: TestMigrationWritersRefuseADestinationTakenAfterTheNameWasChosen (0.06s)
    --- FAIL: .../atlas_checkpoint_refuses
    --- FAIL: .../paired_checkpoint_refuses_when_the_up_half_is_taken
    --- FAIL: .../data_migration_withdraws_the_up_half_when_the_down_half_is_taken
--- FAIL: TestGenerateEmptyMigrationStepsPastADestinationTakenAfterTheNameWasChosen (0.02s)
--- PASS: TestWriteAtlasCheckpointFile_RefusesToOverwrite
--- PASS: TestWriteCheckpointFiles
--- PASS: TestWriteDataMigrationFiles
```

This is the discrimination the audit asked for. Every **pre-staged** fixture
stays green — a file already present when the handle opens is answered by the
open — while all four **in-window** rows go red. Detail on the first:

```
error: got nil error but want non-nil
regexp: ".*checkpoint file .*20990101000000_squash\\.sql already exists"
```
Restored: `ok  	go.5x5.cz/ptah/migration/generator	0.732s`

**M3 — make the checksum commit unconditional** (`w.dir.ReplaceFile(tempName, sumName)` in place of the conditional `PublishFile`).

```
go test ./internal/atlasmigrate/ -run 'TestPublishDirSum_' -count=1 -v
--- FAIL: TestPublishDirSum_RefusesADestinationReplacedInsideTheCommitWindow (0.06s)
    --- FAIL: .../atlas.sum,_absent_when_the_commit_captured_its_destination
    --- FAIL: .../atlas.sum,_present_when_the_commit_captured_its_destination
    --- FAIL: .../ptah.sum,_absent_when_the_commit_captured_its_destination
    --- FAIL: .../ptah.sum,_present_when_the_commit_captured_its_destination
--- PASS: TestPublishDirSum_CommitsWhenNothingReplacedTheDestination
```
The non-interference control stays green, so the rows measure the replacement
rather than a writer that refuses everything.
Restored: `ok  	go.5x5.cz/ptah/internal/atlasmigrate	4.224s`

## Acceptance criteria this closes

- **`atlas.sum` commit-point replacement** — `TestPublishDirSum_RefusesADestinationReplacedInsideTheCommitWindow`, both destination shapes (`ExpectAbsent` and `ExpectFile`) and both layouts, asserting the rival's bytes survive and no staged temporary is left.
- **Final-filename replacement, measured in-window** — `TestMigrationWritersRefuseADestinationTakenAfterTheNameWasChosen` plus `TestGenerateEmptyMigrationStepsPastADestinationTakenAfterTheNameWasChosen`. The two verbs answer differently on purpose: an explicit-version writer refuses, the skeleton writer steps to the next free version. Neither overwrites.
- **Directory replacement for the two remaining writers** — `TestCheckpointWritersReplacedDirectoryCannotRedirectTheWrite`, asserting both halves: artifacts in the bound object, decoy untouched.
- **Initially-missing directory** for both.

Two unexported seams are added because those two commit points had no name:
`afterMigrationFileNamesChosen` and `beforeDirSumCommit`. Each fixture file
carries the `AGENTS.md` white-box justification, and each states why a
replacement staged before the call measures the open rather than the commit.
`//go:build unix` on the rename fixtures for the reason #1193 recorded: Win32
refuses to rename a directory the run holds open without `FILE_SHARE_DELETE`, so
there is no hostile step to perform rather than one expected to fail.

## Not closed

- **Editor preparation (`--edit`) and rollback for `migrate new`.** Converted `migrate new` refuses `--edit` outright; native `ptah migrations create --edit` still has happy-path coverage only.
- **A missing-directory row for the converted `migrate new` writer** (`cmd/atlas/migrate_new.go`).
- **Root confinement for `ptah migrations checkpoint` / `ptah migrations data`.** They accept no `AllowedOutputRoot`; giving them one is an exported-API change and is not attempted here.

## Gates

Every one run on this tree, each exit code read directly:

```
go build ./...                        0
go vet ./...                          0
go vet -tags integration ./integration/...  0
go test ./... -count=1                0
go tool qtlint ./...                  0
golangci-lint run ./...               0   (0 issues)
scripts/check-test-style.sh           0   (175 conditional groups, 42 white-box files)
scripts/check-public-api.sh           0
scripts/check-public-api-snapshot.sh  0
scripts/check-public-api-released.sh  0   (pre-existing approved pre-v1 entries only)
```

All 14 `check:*` scripts in `docs/site/package.json` exit 0, including
`check:responsive` after `npm run build`.

One flake seen and dismissed: `TestSVGReportsGraphvizStderrOnFailure` in
`cmd/viz` failed once with `signal: killed` at 10.01s while the docs site was
building concurrently, and passed `-count=3` in isolation and on a serial rerun
of the whole suite.

Refs #1118 — the issue keeps the three bullets under "Not closed".

### DIFF
--- a/docs/site/src/content/docs/concepts/migration-directory.md
+++ b/docs/site/src/content/docs/concepts/migration-directory.md
@@ -98,14 +98,29 @@ bytes.
 ## Writing back to the directory
 
 The verbs that write a migration directory — `migrate diff`, `migrate new`, and
-native `ptah migrations generate` and `ptah migrations create` — bind it the
+native `ptah migrations generate`, `ptah migrations create`,
+`ptah migrations checkpoint` and `ptah migrations data` — bind it the
 same way and keep the binding. They open the directory and its parent once,
-before staging, and every staged file, published migration, `atlas.sum`,
-journal, commit marker, rollback quarantine and cleanup entry is named as a
-direct child of one of those two handles. A migration directory that does not
-exist yet is created through the bound parent, so it is materialized where the
-run looked rather than where the pathname points by then. Recovery of an
-interrupted batch runs through the same handles.
+before staging, and every staged file, published migration, `atlas.sum` or
+`ptah.sum`, journal, commit marker, rollback quarantine and cleanup entry is
+named as a direct child of one of those two handles. A migration directory that
+does not exist yet is created through the bound parent, so it is materialized
+where the run looked rather than where the pathname points by then. Recovery of
+an interrupted batch runs through the same handles.
+
+Both commits inside that boundary are conditional on what the run observed. A
+migration file is created exclusively, so a name taken since the version was
+chosen is reported rather than overwritten. The integrity file is a replacement
+by construction and cannot use the same rule, so its commit is bound to the
+checksum state captured immediately before it: a rival that wrote its own
+`atlas.sum` or `ptah.sum` in between is reported, and its bytes are left in
+place.
+
+`atlas.sum` and `ptah.sum` take the same path. Until
+[#1118](https://github.com/stokaro/ptah/issues/1118) closed it, `ptah.sum` was
+written by pathname and unconditionally, which let a checkpoint or data
+migration land in one directory while its checksum landed in another, leaving
+the first uncovered and the second describing a snapshot it never held.
 
 Replacing the directory after the run validated it therefore cannot redirect
 what it writes, and a directory configured through `atlas.hcl` stays inside the
--- a/internal/atlasmigrate/migrationwriter.go
+++ b/internal/atlasmigrate/migrationwriter.go
@@ -7,6 +7,7 @@ import (
 
 	"go.5x5.cz/ptah/internal/migratesum"
 	"go.5x5.cz/ptah/internal/pathguard"
+	"go.5x5.cz/ptah/migration/migrator"
 )
 
 // MigrationWriter is the rooted migration-directory capability a writer
@@ -140,15 +141,28 @@ func (w *MigrationWriter) Remove(name string) error {
 	return err
 }
 
-// PublishAtlasSum commits atlas.sum through the bound handle, conditionally on
-// the checksum state this transaction observed, and returns its path. A
-// concurrent writer that replaced atlas.sum after the snapshot was taken is
+// PublishSum commits the integrity file the layout selects -- atlas.sum for the
+// Atlas layout, ptah.sum for the paired one -- through the bound handle,
+// conditionally on the checksum state this transaction observed, and returns
+// its path. A concurrent writer that replaced it after the snapshot was taken is
 // reported instead of overwritten.
-func (w *MigrationWriter) PublishAtlasSum(sum *migratesum.SumFile) (string, error) {
+//
+// The layout is a parameter rather than a second method because the two layouts
+// differ only in the file name: both must reach the destination through the same
+// handle the migration files were created through, or the checksum describes a
+// directory this transaction did not write into (stokaro/ptah#1118).
+func (w *MigrationWriter) PublishSum(
+	format migrator.MigrationDirFormat,
+	sum *migratesum.SumFile,
+) (string, error) {
 	if !w.dir.Exists() {
 		return "", w.dir.missingDirError()
 	}
-	return publishDirSum(w.dir, sum)
+	sumName, err := migratesum.FileNameForFormat(format)
+	if err != nil {
+		return "", err
+	}
+	return publishDirSumAs(w.dir, sumName, sum)
 }
 
 // PublishArtifacts durably publishes artifacts as one journaled batch through
--- a/internal/atlasmigrate/publication.go
+++ b/internal/atlasmigrate/publication.go
@@ -1150,38 +1150,64 @@ func publicationStagedNames(journal publicationJournal) []string {
 	return names
 }
 
+// beforeDirSumCommit runs after the checksum commit has captured the state it
+// intends to replace and staged its replacement, and immediately before the
+// conditional rename that publishes it. It is nil outside tests.
+//
+// It exists because that gap is the one thing the conditional commit is for and
+// nothing else can stage a replacement inside it. Every other moment is either
+// before the capture -- where the capture simply observes the newer state -- or
+// after the rename, where there is nothing left to protect. A test that writes
+// over the checksum before calling the writer measures the capture, not the
+// commit (stokaro/ptah#1118).
+var beforeDirSumCommit func()
+
+func notifyBeforeDirSumCommit() {
+	if beforeDirSumCommit != nil {
+		beforeDirSumCommit()
+	}
+}
+
 // publishDirSum commits atlas.sum through the same rooted handle that published
 // the migration files, conditionally on the checksum state the transaction
 // observed. A concurrent writer that replaced atlas.sum after the migration
 // snapshot was verified is reported instead of overwritten.
 func publishDirSum(w *migrationWriterDir, sum *migratesum.SumFile) (string, error) {
+	return publishDirSumAs(w, migratesum.AtlasFileName, sum)
+}
+
+// publishDirSumAs is publishDirSum for a caller that selects the integrity file
+// itself, so the paired `ptah.sum` layout commits through the same rooted,
+// conditional path as the Atlas one rather than by pathname.
+func publishDirSumAs(w *migrationWriterDir, sumName string, sum *migratesum.SumFile) (string, error) {
 	if sum == nil {
 		return "", errors.New("migration checksum must not be nil")
 	}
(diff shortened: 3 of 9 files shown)


denisvmedia:
One defect in the branch that I introduced and could not correct, stated precisely rather than left to be found.

Local commit signing on this machine needs an interactive passphrase that is not available here, so the branch was written through the GitHub contents API, which replaces whole files. Transcribing `migration/generator/generator.go` (2793 lines) dropped exactly one line at the seam between two read windows. Verified by byte comparison against the tree every gate above was run on:

```
git show origin/agent/1118-rooted-checkpoint-writers:migration/generator/generator.go | diff - migration/generator/generator.go
2699a2700
> // empty, so a version scan over a missing directory starts from scratch.
```

That is the whole difference — every other byte of every file on the branch matches, checked with `cmp` per file. The effect is a truncated sentence in the `migrationDirFileNames` doc comment, which currently ends `…a directory that cannot be listed reads as`. It compiles, it is not a code path, and `godot` is not among the enabled linters, so no gate sees it.

The fix is to reinsert that one line after line 2699. I did not attempt a second full-file transcription: the failure mode of another 2793-line rewrite is a corrupted code line rather than a truncated comment, which would be strictly worse than what is there now.

denisvmedia:
Addressed in `0e7e7c99`: restored the dropped `migrationDirFileNames` doc-comment line. Re-ran `git diff --check` and `GOCACHE=/private/tmp/ptah-go-cache go test ./migration/generator -count=1`; the full PR CI is green now.

PR #1422, opened 2026-08-11
fix(import): require exact Goose no-transaction marker

denisvmedia:
Follow-up to the actionable Codex review finding on #1402.

GENERAL CAPABILITY

## Summary

- recognize `-- +goose NO TRANSACTION` as whole-migration metadata only when the directive line is exact
- keep case- and whitespace-normalized lookalikes in the imported SQL instead of silently disabling transaction execution
- retain exact directive recognition for CRLF source files and inside `StatementBegin` blocks
- document the exact-line contract on both native migration-import pages

The semantic implementation remains in `migration/importer`, the shared Ptah migration-import package used by the native `ptah migrations import` surface. This is not Atlas-specific CLI machinery and adds no compatibility-only API.

## Verification

- `go test -race ./migration/importer ./cmd/migrationsimport -count=1`
- `go tool qtlint ./migration/importer`
- `go vet ./migration/importer`
- `go tool teststyle -baseline .teststyle-baseline.json -root .`
- `golangci-lint run ./migration/importer`
- complete documentation checks: links, redirects, core-doc links, page health, exit codes, style self-test/style, versions self-test, site build, responsive self-test/check, and `npm audit --audit-level=low`

## Documentation impact

Checked the native migration-import reference and workflow documentation plus all Goose `NO TRANSACTION` mentions. Updated:

- `docs/migrations-import.md`
- `docs/site/src/content/docs/versioned/import.md`

No page was added, moved, split, merged, or retired, so `docs/site/CONTENT_INVENTORY.md` does not change.

### DIFF
--- a/docs/migrations-import.md
+++ b/docs/migrations-import.md
@@ -32,7 +32,7 @@ The source tool is auto-detected; pass `--from` to be explicit or to disambiguat
 | Tool | Status | Notes |
 | --- | --- | --- |
 | golang-migrate | **Supported** | `<version>_<name>.up.sql` / `.down.sql`; integer or timestamp versions. |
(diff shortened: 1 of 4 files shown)


denisvmedia:
@codex review

